PR #7, opened 2019-01-18
install_zip.sh: convert sparse images

metti:
Convert *.img files that are detected as 'sparse' to regular image
files.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,6 +1,14 @@
+cuttlefish-common (0.9+nmu5) UNRELEASED; urgency=medium
+
+  * install_zip.sh: exit early in case of errors
+  * install_zip.sh: support partial extraction
+  * install_zip.sh: convert sparse images
+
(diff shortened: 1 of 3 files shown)


Databean:
Can you please add a new version to debian/changelog?

debchange should make it easier to get the format right.

debian/changelog
```diff
@@ -1,6 +1,14 @@
+cuttlefish-common (0.9+nmu5) UNRELEASED; urgency=medium
+
+  * install_zip.sh: exit early in case of errors
+  * install_zip.sh: support partial extraction
```
Databean: I'm not clear on where partial extraction and exiting early is added. The 'find "${destdir}" -name "*.img"' code only seems to do the sparse image conversion.

metti: That happened already here: https://github.com/google/android-cuttlefish/pull/6
This commit updates the change log including the missing entry from the mentioned PR.

Databean: Sounds good, thanks.

PR #14, opened 2019-03-19
Make images sparse where possible

ghartman17022:
BUG: 128526971
Change-Id: Ib53d99e5dbb1e995d6be2f2ddcdbc396af5141c6

### DIFF
--- a/host/deploy/install_zip.sh
+++ b/host/deploy/install_zip.sh
@@ -50,7 +50,7 @@ case $# in
 esac
 
 mkdir -p "${destdir}"
-bsdtar -x -C "${destdir}" -f "${source}" ${files_to_extract}
+bsdtar -x -S -C "${destdir}" -f "${source}" ${files_to_extract}
 
 if [[ " ${files_to_extract[*]} " == *" boot.img "* ]]; then
     /usr/lib/cuttlefish-common/bin/unpack_boot_image.py -boot_img "${destdir}/boot.img" -dest "${destdir}"


ghartman17022:
With this change:

fetch_artifacts.py -branch aosp-pie-cuttlefish-testing -build_target aosp_cf_x86_phone-userdebug
du -sh *.img
6.6M	boot.img
60K	cache.img
915M	system.img
1.2M	userdata.img
58M	vendor.img

ghartman17022:
This may be rendering things unbootable

ghartman17022:
For reference, I'm testing with ab/5384584

ghartman17022:
Yeah, sadly qemu can't figure out the partition sizes.

ghartman17022:
Actually, it looks like a bug in bsdtar. I uploaded with -S, and both qemu and crosvm failed.

I then tried an upload without -S. The file sizes as reported by ls differ:

$ ls -l system.img sparse/system.img 
-rw-rw-r--+ 1 vsoc-01 vsoc-01 2147487744 Jan  1  2008 sparse/system.img
-rw-r--r--  1 vsoc-01 vsoc-01 3221225472 Jan  1  2008 system.img

And it looks like the front of the files actually matches:

cmp system.img sparse/system.img 
cmp: EOF on sparse/system.img

And the offending section of the file is all 0's:

dd if=system.img ibs=2147487744 skip=1 | od 
0000000 000000 000000 000000 000000 000000 000000 000000 000000
*
0+1 records in
2097144+0 records out
7777770000
1073737728 bytes (1.1 GB, 1.0 GiB) copied, 14.1298 s, 76.0 MB/s

It looks like they forgot to extend the file in the case where it ends in a block of all 0's.

ghartman17022:
This is really odd. It appears to be triggered by the way that I'm using ssh to stream the archives up to the instance. It untars correctly on my local system.

ghartman17022:
Ah, the problem appears when the input is a pipe, even in the latest code. That's rather interesting.

ghartman17022:
I don't think we're going to be able to do this one.

PR #18, opened 2019-04-02
Turn on detailed logging in ssh

ghartman17022:
Log every connection and also every attempt to open a channel
(aka connect through a tunnel) on the console. These messages
will then appear on the first serial port, making them available
on TreeHugger.

BUG: 113671281
Test: Built image, logs appear
Change-Id: If3759ee98b86c757b7ad954705af351f4be1a466

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,10 @@
+cuttlefish-common (0.9.5) UNRELEASED; urgency=medium
+
+  * No need to displace instance_configs.cfg.template
+  * Turn on detailed logging in ssh
+
+ -- Greg Hartman <ghartman@google.com>  Tue, 02 Apr 2019 17:48:43 -0700
+
 cuttlefish-common (0.9.4) stable; urgency=medium
 
   * Add binutils for kernel decompresson on crosvm boots
--- a/debian/control
+++ b/debian/control
@@ -48,8 +48,6 @@ Depends:
  socat,
  vde2,
  ${misc:Depends}
-Provides: ${diverted-files}
-Conflicts: ${diverted-files}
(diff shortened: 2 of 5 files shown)


host/packages/cuttlefish-integration/etc/ssh/sshd_config.cuttlefish
```diff
@@ -0,0 +1,123 @@
...
+#HostKey /etc/ssh/ssh_host_rsa_key
+#HostKey /etc/ssh/ssh_host_ecdsa_key
+#HostKey /etc/ssh/ssh_host_ed25519_key
+
+# Ciphers and keying
+#RekeyLimit default none
+
+# Logging
+SyslogFacility LOCAL1
+LogLevel DEBUG1
```
Databean: What do you think of using a transform instead of a displace here? It's hard to tell from this file what's special about cuttlefish-integration and what were the existing options set in the file. With a transform, only the added lines will have to be kept in the android-cuttlefish repository. I think the transform would be something like

/etc/ssh/sshd_config cat - <(echo "SyslogFacility LOCAL1") <(echo "LogLevel DEBUG1")

The dash adds stdin, then the two echo statements add the extra lines.

http://manpages.ubuntu.com/manpages/bionic/man1/dh_configpackage.1.html

The rest of the file could then be removed. That avoids permanently fixing the sshd_config when other parts of debian may want to change it later.

ghartman17022: Thanks for the suggestion. That approach won't work because /etc/ssh/sshd_config doesn't appear in dpkg -L, and the transform checks for this before packaging.

I did something similar in the postinstll script.

PR #21, opened 2019-04-16
install_zip.sh: do not fail when files_to_extract does not match

metti:
Files passed as arguments to install_zip.sh might not exist in a
particular archive. In that case, warn the user and proceed.

Bug: 129743198
Signed-off-by: Matthias Maennich <maennich@google.com>

### DIFF
--- a/host/deploy/install_zip.sh
+++ b/host/deploy/install_zip.sh
@@ -51,7 +51,11 @@ case $# in
 esac
 
 mkdir -p "${destdir}"
-bsdtar -x -C "${destdir}" -f "${source}" ${sparse} ${files_to_extract}
+if ! bsdtar -x -C "${destdir}" -f "${source}" ${sparse} ${files_to_extract}; then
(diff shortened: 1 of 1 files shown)


metti:
@ajs1984 Could you have a look?

ajs1984:
Hmm, so recently we've had a few problems with bsdtar, like 4GB limit bugs and other errors, I'm not sure I'm too happy masking the error. Is there any reason we don't just extract everything, rather than the named list of ${files_to_extract}?

metti:
> Is there any reason we don't just extract everything, rather than the named list of ${files_to_extract}?
`install_zip.sh` might be called several times to extract into the same directory, hence it already overwrites what is potentially there. So, extracting more than is requested is problematic.
Another solution would be to probe the .zip for its content first, but we might only see it as stdin, that is not an option.
I would really like to avoid the 3rd option to extract everything first into a tmp dir and move it around later. That was space and time consuming in the past.

PR #25, opened 2019-06-10
Add a dependency on u-boot-tools.

Databean:
This supports running `mkimage -T script` to create u-boot script images
at launch_cvd runtime which can be invoked by u-boot inside the guest.

Bug: 133797099
Change-Id: I186d3f17314ec811eff52a7807b5bd65f3eb2d84

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -3,6 +3,7 @@ cuttlefish-common (0.9.8) UNRELEASED; urgency=medium
   * Changes to enable arm builds
   * Fix dependencies for buster
   * Add qemu-user-static on non-amd64
+  * Add dependency on u-boot-tools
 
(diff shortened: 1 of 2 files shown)


ghartman17022:
It seems to me that the dependencies for uboot should generally come from the host tar, not the stable host image.

There are two reasons for changing the stable host host image:

1. Something is needed to fetch the build
2. Something needs to run as root to configure the host environment

It seems like uboot stuff should be able to come with the build and wouldn't need to run as root on the host.

ghartman17022:
In this case I believe the code is here in the Android tree:

external/u-boot/Android.bp

Databean:
Fair enough, thanks for the reference. Note that u-boot is on master, Q-preview, N-iot, and O-iot, but not P. We might have to work around this for P support.

PR #31, opened 2019-09-13
Fix dependencies for cuttlefish-common

theanine:
Remove superfluous dependencies for cuttlefish-common

Add in required dependencies for all platforms:
  gcc
  libfdt1
  libx11-6
  libxext6

Add in dependencies for arm64:
  binfmt-support
  libc6:amd64
  qemu-user-static

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -16,13 +16,15 @@ Build-Depends:
  dh-buildinfo,
  config-package-dev (>= 4.5~),
  libnl-3-dev,
- libnl-genl-3-dev
+ libnl-genl-3-dev,
+ gcc
 
 Package: cuttlefish-common
 Architecture: any
 Depends:
  acl,
  binutils,
+ binfmt-support [!amd64],
  bridge-utils,
  device-tree-compiler,
  dnsmasq-base,
@@ -33,18 +35,23 @@ Depends:
  libarchive-tools | bsdtar,
  libdrm2,
  libc6,
+ libc6:amd64 [!amd64],
+ libfdt1,
  libgoogle-glog0v5 | libgoogle-glog0,
  libnl-genl-3-200,
  libstdc++6,
  libudev1,
  libwayland-client0,
  libwayland-server0,
+ libx11-6,
+ libxext6,
  libxml2,
  lzop,
  psmisc,
  python,
  libc6:amd64 [!amd64],
  qemu-user-binfmt [!amd64],
+ qemu-user-static [!amd64],
  qemu-system-common (>= 2.8.0),
  qemu-system-x86 (>= 2.8.0),
  qemu-utils (>= 2.8.0),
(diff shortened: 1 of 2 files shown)


debian/control
```diff
@@ -4,53 +4,24 @@ Section: misc
 Priority: optional
 Standards-Version: 3.9.2
 Build-Depends:
- clang-3.8,
```
Databean: Is this not necessary for adbshell.c?

https://github.com/google/android-cuttlefish/blob/master/host/commands/adbshell/adbshell.c

Though maybe gcc should be listed as a build dependency rather than clang, the Makefile looks like it is implicitly using gcc.

debian/control
```diff
@@ -4,53 +4,24 @@ Section: misc
...
+ libc6:amd64 [!amd64],
  libdrm2,
- libc6,
- libgoogle-glog0v5 | libgoogle-glog0,
- libnl-genl-3-200,
- libstdc++6,
- libudev1,
+ libfdt1,
  libwayland-client0,
- libwayland-server0,
```
Databean: I thought this was necessary for crosvm?

https://github.com/google/android-cuttlefish/commit/8706386b1b053834a929670be58f5a9cb6a0627e

Databean:
> Add in required dependencies for all platforms:
> libfdt1
> libx11-6
> libxext6
> 
> Add in dependencies for arm64:
> binfmt-support
> libc6:amd64
> qemu-user-static

I think the first line of the commit message is being used as the title, so the title just announces deleting dependencies when there are also some being added. Can the title be changed to cover both adding and deleting dependencies?

debian/control
theanine: Yeah, looks like I missed this because the buster build script is installing build-essential so gcc is being used.  I'll add gcc here.

debian/control
theanine: Why would the Wayland server be needed on the rock pi?  Either way, I was able to run under crosvm without it.

theanine:
> I think the first line of the commit message is being used as the title, so the title just announces deleting dependencies when there are also some being added. Can the title be changed to cover both adding and deleting dependencies?

Yeah, I should probably just break this up into two commits.

theanine:
PTAL

debian/control
Databean: You're right, I'm also able to launch without it on GCE.

debian/control
```diff
@@ -4,60 +4,26 @@ Section: misc
...
  python,
- libc6:amd64 [!amd64],
- qemu-user-binfmt [!amd64],
  qemu-user-static [!amd64],
- qemu-system-common (>= 2.8.0),
- qemu-system-x86 (>= 2.8.0),
- qemu-utils (>= 2.8.0),
- rsync,
- socat,
- vde2,
```
Databean: Please remove the "vde" capability together with this:
https://github.com/google/android-cuttlefish/blob/master/host/deploy/capability_query.py#L27

The capability was supposed to signal whether VDE was available on the host, but the implementation that used it never ended up being submitted so that's also part of this legacy.

theanine: Fixed

PR #33, opened 2019-11-05
Add missing libusb-1.0-0 dependency

Databean:
This is loaded by crosvm at runtime, and a system with only cuttlefish-common doesn't otherwise necessarily have this package.

May be obsoleted by b/140963307 in the future.

Originally reported by malchev@.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,15 @@
+cuttlefish-common (0.9.10) UNRELEASED; urgency=medium
+
+  [ Tristan Muntsinger ]
+  * Add in required dependencies for arm64 buster
+  * Remove superfluous dependencies for cuttlefish-common
+  * Add in net-tools dependency
+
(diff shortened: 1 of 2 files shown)


debian/changelog
```diff
@@ -1,3 +1,15 @@
+cuttlefish-common (0.9.10) UNRELEASED; urgency=medium
+
+  [ Tristan Muntsinger ]
+  * Add in required dependencies for arm64 buster
+  * Remove superfluous dependencies for cuttlefish-common
+  * Add in net-tools dependency
```
jemoreira: Is this part of this PR?

Databean: No, this is catching up as
https://github.com/google/android-cuttlefish/pull/32/files
and
https://github.com/google/android-cuttlefish/pull/31/files
did not update the changelog.

Databean:
Thanks for reviewing!

PR #35, opened 2019-12-04
Add missing dnsmasq dependency

jmacnak:
This is needed in create_interface() in debian/cuttlefish-common.init.

Found while investigating for b/128842306


jmacnak:
@googlebot I signed it!

PR #37, opened 2019-12-18
Add a Dockerfile for running cuttlefish inside a container

malchev:
This Dockerfile creates an image based on a systemd-enabled
debian-10-based image.  Containers created from this image have all the
prerequisites needed to run cuttlefish:

-- all necessary debian packages (including cuttlefish-common)
-- ssh setup for password-less nonroot access to the container
-- an empty vsoc-01 account ready to host cuttlefish images

Instructions for building and using are in file BUILDING.md

Signed-off-by: Iliyan Malchev <malchev@google.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,55 @@
+# This file is based on https://hub.docker.com/r/jrei/systemd-debian/.
+
+FROM debian:buster
+
+ENV container docker
+ENV LC_ALL C
+ENV DEBIAN_FRONTEND noninteractive
+
+RUN apt-get update \
+    && apt-get install -y systemd \
+    && apt-get clean \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM jrei/systemd-debian:10
```
Databean: Docker has a concept of "official images", should this be based off of one of those instead?

https://docs.docker.com/docker-hub/official_images/
https://hub.docker.com/_/debian

adelva1984: I wondered the same. The fork w/ systemd made sense when Debian had sysv init but that isn't the default in 10 (or 9?). Maybe this fork does other things too?

PR #42, opened 2020-02-14
Update version numbers in changelog

jmacnak:
A couple images were released without updating the changelog. The
0.9.12 image was just released for the GPU work so let's at least
try to adjust and keep it accurate going forward.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,12 +1,4 @@
-cuttlefish-common (0.9.12) stable; urgency=medium
-
-  [ Alistair Delva ]
-  * Disable masquerading of DNS lookups
-  * Added /etc/default/cuttlefish-common config file
-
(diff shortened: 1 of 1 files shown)


debian/changelog
```diff
@@ -8,7 +8,7 @@ cuttlefish-common (0.9.11) UNRELEASED; urgency=medium
 
  -- Jason Macnak <natsu@google.com> Tue, 11 Feb 2020 14:56:24 -0800
 
-cuttlefish-common (0.9.10) UNRELEASED; urgency=medium
+cuttlefish-common (0.9.12) stable; urgency=medium
```
Databean: This seems like rewriting history a bit, don't understand the reason for this change.

jmacnak: Correct, the goal of this change is to sync the version numbers in this changelog with the released images.

PR #55, opened 2020-07-06
Debian fixes

rpavlik:
This updates the Debian packaging to be a bit more modern and less error/warning filled.

- CDBS no longer used
- Structured copyright file
- rules file now entirely uses modern dh
- Many lintian errors/warnings fixed, though not all
- Un-parseable changelog entries fixed or, in the case of old version lines without release/signature lines, excluded as "Old Changelog"


rpavlik:
@googlebot I signed it!

PR #58, opened 2020-07-25
Docker fixes onto main

malchev:
A few small fixes to the shell functions.  Also adds a shell function to log into a container without having to fish out the IP address.


adelva1984:
We merged this another way. This can be closed.

PR #70, opened 2020-09-25
Improv download aosp script, make sure it's always mapped into the container

malchev:


kwstephenkim:
Will create another branch to modify the BUILDING.md accordingly.

PR #80, opened 2020-11-08
Added vsock_guest_cid option and separated port forwarding logic

kwstephenkim:
1. --vsock_guest_cid enables the new cuttleifsh options in cvd_start_${name}
   and let host services in the container listen on the original cuttlefish ports
   such as 6520, 8443, etc
2. separated port forwarding logic from the rest
   host port & guest port are managed separately
   guest port are managed automatically depending on whether --vsock_guest_cid or not
   host port can be given when the new cvd_publish_${name} is invoked
   the default values of the host ports are the same as before

### DIFF
--- a/expose-port.sh
+++ b/expose-port.sh
@@ -1,0 +1,73 @@
+#
+# includeed by setup.sh
+# utility functions for port publishing for docker
+#
+
+
+# forward host tcp port to $2:$3/tcp
+function socat_tcp() {
+    nohup socat tcp4-listen:"$1",fork TCP:"$2":"$3" > /dev/null 2>&1 &
+}
+
+# forward host udp port to $2:$3/udp
+function socat_udp() {
+    nohup socat udp4-listen:"$1",fork UDP:"$2":"$3" > /dev/null 2>&1 &
+}
+
+# socat a range of ports unconditionally
+#
+# 1: container name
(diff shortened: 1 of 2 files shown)


kwstephenkim:
Besides a little change of indentation, the changes are:
 1. separated port forwarding logic
   - Not port forwarding at provisioning time at all
   - If some users need a quick forwarding solution, we offer socat-based bash function to do so
   - host and guest ports are managed separately
   - host ports are by default equal to the default cuttlefish ports (e.g. 8443 for webRTC), and can be overridden 
   - guest ports are automatically determined
     * If vsock_guest_cid option of new cuttlefish is intended to be used, we automatically decide the cid based on instance_num
       and the guest ports are the same as default cuttlefish ports
    * If users run old host packages that are missing --vsock_guest_cid, we calculated the guest ports based on --base_instance_num of launch_cvd
 2. As only new host binaries have the --vsock_guest_cid option, we offer the command line option for the users to choose to add the option or not to "cvd_start_${name}"

expose-port.sh
```diff
@@ -0,0 +1,73 @@
+#
+# includeed by setup.sh
```
adelva1984, line 2: typo: included

adelva1984 (approved):

PR #84, opened 2020-12-16
Add IPv6 configuration to cuttlefish-common

hsinyichen-google:
This commit configures IPv6 NAT for the WiFi and Ethernet interfaces.
In the subnet, the dnsmasq process provides DNS information via DHCP,
and also sends router advertisement for the virtual machines to
determine their addresses.

### DIFF
--- a/debian/cuttlefish-common.init
+++ b/debian/cuttlefish-common.init
@@ -55,9 +55,16 @@ ebtables=${ebtables:-ebtables}
 
 # Start DHCP server on an interface
 # $1 = interface to bind to
-# $2 = listen address (gateway IP)
+# $2 = listen address (IPv4 gateway)
 # $3 = dhcp_range ("a.b.c.d,w.x.y.z")
+# $4 = IPv6 address prefix ("a:b::")
+# $5 = IPv6 address prefix length
 start_dnsmasq() {
+    if [ -n "${4}" -a -n "${5}" ]; then
+        ipv6_args="--dhcp-range=${4},ra-stateless,${5} --enable-ra"
+    else
+        ipv6_args=""
+    fi
     dnsmasq \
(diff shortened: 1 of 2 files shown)


adelva1984:
I'm curious, what motivated you to add NAT support? We have support already for IPv6 bridging, but you have to enable it in the cuttlefish-common default config. IPv6 NAT is a strange beast and normally not required on a real IPv6 network. Did you test your changes with the IPv6 bridge feature enabled as well (no dhcp, no ra, no static addresses on bridges).

hsinyichen-google:
I want to make the CF instances on my workstation able to access IPv6 network.
I tried to set bridge_interface in /etc/default/cuttlefish-common, but couldn't get it to work. It seems to require extra setup on the host. Could you explain how to enable the IPv6 bridge feature?

debian/cuttlefish-common.init
```diff
@@ -197,13 +222,18 @@ destroy_bridged_interfaces() {
 
 start() {
     # Enable ip forwarding
-    echo 1 >/proc/sys/net/ipv4/ip_forward
+    echo 1 > /proc/sys/net/ipv4/ip_forward
+    echo 1 > /proc/sys/net/ipv6/conf/all/forwarding
```
adelva1984, line 226: If you aren't NAT'ing any more, do you need ipv6 forwarding?

hsinyichen-google: I tried turning off ipv6 forwarding. The device got IPv6 addresses and dropped them in a few seconds. Eventually it had an IPv4 address but no IPv6 ones.
When ipv6 forwarding is on, the device got persistent IPv6 addresses even though it could not access the Internet. I have no idea why ipv6 forwarding made the difference.

PR #85, opened 2021-01-20
Let the user know that using WebRTC requires socat

sirkkalap:
I do not yet know how to install the binary, but after cvd_publish I get constant errors where nohup socat fails to run, so I assume socat is required.

```
petri@test-petri:~/android-cuttlefish$ cvd_publish_cuttlefish 0
[1]   Exit 127                nohup socat tcp4-listen:"$1",fork TCP:"$2":"$3" > /dev/null 2>&1
[2]   Exit 127                nohup socat tcp4-listen:"$1",fork TCP:"$2":"$3" > /dev/null 2>&1
```

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -311,3 +311,5 @@ connect (on the same machine as the docker container) by pointing your browser
 at https://localhost:8443/
 
 WebRTC allows you to control cuttlefish inside the docker container without ssh tunnel.
+
+Using WebRTC requires installing `socat` binary to the host.


sirkkalap:
@googlebot I signed it!

sirkkalap:
I will leave this as a draft, since the edits I made are very rudimentary and likely need more detail to be useful. If you wish I will take a closer look at this, or you may opt to close this with no action too.

kwstephenkim:
Thank you for your comments & pull request. 

The cvd_publish_* function is optional. And, the BUILDING.md file is not up to date, and does not seem to be clear on how the users should use the scripts and/or the containers. When we updated the changes in the .sh files, etc, we did not update BUILDING.md accordingly. Sorry for that, and thank you for reminding us of it.

I will update the BUILDING.md file, etc, accordingly. 

As of now, the users of the containers are recommended that they should assign an address per container, and access each container by the corresponding IP address.

If the users cannot assign an IP address, there would be two categories of use cases:
    1. Access locally, i.e. from the docker host to the docker container
    2. Access remotely, i.e. from anywhere outside the docker host to the docker container
   
In case of no. 1, there is another function called ```get_cvd_ip```. That is used like ```get_cvd_ip ${name}``` and, I guess that it returns the ip address of the docker container ```${name}```, which is valid on the docker host.

```cvd_publish_*``` functions should be optional. When the user cannot assign an IP and does want to remotely access to the container, we temporarily offer a convenience function: ```cvd_publish_${name}```.

The users are free to create ssh tunnel, instead. So, socat is optional. But if the users prefer socat over anything else (e.g. ssh tunneling), then, socat should be installed. 

I'll close this pull request, and as you kindly told us, we will update the BUILDING.md file soon.

PR #93, opened 2021-08-12
Add e2fsprogs package to control file

theanine:
This is for the e2fsck binary so cuttlefish can support an ext4 userdata filesystem.


Databean (approved):

Databean (commented):
Consider updating debian/changelog as well to signal this change.

PR #101, opened 2021-11-04
 Add partial implementation of host orchestrator

jemoreira:
This only adds the code to our repo, but it doesn't build as part of any debian package yet. It also still doesn't use a unix socket endpoint for the device, it uses a websocket instead so it can be easily tested with the current cuttlefish code.

To test you can do:
```
cd host/frontend/host-orchestrator
go get cuttlefish/host-orchestrator
go build
./host-orchestrator
```
The last command starts a server listening for devices in port 8081 and for clients in port 8080, so you can register a device with
```
launch_cvd --nostart_webrtc_sig_server --nowebrtc_sig_server_secure --webrtc_sig_server_port=8081
```
Then simply direct your browser to http://localhost:8080

### DIFF
--- a/host/frontend/host-orchestrator/clients.go
+++ b/host/frontend/host-orchestrator/clients.go
@@ -16,9 +16,17 @@ package main
 
 import (
 	"math/rand"
+	"strings"
 	"sync"
 	"time"
 )
+
+type Client interface {
+	// Send a message to the device
+	Send(msg interface{}) bool
+	// Provides an oportunity for the client to react to the device being disconnected
+	OnDeviceDisconnected()
+}
 
 // Implements the client interface using a websocket connection
 type WsClient struct {
@@ -48,7 +56,7 @@ type PolledClient struct {
 	// The id given to this client by the device
 	clientId int
 	// A reference to the polled set to clean up when the device disconnects
-	set *PolledSet
+	polledSet *PolledSet
 }
 
 // From IClient
@@ -60,7 +68,7 @@ func (c *PolledClient) Send(msg interface{}) bool {
 }
 func (c *PolledClient) OnDeviceDisconnected() {
 	// This object is useless when the device disconnects, release all resourcess associated to it
-	c.set.Destroy(c.id)
+	c.polledSet.Destroy(c.id)
 }
 
 // Sends a message to the device
@@ -78,9 +86,11 @@ func (c *PolledClient) GetMessages(start int, count int) []interface{} {
 	end := start + count
 	if end > len(c.messages) {
 		end = len(c.messages)
+		count = end - start
 	}
-	// Safe to return slice because c.messages is never shrunk
-	return c.messages[start:end]
+	ret := make([]interface{}, count)
+	copy(ret, c.messages[start:end])
+	return ret
 }
 
 func (c *PolledClient) Id() string {
@@ -104,7 +114,12 @@ func (s *PolledSet) NewConnection(d *Device) *PolledClient {
 	for {
 		id := randStr(64)
 		if _, ok := s.connections[id]; !ok {
-			conn := &PolledClient{id: id, device: d, messages: make([]interface{}, 0), set: s}
+			conn := &PolledClient{
+				id:        id,
+				device:    d,
+				messages:  make([]interface{}, 0),
+				polledSet: s,
+			}
 			s.connections[id] = conn
 			conn.clientId = d.Register(conn)
 			return conn
@@ -133,9 +148,9 @@ func NewPolledSet() *PolledSet {
 const runes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
 
 func randStr(l int) string {
-	s := make([]byte, l)
-	for i := range s {
-		s[i] = runes[rand.Intn(len(runes))]
+	var s strings.Builder
+	for i := 0; i < l; i++ {
+		s.WriteByte(runes[rand.Intn(len(runes))])
 	}
-	return string(s)
+	return s.String()
 }
--- a/host/frontend/host-orchestrator/clients.go
+++ b/host/frontend/host-orchestrator/clients.go
@@ -23,7 +23,7 @@ import (
 
 type Client interface {
 	// Send a message to the device
-	Send(msg interface{}) bool
+	Send(msg interface{}) error
 	// Provides an oportunity for the client to react to the device being disconnected
 	OnDeviceDisconnected()
 }
@@ -38,7 +38,7 @@ func NewWsClient(ws *JsonWs) *WsClient {
 }
 
 // From IClient
-func (c *WsClient) Send(msg interface{}) bool {
+func (c *WsClient) Send(msg interface{}) error {
 	return c.ws.Send(msg)
 }
 func (c *WsClient) OnDeviceDisconnected() {
@@ -60,11 +60,11 @@ type PolledClient struct {
 }
 
 // From IClient
-func (c *PolledClient) Send(msg interface{}) bool {
+func (c *PolledClient) Send(msg interface{}) error {
 	c.msgMtx.Lock()
 	defer c.msgMtx.Unlock()
 	c.messages = append(c.messages, msg)
-	return true
+	return nil
 }
 func (c *PolledClient) OnDeviceDisconnected() {
 	// This object is useless when the device disconnects, release all resourcess associated to it
@@ -72,7 +72,7 @@ func (c *PolledClient) OnDeviceDisconnected() {
 }
 
 // Sends a message to the device
-func (c *PolledClient) ToDevice(msg interface{}) bool {
+func (c *PolledClient) ToDevice(msg interface{}) error {
 	return c.device.Send(msg)
 }
 
--- a/host/frontend/host-orchestrator/clients.go
+++ b/host/frontend/host-orchestrator/clients.go
@@ -1,0 +1,141 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+//     Unless required by applicable law or agreed to in writing, software
+//     distributed under the License is distributed on an "AS IS" BASIS,
+//     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//     See the License for the specific language governing permissions and
+//     limitations under the License.
+
+package main
+
+import (
+	"math/rand"
+	"sync"
+	"time"
+)
+
+// Implements the client interface using a websocket connection
+type WsClient struct {
+	ws *JsonWs
+}
+
+func NewWsClient(ws *JsonWs) *WsClient {
+	return &WsClient{ws: ws}
+}
+
+// From IClient
+func (c *WsClient) Send(msg interface{}) bool {
+	return c.ws.Send(msg)
+}
+func (c *WsClient) OnDeviceDisconnected() {
+	// Do nothing
+}
+
+// Implements the client interface using a polled connection
+type PolledClient struct {
+	// The polled connection id
+	id       string
+	device   *Device
+	messages []interface{}
+	// Synchronizes access to the messages list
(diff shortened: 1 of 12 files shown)


host/frontend/host-orchestrator/clients.go
```diff
@@ -0,0 +1,141 @@
...
+func NewPolledSet() *PolledSet {
+	rand.Seed(time.Now().UnixNano())
+	return &PolledSet{connections: make(map[string]*PolledClient)}
+}
+
+// The available characters for the connection ids
+const runes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+
+func randStr(l int) string {
+	s := make([]byte, l)
```
Databean, line 136: Can this use strings.Builder?

https://pkg.go.dev/strings#example-Builder

host/frontend/host-orchestrator/operator.go
```diff
@@ -0,0 +1,400 @@
...
+	if devMux == nil {
+		log.Fatal("Failed to allocate ServeMux")
+	}
+	// http.HandleFunc("/", serveHome)
+	devMux.HandleFunc("/register_device", func(w http.ResponseWriter, r *http.Request) {
+		deviceWs(w, r, pool, config)
+	})
+	// Serve the register_device endpoint in a different thread
+	go func() {
+		err := http.ListenAndServe("127.0.0.1:8081", devMux)
```
Databean, line 57: Listening on this address enforces that it only accepts local connections, right?

host/frontend/host-orchestrator/operator.go
```diff
@@ -0,0 +1,400 @@
...
+	"os"
+	"strconv"
+
+	"github.com/gorilla/mux"
+)
+
+func main() {
+	pool := NewDevicePool()
+	polledSet := NewPolledSet()
+	var config interface{}
```
Databean, line 31: Can this get its own type rather than using `interface{}`? The only requirement is that it eventually serializes into exactly the correct json, but that can be controlled with [annotations](https://pkg.go.dev/encoding/json#Marshal).

host/frontend/host-orchestrator/jsonws.go
```diff
@@ -0,0 +1,159 @@
...
+	send, recv, cl := make(chan []byte, 100), make(chan []byte, 100), make(chan bool)
+	go writePump(conn, send)
+	go readPump(conn, recv, cl)
+	return &JsonWs{send: send, recv: recv, cl: cl, conn: conn}
+}
+
+func (ws *JsonWs) Send(val interface{}) bool {
+	select {
+	case _, _ = <-ws.cl:
+		// A message on this channel means the ws is closed
```
Databean, line 52: nit: This isn't ever reading a message, but is instead getting the "closed indicator":

https://stackoverflow.com/a/16105721

host/frontend/host-orchestrator/operator.go
```diff
@@ -0,0 +1,400 @@
...
+	vars := mux.Vars(r)
+	devId := vars["deviceId"]
+	dev := pool.GetDevice(devId)
+	if dev == nil {
+		http.NotFound(w, r)
+		return
+	}
+	path := vars["path"]
+	if shouldIntercept(path) {
+		http.ServeFile(w, r, fmt.Sprintf("intercept%s", path))
```
Databean, line 185: `shouldIntercept` does `fmt.Sprintf("intercept/%s", ...)` while this does `fmt.Sprintf("intercept%s", ...)`, what's the distinction between these situations?

Databean, line 185: Is it possible for the client to request arbitrary files here by writing `intercept/../../../../whatever`?

Databean (commented):
Would it be possible to add unit tests for the go code? `jsonws.go` especially, the operations there look a bit low-level.

Haven't finished reviewing, but posting some comments early.

host/frontend/host-orchestrator/operator.go
jemoreira: Yes, as opposed to ":8080" which accepts anything.

host/frontend/host-orchestrator/clients.go
```diff
@@ -0,0 +1,141 @@
...
+	// The polled connection id
+	id       string
+	device   *Device
+	messages []interface{}
+	// Synchronizes access to the messages list
+	msgMtx sync.Mutex
+	// The id given to this client by the device
+	clientId int
+	// A reference to the polled set to clean up when the device disconnects
+	set *PolledSet
```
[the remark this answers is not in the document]
jemoreira: Done

host/frontend/host-orchestrator/clients.go
```diff
@@ -0,0 +1,141 @@
...
+	c.msgMtx.Lock()
+	defer c.msgMtx.Unlock()
+	if count < 0 {
+		count = len(c.messages) - start
+	}
+	end := start + count
+	if end > len(c.messages) {
+		end = len(c.messages)
+	}
+	// Safe to return slice because c.messages is never shrunk
```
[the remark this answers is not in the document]
jemoreira: Done

host/frontend/host-orchestrator/clients.go
jemoreira: Done

host/frontend/host-orchestrator/operator.go
jemoreira: Done

host/frontend/host-orchestrator/jsonws.go
jemoreira: Reworded

host/frontend/host-orchestrator/operator.go
jemoreira: Yes
Given that we only care about intercepting one file I've changed the function to check for that file name specifically.

We could use a file server like the one for static and that would guarantee files within the directory only, but just checking the file name is simpler and unlikely to change in the near future anyways.

jemoreira: Fixed
There is no difference, path includes the leading '/' so it's not needed in `shouldIntercept`. It worked because apparently some/path is equivalent to some//path in linux.

jemoreira:
> Would it be possible to add unit tests for the go code? `jsonws.go` especially, the operations there look a bit low-level.

Added unit tests for clients.go and device.go. The other 2 files are not easy to unit tests because of their server nature and instead would benefit more from end to end testing.

host/frontend/host-orchestrator/clients.go
```diff
@@ -0,0 +1,156 @@
...
+	// This object is useless when the device disconnects, release all resourcess associated to it
+	c.polledSet.Destroy(c.id)
+}
+
+// Sends a message to the device
+func (c *PolledClient) ToDevice(msg interface{}) bool {
+	return c.device.Send(msg)
+}
+
+// Gets count messages from the device, skipping the first start messages
```
Databean, line 79: Are messages ever cleared? Can a long-running device pile up messages indefinitely?

host/frontend/host-orchestrator/operator.go
```diff
@@ -0,0 +1,401 @@
...
+		IceServers: []IceServer{
+			IceServer{URLs: []string{"stun:stun.l.google.com:19302"}},
+		},
+	}
+
+	setupDeviceEndpoint(pool, config)
+	r := setupServerRoutes(pool, polledSet, config)
+
+	http.Handle("/", r)
+	if err := http.ListenAndServe(":8080", nil); err != nil {
```
Databean, line 41: Maybe worth taking a port other than 8080 if the package is going to always run this, as many demo http servers run at port 8080 or 8000.

host/frontend/host-orchestrator/operator.go
```diff
@@ -0,0 +1,401 @@
...
+			if msg.Type != "forward" {
+				wsReplyError(ws, fmt.Sprintln("Unrecognized message type: ", msg.Type))
+				return
+			}
+			payload := msg.Payload
+			if payload == nil {
+				wsReplyError(ws, "Client forward message missing payload")
+				return
+			}
+			cMsg := map[string]interface{}{
```
Databean, line 253: Maybe worth creating a type for this message, considering it's used as well by `forward` below.

host/frontend/host-orchestrator/jsonws.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		log.Println(err)
+		return false
+	}
+
+	ws.writeMtx.Lock()
+	defer ws.writeMtx.Unlock()
+
+	err = ws.conn.WriteMessage(websocket.TextMessage, msg)
+	if err != nil {
```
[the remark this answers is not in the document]
jemoreira: Done

host/frontend/host-orchestrator/jsonws.go
```diff
@@ -0,0 +1,82 @@
...
+	writeMtx sync.Mutex
+}
+
+var upgrader = websocket.Upgrader{} // default options
+
+func NewJsonWs(w http.ResponseWriter, r *http.Request) *JsonWs {
+	conn, err := upgrader.Upgrade(w, r, nil)
+	if err != nil {
+		log.Println(err)
+		return nil
```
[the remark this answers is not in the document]
jemoreira: Done

host/frontend/host-orchestrator/devices.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	p.devices[id] = d
+	return true
+}
+
+func (p *DevicePool) Unregister(id string) {
+	p.devicesMtx.Lock()
+	defer p.devicesMtx.Unlock()
+	if d, ok := p.devices[id]; ok {
+		d.DisconnectClients()
```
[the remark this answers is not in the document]
jemoreira: clients have no access to the device pool.
The device pool is only used to keep track of devices by id, it can be accessed by a thread that handles client requests but the actual client won't need a reference to it because it cares about only one device.
Similarly devices don't need to reference the pool either, so the client won't end up calling to the pool indirectly

host/frontend/host-orchestrator/clients_test.go
```diff
@@ -0,0 +1,100 @@
...
+func TestDestroy(t *testing.T) {
+	ps := NewPolledSet()
+	c := ps.NewConnection(NewDevice(nil, 0, ""))
+	c.OnDeviceDisconnected()
+	if ps.GetConnection(c.Id()) != nil {
+		t.Error("connection was not destroyed after device disconnect")
+	}
+}
+
+func TestRandStr(t *testing.T) {
```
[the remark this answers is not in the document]
jemoreira: Done

host/frontend/host-orchestrator/clients.go
jemoreira: Messages are cleared when the device disconnects. In theory a device can pile up messages indefinitely, but devices will be connecting over a unix socket, so they'll be trusted. A malicious user can still create many clients and cause a device to send many messages to each, but that would still be the case if messages were cleared after read.
We could limit the number or clients per device and the number of messages, but what's a good limit?
We could also simply limit the memory consumption of the process (which is what I plan to do when making it start on boot)

host/frontend/host-orchestrator/operator.go
jemoreira: Done

host/frontend/host-orchestrator/operator.go
jemoreira: Done

host/frontend/host-orchestrator/devices.go
```diff
@@ -67,19 +68,20 @@ func (d *Device) Unregister(id int) {
 // Notify the clients that the device has disconnected
 func (d *Device) DisconnectClients() {
 	d.clientsMtx.Lock()
-	defer d.clientsMtx.Unlock()
-	for _, client := range d.clients {
+	clientsCopy := d.clients
```
Databean, line 71: This creates an alias rather than a copy, see this [playground example](https://play.golang.org/p/ZwLCCx6nyV1), and this [stack overflow answer](https://stackoverflow.com/a/23058707).

Databean (approved):

host/frontend/host-orchestrator/devices.go
jemoreira: Thanks for catching that!

PR #103, opened 2021-11-18
Serve orchestrator device endpoint in a unix socket

jemoreira:


host/frontend/host-orchestrator/jsonws.go
```diff
@@ -15,44 +15,86 @@
...
 	"sync"
 
 	"github.com/gorilla/websocket"
 )
 
-// A websocket connection that can send and receive json objects.
+// A websocket connection that can send and receive JSON objects.
 // Only one thread should call Recv() at a time, Send() and Close() are thread safe
-type JsonWs struct {
+type JSONWs struct {
```
Databean, line 29: Should this be [MixedCaps](https://golang.org/doc/effective_go#mixed-caps) style? `JsonWs` already followed the abbreviation rules mentioned below.

Because the Effective Go page is not specific on abbreviations, referencing the [c++ style guide](https://google.github.io/styleguide/cppguide.html#General_Naming_Rules):

> For the purposes of the naming rules below, a "word" is anything that you would write in English without internal spaces. This includes abbreviations, such as acronyms and initialisms. For names written in mixed case (also sometimes referred to as "[camel case](https://en.wikipedia.org/wiki/Camel_case)" or "[Pascal case](https://en.wiktionary.org/wiki/Pascal_case)"), in which the first letter of each word is capitalized, prefer to capitalize abbreviations as single words, e.g., StartRpc() rather than StartRPC().

host/frontend/host-orchestrator/jsonws.go
```diff
@@ -15,44 +15,86 @@
...
+func (ws *JSONWs) Close() {
 	ws.writeMtx.Lock()
 	defer ws.writeMtx.Unlock()
 	ws.conn.WriteMessage(websocket.CloseMessage, []byte{})
 	ws.conn.Close()
 }
+
+// A Unix socket connection (as returned by Accept) that can send recieve JSON objects.
+// Only one thread should call Recv() at a time, Send and Close are thread safe
+type JSONUnix struct {
```
Databean, line 64: `JsonUnix`?

host/frontend/host-orchestrator/operator.go
```diff
@@ -388,10 +391,15 @@ type IceServer struct {
 
 // Utility functions
 
-// Log and reply with an error over a websocket
-func wsReplyError(ws *JsonWs, msg string) {
+// Interface implemented by any connection capable of sending in JSON format
+type JSONConn interface {
```
Databean, line 395: `JsonConn`?

host/frontend/host-orchestrator/jsonws.go
```diff
@@ -15,44 +15,86 @@
...
+		return err
+	}
+	c.writeMtx.Lock()
+	defer c.writeMtx.Unlock()
+	_, err = c.conn.Write(s)
+	return err
+}
+
+func (c *JSONUnix) Recv(val interface{}) error {
+	n, err := c.conn.Read(c.buff)
```
Databean, line 89: The c++ side of this implementation is able to detect when its buffer is too small for the seqpacket messages, is it possible to implement the equivalent here?

Databean: Something that should help with this is to switch out some of the calls for ones that return more specific types:

 - [`net.ListenUnix`](https://pkg.go.dev/net#ListenUnix) rather than `net.Listen` produces a `UnixListener`
 - [`net.UnixListener.AcceptUnix`](https://pkg.go.dev/net#UnixListener) rather than `Accept` produces a `UnixConn`
 - [`net.UnixConn.ReadMsgUnix`](https://pkg.go.dev/net#UnixConn.ReadMsgUnix) produces more information than the `Read` method
 - If necessary, [`net.UnixConn.SyscallConn`](https://pkg.go.dev/net#UnixConn.SyscallConn) produces a [`syscall.RawConn`](https://pkg.go.dev/syscall#RawConn) which can temporarily expose the file descriptor
 -  The raw [`syscall.Recvmsg`](https://pkg.go.dev/syscall#Recvmsg) is available over file descriptors so it's possible to do `MSG_PEEK` and `MSG_TRUNC`

host/frontend/host-orchestrator/jsonws.go
jemoreira: Initialism in Go are supposed to be kept with all caps: JSON, HTTP, URL, etc.

host/frontend/host-orchestrator/jsonws.go
jemoreira: Initialisms, see previous reply

host/frontend/host-orchestrator/operator.go
jemoreira: same

host/frontend/host-orchestrator/jsonws.go
jemoreira: Done

Databean (approved):

PR #104, opened 2021-11-24
Build host-orchestrator with cuttlefish-common

jemoreira:
It just builds it and packages it, but doesn't start it at boot like it does with the cuttlefish-common service.

To test this you need to:
```
sudo service cuttlefish-host-orchestrator start
```
And then just launch_cvd from a recent master build, the device will be accessible in port 1080 over HTTP instead of port 8443 over HTTPS.

### DIFF
--- a/.github/workflows/test_build_deb.yaml
+++ b/.github/workflows/test_build_deb.yaml
@@ -11,7 +11,7 @@ jobs:
     - name: setup apt
       run: apt update -y && apt upgrade -y
     - name: install debuild dependencies
-      run: apt install -y git devscripts config-package-dev debhelper-compat
+      run: apt install -y git devscripts config-package-dev debhelper-compat golang
     - name: checkout repository
       uses: actions/checkout@a81bbbf8298c0fa03ea29cdc473d45769f953675 # aka v2
       with:
--- a/debian/cuttlefish-common.cuttlefish-host-orchestrator.init
+++ b/debian/cuttlefish-common.cuttlefish-host-orchestrator.init
@@ -98,14 +98,11 @@ else
         start|stop|status)
             "$1"
             ;;
-        restart)
+        restart|force-reload|condrestart|try-restart)
             stop && start
             ;;
-        condrestart|try-restart)
-            stop && start
-            ;;
-        reload|force-reload)
-            # Nothing to do; we reread configuration on each invocation
+        reload)
+            # Nothing to do
             ;;
         shutdown)
             stop
--- a/debian/cuttlefish-common.cuttlefish-host-orchestrator.init
+++ b/debian/cuttlefish-common.cuttlefish-host-orchestrator.init
@@ -1,0 +1,118 @@
+#!/bin/sh
+#
+### BEGIN INIT INFO
+# Provides: cuttlefish-host-orchestrator
+# Required-Start: $network $remote_fs
+# Required-Stop: $network $remote_fs
+# Default-Start:
+# Default-Stop: 0 1 2 3 4 5 6
+# Short-Description: Cuttlefish Host Orchestrator service
+# Description: The Host Orchestrator service provides the signaling
+#              server used by all cuttlefish instances running in this
+#              host as well as orchestration capabilities.
+### END INIT INFO
+#
+# Copyright (C) 2021 The Android Open Source Project
+#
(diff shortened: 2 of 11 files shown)


.github/workflows/test_build_deb.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
     - name: setup apt
       run: apt update -y && apt upgrade -y
     - name: install debuild dependencies
-      run: apt install -y git devscripts config-package-dev debhelper-compat
+      run: apt install -y git devscripts config-package-dev debhelper-compat golang
```
Databean, line 14: The [build instructions](https://cs.android.com/android/platform/superproject/+/master:device/google/cuttlefish/README.md;l=29;drc=9dba51e13a40b3b0eebbf3ec1ba7492b22452674) should probably be updated as well for this.

debian/cuttlefish-common.cuttlefish-host-orchestrator.init
```diff
@@ -0,0 +1,118 @@
...
+  chown _cutf-operator:cvdnetwork "${RUN_DIR}"
+  chmod 775 "${RUN_DIR}"
+
+  ORCHESTRATOR_PORT="${orchestrator_port}" \
+  ORCHESTRATOR_SOCKET_PATH="${RUN_DIR}"/operator \
+  start-stop-daemon --start \
+    --pidfile "${PIDFILE}" \
+    --chuid _cutf-operator:cvdnetwork \
+    --chdir "${ASSET_DIR}" \
+    --background --no-close \
```
Databean, line 55: The man page for `start-stop-daemon` writes

> **`-b`, `--background`**
> ...
> **Warning: `start-stop-daemon`** cannot check the exit status if the process fails to execute for any reason. This is a last resort, and is only meant for programs that either make no sense forking on their own, or where it's not feasible to add the code for them to do this themselves.

Since we control the code of the operator, should we add the forking code so `start-stop-daemon` can tell if the process ran successfully or not?

It looks like this might also require implementing the pidfile functionality internally as well though.

debian/cuttlefish-common.cuttlefish-host-orchestrator.init
```diff
@@ -0,0 +1,118 @@
...
+            "$1"
+            ;;
+        restart)
+            stop && start
+            ;;
+        condrestart|try-restart)
+            stop && start
+            ;;
+        reload|force-reload)
+            # Nothing to do; we reread configuration on each invocation
```
Databean, line 108: The [debian init script policy](https://www.debian.org/doc/debian-policy/ch-opersys.html#s-writing-init) writes
> - `reload`
> cause the configuration of the service to be reloaded without actually stopping and restarting the service,
> - `force-reload`
> cause the configuration to be reloaded if the service supports this, otherwise restart the service.
>
> ...
> The `start`, `stop`, `restart`, and `force-reload` options should be supported by all init scripts. Supporting `status` is encouraged. The `reload` and `try-restart` options are optional.

If I understand correctly, the configuration includes the `orchestrator_port` variable in `/etc/default/cuttlefish-common` which requires a process restart. `reload` is optional and might not be worth the trouble initially, but `force-reload` looks like it should restart the process to be able to pick up configuration changes.

debian/cuttlefish-common.cuttlefish-host-orchestrator.init
jemoreira: The operator is written in go, which doesn't seem to be too friendly to demonizing (no support for fork out of the box) although it can be done through adding a dependency (https://github.com/sevlyar/go-daemon) I read that the preferred way is to use an external tool like demonize.
The main benefit of checking the exit code would be to report whether the process was able to grab the resources it needs (port, socket, etc), which is certainly useful, but I think it's best to add it later as new feature and rely on start-stop-demon for now.

.github/workflows/test_build_deb.yaml
jemoreira: Done: aosp/1938906

debian/cuttlefish-common.cuttlefish-host-orchestrator.init
jemoreira: Done

Databean (approved):

PR #106, opened 2022-02-01
Supports https communication in host orchestrator.

ser-io:


host/frontend/host-orchestrator/operator.go
```diff
@@ -46,12 +51,26 @@ func main() {
...
-		log.Fatal("ListenAndServe client: ", err)
-	}
+
+	wg := new(sync.WaitGroup)
+	wg.Add(2)
+
+	go func() {
+		log.Println("Client http endpoint created")
+		log.Fatal(http.ListenAndServe(fmt.Sprint(":", httpPort), nil))
+		wg.Done()
```
jemoreira, line 62: this should be done with `defer` instead, in case log.Fatal panics for example.

host/frontend/host-orchestrator/operator.go
```diff
@@ -46,12 +51,26 @@ func main() {
...
+	}()
+
+	go func() {
+		log.Println("Client https endpoint created.")
+		certPath := tlsCertDir + "/cert.pem"
+		keyPath := tlsCertDir + "/key.pem"
+		log.Fatal(http.ListenAndServeTLS(fmt.Sprint(":", httpsPort), certPath, keyPath, nil))
+		wg.Done()
```
jemoreira, line 70: defer

host/frontend/host-orchestrator/operator.go
```diff
@@ -46,12 +51,26 @@ func main() {
...
 	r := setupServerRoutes(pool, polledSet, config)
-
 	http.Handle("/", r)
-	log.Println("Client endpoint created")
-	if err := http.ListenAndServe(fmt.Sprint(":", port), nil); err != nil {
-		log.Fatal("ListenAndServe client: ", err)
-	}
+
+	wg := new(sync.WaitGroup)
+	wg.Add(2)
```
jemoreira, line 57: This makes me nervous: What happens if someone wants to add a new go routine after line 71? They will most likely copy from 65 to 70 and modify as needed, but then they may not change this to add 3 instead.

I know this is the way recommended almost anywhere and there is nothing we can do to change the WaitGroup api, but we can put the call to Add next to the go routines, and do it one by one, so when someone copies they will see the Add() call.

Ideally, I think WaiGroup's Add() should take a closure to be executed in a routine instead of a number and you simply call wait after adding all the closures.

host/frontend/host-orchestrator/operator.go
```diff
@@ -46,12 +51,26 @@ func main() {
...
+		log.Println("Client http endpoint created")
+		log.Fatal(http.ListenAndServe(fmt.Sprint(":", httpPort), nil))
+		wg.Done()
+	}()
+
+	go func() {
+		log.Println("Client https endpoint created.")
+		certPath := tlsCertDir + "/cert.pem"
+		keyPath := tlsCertDir + "/key.pem"
+		log.Fatal(http.ListenAndServeTLS(fmt.Sprint(":", httpsPort), certPath, keyPath, nil))
```
jemoreira, line 69: Please leave a comment with a pointer to the docs that say it's safe to use http.ListenAndServe...() from different go routines.

debian/cuttlefish-common.cuttlefish-host-orchestrator.init
```diff
@@ -34,19 +34,34 @@
...
 RUN_DIR="/run/cuttlefish"
 ASSET_DIR="/usr/share/cuttlefish-common/host-orchestrator"
 DAEMON="/usr/lib/cuttlefish-common/bin/host-orchestrator"
+GENERATE_CERT="/usr/lib/cuttlefish-common/bin/generate_cert"
 PIDFILE="${RUN_DIR}"/host-orchestrator.pid
+CERT_DIR="/etc/cuttlefish-common/host-orchestrator/cert"
+
+function gen_cert {
+  mkdir -p "${CERT_DIR}"
+  cd "${CERT_DIR}"
```
jemoreira, line 49: nit: pushd and popd are a better choice, in case something does another cd

debian/cuttlefish-common.cuttlefish-host-orchestrator.init
```diff
@@ -34,19 +34,34 @@
...
+  # https://go.dev/src/crypto/tls/generate_cert.go
+  ${GENERATE_CERT} --host="localhost" --duration="867240h0m0s"
```
jemoreira, line 51: this should check that the certificates don't exist before trying to create them, otherwise you'll create a new certificate on every boot.
Also, checking before creating will allow to provide your own certificate if you want to.

debian/cuttlefish-common.cuttlefish-host-orchestrator.init
ser-io: Done

debian/cuttlefish-common.cuttlefish-host-orchestrator.init
ser-io: Done

host/frontend/host-orchestrator/operator.go
ser-io: great point, changed the way the WaitGroup is incremented in order to avoid what you pointed out, PTAL.

host/frontend/host-orchestrator/operator.go
ser-io: Done

host/frontend/host-orchestrator/operator.go
ser-io: made explicit creation of server instances per each go-routine to avoid the concurrency concerns that using `http.ListenAndServe` from different go-routines raise.

host/frontend/host-orchestrator/operator.go
```diff
@@ -53,23 +67,16 @@ func main() {
...
-		wg.Done()
-	}()
-
+	wg.Add(len(starters))
+	for _, starter := range starters {
+		starter_ := starter
+		go func(f func()) {
+			defer wg.Done()
+			f()
+		}(starter_)
```
jemoreira, line 78: by putting the function in the local variable you don't need to pass it as a parameter anymore, you can simply grab it from the closure.

host/frontend/host-orchestrator/operator.go
jemoreira: The server objects were not the problem, the problem is sharing the mux object, which this code still does. All you need to do is add a comment mentioning that object is thread safe. The code as written before was OK, and simpler and more readable than the new one.

ser-io: added a comment mentioning that DefaultServerMux is used by both servers and that http.ServeMux instances are thread-safe.

host/frontend/host-orchestrator/operator.go
ser-io: passed it as a parameter.

host/frontend/host-orchestrator/operator.go
```diff
@@ -53,23 +67,16 @@ func main() {
...
-		log.Println("Client https endpoint created.")
-		certPath := tlsCertDir + "/cert.pem"
-		keyPath := tlsCertDir + "/key.pem"
-		log.Fatal(http.ListenAndServeTLS(fmt.Sprint(":", httpsPort), certPath, keyPath, nil))
-		wg.Done()
-	}()
-
+	wg.Add(len(starters))
+	for _, starter := range starters {
+		starter_ := starter
```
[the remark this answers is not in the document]
ser-io: ack, removed the local variable and passed the for loop variable as parameter instead.

jemoreira (approved):

PR #113, opened 2022-03-31
Don't require the deviceId query parameter in host orchestrator

jemoreira:


host/frontend/host-orchestrator/static/js/index.js
```diff
@@ -89,8 +89,7 @@ window.addEventListener('load', e => {
   let selectDeviceCb = deviceId => {
     return new Promise((resolve, reject) => {
       let client = window.open(
-          `/devices/${deviceId}/files/client.html?deviceId=${deviceId}`,
-          deviceId);
+          `/devices/${deviceId}/files/client.html`, deviceId);
```
ser-io, line 92: can this line be moved up as it's shorter?

host/frontend/host-orchestrator/intercept/js/server_connector.js
```diff
@@ -29,8 +29,16 @@
...
-  const urlParams = new URLSearchParams(window.location.search);
-  return urlParams.get('deviceId');
+  // The server connector is loaded from the client.html file, which is loaded
+  // with a path like "/devices/{deviceId}/files/"
+  let path = location.pathname;
+  let beginIdx = path.indexOf("/devices/") + "/devices/".length;
+  let endIdx = path.indexOf("/files/");
+  if (beginIdx < 0 || endIdx < 0) {
+    // The path doesn't match our expectations
+    throw "server connector is incompatible with this server";
```
Databean, line 39: Can this throw an [Error instance](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) rather than a string?

host/frontend/host-orchestrator/intercept/js/server_connector.js
```diff
@@ -29,8 +29,16 @@
...
 // In this case the device id is passed as a parameter in the url.
 export function deviceId() {
-  const urlParams = new URLSearchParams(window.location.search);
-  return urlParams.get('deviceId');
+  // The server connector is loaded from the client.html file, which is loaded
+  // with a path like "/devices/{deviceId}/files/"
+  let path = location.pathname;
+  let beginIdx = path.indexOf("/devices/") + "/devices/".length;
+  let endIdx = path.indexOf("/files/");
+  if (beginIdx < 0 || endIdx < 0) {
```
Databean, line 37: What do you think of splitting `path` on `/` characters and then searching through the resulting array instead?

Databean (approved):

host/frontend/host-orchestrator/static/js/index.js
jemoreira: ran clang-format on the line

ser-io (approved):

host/frontend/host-orchestrator/intercept/js/server_connector.js
jemoreira: Done

host/frontend/host-orchestrator/static/js/index.js
jemoreira: Ran cllang-format

host/frontend/host-orchestrator/intercept/js/server_connector.js
jemoreira: Doing that would break if the device id has `/` characters on it. I don't think we'll allow those in valid device ids, but still it's nice not to have a bug in JS in case we do.
Performance wise I don't think there will be a significant difference: Splitting the entire string by `/` does more work than necessary, but what I did searches for a string from the beginning twice, which is more costly than searching for "devices" first and then for "files". I guess I could change the second search to not start at 0.
Finally, the code would look like this, I'm not sure it's clearer:
```
let pathElements = location.pathname.split('/');
let devIdx = pathElements.indexOf("devices");
let filesIdx = devIdx + 2;
if (devIdx < 0 || filesIdx >= pathElements.length || pathElements[filesIdx] != "files") {
    throw...
}
return pathElements[devIdx + 1];
```
changed the search to not start at 0, let me know if you had other reasons in mind when suggesting this.

jemoreira: discussed offline and decided to switch to splitting

PR #114, opened 2022-04-07
Enables passing the tls cert directory through environment variables.

ser-io:


host/frontend/host-orchestrator/operator.go
```diff
@@ -45,6 +45,7 @@ func startHttpServer() {
 }
 
 func startHttpsServer() {
+	tlsCertDir := fromEnvOrDefault("ORCHESTRATOR_TLS_CERT_DIR", defaultTLSCertDir)
```
jemoreira, line 48: please add a commented out example for this configuration to debian/cuttlefish-common.default

jemoreira (changes requested):

host/frontend/host-orchestrator/operator.go
ser-io: done

jemoreira (approved):

PR #126, opened 2022-05-14
Adds operations management logic.

ser-io:


host/frontend/host-orchestrator/operation.go
```diff
@@ -0,0 +1,102 @@
...
+
+func NewMapOM(uuidFactory func() string) *MapOM {
+	return &MapOM{
+		uuidFactory: uuidFactory,
+		mutex:       sync.RWMutex{},
+		operations:  make(map[string]OperationData),
+	}
+}
+
+func (m *MapOM) NewOperation() (OperationData, error) {
```
jemoreira, line 68: Creating an operation should never fail, there are infinite valid strings for the operation name and the only reason this fails is because the generated one is already chosen, if that's the case you should generate a new one until an unused one appears.

This would break the test where you use a different factory that always returns the same name, but that's easily fixable: Don't use that factory and get the name from the returned operation, which is guaranteed to have a unique name.

Given that you don't need the factory to ensure a specific name in the test and that you need to guarantee different names for different operations, you should get rid of the factory altogether and use UUID directly from this code.

host/frontend/host-orchestrator/operation.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (m *MapOM) GetOperation(name string) (OperationData, bool) {
+	m.mutex.RLock()
+	defer m.mutex.RUnlock()
+	op, ok := m.operations[name]
+	return op, ok
+}
+
+func (m *MapOM) CompleteOperation(name string, result OperationResultData) {
+	m.mutex.Lock()
```
jemoreira, line 88: This blocks the entire operations manager object (so, all operations) just to change the state of one operation. Synchronization on the operations should be done on the operations object directly and this mutex should only guard the map, not the operations inside (you can avoid this by following my other comment about changing the OperationData struct and the map)

host/frontend/host-orchestrator/operation.go
```diff
@@ -0,0 +1,102 @@
...
+	"sync"
+)
+
+type NewOperationError string
+
+func (s NewOperationError) Error() string {
+	return fmt.Sprintf("failure creating a new operation: %s", string(s))
+}
+
+type OperationData struct {
```
jemoreira, line 29: I suggested elsewhere to add a WaitGroup in this struct before looking at how it was used, some modifications are required to enable that:
- The operation manager needs to work with pointers to operations
- The OperationData structs needs to provide proper synchronization to access its fields since it's going to be shared among several threads
- Since the Operation only changes data from NotDone to Done, you can probably guarantee proper synchronization with just atomics and the correct order of updates (set result first and done later). Although a mutex would be correct too.

jemoreira (changes requested):

host/frontend/host-orchestrator/instancemanager_test.go
```diff
@@ -52,7 +52,7 @@ func TestCreateCVDInvalidRequestsEmptyFields(t *testing.T) {
 
 	for _, test := range tests {
 		req := validRequest()
```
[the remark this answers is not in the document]
ser-io: done, good catch, yes this is possible now.

host/frontend/host-orchestrator/instancemanager_test.go
```diff
@@ -65,6 +65,79 @@ func TestCreateCVDInvalidRequestsEmptyFields(t *testing.T) {
...
+	var newOperationErr NewOperationError
+	if !errors.As(err, &newOperationErr) {
+		t.Errorf("error type <<\"%T\">> not found in error chain", newOperationErr)
+	}
+	im.launchCVDWG.Wait()
+	expectedOp := OperationData{
+		Name: opName,
+		Done: false,
+	}
+	op, _ := om.GetOperation(opName)
```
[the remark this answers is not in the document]
ser-io: the idea here was to make that there was an operation created by the first `CreateCVD`call with the relevant name, even though the second call failed. I changed the test a little bit just to reflect that, rather than verifying the complete operation values.

There are independent tests for each of the `MapOM` methods in `operation_test.go` file.

host/frontend/host-orchestrator/operation.go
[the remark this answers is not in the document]
ser-io: changed the name to `Operation` as we have a new package for communication objects.

ser-io: acknowledged, moved the WaitGroup to the `MapOM` values, as I'd like to keep the Operation related structs to holding data only, I created an `operationEntry` type which is the one will have the WaitGroup and will managed with pointers within the `MapOM` scope only.

host/frontend/host-orchestrator/operation.go
```diff
@@ -0,0 +1,102 @@
...
+	Name   string
+	Done   bool
+	Result OperationResultData
+}
+
+type OperationResultData struct {
+	Error OperationErrorData
+	OK    OperationOKData
```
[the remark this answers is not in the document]
ser-io: removed the OperationOK part is is not needed right now, but it should encapsulate the result of the operation, for example if the operation was to created cvds, it should return the created cvds.

Error is only a string right now, but I think we should add error code canonical to http error codes for alerting and data processing purposes.

host/frontend/host-orchestrator/instancemanager_test.go
```diff
@@ -65,6 +65,79 @@ func TestCreateCVDInvalidRequestsEmptyFields(t *testing.T) {
...
+	if op != expectedOp {
```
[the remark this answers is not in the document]
ser-io: done.

host/frontend/host-orchestrator/instancemanager_test.go
```diff
@@ -65,6 +65,79 @@ func TestCreateCVDInvalidRequestsEmptyFields(t *testing.T) {
...
+		},
+	}
+
+	op, _ := im.CreateCVD(req)
+	if op != returnedOp {
+		t.Errorf("expected <<%+v>>, got %+v", returnedOp, op)
+	}
+	im.launchCVDWG.Wait()
+	op, _ = om.GetOperation(opName)
+	if op != completedOp {
```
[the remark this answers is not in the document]
ser-io: done comparing only the relevant fields. Operation stores the error message as string, which is going to be presented to the client as is, rather than a Golang error object therefore the plain string comparison.

host/frontend/host-orchestrator/operation.go
ser-io: i think being defensive about uiid generation makes the code more robust, why should we trust 100% the uuid generation library, suddenly this library has a bug that returns the same id every time under some special circumstances and we have no protection against it. Having said so, I though of having a retry logic, but still we'll need to test the retry logic so I decided to fail the first time and and have a simpler test as I recognize this could be a very very edge case.

host/frontend/host-orchestrator/operator.go
```diff
@@ -247,15 +253,25 @@ func createDevices(w http.ResponseWriter, r *http.Request, im *InstanceManager)
 	var msg CreateCVDRequest
 	err := json.NewDecoder(r.Body).Decode(&msg)
 	if err != nil {
-		replyJSONErr(w, NewBadRequestError("Malformed JSON in request", err))
+		replyJSONErr(w, NewBadRequestError("malformed JSON in request", err))
```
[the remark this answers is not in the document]
ser-io: reverted this, used a lowercase to be inline with golang error messages, but it should be in a different cl.

host/frontend/host-orchestrator/operation.go
```diff
@@ -0,0 +1,102 @@
...
+type OperationManager interface {
+	NewOperation() (OperationData, error)
+
+	GetOperation(string) (OperationData, bool)
+
+	CompleteOperation(string, OperationResultData)
+}
+
+type MapOM struct {
+	uuidFactory func() string
```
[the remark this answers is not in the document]
ser-io: discussed offline and decided to keep the uuid prefix in this case.

host/frontend/host-orchestrator/operation.go
```diff
@@ -0,0 +1,143 @@
...
+	m.operations[name] = op
+	return op.data
+}
+
+func (m *MapOM) Get(name string) (Operation, bool) {
+	op, ok := m.getOperationEntry(name)
+	if !ok {
+		return Operation{}, false
+	}
+	op.mutex.RLock()
```
Databean, line 108: It's unclear what this lock is protecting. The only statement actually covered is the `return` line and is it possible for that to race with anything else?

It's also a bit strange that the lock is effectively acquired twice in most of these methods, first by the inner call to`getOperationEntry` and then later by doing it itself. Is it possible to say "the mutex must be at least read-locked to enter `getOperationEntry`" and then only lock the mutex a single time for these methods?

host/frontend/host-orchestrator/operation.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+type OperationResultError struct {
+	ErrorMsg string
+}
+
+type OperationManager interface {
+	New() Operation
+
+	Get(string) (Operation, bool)
```
Databean, line 46: Can this return an error instead of a bool? A `nil` error can mean success, and a non-`nil` error is more informative than a boolean.

host/frontend/host-orchestrator/operation.go
ser-io: we have the map mutex and different mutexes per each operation entry.

1. the map mutex is for races between New and (Get, Complete), note that Get and Complete do not race against each other as both are read from the map perspective.
2. the operation mutexes are for races between Get and Complete for the same operation.

the two-layers solution is used to avoid blocking Get when Completing each operation.

Databean: Ah I didn't see that they were distinct mutes, my mistake. I'm still confused about what this particular lock is protecting as I don't see how a return can race with something else.

host/frontend/host-orchestrator/operation.go
ser-io: agree, I think error here makes the interface more extensible.

host/frontend/host-orchestrator/operation.go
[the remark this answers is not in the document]
jemoreira: I thought the conclusion of our offline discussion was to get rid of the factory altogether: The MapOM would always use the uuid library to generate operation names, so it didn't need the factory parameter. The only place where this factory is used without uuid is in a test, where you could simply take the name of the returned operation and assign it to the local variable instead of forcing the value of the local variable to be the name of the operation.

If MapOM receives the factory as a parameter then it doesn't make as much sense to guarantee New would never fail.

host/frontend/host-orchestrator/operation.go
```diff
@@ -20,83 +20,126 @@ import (
...
+	name, retryCount := m.uuidFactory(), 0
+	_, found := m.operations[name]
+	for found && retryCount < mapOMNewUUIDRetryLimit {
+		name, retryCount = m.uuidFactory(), retryCount+1
+		_, found = m.operations[name]
+	}
+	if retryCount == mapOMNewUUIDRetryLimit {
+		panic("uuid retry limit reached")
+	}
+	wg := sync.WaitGroup{}
```
jemoreira, line 88: https://pkg.go.dev/sync#WaitGroup says "A WaitGroup must not be copied after first use". It's not clear what first use means (does Add(1) count as first use?), but it's probably safer to just instantiate it as part of the struct directly and increment it after.

host/frontend/host-orchestrator/operation.go
```diff
@@ -20,83 +20,126 @@ import (
...
 
-func (m *MapOM) CompleteOperation(name string, result OperationResultData) {
-	m.mutex.Lock()
-	defer m.mutex.Unlock()
-	op, ok := m.operations[name]
+func (m *MapOM) Complete(name string, result OperationResult) {
+	op, ok := m.getOperationEntry(name)
 	if !ok {
 		log.Println("complete operation error: attempting to complete an operation which does not exist")
 		return
```
jemoreira, line 119: nit: an error should be returned in this case

host/frontend/host-orchestrator/operation.go
```diff
@@ -20,83 +20,126 @@ import (
...
-	op.Result = result
-	m.operations[name] = op
+	op.mutex.Lock()
+	defer op.mutex.Unlock()
+	op.wait.Done()
+	op.data.Done = true
+	op.data.Result = result
+}
+
+func (m *MapOM) Wait(name string) {
```
jemoreira, line 128: The interface would be more useful if the (new) operation data were to be returned when `Wait()` finishes. It seems to me in almost every case calling this function would be immediately followed with a call to `Get()` otherwise.

host/frontend/host-orchestrator/operation_test.go
```diff
@@ -17,37 +17,32 @@ package main
...
+	}
 
-		op, _ := om.NewOperation()
+	op := om.New()
 
-		if op != expectedOp {
-			t.Errorf("expected <<%+v>>, got %+v", expectedOp, op)
-		}
-	})
+	if op != expectedOp {
```
jemoreira, line 33: comparing the entire struct here has the same problem as elsewhere: If you add fields to the struct the test will start failing.

host/frontend/host-orchestrator/operation_test.go
```diff
@@ -117,48 +116,97 @@ func TestMapOMCompleteOperation(t *testing.T) {
...
+
+		om.Complete(opName, result)
+
+		select {
+		case <-waitDoneCh:
+			return
+		case <-time.After(1 * time.Second):
+			t.Errorf("expected stop waiting as operatioin was completed")
+			return
+		}
```
jemoreira, line 158: most of this code could be encapsulated in a function to avoid repetition with the other test

jemoreira (changes requested):

host/frontend/host-orchestrator/operation_test.go
ser-io: acknowledged, done

host/frontend/host-orchestrator/instancemanager_test.go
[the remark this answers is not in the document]
ser-io: acknowledged, created and used a constant to avoid updating the test when the text changes.

host/frontend/host-orchestrator/instancemanager.go
```diff
@@ -36,27 +36,37 @@ func (s EmptyFieldError) Error() string {
...
-	}()
-	return &apiv1.Operation{}, nil
+		m.om.Complete(op.Name, result)
+		return
+	}
+	m.om.Complete(op.Name, OperationResult{})
 }
 
-func validateRequest(r *apiv1.CreateCVDRequest) error {
+func validateRequest(r apiv1.CreateCVDRequest) error {
```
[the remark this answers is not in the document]
ser-io: good catch, reverted this.

host/frontend/host-orchestrator/operation.go
ser-io: changed the approach of `Named result parameters` but instead used a classic `Lock Read Unlock` approach to improve to have a very clear reading.

host/frontend/host-orchestrator/operation.go
ser-io: great catch, done.

host/frontend/host-orchestrator/operation.go
ser-io: done

host/frontend/host-orchestrator/operation_test.go
ser-io: changed tests TestMapOMGetOperation and TestMapOMCompleteOperation following this suggestion: https://github.com/google/android-cuttlefish/pull/126/commits/7e98479450f2daf884bf4916d1376eaa87d8f7ae

However, in this specific test case, I think it's best to make sure what a brand new operation fully contains, even though it means that we need to update the test in case we decide to add new information when creating a new information.

host/frontend/host-orchestrator/operation.go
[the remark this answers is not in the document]
ser-io: wanted to keep the parameter for testing purposes to better control the generated operation name to test the panic logic. hid the parameter as suggested and kept as a field to override it in the specific unit test that test the panic logic.

host/frontend/host-orchestrator/instancemanager_test.go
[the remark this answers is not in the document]
jemoreira: Even with the constant I don't see any value in checking the string, but at least it shouldn't hurt now.

jemoreira (approved):

host/frontend/host-orchestrator/operation_test.go
ser-io: changed the test checking for the relevant fields this time: the name is not empty, done is false and result is empty.

PR #129, opened 2022-06-02
Set execute permission to fetch_cvd.

ser-io:
- Remove the use of `t.Run` in FetchCVDHandler unit tests as its forcing
  keeping track of already used filenames.

### DIFF
--- a/host/frontend/host-orchestrator/instancemanager_test.go
+++ b/host/frontend/host-orchestrator/instancemanager_test.go
@@ -19,7 +19,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/fs"
 	"io/ioutil"
 	"net/http"
 	"os"
@@ -156,7 +155,7 @@ func TestFetchCVDHandlerDownload0750FileAccessIsSet(t *testing.T) {
 	h.Download("1")
 
 	stats, _ := os.Stat(BuildFetchCVDFileName(dir, "1"))
-	var expected fs.FileMode = 0750
+	var expected os.FileMode = 0750
 	if stats.Mode() != expected {
 		t.Errorf("expected <<%+v>>, got %+v", expected, stats.Mode())
 	}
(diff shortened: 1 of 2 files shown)


Databean (approved):

host/frontend/host-orchestrator/instancemanager_test.go
```diff
@@ -102,53 +102,79 @@ func (d *FakeFetchCVDDownloader) Download(dst io.Writer, buildID string) error {
...
+
+	if err != nil {
+		t.Errorf("epected <<nil>> error, got %#v", err)
+	}
+	content, err := ioutil.ReadFile(BuildFetchCVDFileName(dir, "1"))
+	if err != nil {
+		t.Fatal(err)
+	}
+	actual := string(content)
+	expected := "bar"
```
jemoreira, line 129: please use a variable or constant local to this function for "bar" here and in line 112

host/frontend/host-orchestrator/instancemanager_test.go
```diff
@@ -102,53 +102,79 @@ func (d *FakeFetchCVDDownloader) Download(dst io.Writer, buildID string) error {
...
+	downloader := &FakeFetchCVDDownloader{t, "foo"}
+	h := NewFetchCVDHandler(dir, downloader)
+	expectedErr := errors.New("error")
+	h.osChmod = func(_ string, _ os.FileMode) error {
+		return expectedErr
+	}
+
+	err := h.Download("1")
+
+	if err != expectedErr {
```
jemoreira, line 175: use errors.Is() to check this. That leaves open the door for Download to wrap the error returned by osChmod.

jemoreira (changes requested):

host/frontend/host-orchestrator/instancemanager_test.go
ser-io: done

host/frontend/host-orchestrator/instancemanager_test.go
ser-io: good catch! done

jemoreira (approved):

PR #130, opened 2022-06-02
Split cuttlefish-common

jemoreira:
into cuttlefish-base and cuttlefish-common (the latter depends on the former)
cuttlefish-common will later be split into cuttlefish-user and cuttlefish-orchestration, with cuttlefish-user "providing" cuttlefish-common.


jemoreira:
Issues fixed, ready for review now

PR #131, opened 2022-06-02
Split cuttlefish-common package

jemoreira:
into cuttlefish-base and cuttlefish-common (the latter depends on the former)
cuttlefish-common will later be split into cuttlefish-user and cuttlefish-orchestration, with cuttlefish-user "providing" cuttlefish-common.


.github/workflows/test_build_deb.yaml
```diff
@@ -19,5 +19,7 @@ jobs:
         path: "android-cuttlefish"
     - name: build debian package
       run: cd android-cuttlefish && debuild -i -us -uc -b
-    - name: install debian package
+    - name: install base debian package
```
Databean, line 22: Will all of our instructions have to be updated to be "install `cuttlefish-base...`" and "install `cuttlefish-common...`" rather than just "install `cuttlefish-common...`"?

I believe this was the reason adelva@ suggested using a metapackage in the design, so the internal packages are named something more descriptive but we don't have to change any of the install instructions.

jemoreira:
Renamed the cuttlefish-common service as cuttlefish-host-resources too

debian/cuttlefish-base.cuttlefish-host-resources.default
```diff
@@ -1,4 +1,4 @@
-# defaults for cuttlefish-common
+# defaults for cuttlefish-base
```
ser-io, line 1: nit: should it cuttlefish-host-resources here?

ser-io (approved):

debian/cuttlefish-base.cuttlefish-host-resources.default
jemoreira: Done

PR #133, opened 2022-06-09
Extract common signaling server code into a library

jemoreira:


ser-io (approved):

host/frontend/liboperator/operator/operator.go
```diff
@@ -135,20 +62,32 @@ func setupDeviceEndpoint(pool *DevicePool, config apiv1.InfraConfig, path string
...
+// GET  /infra_config
+// GET  /devices
+// GET  /devices/{deviceId}/files/{path}
+// GET  /polled_connections
+// GET  /polled_connections/{connId}/messages
+// POST /polled_connections/{connId}/:forward
+// The maybeIntercept parameter is a function that accepts the
+// requested device file and returns a path to a file to be returned instead or
+// nil if the request should be allowed to proceed to the device.
+func SetupHttpEnpoints(
```
Databean, line 82: Rather than mutating the input `mux.Router`, what do you think of creating these functions to return their own `mux.Router` instances, and having the caller attach them with [`Router.Handle`](https://pkg.go.dev/github.com/gorilla/mux#Router.Handle)?

I can't find anything in Effective Go against mutating argument values, so I'll fall back to [more general sources](https://softwareengineering.stackexchange.com/questions/322490/what-is-an-output-argument-as-refered-to-in-martins-clean-code).

It looks like the downside of this strategy (as opposed to [subrouters](https://softwareengineering.stackexchange.com/questions/322490/what-is-an-output-argument-as-refered-to-in-martins-clean-code), which would have the same mutating-argument problem as you need to create them from the base `mux.Router`) is that parts of the path might have to be repeated between both routers.

Databean (approved):

host/frontend/liboperator/operator/operator.go
jemoreira: POST to /devices and GET to /devices are going to be implemented in different places (orchestrator binary and liboperator respectively), so, at least one of them will have to accept a partially configured router and add its functionality. Even if we decide to create and return the router on the first function, in the calling function we'll likely want to encapsulate configuring it, so we'll end up with the same pattern: a function that takes an output parameter and modifies it.
That being said, I like your suggestion for a different reason: if liboperator receives the router to configure it should know what NOT to modify/break, meaning the library would need to know about the caller. Going the other way around it means the library can't break anything and responsibility falls on the caller not to break what the library did.

Databean (approved):

PR #135, opened 2022-06-10
Add a cvd script for invoking the commands without sourcing.

pcc:
This is ideal for people who don't use bash, or those
who prefer not to have to source a shell script to use
the cvd_ commands.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -21,9 +21,16 @@ cd android-cuttlefish
 
 Set up your environment:
 
+*Option 1*: Use `setup.sh`:
 ```bash
 source setup.sh # once in each terminal window
 ```
+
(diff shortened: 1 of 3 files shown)


Databean:
Sorry that I missed this before. The `cvd xyz` commands actually have the same name as this other tool [we are developing in AOSP](https://cs.android.com/android/platform/superproject/+/master:device/google/cuttlefish/host/commands/cvd/server_command.cpp;l=53;drc=3a8b5a8815d5f8b72641099b711e0ae159ef2b3f), is there a way we can fix the conflicts on this naming?

Databean:
I've uploaded a change to rename it to `cf` but would be fine with something else too.

https://github.com/google/android-cuttlefish/pull/137

PR #136, opened 2022-06-15
Put orchestration specific feature in its own package

jemoreira:


host/frontend/operator/tags
```diff
@@ -0,0 +1,8 @@
+Mmain.go	main.go	15
```
ser-io, line 1: wondering what's this file for? is it vim tags?

host/frontend/liboperator/operator/operator.go
```diff
@@ -28,11 +28,20 @@ import (
 	"github.com/gorilla/mux"
 )
 
+const (
+	DefaultSocketPath     = "/run/cuttlefish/operator"
```
ser-io, line 32: I think is clearer if the binaries define their own default values. the problem I see here is that the library's users do not need to to use these defaults to use the library and the library does not use these values either so the library is exposing and being responsible of something that's not necessary needed.

host/frontend/operator/tags
jemoreira: yes, removed it.

host/frontend/liboperator/operator/operator.go
jemoreira: done

Databean (approved):

PR #141, opened 2022-06-29
Adds stage that creates the artifacts and homes root directories

ser-io:


host/frontend/host-orchestrator/orchestrator/instancemanager.go
```diff
@@ -168,6 +172,20 @@ func (s *StageStartCVDServer) Run() error {
 	return err
 }
 
+type StageCreateDirIfNotExist struct {
+	Dir string
+}
+
+func (s *StageCreateDirIfNotExist) Run() error {
+	if err := os.Mkdir(s.Dir, 0755); os.IsNotExist(err) {
```
Databean, line 180: The documentation for [`os.IsNotExist`](https://pkg.go.dev/os#IsNotExist) suggests using [`errors.Is(err, fs.ErrNotExist)`](https://pkg.go.dev/errors#example-Is).

host/frontend/host-orchestrator/orchestrator/instancemanager.go
```diff
@@ -168,6 +172,20 @@ func (s *StageStartCVDServer) Run() error {
...
+		return err
+	} else if err == nil {
+		// Mkdir set the permission bits (before umask)
+		return os.Chmod(s.Dir, 0755)
+	}
+	return nil
```
Databean, line 186: This returns a `nil` `error` for errors other than `ErrNotExist`. If `os.Mkdir` returns `ErrPermission` for permission denied, it won't enter the first or second conditional.

Was the first conditional intended to be `!os.IsNotExist(err)`? If the directory does already exist, you may want to attempt to `chown` / `chgrp` / `chmod` the directory to make sure the directory satisfies any requirements.

host/frontend/host-orchestrator/orchestrator/instancemanager_test.go
```diff
@@ -200,31 +200,26 @@ func TestProcedureExecuteInnerStageWithFails(t *testing.T) {
 
 func TestLaunchCVDProcedureBuilder(t *testing.T) {
 	cvdBuildID := "1"
-	cvdBin := "bin/cvd"
+	paths := IMPaths{
```
Databean, line 203: What do you think of making these absolute paths?

Databean (approved):

host/frontend/host-orchestrator/orchestrator/instancemanager.go
ser-io: good observation, created b/238431258 for it as right now we are not controlling the version of `go` used when the image is built and the `fs` package was introduced in go 1.16. 

I faced the same issue before as go 1.15 is getting used when building the image because right now it's tied to the version the package manager has. I created b/236976427 to address this problem.

host/frontend/host-orchestrator/orchestrator/instancemanager.go
ser-io: for errors other than `ErrNotExist` the first condition `os.IsNotExist(err)` will test true reaching the early exit returning the error itself.

the attempt to `chmod` was meant to be done once when the directory was created the first time. But I see your point, if the directory was already there we are not going to set the relevant permission, so modified the code to set the relevant permission even if the code was there.

host/frontend/host-orchestrator/orchestrator/instancemanager_test.go
ser-io: done. definitely better for consistency purposes.

host/frontend/host-orchestrator/orchestrator/instancemanager.go
Databean: > for errors other than `ErrNotExist` the first condition `os.IsNotExist(err)` will test true reaching the early exit returning the error itself.

My reading of [the `os.IsNotExist(err)` documentation](for errors other than ErrNotExist the first condition os.IsNotExist(err) will test true reaching the early exit returning the error itself.) is that the opposite is true:

> IsNotExist returns a boolean indicating whether the error is known to report that a file or directory does not exist. It is satisfied by ErrNotExist as well as some syscall errors.

ser-io: I apologize as I was wrong, got confused with the name. Your reading was the correct one.

Changed the condition to `err != nil && !os.IsExist(err)` and added a unit test showcasing the behavior where Mkdir fails with an error other than os.ErrExist.

Thanks for the review!

PR #142, opened 2022-07-11
Removes build id used to download the cvd binary from the `create cvd` API method.

ser-io:


host/frontend/host-orchestrator/orchestrator/instancemanager.go
```diff
@@ -29,6 +29,9 @@ import (
 	"cuttlefish/liboperator/operator"
 )
 
+// The Build Id used to download the cvd binary from.
+const CVDBinBuildID = "8817521"
```
Databean, line 33: What build target is it pulling the file from? I found [this code](https://github.com/google/android-cuttlefish/blob/107db3d8b6326294ad750e81d104c8e1b8342d9f/host/frontend/host-orchestrator/orchestrator/instancemanager.go#L288) that defaults to `aosp_cf_x86_64_phone-userdebug` but that constant should probably be moved here as well if it's applicable to this build id. Only a build id + build target uniquely identifies a build. [These best practices](https://docs.google.com/document/d/1Xni38mX5Jssj-DOgE4oF9unWj7C9vDEFfaMIlUzCSeE/preview) might be applicable to that function.

Databean (approved):

host/frontend/host-orchestrator/orchestrator/instancemanager.go
ser-io: Acknowledged, now handling the build id and target for the cvd binary together passing the values as parameters as suggested in the [best practice](https://docs.google.com/document/d/1Xni38mX5Jssj-DOgE4oF9unWj7C9vDEFfaMIlUzCSeE/preview) article. I also created made the cvd binary build id and target values config parameters of the host orchestrator. 

Thanks for the quick review!

PR #143, opened 2022-07-12
Creates single instances of first stages when building the launch cvd procedure

ser-io:


host/frontend/host-orchestrator/orchestrator/instancemanager.go
```diff
@@ -127,50 +142,41 @@ func (b *LaunchCVDProcedureBuilder) Build(input interface{}) Procedure {
...
+		b.stageCreateHomesDir,
 	}
 }
 
 type StageDownloadCVD struct {
 	CVDBin     string
 	Build      AndroidBuild
 	Downloader *CVDDownloader
-	Mutex      *sync.Mutex
+	mutex      sync.Mutex
```
jemoreira, line 156: nit: This is dangerous, if you change the enclosing struct to be passed by value instead of by reference then the mutex will get copied everytime instead of shared.

Might be worth adding a call of `go tool vet` at build time to get a warning about this.

jemoreira (approved):

host/frontend/host-orchestrator/orchestrator/instancemanager.go
ser-io: instances of this class are not meant to be passed by value, if so `go tool vet` will comply about it as the mutex is not a pointer. Now, if having the mutex as a pointer `go tool vet` won't comply and the class will be miss-used with no feedback sharing a mutex not meant to be shared which could lead to deadlock.

PR #152, opened 2022-07-28
Adds presubmit action building the debian package with go1.13.

ser-io:


.github/workflows/test_build_deb_go1_13.yaml
```diff
@@ -0,0 +1,24 @@
...
+    - name: checkout repository
+      uses: actions/checkout@a81bbbf8298c0fa03ea29cdc473d45769f953675 # aka v2
+      with:
+        path: "android-cuttlefish"
+    - name: build debian package
+      run: cd android-cuttlefish && debuild -i -us -uc -b
+    - name: install base debian package
+      run: dpkg -i cuttlefish-base_*_*64.deb || apt-get install -f -y
+    - name: install common debian package
+      run: dpkg -i cuttlefish-common_*_*64.deb || apt-get install -f -y
```
jemoreira, line 24: This should be cuttlefish-user or cuttlefish-orchestration. cuttlefish-common currently depends on cuttlefish-user which isn't going to be found by dpkg, I think.

ser-io: done

.github/workflows/test_build_deb_go1_13.yaml
```diff
@@ -0,0 +1,24 @@
...
+    - name: checkout repository
+      uses: actions/checkout@a81bbbf8298c0fa03ea29cdc473d45769f953675 # aka v2
+      with:
+        path: "android-cuttlefish"
+    - name: build debian package
+      run: cd android-cuttlefish && debuild -i -us -uc -b
+    - name: install base debian package
+      run: dpkg -i cuttlefish-base_*_*64.deb || apt-get install -f -y
+    - name: install cuttlefish-orchestration debian package
+      run: dpkg -i cuttlefish-orchestration_*_*64.deb || apt-get install -f -y
```
jemoreira, line 24: using the `-f` option with `apt-get` might be working against us, since according to the manpage it "can omit any packages to permit APT to deduce a likely solution". Can we add another command to check that the package was actually installed correctly? Something like `dpkg -i ... || apt-get ... && dpkg -s <pkg>`, please double check that this indeed fails when installing cuttlefish-common instead of orchestration.

jemoreira (changes requested):

.github/workflows/test_build_deb_go1_13.yaml
ser-io: Done, used `apt-get install -y <package_name>` only. test it fails with  package `-common` here https://github.com/google/android-cuttlefish/runs/7587993314?check_suite_focus=true

jemoreira (approved):

PR #155, opened 2022-08-01
ReverseProxy OpenWrt luci page

ikicha:
Cloud Android uses OpenWrt as an AP, and it has web configuration page which is served in 192.168.96.2.
Before this change, a user should use web browser in cuttlefish instance or a workstation which cloud android is running in, but accessing by IP address is painful for users.

So I think it would be great if we could serve openwrt configuration page inside cuttlefish web ui as iframe like device page. For this, I set up reverse proxy between 192.168.96.2:80 and /openwrt

It doesn't change a kind of access level, because it is already accessible in cuttlefish instance.

Test: try to connect /openwrt and check if luci page shows

### DIFF
--- a/host/frontend/liboperator/operator/operator.go
+++ b/host/frontend/liboperator/operator/operator.go
@@ -30,6 +30,10 @@ import (
 	"github.com/gorilla/mux"
 )
 
+const (
+	DefaultOpenWrtLuciUrl = "http://192.168.96.2"
+)
+
 // Sets up a unix socket for devices to connect to and returns a function that listens on the
 // socket until an error occurrs.
 func SetupDeviceEndpoint(pool *DevicePool, config apiv1.InfraConfig, path string) func() error {
@@ -99,18 +103,17 @@ func CreateHttpHandlers(
 	router.HandleFunc("/infra_config", func(w http.ResponseWriter, r *http.Request) {
 		ReplyJSONOK(w, config)
 	}).Methods("GET")
-	remoteUrl, proxy, err := getOpenWrtProxy()
+	remoteUrl, proxy, err := getOpenWrtProxy(DefaultOpenWrtLuciUrl)
 	if err == nil {
-		router.HandleFunc("/openwrt", forwardWithoutUrl(remoteUrl, proxy))
+		router.HandleFunc("/openwrt", getForwarder(remoteUrl, proxy, false /* urlPreserved */))
 		// We cannot set up the base url for luci, so just add necessary paths as well.
-		router.HandleFunc("/cgi-bin/{rest:.*}", forwardWithPreservedUrl(remoteUrl, proxy))
-		router.HandleFunc("/luci-static/{rest:.*}", forwardWithPreservedUrl(remoteUrl, proxy))
+		router.PathPrefix("/cgi-bin").HandlerFunc(getForwarder(remoteUrl, proxy, true /* urlPreserved */))
+		router.PathPrefix("/luci-static").HandlerFunc(getForwarder(remoteUrl, proxy, true /* urlPreserved */))
 	}
 	return router
 }
 
-func getOpenWrtProxy() (*url.URL, *httputil.ReverseProxy, error) {
-	luciUrl := "http://192.168.96.2"
+func getOpenWrtProxy(luciUrl string) (*url.URL, *httputil.ReverseProxy, error) {
 	remoteUrl, err := url.Parse(luciUrl)
 	if err != nil {
 		log.Fatal(err)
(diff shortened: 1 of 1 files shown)


host/frontend/liboperator/operator/operator.go
```diff
@@ -97,9 +99,45 @@ func CreateHttpHandlers(
 	router.HandleFunc("/infra_config", func(w http.ResponseWriter, r *http.Request) {
 		ReplyJSONOK(w, config)
 	}).Methods("GET")
+	remoteUrl, proxy, err := getOpenWrtProxy()
+	if err == nil {
+		router.HandleFunc("/openwrt", forwardWithoutUrl(remoteUrl, proxy))
+		// We cannot set up the base url for luci, so just add necessary paths as well.
+		router.HandleFunc("/cgi-bin/{rest:.*}", forwardWithPreservedUrl(remoteUrl, proxy))
+		router.HandleFunc("/luci-static/{rest:.*}", forwardWithPreservedUrl(remoteUrl, proxy))
```
jemoreira, line 107: Looks like you're not using the `rest` variable anywhere so the matcher for this route could simply be `"/luci-static/"` I think.
The same applies for cgi-bin.

host/frontend/liboperator/operator/operator.go
```diff
@@ -97,9 +99,45 @@ func CreateHttpHandlers(
...
+	return remoteUrl, httputil.NewSingleHostReverseProxy(remoteUrl), nil
+}
+
+func forwardWithoutUrl(ur *url.URL, p *httputil.ReverseProxy) func(http.ResponseWriter, *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		r.URL.Host = ur.Host
+		r.URL.Scheme = ur.Scheme
+		r.Host = ur.Host
+		r.URL.Path = ""
+		p.ServeHTTP(w, r)
```
jemoreira, line 128: nit: you could avoid code duplication if you make the `PreservedUrl` part a `bool` parameter. Something like this:
```
func forward(ur *url.URL, p *httputil.ReverseProxy, resetUrlPath bool) func(http.ResponseWriter, *http.Request) {
	return func(w http.ResponseWriter, r *http.Request) {
		r.URL.Host = ur.Host
		r.URL.Scheme = ur.Scheme
		r.Host = ur.Host
                if resetUrlPath {
                        r.URL.Path = ""
                }
		p.ServeHTTP(w, r)
	}
}
```

host/frontend/liboperator/operator/operator.go
```diff
@@ -97,9 +99,45 @@ func CreateHttpHandlers(
...
+		router.HandleFunc("/openwrt", forwardWithoutUrl(remoteUrl, proxy))
+		// We cannot set up the base url for luci, so just add necessary paths as well.
+		router.HandleFunc("/cgi-bin/{rest:.*}", forwardWithPreservedUrl(remoteUrl, proxy))
+		router.HandleFunc("/luci-static/{rest:.*}", forwardWithPreservedUrl(remoteUrl, proxy))
+	}
 	return router
 }
 
+func getOpenWrtProxy() (*url.URL, *httputil.ReverseProxy, error) {
+	luciUrl := "http://192.168.96.2"
```
jemoreira, line 113: The ip address should be in a constant or, even better, an argument to the function

host/frontend/liboperator/operator/operator.go
```diff
@@ -97,9 +99,45 @@ func CreateHttpHandlers(
 	router.HandleFunc("/infra_config", func(w http.ResponseWriter, r *http.Request) {
 		ReplyJSONOK(w, config)
 	}).Methods("GET")
+	remoteUrl, proxy, err := getOpenWrtProxy()
+	if err == nil {
+		router.HandleFunc("/openwrt", forwardWithoutUrl(remoteUrl, proxy))
+		// We cannot set up the base url for luci, so just add necessary paths as well.
+		router.HandleFunc("/cgi-bin/{rest:.*}", forwardWithPreservedUrl(remoteUrl, proxy))
+		router.HandleFunc("/luci-static/{rest:.*}", forwardWithPreservedUrl(remoteUrl, proxy))
+	}
```
jemoreira, line 108: The proxying to operwrt should be parameterized in the library and allow the library users to decide whether to enable it or not.

jemoreira (changes requested):

host/frontend/liboperator/operator/operator.go
```diff
@@ -97,9 +99,45 @@ func CreateHttpHandlers(
...
 	return router
 }
 
+func getOpenWrtProxy() (*url.URL, *httputil.ReverseProxy, error) {
```
Databean, line 112: This looks like it is assuming there is at most one OpenWRT instance per host. What happens if a user runs multiple `launch_cvd` invocations that each create their own OpenWRT router?

host/frontend/liboperator/operator/operator.go
ikicha: yeah, thanks for catching that, I replaced it with PathPrefix

host/frontend/liboperator/operator/operator.go
ikicha: For now, it has limitation of OpenWrt instance count because some paths (such as `/cgi-bin/luci` or `/luci-static`) are duplicated. If we have a need to have multiple OpenWrt instances, OpenWrt instances should have different path config in uHttpd for Luci respectively. (BTW, multiple AP could be supported without multiple OpenWrt instances)

host/frontend/liboperator/operator/operator.go
ikicha: thanks, done!

host/frontend/liboperator/operator/operator.go
ikicha: made it constant, and added an argument as well.

host/frontend/liboperator/operator/operator.go
ikicha: I think it would be exposed regardless of an operator mode. (that page is already accessible in cuttlefish instance)

jemoreira: The current two library users are not the only ones. We may use this library from the orchestration project too where this is not necessary..

jemoreira:
So, your end goal is to be able to show the luci page alongside cuttlefish devices in the multi-device view? Possibly with an entry in the sidebar too?
If that's the case you don't need to modify this project at all:
You should start a companion process to the openwrt instance that registers as another "device" with the host orchestrator. As part of that registration you'll provide a port and a "device" name and the webui will automatically include the entry in the sidebar and the proxying to that port, when the user clicks on the sidebar the http server on the port you provided will receive a request for "/client.html", which you can redirect to "/luci". Every other page that gets requested from the luci page should work if the paths are relative, but even if they are not you can still modify it as you want to make it work.
The only remaining piece is that port. The host orchestrator only accepts a port and it will proxy requests to localhost:port, which is probably not the address you need, so in the same companion process where you register with the HO you should start your own reverse proxy that sends the requests to http://192.168.96.2:80. You also need that proxy because the webui will try to load "/client.html" first which you need to rewrite as "/luci" before forwarding.

ikicha:
I will re-open the pull request after making openwrt related url be under the its namespace(such as /devices/openwrt-1) instead of root.

PR #156, opened 2022-08-09
Refactors commands execution logic.

ser-io:
- Set environment variables per execution rather than polluting the
  server's binary environment.


jemoreira (approved):
This did a bit more than just change when the environment is set. It would have been nice to separate the refactoring of the structs holding commands and the environment stuff into different commits

PR #157, opened 2022-08-13
Launch cvd stage

ser-io:


host/frontend/host-orchestrator/orchestrator/instancemanager.go
```diff
@@ -270,6 +290,40 @@ func (s *StageFetchCVD) Run() error {
 	return err
 }
 
+const (
+	daemonArg                    = "--daemon"
+	reportAnonymousUsageStatsArg = "--report_anonymous_usage_stats=y"
```
jemoreira, line 295: Reporting should be configurable, at least leave a TODO and file a bug internally to do it later

jemoreira (approved):

host/frontend/host-orchestrator/orchestrator/instancemanager.go
ser-io: acknowledged, added the TODO and filed the bug.

PR #160, opened 2022-08-17
Adds endpoint to retrieve the host orchestrator service logs

ser-io:


jemoreira:
So, the host orchestrator is supposed to retrieve its own logs? That's a bit counterintuitive isn't it?
Are you planning to make this available from the cloud orchestrator too? or only from inside the GCE network?

ser-io:
yes, it should be reachable from the cloud orchestrator 

The context here is that with the current GCE network config it's not straightforward to ssh into these machines, having an accessible http endpoint will make things way easier when debugging issues.

jemoreira:
it's not straightforward for users, but it should be for admins (us).
Debugging the host orchestrator is not something we should ask the users to do, it's our job.
The issues I see with this are:
* the _cutf-operator user has access to ALL system logs now, don't know how sensitive those are, but it gives it access to information it doesn't need.
* If we suspect issues with the host orchestrator, depending on the host orchestrator to get those logs is not necessarily a good idea
I don't completely oppose this, I just don't think it provides enough value for the risks. We should, at some point, send those logs to a central location that makes it more available than syslog, such as GCP logs for example.

jemoreira (approved):

ser-io:
Agree, I was also concerned about giving _cutf-operator access to ALL the system logs, let's cancel this.

PR #161, opened 2022-08-17
Moves host orchestrator API objects to the host-orchestrator directory

ser-io:


jemoreira:
We shouldn't merge this:

We've been having discussions whether to allow creating CVDs in a user workstation or not. The current plan, subject to change, is to allow it while making it only listen on localhost and ignore request from other hosts (for security). That's why those API endpoints are in liboperator, to support that use case when implemented.

PR #168, opened 2022-08-29
Add `/devices/{deviceId}` API & WebUI service method for getting device info

k311093:
To determine initial view size at Web UI, it needs to get display info of a device.

Add API for getting specific device's info, and also add API client at DeviceService of web UI.

### DIFF
--- a/host/frontend/liboperator/operator/operator.go
+++ b/host/frontend/liboperator/operator/operator.go
@@ -65,6 +65,7 @@ func SetupDeviceEndpoint(pool *DevicePool, config apiv1.InfraConfig, path string
 // Creates a router with handlers for the following endpoints:
 // GET  /infra_config
 // GET  /devices
+// GET  /devices/{deviceId}
 // GET  /devices/{deviceId}/files/{path}
 // GET  /polled_connections
 // GET  /polled_connections/{connId}/messages
@@ -84,6 +85,9 @@ func CreateHttpHandlers(
 	}).Methods("GET")
 	router.HandleFunc("/devices", func(w http.ResponseWriter, r *http.Request) {
 		listDevices(w, r, pool)
+	}).Methods("GET")
+	router.HandleFunc("/devices/{deviceId}", func (w http.ResponseWriter, r *http.Request) {
+		deviceInfo(w, r, pool);
 	}).Methods("GET")
 	router.HandleFunc("/polled_connections/{connId}/:forward", func(w http.ResponseWriter, r *http.Request) {
 		forward(w, r, polledSet)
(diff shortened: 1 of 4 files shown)


jemoreira (commented):
LGTM but I'd like to get Sergio's opinion on this before merging since he's been working on this API lately

host/frontend/liboperator/operator/operator.go
```diff
@@ -173,6 +177,19 @@ func listDevices(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
...
+
+func deviceInfo(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
+	vars := mux.Vars(r)
+	devId := vars["deviceId"]
+	dev := pool.GetDevice(devId)
+	if dev == nil {
+		http.NotFound(w, r)
+		return
+	}
+	ReplyJSONOK(w, dev.info)
```
ser-io, line 190: can we define and use a device info object to be used here, if so let's add it to [host/frontend/liboperator/api/v1/messages.go](https://github.com/google/android-cuttlefish/blob/83f2bb32176995f3aaec76b42a8a7bd6c68938de/host/frontend/liboperator/api/v1/messages.go), which is the central place for clients to use as reference and documentation when trying to communicate with the api.

jemoreira: The device info object is provided by the device and consumed by the client. It's meant to be a black box to the signaling server.
This particular API is being consumed by the signaling server UI, so that's an argument to make it an object with a known structure here, but if we do that we'll lose any properties provided by the device's JSON that are not defined in the struct. New properties could be added at any point by the devices, in fact different devices are allowed to have different info types.
With the above in mind, I think we should still return the entire object as it's done in this PR, but we could validate that it can be parsed as a particular type in messages.go. That type would specify the attributes that any device should include in their info and that the signaling server depends on (i.e the display descriptions)

ser-io: I see, let's add documentation explaining this behavior. As  a final client, where should I look into to see the expected properties of the object returned here.

jemoreira: Sergio and I discussed this over chat, our conclusion is: the `/devices/{deviceId}` endpoint should not return a "black box" (the device info) from the orchestration API point of view. Instead it should return an object (with a well defined type in `messages.go`) that includes some useful device information (probably just the device name for now) and the device info. Something like this:
```
GET /devices/{deviceId}
{
    id: string
    registration_info: interface{}
}
```
This way, if we need to add more information from the orchestration api we have a place to do so.

k311093: I uploaded the patch reflecting comments. Thanks!

ser-io (approved):

jemoreira (approved):

PR #169, opened 2022-09-01
Refactor directory structure

jemoreira:
The debian packages are split into the base and fronted sources and moved into subdirectories with those names.
The docker utilities are moved to the docker subdirectory.


jemoreira:
This shouldn't be submitted until the create_base_image.go script is fixed in AOSP, but it's ready for review at least

frontend/build-webui.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+. ./setup-nodejs-env.sh
```
ser-io, line 2: typo?

ser-io (approved):

frontend/build-webui.sh
jemoreira: I didn't touch this line, it looks like it's sourcing the setup-nodejs-env.sh file instead of executing it. I don't think it's a typo.

kwstephenkim:
There was a bug related to the directory location; however, that is not related with these set of changes. Will fix it soon once these changes land.

kwstephenkim (approved):

PR #174, opened 2022-09-08
Debian packaging boilerplate for cvdserver-bootstrapper.

ser-io:


frontend/src/host-orchestrator/cmd/cvdserver-bootstrapper/go.mod
```diff
@@ -0,0 +1,3 @@
+module cvdserver-bootstrapper
```
jemoreira, line 1: Please move `cvdserver-bootstrapper` to the `frontend/src` directory.

It shouldn't be under host-orchestrator since it's a different binary. The `cmd` directory should only exist if we create directories for the other modules, such as 'libs' or 'services', we only have 4 go modules IIRC so don't bother with that just yet.

frontend/debian/cuttlefish-orchestration.cuttlefish-cvdserver-bootstrapper.init
```diff
@@ -0,0 +1,110 @@
+#!/bin/sh
+#
+### BEGIN INIT INFO
+# Provides: cuttlefish-host-orchestrator
```
jemoreira, line 4: This seems wrong, it should probably "Provide" cvdserver-bootstrapper.

frontend/debian/cuttlefish-orchestration.cuttlefish-cvdserver-bootstrapper.init
```diff
@@ -0,0 +1,110 @@
...
+RUN_DIR="/run/cuttlefish-cvdserver-bootstrapper"
+DAEMON="/usr/lib/cuttlefish-common/bin/cvdserver-bootstrapper"
+PIDFILE="${RUN_DIR}"/cvdserver-bootstrapper.pid
+
+start() {
+  mkdir -p "${RUN_DIR}"
+  chown _cvd-executor:cvdnetwork "${RUN_DIR}"
+  chmod 775 "${RUN_DIR}"
+
+  mkdir -p  "${download_dir}"
```
jemoreira, line 49: where is `download_dir` defined?

frontend/debian/cuttlefish-orchestration.cuttlefish-cvdserver-bootstrapper.init
```diff
@@ -0,0 +1,110 @@
...
+    --exec "${DAEMON}"
+}
+
+status() {
+  # Return
+  #   0 if daemon is running
+  #   1 if daemon is dead and pid file exists
+  #   3 if daemon is not running
+  #   4 if daemon status is unknown
+  start-stop-daemon --start --quiet --pidfile "${PIDFILE}" --exec ${DAEMON} --test > /dev/null
```
jemoreira, line 74: `--start` will start the daemon if it's not running. `--status` seems like a more appropriate option for this purpose.

jemoreira: `--test` should also be dropped, it only prints information which is immediately discarded in `/dev/null`

frontend/debian/cuttlefish-orchestration.postinst
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+set -e
+
+case "$1" in
+    (configure)
```
jemoreira, line 6: shouldn't it be `configure)`?
Also, a `switch` with only one case should probably be an `if`.

frontend/src/host-orchestrator/cmd/cvdserver-bootstrapper/main.go
```diff
@@ -0,0 +1,43 @@
+// Copyright 2021 Google LLC
```
jemoreira, line 1: 2022

frontend/src/host-orchestrator/cmd/cvdserver-bootstrapper/main.go
```diff
@@ -0,0 +1,43 @@
...
+package main
+
+import (
+	"fmt"
+	"log"
+	"net/http"
+	"os"
+)
+
+func hello(w http.ResponseWriter, req *http.Request) {
```
jemoreira, line 26: why is an http server needed here?
This should just do the bootstrapping as soon as it starts. If this waits for an http request to come from elsewhere then that elsewhere will need to know when it's a good time to issue that request. For example, if the cloud orchestrator is the one making the request, it needs to know when this service is started and ready to run before issuing the request, which is not immediately after creating the host. Something else to take into account is that this port will need to be open in the firewall as well.

jemoreira (changes requested):

frontend/debian/cuttlefish-orchestration.cuttlefish-cvdserver-bootstrapper.init
ser-io: done

frontend/debian/cuttlefish-orchestration.cuttlefish-cvdserver-bootstrapper.init
ser-io: `--test` was added here so no action is taken, so `--start` won't start the daemon if it's not running. It seems this combination was used for simulating `--status` which was added in version 1.16.1.

started using `--status`

frontend/src/host-orchestrator/cmd/cvdserver-bootstrapper/main.go
ser-io: done

frontend/debian/cuttlefish-orchestration.cuttlefish-cvdserver-bootstrapper.init
ser-io: removed this.

frontend/debian/cuttlefish-orchestration.cuttlefish-cvdserver-bootstrapper.init
jemoreira: oh! is that version available in ubuntu 18.04? Maybe we should just leave it with `--start --test` to avoid breaking anything.

frontend/debian/cuttlefish-orchestration.postinst
ser-io: it seems the open parenthesis is optional, I copied/pasted from other postint script without noticing.

I think is better to keep the `case` statement as it's kind of standard in `postint` scripts making easier the future job of managing new options.

frontend/src/host-orchestrator/cmd/cvdserver-bootstrapper/go.mod
jemoreira: But I don't think the cvdserver-bootstrapper is a companion command to the host-orchestrator. The HO may depend on it, but won't execute it directly.

More importantly, I got a different interpretation of the purpose of the `/cmd` directory from the [golang-standards docs](https://github.com/golang-standards/project-layout). From reading that our directory structure should be
```
src
  |  cmd
  |    |  host-orchestrator
  |    |  host-operator
  |    |  cvdserver-bootstrapper
  |  pkg
  |    |  liboperator
  |  web
  |    |  webui
```
That would be a much more involved refactor that doesn't belong in this PR though.

frontend/src/host-orchestrator/cmd/cvdserver-bootstrapper/main.go
ser-io: the service won't be called by the cloud orchestrator, and the port does not have to be exposed, it only need to be called by the host-orchestrator.

I wanted to have a live service, cause I want the `cvd create` workflow to be aware of the `bootstrapping` stage rather than assume that there's going to be a cvd_server already up and running.

Having a live service managing the cvd_server has the extra benefit of updating the cvd_server itself which is a feature we could support once cvd_server can update itself.

frontend/debian/cuttlefish-orchestration.postinst
jemoreira: SG

frontend/src/host-orchestrator/cmd/cvdserver-bootstrapper/main.go
jemoreira: > the service won't be called by the cloud orchestrator, and the port does not have to be exposed, it only need to be called by the host-orchestrator.

You should probably use unix sockets instead of the network for communication between processes in the same host. You should be able to still use HTTP over unix sockets if you really want that.

> I wanted to have a live service, cause I want the cvd create workflow to be aware of the bootstrapping stage rather than assume that there's going to be a cvd_server already up and running.

But then you're assuming there is a cvdserver-bootstrapper already up and running, I don't think that's any different

> Having a live service managing the cvd_server has the extra benefit of updating the cvd_server itself which is a feature we could support once cvd_server can update itself.

That's actually a nice feature to have. I assume we'd ask to update the cvd server from the same place we ask to bootstrap it (likely the host-orchestrator), so why don't we wait until we decide to add that feature to add the http server and the request from the HO? The `cvd server` is still a work in progress, I'm worried they may change the way it's supposed to be updated and we end up doing this work for nothing.

Something else that I just remembered: The `cvd server` can not be correctly updated if there are instances running, so this may end up being harder than you think.

frontend/src/host-orchestrator/cmd/cvdserver-bootstrapper/go.mod
ser-io: yes, that would be project structure for our Golang project if following those practices. So as you mention, it would be better to have it on other PR and for now follow the current structure. so I moved up `cvdserver-bootstrapper` to `frontend/src`.

frontend/src/host-orchestrator/cmd/cvdserver-bootstrapper/main.go
ser-io: great call, started using unix sockets instead.

> But then you're assuming there is a cvdserver-bootstrapper already up and running, I don't think that's any different

`cvdserver-bootstrapper` is installed together with `host-orchestrator` so it's a safer assumptions.

jemoreira: I think there is an even simpler way to do this:

Put a `cvd` executable in the image that doesn't really contain the cvd executable, but instead contains the script that will download it and run it.

When the HO executes `cvd` the first time it will download the actual `cvd` binary from ci.android.com and execute that instead with whatever parameters HO gave it. If there is an error, it can communicate those back to the HO through stderr and/or the exit code. Further executions of `cvd` from the HO will call the actual `cvd` binary.

Problem 1: cvd server user
Solution: Use the setuid bit on both the bootstrapper and the cvd binary to ensure they always run as _cutf-executor. This is something you needed to do anyways in your http design to work around the case of the cvd server crashing and being restarted by an execution of the `cvd` binary from HO with the wrong user.

Problem 2: Updates
Solution: Don't delete the bootstrapper after downloading cvd, you can keep it in a different safe location and copy it over the cvd binary to force a new download the next time `cvd` is run or you can run the bootstrapper directly from that other location.

ser-io: tried the seuid bit idea, however it has an interesting drawback: the dynamic linker won't load shared libraries for cvd binaries spawned from cvd_server, i.e: cvd_internal_start, resulting in errors like:

```
/var/lib/cuttlefish-common/artifacts/8673413_aosp_cf_x86_64_phone-userdebug/bin/cvd_internal_start: error while loading shared libraries: libext2_blkid-host.so: cannot open shared object file: No such file or directory    
```

this is because the linker detects these binaries are being run in secure mode hence it does ignore environment variables like: LD_LIBRARY_PATH, see [ld.so.8](https://man7.org/linux/man-pages/man8/ld.so.8.html)

jemoreira (approved):

PR #184, opened 2022-09-28
Run cvd commands as _cvd-executor user.

ser-io:


frontend/debian/cuttlefish-orchestration.postinst
```diff
@@ -15,6 +15,8 @@ case "$1" in
         adduser --system --disabled-password --disabled-login --home /var/empty \
                 --no-create-home --quiet --force-badname --ingroup cvdnetwork _cvd-executor
     fi
+    # Allow `_cutf-operator` to run cvd commands as `_cvd-executor` using sudo without password.
+    echo "_cutf-operator ALL=(_cvd-executor) NOPASSWD:ALL" > /etc/sudoers.d/cuttlefish-orchestration
```
jemoreira, line 19: This should probably be an entry in `debian/cuttlefish-orchestration.install`. That way it gets removed on `purge`.

ser-io: good point, done!

jemoreira (approved):

PR #185, opened 2022-09-28
Adds `_cvd-executor` to the `kvm` group.

ser-io:


frontend/debian/cuttlefish-orchestration.postinst
```diff
@@ -14,6 +14,7 @@ case "$1" in
     then
         adduser --system --disabled-password --disabled-login --home /var/empty \
                 --no-create-home --quiet --force-badname --ingroup cvdnetwork _cvd-executor
+        usermod -aG kvm _cvd-executor
```
jemoreira, line 17: Using `adduser`'s `--ingroup` makes cvdnetwork the primary group of the user. You probably don't want this, so it should be in a `usermod` line like kvm.

ser-io: this is a good point, let's doing it in a new PR specific to this concern.

jemoreira (approved):

PR #191, opened 2022-09-30
Supports deadlines in operation's wait method

ser-io:


frontend/src/host_orchestrator/orchestrator/operation.go
```diff
@@ -44,18 +45,19 @@ type OperationResultError struct {
...
+	Wait(name string, dt time.Duration) (Operation, error)
 }
 
 type mapOMOperationEntry struct {
-	data      Operation
-	mutex     sync.RWMutex
-	waitGroup sync.WaitGroup
+	data  Operation
+	mutex sync.RWMutex
+	done  chan struct{}
```
jemoreira, line 60: Using a channel here suggests information is going to be sent through it, but whatever that is it's ignored on line 134. Also, not sure if this applies, but using a channel means the operation can only be waited on as many times as the channel is written into (usually once).
Maybe it's better to use a conditional variable for this (`sync.Cond`)

ser-io: After a channel has been closed (and drained of all sent values, not relevant here), subsequent receive operations proceed immediately, yielding zero values. This capability  of channels is common practice in Golang to create a broadcast mechanism: don't send a value on the channel, just close it.

jemoreira (approved):

PR #193, opened 2022-10-05
[HO] Makes web controller testable.

ser-io:


frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -48,12 +52,13 @@ type IMPaths struct {
 	HomesRootDir     string
 }
 
-type InstanceManager struct {
+// Instance manager implementation based on execution of `cvd` tool commands.
+type CVDToolWrapperIM struct {
```
jemoreira, line 56: nit: I like having the InstanceManager as an interface, but I found the name "CVDToolWrapperIM" a bit confusing when I saw it in the `main` file first. Maybe "CVDCmdInstanceManager" or "CVDToolInstanceManager" are better?
The main thing for me is that by seeing Wrapper in the name I immediately thought of something else. The names I came up with put more emphasis on the fact that it is an InstanceManager and less on the fact it wraps some tool.
It's up to you if you want to change or leave it as it is, I don't have a strong opinion either way.

jemoreira (approved):

frontend/src/host_orchestrator/orchestrator/controller_test.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	controller := Controller{InstanceManager: &testIM{}}
+	controller.AddRoutes(router)
+
+	router.ServeHTTP(rr, req)
+
+	expected := http.StatusOK
+	if rr.Code != expected {
+		t.Errorf("unexpected status code <<%d>>, want: %d", rr.Code, expected)
+	}
```
jemoreira, line 49: This test is almost the same as the one below and you're likely to add similar tests in the future. I suggest you refactor like this:
```
func requestFromTestRouter(t *testing.T, r *http.Request, expected http.StatusCode) {
	rr := httptest.NewRecorder()
	router := buildTestRouter()
	
        router := mux.NewRouter()
        controller := Controller{InstanceManager: &testIM{}}
	controller.AddRoutes(router)

	if rr.Code != expected {
		t.Errorf("unexpected status code <<%d>>, want: %d", rr.Code, expected)
	}
}

func TestCreateCVDSucceeds(t *testing.T) {
	req, err := http.NewRequest("POST", "/cvds", strings.NewReader("{}"))
	if err != nil {
		t.Fatal(err)
	}
        requestFromTestRouter(t, router, req, http.StatusOk)
}

func TestGetOperationSucceeds(t *testing.T) {
	req, err := http.NewRequest("GET", "/operations/"+op.Name, nil)
	if err != nil {
		t.Fatal(err)
	}
        requestFromTestRouter(t, router, req, http.StatusOk)
}
```

frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -24,35 +24,44 @@ import (
...
 )
 
-func SetupInstanceManagement(router *mux.Router, im *InstanceManager, om OperationManager) {
-	router.HandleFunc("/cvds", func(w http.ResponseWriter, r *http.Request) {
-		createDevices(w, r, im)
-	}).Methods("POST")
-	router.HandleFunc("/operations/{name}", func(w http.ResponseWriter, r *http.Request) {
-		getOperation(w, r, om)
-	}).Methods("GET")
+type Controller struct {
```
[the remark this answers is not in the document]
ser-io: done

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: agree, done!

frontend/src/host_orchestrator/orchestrator/controller_test.go
```diff
@@ -0,0 +1,70 @@
...
+		t.Fatal(err)
+	}
+	controller := Controller{OperationManager: om}
+	controller.AddRoutes(router)
+
+	router.ServeHTTP(rr, req)
+
+	expected := http.StatusOK
+	if rr.Code != expected {
+		t.Errorf("unexpected status code <<%d>>, want: %d", rr.Code, expected)
```
jemoreira, line 68: The only value I see in these tests is signaling to the developer changing this that there are components outside the project that depend on this interface and will be broken by the change. This message should explicitly say that. How about `"Unexpected status code <code>, want <other code>. API may no longer be backwards compatible, client code may stop working"`.

jemoreira (approved):

jemoreira (changes requested):
Ignore this comment, have to write something for github to let me change my review

frontend/src/host_orchestrator/orchestrator/controller.go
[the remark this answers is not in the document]
ser-io: reverted back to `Controller`, although used `imController` variable name in main.go

frontend/src/host_orchestrator/orchestrator/controller_test.go
ser-io: with this refactoring you're always using the same Controller instance and each test should create its own controller instance to test its specific behavior.

I agree there's room for refactoring, so I created a helper method to make the requests which do not obscure the relationship between arrangement and assertions.

jemoreira: Actually, each test would create a new controller instance, it would just happen to be created in the same way.

With the proposed refactor each controller instance would have an actual (test) object on each of its attributes, which is arguably better than having `nil` on some, so I think it's a net improvement.

ser-io: the proposed `requestFromTestRouter` helper method obscures how the controller is created which is what we are unit testing, this is a problem for writing new tests that requires a change in the controller, i.e a new field is added and needs to be set, in such cases you'll need to search for, understand and modify `requestFromTestRouter` which is not ideal, this also affects the existing tests.

jemoreira: SG

frontend/src/host_orchestrator/orchestrator/controller_test.go
ser-io: done.

jemoreira (approved):

PR #214, opened 2022-12-06
[orchestrator] Fails fast if host does not support nested virtualization

ser-io:
…lization

### DIFF
--- a/frontend/src/host_orchestrator/orchestrator/instancemanager.go
+++ b/frontend/src/host_orchestrator/orchestrator/instancemanager.go
@@ -609,10 +609,8 @@ type HostValidator struct {
 }
 
 func (v *HostValidator) Validate() error {
-	cmd := v.ExecContext("grep", "-cw", "vmx", "/proc/cpuinfo")
-	err := cmd.Run()
-	if err != nil {
-		return operator.NewInternalError("Nested virtualization is not enabled.", err)
+	if ok, _ := fileExist("/dev/kvm"); !ok {
+		return operator.NewInternalError("Nested virtualization is not enabled.", nil)
 	}
 	return nil
 }
--- a/frontend/src/host_orchestrator/orchestrator/instancemanager.go
+++ b/frontend/src/host_orchestrator/orchestrator/instancemanager.go
@@ -36,6 +36,10 @@ import (
 
 type ExecContext = func(name string, arg ...string) *exec.Cmd
 
+type Validator interface {
+	Validate() error
+}
+
 type InstanceManager interface {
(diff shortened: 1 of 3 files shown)


frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -592,3 +602,17 @@ func buildCvdCommand(execContext ExecContext,
...
 	return execContext("sudo", finalArgs...)
 }
+
+// Validates whether the current host is valid to run CVDs.
+type HostValidator struct {
+	ExecContext ExecContext
+}
+
+func (v *HostValidator) Validate() error {
+	cmd := v.ExecContext("grep", "-cw", "vmx", "/proc/cpuinfo")
```
jemoreira, line 612: I'm worried about the forward compatibility of this check. For example, is this still correct if the host is ARM based instead of intel?
Cuttlefish only requires (and checks for) the availability of kvm. Could we do the same?

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -105,6 +112,9 @@ func (m *CVDToolInstanceManager) CreateCVD(req apiv1.CreateCVDRequest) (apiv1.Op
 	if err := validateRequest(&req); err != nil {
 		return apiv1.Operation{}, operator.NewBadRequestError("invalid CreateCVDRequest", err)
 	}
+	if err := m.hostValidator.Validate(); err != nil {
```
jemoreira, line 115: Is this check only done on `CreateCVD`? Whether the host is valid or not can be determined before a CVD is created. We can fail even faster if we add a new route to the host orchestrator that returns whether the host is valid or not (200-OK = valid, 503 = invalid). This endpoint could be checked as part of the `CreateHost` operation in the cloud orchestrator and fail it if appropriate.

The check should probably still happen on `CreateCVD` too, since it would avoid extra work in cases where the host was not created by the cloud orchestrator.

jemoreira (changes requested):

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: good catch! did as suggested.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: yes, I was thinking on a new endpoint to get the relevant specs (nested virtualization enabled among them) of the host to know whether valid or not. still we need this check on the create cvd logic.

jemoreira (approved):

PR #216, opened 2022-12-07
[orchestrator] List upload directories

ser-io:


frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -183,3 +185,17 @@ func (h *createUploadDirectoryHandler) ServeHTTP(w http.ResponseWriter, r *http.
...
 }
+
+type listUploadDirectoriesHandler struct {
+	m UserArtifactsManager
+}
+
+func (h *listUploadDirectoriesHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	res, err := h.m.ListDirs()
+	if err != nil {
+		log.Printf("error retrieving upload tokens: %+v", err)
```
jemoreira, line 196: This is the only place with logging in this file, is that intentional?

jemoreira (approved):

frontend/src/host_orchestrator/orchestrator/controller.go
ser-io: good point, removed it from here.

PR #218, opened 2022-12-08
[orchestrator] Upload user artifacts

ser-io:


frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -209,3 +213,24 @@ type listUploadDirectoriesHandler struct {
...
+		if err == multipart.ErrMessageTooLarge {
+			return nil, &operator.AppError{
+				StatusCode: http.StatusInsufficientStorage,
+				Err:        err,
+			}
+		}
+		return nil, operator.NewBadRequestError("Invalid multipart form request", err)
+	}
+	defer r.MultipartForm.RemoveAll()
+	return nil, h.m.CreateUpdateArtifact(dir, fheader.Filename, f)
```
jemoreira, line 235: Is it safe to use `fheader.Filename`? For example, is it guaranteed to not be something like "../../../usr/..."?

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -79,3 +84,32 @@ func (m *UserArtifactsManagerImpl) ListDirs() (*apiv1.ListUploadDirectoriesRespo
...
+}
+
+func (m *UserArtifactsManagerImpl) CreateUpdateArtifact(dir, filename string, src io.Reader) error {
+	dir = m.RootDir + "/" + dir
+	if ok, err := fileExist(dir); err != nil {
+		return err
+	} else if !ok {
+		return operator.NewBadRequestError("upload directory %q does not exist", err)
+	}
+	dst, err := ioutil.TempFile("", "cutfArtifact")
```
jemoreira, line 99: why copy to a tmp file and then move to final dest? Wouldn't it be better/faster to just copy directly to final dest?

frontend/src/host_orchestrator/orchestrator/controller.go
ser-io: Specs required that the path information should be stripped and Go complies with it: https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/mime/multipart/multipart.go;l=84;drc=7c7cd56870ba617f964014fa4694e9b61e29cf97

frontend/src/host_orchestrator/orchestrator/userartifacts.go
ser-io: performance wise should be quite similar, used a temp file just to avoid having some corrupted file if copying fails for any reason.

jemoreira: wouldn't it be the same if the rename fails? Rename is basically a copy and delete if done across file systems in linux IIRC (and tmp is commonly an in-memory file system different from the root disk)

ser-io: good point, didn't considered moving across file system, so I created the temp file in the same directory so rename does not result on copy/delete.

continued using the tmp file idea to allow concurrent updates for the same filename.

jemoreira (approved):

PR #219, opened 2022-12-09
[orchestrator] Extract cvd-host_package.tar.gz

ser-io:


jemoreira (approved):
I failed to mention this on the previous PR:

Please use the filename in the pattern passed to `ioutil.TempFile`, it may make debugging easier.

ser-io:
Good point about using the filename as the temp pattern, done.

PR #222, opened 2022-12-10
[orchestrator] Code health change.

ser-io:
* Following with feedback in https://github.com/google/android-cuttlefish/pull/215#discussion_r1042578681

### DIFF
--- a/frontend/src/host_orchestrator/orchestrator/instancemanager.go
+++ b/frontend/src/host_orchestrator/orchestrator/instancemanager.go
@@ -194,16 +194,14 @@ func (m *CVDToolInstanceManager) launchCVD(req apiv1.CreateCVDRequest, op apiv1.
 	result = OperationResult{Value: cvd}
 }
 
-func (m *CVDToolInstanceManager) launchCVD_(
-	req apiv1.CreateCVDRequest,
-	op apiv1.Operation) (*apiv1.CVD, error) {
+func (m *CVDToolInstanceManager) launchCVD_(req apiv1.CreateCVDRequest, op apiv1.Operation) (*apiv1.CVD, error) {
 	if err := m.downloadCVDHandler.Download(); err != nil {
 		return nil, err
 	}
-	if err := createDir(m.paths.ArtifactsRootDir, false); err != nil {
+	if err := createDir(m.paths.ArtifactsRootDir); err != nil {
 		return nil, err
 	}
-	if err := createDir(m.paths.HomesRootDir, false); err != nil {
+	if err := createDir(m.paths.HomesRootDir); err != nil {
 		return nil, err
 	}
 	artifactsDir, err := m.fetchCVDHandler.Fetch(req.CVD.BuildSource.AndroidCIBuild)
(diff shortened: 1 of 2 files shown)


frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -366,11 +367,18 @@ func (h *startCVDHandler) Launch(instanceNumber uint32, artifactsDir, homeDir st
...
-func createDir(dir string, failIfExist bool) error {
-	// TODO(b/238431258) Use `errors.Is(err, fs.ErrExist)` instead of `os.IsExist(err)`
-	// once b/236976427 is addressed.
+func createDirFailIfExist(dir string) error {
+	err := os.Mkdir(dir, 0774)
+	if err != nil {
+		return err
+	}
+	// Mkdir set the permission bits (before umask)
+	return os.Chmod(dir, 0774)
```
jemoreira, line 376: nit: "regardless of `umask`" is more accurate than "before `umask`" since `chmod` doesn't take `umask` into account at all.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -366,11 +367,18 @@ func (h *startCVDHandler) Launch(instanceNumber uint32, artifactsDir, homeDir st
...
+}
+
+func createDir(dir string) error {
```
jemoreira, line 379: nit: this function could be implemented as:
```
err := createDirFailIfExists(dir)
if os.IsExist(err) {
    return nil
} else {
    return err
}
```

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -366,11 +367,18 @@ func (h *startCVDHandler) Launch(instanceNumber uint32, artifactsDir, homeDir st
 	return nil
 }
 
-func createDir(dir string, failIfExist bool) error {
-	// TODO(b/238431258) Use `errors.Is(err, fs.ErrExist)` instead of `os.IsExist(err)`
-	// once b/236976427 is addressed.
+func createDirFailIfExist(dir string) error {
```
jemoreira, line 370: nit: `createNewDir` ?

jemoreira (approved):

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: I see, the comment was referring to the `os.Mkdir` call above which aims to set the same permission bits as well, clarifying the need for the `os.Chmod` call. I rephrased the comment a little bit. WDYT?

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: good call! done

frontend/src/host_orchestrator/orchestrator/instancemanager.go
jemoreira: I didn't realized the comment said "Mkdir sets...". This comment could just be "Set dir permission regardless of umask".

jemoreira (approved):

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: done!

PR #230, opened 2023-01-13
Kwstephenkim allocd work

kwstephenkim:
The main question is if I could parasite in the allocd-port directory while the allocd is ported from AOSP. We are going to port:
  1. allocd
  2. common/utils/result.h
  3. common/libs/fs
  
  At this point, the pull request is only for result.h in cuttlefish of AOSP. Fs is ready to be made a pull request but only if result.h landed first.


allocd-port/build/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Ignore everything in this directory
+*
+# Except this file
+!.gitignore
```
jemoreira, line 4: wouldn't it be easier to put the following in `allocd-port/.gitignore`:
```
build/*
bin/*
```

allocd-port/srcs/testings/result_test.cc
```diff
@@ -0,0 +1,45 @@
+/*
```
jemoreira, line 1: rename directory `testings` as `test`

jemoreira (changes requested):

allocd-port/build/.gitignore
kwstephenkim: The intention is to commit an empty directory. I believe git allows me to commit files or directories that contains the files but not an empty directory. I guess this is a common trick to bypass it. I've tried the suggestion. I think it doesn't allow me to add bin or build directories.

allocd-port/srcs/testings/result_test.cc
kwstephenkim: Done

allocd-port/build/.gitignore
jemoreira: What's the purpose of those empty directories?
My guess is that some script or tool will use them, would it be possible to have the script/tool create them if they don't already exist?

PR #233, opened 2023-01-19
[Web UI] Sort device ID with instance number

0405ysj:
Currently device ID is sorted by default rule for string type.

Newly added function compares instance number instead when having a prefix `cvd-`.

### DIFF
--- a/frontend/src/operator/webui/src/app/device.service.ts
+++ b/frontend/src/operator/webui/src/app/device.service.ts
@@ -18,12 +18,21 @@ export class DeviceService {
     private sanitizer: DomSanitizer
   ) {}
 
+  compareDeviceId(x: string, y: string): number {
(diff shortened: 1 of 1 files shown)


frontend/src/operator/webui/src/app/device.service.ts
```diff
@@ -18,12 +18,21 @@ export class DeviceService {
     private sanitizer: DomSanitizer
   ) {}
 
+  compareDeviceId(x: string, y: string): number {
```
ikicha, line 21: Instead of implementing it, how about using localCompare for natural order sort? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

0405ysj: Sounds good! I replaced it with using localeCompare.

ikicha (approved):

PR #240, opened 2023-01-27
[orchestrator] Boilerplate for exposing system variables.

ser-io:
- Use /varz endpoint to expose variables that helps with debugging tasks.

### DIFF
--- a/frontend/src/host_orchestrator/orchestrator/controller.go
+++ b/frontend/src/host_orchestrator/orchestrator/controller.go
@@ -23,6 +23,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/google/android-cuttlefish/frontend/src/host_orchestrator/orchestrator/debug"
 	apiv1 "github.com/google/android-cuttlefish/frontend/src/liboperator/api/v1"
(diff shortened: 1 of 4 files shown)


frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -57,7 +59,7 @@ func (c *Controller) AddRoutes(router *mux.Router) {
 		httpHandler(&listUploadDirectoriesHandler{c.UserArtifactsManager})).Methods("GET")
 	router.Handle("/userartifacts/{name}",
 		httpHandler(&createUpdateUserArtifactHandler{c.UserArtifactsManager})).Methods("PUT")
-
+	router.Handle("/varz", httpHandler(&getDebugVariablesHandler{c.DebugVariablesManager})).Methods("GET")
```
jemoreira, line 62: can the url be /debug/varz instead?

jemoreira (approved):

frontend/src/host_orchestrator/orchestrator/controller.go
ser-io: done, used `/_debug/` instead.

PR #243, opened 2023-02-01
[orchestrator] Create cvd using latest green build id.

ser-io:


frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -219,9 +222,31 @@ func (m *CVDToolInstanceManager) launchCVD(req apiv1.CreateCVDRequest, op apiv1.
 	result = OperationResult{Value: cvd}
 }
 
+const (
+	defaultBranch = "aosp-master"
+	defaultTarget = "aosp_cf_x86_64_phone-userdebug"
+)
```
jemoreira, line 228: these defaults should be configurable

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -219,9 +222,31 @@ func (m *CVDToolInstanceManager) launchCVD(req apiv1.CreateCVDRequest, op apiv1.
...
+
 func (m *CVDToolInstanceManager) launchFromAndroidCI(
 	req apiv1.CreateCVDRequest, op apiv1.Operation) (*apiv1.CVD, error) {
-	artifactsDir, err := m.fetchCVDHandler.Fetch(req.CVD.BuildSource.AndroidCIBuild)
+	build := req.CVD.BuildSource.AndroidCIBuild
```
jemoreira, line 232: nit: could this be `req.BuildID()`?

That would decouple the internals of the req object from where it's used.

I'm also thinking about supporting multiple build sources: In that scenario the right property to read from the request is probably not `AndroidCIBuild`. That actually means it's probably a better idea to do `buildAPI.GetBuildIDFromRequest(req)`, which would also make `defaultBranch` and `defaultTarget` be properties of the build API too.

frontend/src/host_orchestrator/orchestrator/buildapi.go
```diff
@@ -0,0 +1,75 @@
...
+	GetLatestGreenBuildID(branch, target string) (string, error)
+}
+
+func NewBuildAPI(baseURL string) BuildAPI {
+	return &buildAPIImpl{
+		BaseURL: baseURL,
+	}
+}
+
+type buildAPIImpl struct {
```
jemoreira, line 35: This should be `AndroidCIBuildAPI`

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -219,9 +222,31 @@ func (m *CVDToolInstanceManager) launchCVD(req apiv1.CreateCVDRequest, op apiv1.
...
+		branch = defaultBranch
+	}
+	target := build.Target
+	if target == "" {
+		target = defaultTarget
+	}
+	buildID := build.BuildID
+	if buildID == "" {
+		var err error
+		buildID, err = m.buildAPI.GetLatestGreenBuildID(branch, target)
```
jemoreira, line 244: Isn't `fetch_cvd` capable of figuring out the last green build from a branch? If that's the case the buildAPI wouldn't be needed at all, right?.

jemoreira: With this in mind, it should probably be `fetchCVDHandler` the one to extract the build id from the request, at least from the `BuildSource` property (see comment on line 232)

jemoreira (changes requested):

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: agree, added a TODO.

frontend/src/host_orchestrator/orchestrator/buildapi.go
ser-io: good point, done

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: I think it's fine using the full struct path vs having shortcut methods for accessing specific properties. 

We already support two different build sources, this method is  the one called when the AndroidCIBuild build source is set.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: fetch_cvd could do that, however fetchCVDHandler has a a logic in-place based on key `build_id/target` to do not download the same artifacts twice, to continue taking advantage of this logic I created this PR, in which we could get the last green `build id` the same way fetch_cvd does beforehand.

jemoreira: can you move the "get last id" logic to fetchCVDHandler then?

ser-io: the build id value is needed in this logic as well.

jemoreira: can fetchCVDHandler return it?

ser-io: I prefer to keep the fetchCVDHandler api as minimal as possible where the passed parameters "build_id/target" correspond to a concrete target leaving the responsibility to determine the `build_id` if none was set to the instance manager itself.

jemoreira: I agree that a wrapper around `fetch_cvd` (i.e `fetchCVDHandler`) should provide the functionality of `fetch_cvd` and nothing more. However I disagree that it's the responsibility of the `InstanceManager` to interact with one particular build API. I think we're missing a layer between those two. That's something we can brainstorm on later since it's beyond the context of this PR.

jemoreira (approved):

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: agree.

PR #245, opened 2023-02-07
Use OpenJDK 17 in the cuttlefish images

sbgoog:


base/debian/changelog
```diff
@@ -1,3 +1,9 @@
+cuttlefish-common (0.9.27) unstable; urgency=medium
+
+  * Use OpenJDK 17
```
jemoreira, line 3: was the actual change done in a previous pull request? Just changing the changelog doens't have any effects.

jemoreira (changes requested):

base/debian/changelog
sbgoog: The actual change is in the image creation script in https://android-review.git.corp.google.com/c/device/google/cuttlefish/+/2422783. It is my understanding that if we want the new image to be used we have to release a new version.

PR #246, opened 2023-02-09
[orchestrator] Adds a ping endpoint.

ser-io:
- Adds a ping endpoint to check wether the orchestrator is running.


frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -61,6 +61,7 @@ func (c *Controller) AddRoutes(router *mux.Router) {
 		httpHandler(&createUpdateUserArtifactHandler{c.UserArtifactsManager})).Methods("PUT")
 	// Debug endpoints.
 	router.Handle("/_debug/varz", httpHandler(&getDebugVariablesHandler{c.DebugVariablesManager})).Methods("GET")
+	router.Handle("/_debug/statusz", okHandler()).Methods("GET")
```
jemoreira, line 64: Any reason why the device list couldn't be used as the ping endpoint?

jemoreira (approved):

frontend/src/host_orchestrator/orchestrator/controller.go
ser-io: it could, however the logic of list devices could return status codes other than 200, from a client implementation POV using list as a ping endpoint would result hacky.

PR #248, opened 2023-02-20
Make sure IPv6 is enabled for bridge interfaces to enable IPv6 for cuttlefish-specific intrefaces inside docker container

dimorinny:
Make sure IPv6 link-local address is generated inside docker container to ensure fastboot will work with oxygen and other docker-centered flows.

Bug: b/270040723


jemoreira (approved):
If this is a docker problem, shouldn't the solution be in the docker scripts?
I don't think there is anything wrong with this change, so I approved it. But, it either makes sense to enable ipv6 for every installation of the package in every environment and the comment is unnecessary or this is docker-specific and therefore the fix should only affect docker usages and not every environment where the service is run.

dimorinny:
You're right. The fact we found this bug on docker doesn't make it docker-specific, so this comment introduced some confusion. Got rid of it.

PR #250, opened 2023-02-27
[orchestrator] Moves download artifact logic to build api.

ser-io:


frontend/src/host_orchestrator/orchestrator/buildapi.go
```diff
@@ -58,18 +66,79 @@ func (s *AndroidCIBuildAPI) GetLatestGreenBuildID(branch, target string) (string
...
+
+func parseErrorResponse(body io.ReadCloser) error {
+	decoder := json.NewDecoder(body)
+	errRes := struct {
+		Error struct {
+			Message string
+			Code    int
+		}
+	}{}
+	err := decoder.Decode(&errRes)
```
jemoreira, line 130: nit: `if err := ...; err != nil {`

jemoreira (approved):

frontend/src/host_orchestrator/orchestrator/buildapi.go
ser-io: done

PR #252, opened 2023-02-28
[orchestrator] Remove unreachable code

ser-io:


jemoreira (approved):
you'd think the compiler could warn that...

ser-io:
yeah, suggestion is to use `go vet` tool to identify this.

PR #253, opened 2023-03-01
[orchestrator] Start devices with speceific kernel build

ser-io:


frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -220,55 +224,74 @@ func (m *CVDToolInstanceManager) launchCVD(req apiv1.CreateCVDRequest, op apiv1.
...
 	}
-	buildID := build.BuildID
-	if buildID == "" {
-		var err error
-		buildID, err = m.buildAPI.GetLatestGreenBuildID(branch, target)
-		if err != nil {
-			return nil, err
+	var mainBuildDir, kernelBuildDir string
+	var mainBuildDwErr, kernelBuildDwErr error
+	downloads := []struct {
```
jemoreira, line 246: nit: I think this slice is making the code harder to read. Consider instead:

```
var wg sync.WaitGroup
wg.Add(1)
go func() {
    mainBuildDir, mainBuildDwErr = m.artifactsMgr.Download(mainBuild.BuildID, mainBuild.Target, cvdBundleType)
    wg.Done()
}
wg.Add(1)
go func() {
    kernelBuildDir, kernelBuildDwErr = m.artifactsMgr.Download(kernelBuild.BuildID, kernelBuild.Target, kernelBundleType)
    wg.Done()
}
wg.Wait()
```

or you could get rid of the waitgroup entirely and have `Download` return a channel instead.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -220,55 +224,74 @@ func (m *CVDToolInstanceManager) launchCVD(req apiv1.CreateCVDRequest, op apiv1.
...
+		{&kernelBuildDir, &kernelBuildDwErr, kernelBuild, kernelBundleType},
+	}
+	var wg sync.WaitGroup
+	for _, d := range downloads {
+		if d.build != nil {
+			wg.Add(1)
+			d := d
+			go func() {
+				defer wg.Done()
+				dir, err := m.artifactsMngr.Download(d.build.BuildID, d.build.Target, d.bundleType)
```
jemoreira, line 262: Is it mandatory to provide a kernel build id to launch from Android CI? I don't see where we enforce it or allow for one not to be downloaded.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -365,44 +416,97 @@ func (h *downloadCVDHandler) download() error {
...
-	cvdBin       string
-	artifactsDir string
-	map_         map[string]*fetchCVDMapEntry
-	mapMutex     sync.Mutex
+type artifactsBundleType int
+
+const (
+	cvdBundleType artifactsBundleType = iota
+	kernelBundleType
+)
```
jemoreira, line 448: this type enum should really be two types/functions that do the respective things and share a bunch of code. For example: `artifactsManager` should have an `androidDownloader` and a `kernelDownloader` that both implement `Download`, which in turn is called from `artifactsManager.DownloadAndroidBuild()` and `artifactsManager.DownloadKernelBuild()` respectively.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -412,25 +516,31 @@ func (h *fetchCVDHandler) Fetch(buildID, target string) (string, error) {
...
+			filename := outDir + "/" + name
+			if err := downloadArtifactToFile(h.buildAPI, filename, name, buildID, target); err != nil {
+				ch <- err
+			}
+			close(ch)
+		}()
+	}
+	for _, ch := range chans {
+		for err := range ch {
+			return err
```
jemoreira, line 540: returning early without reading from all channels has two defects:
1. some errors are lost
2. you'll leave behind a bunch of routines blocked trying to send on the channel

jemoreira (changes requested):

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -220,55 +224,74 @@ func (m *CVDToolInstanceManager) launchCVD(req apiv1.CreateCVDRequest, op apiv1.
...
-		build = *req.CVD.BuildSource.AndroidCIBuildSource.MainBuild
-	}
-	branch := build.Branch
-	if branch == "" {
-		branch = defaultBranch
+	var mainBuild, kernelBuild *apiv1.AndroidCIBuild = &apiv1.AndroidCIBuild{}, nil
+	if req.CVD.BuildSource != nil && req.CVD.BuildSource.AndroidCIBuildSource != nil {
+		buildSource := req.CVD.BuildSource.AndroidCIBuildSource
+		if buildSource.MainBuild != nil {
+			mainBuild = &(*req.CVD.BuildSource.AndroidCIBuildSource.MainBuild)
```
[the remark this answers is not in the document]
ser-io: (*foo) would create a copy, as this could cause confusion, used `*foo = *bar` instead.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -220,55 +224,74 @@ func (m *CVDToolInstanceManager) launchCVD(req apiv1.CreateCVDRequest, op apiv1.
...
-	if req.CVD.BuildSource != nil &&
-		req.CVD.BuildSource.AndroidCIBuildSource != nil &&
-		req.CVD.BuildSource.AndroidCIBuildSource.MainBuild != nil {
-		build = *req.CVD.BuildSource.AndroidCIBuildSource.MainBuild
-	}
-	branch := build.Branch
-	if branch == "" {
-		branch = defaultBranch
+	var mainBuild, kernelBuild *apiv1.AndroidCIBuild = &apiv1.AndroidCIBuild{}, nil
+	if req.CVD.BuildSource != nil && req.CVD.BuildSource.AndroidCIBuildSource != nil {
```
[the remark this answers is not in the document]
ser-io: there's an issue with having non pointers struct fields in API objects, if the field is not set, in the marshaled the json the field will be set marshaling an empty instance.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -220,55 +224,74 @@ func (m *CVDToolInstanceManager) launchCVD(req apiv1.CreateCVDRequest, op apiv1.
...
+		buildSource := req.CVD.BuildSource.AndroidCIBuildSource
+		if buildSource.MainBuild != nil {
+			mainBuild = &(*req.CVD.BuildSource.AndroidCIBuildSource.MainBuild)
+		}
+		kernelBuild = buildSource.KernelBuild
 	}
-	target := build.Target
-	if target == "" {
-		target = defaultTarget
+	if err := updateBuildWithDefaults(mainBuild, m.buildAPI, mainBuildDefaultBranch, mainBuildDefaultTarget); err != nil {
```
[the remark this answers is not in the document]
ser-io: agree, changed the logic following this feedback. PTAL

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: agree, done.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: it's not mandatory, the code gets executed only if the relevant build reference is not nil

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: removed the enum and made some changes following the feedback, however didn't create different downloader structs.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: true, used multierror here to capture all the errors.

jemoreira (approved):

PR #255, opened 2023-03-03
Uses `go vet` for orchestrator code in presubmit

ser-io:


.github/workflows/presubmit.yaml
```diff
@@ -1,10 +1,24 @@
...
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@a81bbbf8298c0fa03ea29cdc473d45769f953675 # aka v2
+    - name: Install dependencies
+      uses: actions/setup-go@v3
+      with:
+        go-version: '1.13.15'
+    - run: go version
+    - name: Vet Test Build
+      run: cd frontend/src/host_orchestrator && go vet ./... && go test ./... && go build ./...
```
jemoreira, line 17: nit: `go doc cmd/vet` says that the `./...` is not required, it uses CWD by default.

It also says that not everything it detects are genuine problems, that could be an issue when added to presubmit checks.

jemoreira (approved):

PR #258, opened 2023-03-09
Upgrade to Debian Bullseye and install rsync

pcc:
The rsync tool can be used to efficiently copy images from an Android build directory into the container by sending only the file differences. rsync needs to be installed on both ends of the transfer. Since super.img and userdata.img are sparse in the build directory it's useful to pass the --sparse --inplace flags so that they end up being sparse in the container as well. Unfortunately there's a bug in older versions of rsync (such as the one in Debian Buster) that causes transfers with these flags to fail. See: https://github.com/WayneD/rsync/issues/53

Buster is already obsolete at this point, so let's upgrade to Bullseye which has a fixed version of rsync and install rsync into the container so that it doesn't need to be installed manually.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,6 @@
 # This file is based on https://hub.docker.com/r/jrei/systemd-debian/.
 
-FROM debian:buster-slim AS cuttlefish-softgpu
+FROM debian:bullseye-slim AS cuttlefish-softgpu
 
 ENV container docker
 ENV LC_ALL C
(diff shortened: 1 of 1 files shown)


pcc:
Unfortunately there's still an rsync bug that tends to affect super.img files when passing those two flags: https://github.com/WayneD/rsync/issues/450

This is still useful though because --sparse on its own isn't affected, so we can still get the disk space benefits of sparse files on the container end.

kwstephenkim:
Looks like we are trying to bump to bookworm.

cjreynol:
Closing older pull requests.  Please re-open or create a new PR if this is still relevant :)

Looks like we are already on Debian "bookworm".

PR #265, opened 2023-03-28
Temporary fix build error with some environments

k311093:
At some environments, it was not able to import server_connector.js and failed to build. So temporary fix such build error.

This is a temporary fix and will fix fundamental problems later.


ikicha (approved):
To unblock internal build, this temporary fix looks good to me.

PR #266, opened 2023-04-15
[orchestrator] Supports `acloud pull`.

ser-io:
- Allows `acloud pull` to work with cvd instances created with the host orchestrator.

### DIFF
--- a/frontend/src/host_orchestrator/orchestrator/instancemanager.go
+++ b/frontend/src/host_orchestrator/orchestrator/instancemanager.go
@@ -223,6 +223,10 @@ func (m *CVDToolInstanceManager) launchCVD(req apiv1.CreateCVDRequest, op apiv1.
 		log.Printf("failed to launch cvd with error: %v", err)
 		return
 	}
+	// TODO: Remove once `acloud CLI` gets deprecated.
+	if cvd.Name == "cvd-1" {
+		runAcloudSetup(m.paths.RuntimesRootDir)
+	}
 	result = OperationResult{Value: cvd}
 }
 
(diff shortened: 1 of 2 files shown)


frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -852,3 +856,18 @@ func downloadArtifactToFile(buildAPI BuildAPI, filename, artifactName, buildID,
 	downloadErr = buildAPI.DownloadArtifact(artifactName, buildID, target, f)
 	return downloadErr
 }
+
+// Acloud only works with the first instance: cvd-1
+func runAcloudSetup(runtimesRootDir string) {
```
jemoreira, line 861: I prefer to not do this in the instance manager or any other component of the host orchestrator. Also, it's basically running a command as another user, so a shell script is a better place to put this.
For example, it's possible to create a symlink to a target that doesn't exists, so, if `runtimesRootDir` is known at boot time, you could add something to the script that starts the host-orchestrator to create the link. That script runs as root IIRC, so you wouldn't need to give `_cutf-operator` permissions to execute as another user.

ser-io: The caveat with this approach is user `vsoc-01` and `/home/vsoc-01` directory might not exist when the package is installed.

jemoreira: the script could create the user, which would guarantee the directory's existence, but that may break the part that creates it today, so it's probably not worth it.

jemoreira (approved):

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: agree.

PR #267, opened 2023-04-18
[orchestrator] Do not run acloud setup in unit tests.

ser-io:


frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -859,6 +860,10 @@ func downloadArtifactToFile(buildAPI BuildAPI, filename, artifactName, buildID,
 
 // Acloud only works with the first instance: cvd-1
 func runAcloudSetup(runtimesRootDir string) {
+	// Do not run this code in unit tests.
```
jemoreira, line 863: why not? this code is meant to run in production, so it should run in tests as well. If you don't want to modify the host you can always stub/mock/fake it.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -859,6 +860,10 @@ func downloadArtifactToFile(buildAPI BuildAPI, filename, artifactName, buildID,
...
+	if flag.Lookup("test.v") != nil {
```
jemoreira, line 864: This sort of global state check to enable/disable a feature is not right. Use a config entry or something similar at the call site instead.

jemoreira (changes requested):

PR #269, opened 2023-04-19
[orchestrator] Fixes get logs files endpoint.

ser-io:
- Gets the instance directory from the execution of `cvd fleet` rather than using a fixed expected value.


frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -132,7 +132,12 @@ func (h *getCVDLogsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
 	name := vars["name"]
 	pathPrefix := "/cvds/" + name + "/logs"
-	logsDir := h.im.GetLogsDir(name)
+	logsDir, err := h.im.GetLogsDir(name)
+	if err != nil {
+		log.Printf("request %q failed with error: %v", r.Method+" "+r.URL.Path, err)
```
jemoreira, line 137: this log line wouldn't be needed if `getCVDLogsHandler` implemented `handler` and were wrapped with `httpHandler`

frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -132,7 +132,12 @@ func (h *getCVDLogsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		replyJSONErr(w, err)
```
jemoreira, line 138: The caller is expecting to receive a file. Is JSON the best format for the reply?

jemoreira (changes requested):

frontend/src/host_orchestrator/orchestrator/controller.go
ser-io: Agree here, this endpoint works with plain html only, putting JSON in failed request is confusing.

jemoreira (approved):

PR #273, opened 2023-04-24
[orchestrator] Remove double default definitions.

ser-io:
- Defaults are already set in frontend/debian/cuttlefish-orchestration.cuttlefish-host_orchestrator.init


frontend/src/host_orchestrator/main.go
```diff
@@ -34,18 +34,8 @@ import (
...
-	DefaultHttpsPort      = "1443"
-	DefaultTLSCertDir     = "/etc/cuttlefish-common/host_orchestrator/cert"
-	DefaultStaticFilesDir = "static"    // relative path
-	DefaultInterceptDir   = "intercept" // relative path
-	DefaultWebUIUrl       = ""
-
-	defaultAndroidBuildURL          = "https://androidbuildinternal.googleapis.com"
-	defaultCVDBinAndroidBuildID     = "9802869"
-	defaultCVDBinAndroidBuildTarget = "aosp_cf_x86_64_phone-userdebug"
-	defaultCVDArtifactsDir          = "/var/lib/cuttlefish-common"
```
jemoreira, line 48: Removing these forces all those env variables to be defined if you run the binary directly, such as when you're modifying the source code.
Can we keep the defaults here and remove the duplication from the other file instead?

ser-io: Agree, done.

frontend/debian/cuttlefish-orchestration.cuttlefish-host_orchestrator.init
```diff
@@ -79,16 +77,16 @@ start() {
...
 
-  ORCHESTRATOR_HTTP_PORT="${orchestrator_http_port}" \
-  ORCHESTRATOR_HTTPS_PORT="${orchestrator_https_port}" \
-  ORCHESTRATOR_TLS_CERT_DIR="${orchestrator_tls_cert_dir}" \
-  ORCHESTRATOR_INSTANCE_MANAGER_ENABLED="${orchestrator_instance_manager_enabled}" \
-  ORCHESTRATOR_ANDROID_BUILD_URL="${orchestrator_android_build_url}" \
-  ORCHESTRATOR_CVDBIN_ANDROID_BUILD_ID="${orchestrator_cvdbin_android_build_id}" \
-  ORCHESTRATOR_CVDBIN_ANDROID_BUILD_TARGET="${orchestrator_cvdbin_android_build_target}" \
-  ORCHESTRATOR_CVD_ARTIFACTS_DIR="${orchestrator_cvd_artifacts_dir}" \
+  set_config ORCHESTRATOR_HTTP_PORT ${orchestrator_http_port}
```
jemoreira, line 80: nit: put quotes around the second parameters.
you're checking that $2 is non-empty, not that it's defined or not, you might as well ensure it's defined

frontend/src/host_orchestrator/main.go
```diff
@@ -68,11 +68,10 @@ func startHttpsServer(port string, certPath string, keyPath string) error {
...
 
 func fromEnvOrDefault(key string, def string) string {
-	val := os.Getenv(key)
-	if val == "" {
-		return def
+	if val, ok := os.LookupEnv(key); ok {
+		return val
 	}
-	return val
+	return def
```
jemoreira, line 74: nice catch!

jemoreira (approved):

frontend/debian/cuttlefish-orchestration.cuttlefish-host_orchestrator.init
ser-io: good point, done. PTAL

jemoreira (approved):

PR #278, opened 2023-05-16
Support reverse proxy to openwrt luci page

0405ysj:


frontend/src/liboperator/operator/operator.go
```diff
@@ -169,6 +174,38 @@ func deviceEndpoint(c *JSONUnix, pool *DevicePool, config apiv1.InfraConfig) {
...
+		openwrtAddr, ok := dev.info.(map[string]interface{})["openwrt_addr"]
+		if !ok {
+			http.Error(w, "Openwrt address not found", http.StatusNotFound)
+			return
+		}
+		url, _ := url.Parse("http://" + openwrtAddr.(string))
+		proxy := httputil.NewSingleHostReverseProxy(url)
+		r.URL.Path = "/devices/" + devId + "/openwrt" + path
+		proxy.ServeHTTP(w, r)
+	} else {
```
jemoreira, line 203: please invert this `if`:
```
if devId != openwrtDevId {
    http.Redirect(...)
    return
}
...
```

frontend/src/liboperator/operator/operator.go
```diff
@@ -169,6 +174,38 @@ func deviceEndpoint(c *JSONUnix, pool *DevicePool, config apiv1.InfraConfig) {
...
+		http.Redirect(w, r, "/devices/" + openwrtDevId.(string) + "/openwrt" + path, http.StatusFound)
```
jemoreira, line 204: if `devId` is not the same as `openwrtDevId` wouldn't `pool.GetDevice(devId)`(line 180) fail to find the device when this redirect is called?

It seems to me the correct approach would be to not redirect and always use `openwrtDevId` instead of `devId` when building the URL to proxy.

0405ysj: Originally openwrt luci interface has some paths like `/cgi-bin/luci`, `/luci-static/*`, and `/ubus`. While accessing to luci interface, it tries to access and redirects to each other. It's not an issue when accessing openwrt IP address itself like `https://192.168.94.2`, but it failes to access some paths via redirection while using reverse proxy. For example, when I access `https://localhost:1443/devices/cvd-1/openwrt/cgi-bin/luci` then luci interfaces tries to access `https://localhost:1443/luci-static` but that path is not for that openwrt luci interface.

So I applied a change to AOSP adding prefix for all paths in luci interface, like modifying `/luci-static` into `/devices/cvd-1/openwrt/luci-static`. However, it still have an issue when running multiple CF instances with only one openwrt instance such as `cvd start -instance_nums=1,2`. In this case, the openwrt luci interface uses the path `/devices/cvd-1/openwrt/*` That's the reason I applied redirection from `/devices/cvd-2/openwrt/*` into `/devices/cvd-1/openwrt/*` because the actual openwrt luci interface runs with the path `/devices/cvd-1/openwrt/*`.

Not using redirection in operator.go still works so I applied your suggestions, but functionally it still redirects to the path based on device id obtaining openwrt instance because of the redirection within luci interface itself.

jemoreira (approved):

frontend/src/liboperator/operator/operator.go
0405ysj: Sure! The related AOSP change is aosp/2585253. When OpenWRT instance is launched with crosvm, the instance name like `cvd-1` is delivered via `--params` argument of crosvm. I'll add you as cc of that change.

frontend/src/liboperator/operator/operator.go
```diff
@@ -169,6 +174,36 @@ func deviceEndpoint(c *JSONUnix, pool *DevicePool, config apiv1.InfraConfig) {
...
+	}
+
+	path := vars["path"]
+	openwrtAddr, ok := devInfo["openwrt_addr"].(string)
+	if !ok {
+		http.Error(w, "Openwrt address not found", http.StatusNotFound)
+		return
+	}
+
+	url, _ := url.Parse("http://" + openwrtAddr)
```
ikicha, line 200: does http works? IIRC, only https works for luci...

0405ysj: Luci with http works as default, with assuming there's no proxy extensions in chrome.

ikicha (approved):

PR #280, opened 2023-05-18
Increase file descriptor limit

jmacnak:
Bug: b/281897851
Test: m create_base_image
Test: create_base_image
      --launch_instance=natsu-test
      --repository_url=https://github.com/jmacnak/android-cuttlefish.git
Test: gcloud compute ssh natsu@natsu-test -- ulimit -n Test: ./device/google/cuttlefish/tools/upload_to_gce_and_run.py
      -instance=natsu-test
Test: cvd start --gpu_mode=gfxstream_guest_angle_host_swiftshader

### DIFF
--- a/base/host/packages/cuttlefish-base/etc/security/limits.d/1_cuttlefish.conf
+++ b/base/host/packages/cuttlefish-base/etc/security/limits.d/1_cuttlefish.conf
@@ -1,0 +1,4 @@
+# Cuttlefish with GPU passthrough will use additional file descriptors on the
+# host for Gralloc buffers.
+@cvdnetwork hard nofile 4096
+@cvdnetwork soft nofile 4096


base/host/packages/cuttlefish-base/etc/security/limits.d/1_cuttlefish.conf
```diff
@@ -0,0 +1,4 @@
+# Cuttlefish with GPU passthrough will use additional file descriptors on the
+# host for Gralloc buffers.
+* hard nofile 4096
```
Databean, line 3: Rather than `*` what do you think of using the group `@cvdnetwork`?

From `/etc/security/limits.conf`:
```
#Each line describes a limit for a user in the form:
#
#<domain>        <type>  <item>  <value>
#
#Where:
#<domain> can be:
#        - a user name
#        - a group name, with @group syntax
#        - the wildcard *, for default entry
#        - the wildcard %, can be also used with %group syntax,
#                 for maxlogin limit
```

Databean (approved):

base/host/packages/cuttlefish-base/etc/security/limits.d/1_cuttlefish.conf
jmacnak: Done!

PR #288, opened 2023-06-01
Discard stderr for `cvd fleet`.

ser-io:
- Discard stderr on successful executions of `cvd fleet` to avoid parsing errors.

### DIFF
--- a/frontend/src/host_orchestrator/orchestrator/instancemanager.go
+++ b/frontend/src/host_orchestrator/orchestrator/instancemanager.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"log"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -690,6 +690,26 @@ func (e *cvdCommandTimeoutErr) Error() string {
 	return fmt.Sprintf("cvd execution with args %q timed out", strings.Join(e.args, " "))
 }
 
+type filterBuffer struct {
```
jemoreira, line 693: nit: Please add a small comment with the purpose of this struct. It's not immediately obvious from the name.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -690,6 +690,26 @@ func (e *cvdCommandTimeoutErr) Error() string {
...
+func newFilterBuffer(sink *bytes.Buffer) *filterBuffer {
+	return &filterBuffer{b: &bytes.Buffer{}, sink: sink}
+}
+
+func (b *filterBuffer) Write(p []byte) (n int, err error) {
+	if n, err := b.b.Write(p); err != nil {
+		return n, err
+	}
+	return b.sink.Write(p)
+}
```
jemoreira, line 707: Could this be achieved using [io#MultiWriter](https://pkg.go.dev/io#MultiWriter) instead?

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -702,18 +722,22 @@ func (c *cvdCommand) Run() error {
 		// Need to call this once the work is done to release resources.
 		defer cancelFn()
 	}
+	combinedOut := &bytes.Buffer{}
+	stdOut := newFilterBuffer(combinedOut)
 	cmd := buildCvdCommand(ctx, c.execContext, c.androidHostOut, c.home, c.cvdBin, c.args...)
-	out, err := cmd.CombinedOutput()
+	cmd.Stdout = stdOut
+	cmd.Stderr = combinedOut
```
jemoreira, line 729: IIUC this sets up stdout to write to two writers: `stdOut` and `combinedOut`, while stderr writes to `combinedOut` only.

I think this introduces a race:

`Writer`s are not routine-safe. `Command` guarantees that writes will be synchronized if `Stdout` and `Stderr` are the same `Writer` when compared with `==`. But by using a `filterBuffer` the precondition no longer holds.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
jemoreira: Actually you could just rename it to `Tee` (see the `tee` command). I don't think `filterBuffer` matches the function of this struct.

jemoreira (changes requested):

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: removed the struct.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: we ended up removing the struct but io.MultiWriter was the way to go.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: good catch!

used two different buffers instead.

jemoreira (approved):

PR #290, opened 2023-06-02
Create multiple homogeneous instances.

ser-io:


frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -227,13 +232,13 @@ func (m *CVDToolInstanceManager) launchCVD(req apiv1.CreateCVDRequest, op apiv1.
 			log.Printf("failed to complete operation with error: %v", err)
 		}
 	}()
```
jemoreira, line 233: This pattern of setting the `result` in multiple places in the function and processing it in a deferred function is error prone. There might be a code path, especially as the code evolves, where the `result` is not set before the function returns.

A better approach is to call another function that does the work and returns the `result` and then process it. That way the compiler will force the initialization of the result or at least make it a conscious decision to return `nil`.

This was not introduced in this PR so it's ok to fix it in another one.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -343,60 +366,50 @@ func (m *CVDToolInstanceManager) launchFromAndroidCI(
...
 	}
-	instanceNumber := atomic.AddUint32(&m.instanceCounter, 1)
-	cvdName := fmt.Sprintf("cvd-%d", instanceNumber)
-	runtimeDir := m.paths.RuntimesRootDir + "/" + cvdName
-	if err := createNewDir(runtimeDir); err != nil {
-		return nil, err
-	}
 	startParams := startCVDParams{
-		InstanceNumber:   instanceNumber,
+		InstanceNumbers:  m.newInstanceNumbers(1 + req.AdditionalInstancesNum),
```
jemoreira, line 369: nit: `+1` is error prone, so it should be done as few times as necessary, yet it is done in two different functions. Given that both `launchFromAndroidCI` and `launchFromUserBuild` only access `req.BuildSource` and `req.AdditionalIntancesNum` it would be better to pass to both functions `numInstances` and `buildSource` and compute `numInstances` only once at the call site.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -864,3 +886,33 @@ func runAcloudSetup(execContext ExecContext, artifactsRootDir, artifactsDir, run
 	go run(execContext(context.TODO(), "sudo", "-u", cvdUser, "chmod", "644", runtimeDir+"/cuttlefish_runtime/logs/metrics.log"))
 	go run(execContext(context.TODO(), "sudo", "-u", cvdUser, "chmod", "644", runtimeDir+"/cuttlefish_runtime/logs/modem_simulator.log"))
 }
+
+func sliceItoa(s []uint32) []string {
+	result := make([]string, len(s))
+	for i := range s {
+		result[i] = strconv.Itoa(i)
+	}
```
jemoreira, line 893: This looks wrong, it should be:
```
for i, v := range s {
    result[i] = strconv.Itoa(v)
}
```

jemoreira: worth a unit test :)

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -864,3 +886,33 @@ func runAcloudSetup(execContext ExecContext, artifactsRootDir, artifactsDir, run
...
+		return 0, fmt.Errorf("failed parsing instance name %q", s)
+	}
+	n, err := strconv.Atoi(parts[1])
+	if err != nil {
+		return 0, fmt.Errorf("failed parsing instance name %q: %w", s, err)
+	}
+	return n, nil
+}
+
+func contains(s []uint32, e uint32) bool {
```
jemoreira, line 910: nit: this could be generic

jemoreira (changes requested):

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: Agree, this is becoming problematic, I'll handle it in a follow up PR.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: agree, done

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: good catch! done, added some unit tests.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: agree, however in host orchestrator we are contrained to `go 1.13`

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -343,60 +363,51 @@ func (m *CVDToolInstanceManager) launchFromAndroidCI(
...
 	if err := m.startCVDHandler.Start(startParams); err != nil {
 		return nil, err
 	}
-	cvd := &apiv1.CVD{}
-	*cvd = *req.CVD
-	cvd.Name = cvdName
 	// TODO: Remove once `acloud CLI` gets deprecated.
-	if cvdName == "cvd-1" {
-		go runAcloudSetup(m.execContext, m.paths.ArtifactsRootDir, mainBuildDir, runtimeDir)
+	if contains(startParams.InstanceNumbers, 1) {
```
jemoreira, line 377: nit: This is common to all `launch...` functions, it could be done in the calling function instead.

jemoreira (approved):

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: good observation, however the acloud script requires the artifact directory which is computed differently in each of the `launch` functions which makes it not trivial to move it to the calling function.

PR #294, opened 2023-06-13
Use SIGINT signal to cancel cvd commands when timing out. 

ser-io:
- Using SIGKILL only terminates the `sudo` parent process leaving the
  file descriptors open. This causes Wait() to never complete as
  goroutines copying data from file descriptors never stop.

### DIFF
--- a/frontend/src/host_orchestrator/orchestrator/instancemanager.go
+++ b/frontend/src/host_orchestrator/orchestrator/instancemanager.go
@@ -29,6 +29,7 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	apiv1 "github.com/google/android-cuttlefish/frontend/src/liboperator/api/v1"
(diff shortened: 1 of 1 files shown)


frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -734,14 +734,19 @@ func (c *cvdCommand) Run() error {
...
+	doneCh := make(chan error)
+	go func() { doneCh <- cmd.Wait() }()
+	select {
+	case err := <-doneCh:
+		logStderr(cmd, stderr.String())
+		if err != nil {
+			return &cvdCommandExecErr{c.args, stderr.String(), err}
+		}
+	case <-ctx.Done():
+		return &cvdCommandTimeoutErr{c.args}
```
jemoreira, line 749: This doesn't stop the underlying process on timeout, potentially leaving background processes running indefinitely. Consider using [CommandContext](https://pkg.go.dev/os/exec#CommandContext) instead.

jemoreira (changes requested):

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: yes, the caller would be the responsible to kill those processes. this is in coming commits. This commit solves the blocking issue only so the client  could get a "time out" response, which right now it does not get.

We are already using CommandContext, introduced here https://github.com/google/android-cuttlefish/commit/e3f20ad5d210ab7902c78c01d80e6e576f44c43e.

jemoreira: 1. The caller shouldn't be responsible of killing processes started by this function.
2. You can return "time out" by checking `context.Err()` after `Wait` returns an error, if it's `context.DeadlineExceeded` then return "timeout", otherwise it failed for some other reason. It can return `context.Cancelled` but there are no code paths here that would cause that.

jemoreira (changes requested):

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: 1. It's the caller that sets the cvd subcommand to be used, for example the caller using `cvd start` is aware that running `cvd start` might hang out leaving processes running after timeout therefore it should clean it up if time out happened. the caller using `cvd fleet` is aware no process will be left running.
2. Wait() or Run() blocks as stderr is passed to the `cvd_server` daemon is not closed. This is the reason we changed to  `select` to listen on `ctx.Done()` channel to avoid getting blocked on either `Run()` or `Wait()`. Here's more context on why it blocks https://github.com/google/android-cuttlefish/blob/main/frontend/src/host_orchestrator/orchestrator/instancemanager.go#L751-L756 even using a CommandContext with timeout.

jemoreira: `Wait` won't block if you add a context with a timeout. When the context is done it will kill the process and `Wait` will return indicating it received signal 6 (abort) on linux. At least that what happened when I tested it with a time of 1 second and the command `sleep 20`.

jemoreira (approved):

PR #298, opened 2023-06-16
Makes cvd runtime artifacts owned by `cvdnetwork` group.

ser-io:


frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -154,6 +157,21 @@ func (m *CVDToolInstanceManager) CreateCVD(req apiv1.CreateCVDRequest) (apiv1.Op
 	return op, nil
 }
 
+// Makes runtime artifacts readable by `cvdnetwork` group.
+func createCuttlefishRuntimeDir(execContext ExecContext, parent string) error {
+	name := filepath.Join(parent, "cuttlefish")
+	if err := createDir(name); err != nil {
+		return err
+	}
+	chmodCmd := execContext(context.TODO(), "chmod", "g+s", name)
```
jemoreira, line 166: According to the title this should be `g+r`

jemoreira (changes requested):

frontend/src/host_orchestrator/orchestrator/instancemanager.go
jemoreira: Also, could this be done with https://pkg.go.dev/os#Chmod instead?

ser-io: agree, changed the title using the term `owned` instead. Use `os.Chmod` as  suggested as well.

jemoreira (approved):

PR #300, opened 2023-06-16
Use a default timeout for cvd command executions.

ser-io:


frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -676,6 +676,8 @@ func fileExist(name string) (bool, error) {
 	}
 }
 
+const CVDCommandDefaultTimeout = 30 * time.Second
```
jemoreira, line 679: 30 seconds is too short for `cvd start`, make it at least 2 minutes.

jemoreira (changes requested):

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: `cvd start` executions won't use this default, it will use 5 minutes, which I wanted to set to 1 minute in a another PR, but it's good to know 2 minutes it's a better number.

jemoreira (approved):

PR #304, opened 2023-06-21
Create fake runtime dir programmatically.

ser-io:


frontend/src/host_orchestrator/orchestrator/cvd/runtimeartifacts_test.go
```diff
@@ -49,21 +47,56 @@ func TestTar(t *testing.T) {
...
+		case dir:
+			if err := os.Mkdir(name, 0755); err != nil {
+				t.Fatal(err)
+			}
+		case file:
+			if err := orchtesting.WriteFile(name, []byte("foo\n"), 0644); err != nil {
+				t.Fatal(err)
+			}
+		case symlink:
+			if err := os.Symlink("foo", name); err != nil {
```
jemoreira, line 97: symlink pointing to "foo" is not realistic, please make it point to where it would in the real runtime dirs

jemoreira (changes requested):

frontend/src/host_orchestrator/orchestrator/cvd/runtimeartifacts_test.go
ser-io: done

jemoreira (approved):

PR #307, opened 2023-06-27
Use `cvd host_bugreport` to build runtime artifacts bundle.

ser-io:


frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -43,12 +43,17 @@ type Validator interface {
 	Validate() error
 }
 
+// TODO(b/289117411): Refactors `cvd executions` out of instance managemente.
```
jemoreira, line 46: typo in "management"

jemoreira (approved):

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: done

PR #308, opened 2023-07-04
Support protobuf and grpc for golang in liboperator

0405ysj:
To communicate with grpc server which would be implemented in Gerrit someday, this repository needs `protoc-gen-go` and `protoc-gen-go-grpc` to use `.proto` files.


frontend/install-protoc.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+sudo apt install protobuf-compiler
```
ikicha, line 2: https://github.com/google/android-cuttlefish/blob/main/frontend/debian/control#L5

Move it into `Build-Depends` in control file.

https://github.com/google/android-cuttlefish/blob/main/docker/Dockerfile#L20

And docker file as well.

ikicha: https://github.com/google/android-cuttlefish/blob/8927f04c23dd7264e90df78125bb110b7d7d88ae/.github/workflows/presubmit.yaml
as well

PR #311, opened 2023-07-12
Stop passing -q flag to docker ps together with a format

pcc:
Newer versions of Docker print the following warning:

WARNING: Ignoring custom format, because both --format and --quiet are set.

### DIFF
--- a/docker/setup.sh
+++ b/docker/setup.sh
@@ -472,7 +472,7 @@ function cf_docker_rm {
         local ip_addr_var_name="ip_${name}"
         unset ${ip_addr_var_name}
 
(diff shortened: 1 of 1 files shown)


luke-ingle:
I had to make this change too for this same reason. It means that if you open a new shell and source setup.sh you won't have access to your container specific commands. e.g. cf_start_<name>, cf_login_<name> etc.

pcc:
@ser-io @jemoreira can you please take a look?

ser-io (approved):

pcc:
(I think this still needs someone to merge the change.)

PR #313, opened 2023-07-14
Fix init script of host orchestrator to directly set env variables

retroinspect:
- `set_config_expr` does not set env vars as in `cuttlefish-orchestration.cuttlefish-host_orchestrator.default` 
- Thus here replaces `eval` & `set_config_expr` to direct setting env vars as `cuttlefish-user` does in  `cuttlefish-user.cuttlefish-operator.init`
- Default values are from comments of `cuttlefish-orchestration.cuttlefish-host_orchestrator.default`

### DIFF
--- a/frontend/debian/cuttlefish-orchestration.cuttlefish-host_orchestrator.init
+++ b/frontend/debian/cuttlefish-orchestration.cuttlefish-host_orchestrator.init
@@ -34,9 +34,14 @@
 if [ -f /etc/default/cuttlefish-host_orchestrator ]; then
     . /etc/default/cuttlefish-host_orchestrator
 fi
-
+orchestrator_http_port=${orchestrator_http_port:-"1080"}
+orchestrator_https_port=${orchestrator_https_port:-"1443"}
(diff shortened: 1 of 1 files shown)


ser-io:
This was intended in https://github.com/google/android-cuttlefish/pull/273/commits/adbca6738f3fb853daf4b49710e434c2d97796f6#diff-3cb0ebf5f62133d0f3155ee84a1ebaa68ef8ec611c8ebaa4cd04283fa0ac3852. 

The idea behind it was to remove duplication of default definitions in https://github.com/google/android-cuttlefish/blob/main/frontend/src/host_orchestrator/main.go and in https://github.com/google/android-cuttlefish/blob/main/frontend/debian/cuttlefish-orchestration.cuttlefish-host_orchestrator.init, deciding to set default values only in the Go binary itself. The Go binary will use the defaults set in `cuttlefish-orchestration.cuttlefish-host_orchestrator.default` if not commented out.

retroinspect:
Uncommenting `cuttlefish-orchestration.cuttlefish-host_orchestrator.default` did not set env as expected.
I will make new PR to address the issue, fixing https://github.com/google/android-cuttlefish/blob/main/frontend/src/host_orchestrator/main.go

retroinspect:
Never mind. Checking again it works as expected 😅

PR #314, opened 2023-07-14
Per-group web UI

retroinspect:
- Add view by group feature to operator & frontend
- `localhost:1443?groupId={groupId}` now shows devices only belong to group of `groupId`

### DIFF
--- a/frontend/src/liboperator/operator/devices.go
+++ b/frontend/src/liboperator/operator/devices.go
@@ -124,18 +124,14 @@ func GetGroupId(info interface{}) string {
 	}
 
 	groupId, ok := deviceInfo["group_id"].(string)
-	if !ok {
-		return DEFAULT_GROUP_ID
-	}
-
-	if len(groupId) == 0 {
+	if !ok || len(groupId) == 0 {
 		return DEFAULT_GROUP_ID
 	}
 
 	return groupId
 }
 
-func GetGroup(p *DevicePool, groupId string) *Group {
+func CreateOrGetGroup(p *DevicePool, groupId string) *Group {
 	_, ok := p.groups[groupId]
 	if !ok {
 		p.groups[groupId] = &Group{}
@@ -155,7 +151,7 @@ func (p *DevicePool) Register(d *Device, id string) bool {
 	p.devices[id] = d
 
 	groupId := GetGroupId(d.info)
-	group := GetGroup(p, groupId)
+	group := CreateOrGetGroup(p, groupId)
 	group.deviceIds = append(group.deviceIds, id)
 
 	return true
@@ -166,7 +162,7 @@ func (p *DevicePool) Unregister(id string) {
 	defer p.devicesMtx.Unlock()
 	if d, ok := p.devices[id]; ok {
 		groupId := GetGroupId(d.info)
-		group := GetGroup(p, groupId)
+		group := CreateOrGetGroup(p, groupId)
 
 		for i, deviceId := range group.deviceIds {
 			if id == deviceId {
--- a/frontend/src/liboperator/operator/devices.go
+++ b/frontend/src/liboperator/operator/devices.go
@@ -24,8 +24,9 @@ import (
 )
 
 type Device struct {
-	info interface{}
-	conn *JSONUnix
+	info    interface{}
+	groupId string
+	conn    *JSONUnix
 	// Reverse proxy to the client files
 	Proxy *httputil.ReverseProxy
 	// Synchronizes access to the client list and the client count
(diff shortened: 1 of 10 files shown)


frontend/src/liboperator/operator/devices.go
```diff
@@ -93,12 +107,41 @@ func (d *Device) ToClient(id int, msg interface{}) error {
...
+}
+
+func GetGroupId(info interface{}) string {
+	deviceInfo, ok := info.(map[string]interface{})
+	if !ok {
+		return DEFAULT_GROUP_ID
+	}
+
+	groupId, ok := deviceInfo["group_id"].(string)
+	if !ok {
```
ikicha, line 127: if !ok || len(groupId) == 0?

frontend/src/liboperator/operator/devices.go
```diff
@@ -93,12 +107,41 @@ func (d *Device) ToClient(id int, msg interface{}) error {
...
+
+	if len(groupId) == 0 {
+		return DEFAULT_GROUP_ID
+	}
+
+	return groupId
+}
+
+func GetGroup(p *DevicePool, groupId string) *Group {
+	_, ok := p.groups[groupId]
```
ikicha, line 139: is it okay to return nil if the groupId is not found? if so, we can just return p.groups[groupId]

frontend/src/liboperator/operator/devices.go
```diff
@@ -93,12 +107,41 @@ func (d *Device) ToClient(id int, msg interface{}) error {
...
+	}
+
+	if len(groupId) == 0 {
+		return DEFAULT_GROUP_ID
+	}
+
+	return groupId
+}
+
+func GetGroup(p *DevicePool, groupId string) *Group {
```
ikicha, line 138: It is actually CreateOrGetGroup, please rename it.

frontend/src/liboperator/operator/devices.go
```diff
@@ -138,3 +211,28 @@ func (p *DevicePool) DeviceIds() []string {
...
+	p.devicesMtx.Lock()
+	defer p.devicesMtx.Unlock()
+	ret := make([]*DeviceInfo, 0)
+	for id, device := range p.devices {
+		ret = append(ret, &DeviceInfo{DeviceId: id, GroupId: device.groupId})
+	}
+	return ret
+}
+
+func (p *DevicePool) GetDeviceInfoListByGroupId(groupId string) []*DeviceInfo {
```
ikicha, line 225: can we add GetDeviceInfoListByPredicate or something, and let GetDeviceInfoList and GetDeviceInfoListByGroupId use it?

frontend/src/operator/webui/src/app/device.service.ts
```diff
@@ -1,37 +1,59 @@
...
 
-  private devices = merge(
-    this.deviceFromServer,
-    this.refreshSubject.pipe(mergeMap(() => this.deviceFromServer))
-  ).pipe(shareReplay(1));
+  private devices = this.groupIdSubject.pipe(
+    combineLatestWith(this.refreshSubject),
+    mergeMap(([groupId, _]) => {
+      if (groupId === null) {
+        return this.allDeviceFromServer;
```
ikicha, line 21: does refresh actually work for non-group scene?

frontend/src/liboperator/operator/devices.go
retroinspect: To do that I think GetDeviceInfoList should be first refactored to refer group map of the pool(p.group) to get all devices:

```
func (p *DevicePool) GetDeviceInfoList() []*DeviceInfo {
	p.devicesMtx.Lock()
	defer p.devicesMtx.Unlock()
	ret := make([]*DeviceInfo, 0)

	for groupId, group := range p.groups {
		for _, deviceId := range group.deviceIds {
			ret = append(ret, &DeviceInfo{DeviceId: deviceId, GroupId: groupId})
		}
	}
	return ret
}
```

The only common part between GetDeviceInfoList and GetDeviceInfoListByGroupId becomes the following snippet,
thus GetDeviceInfoListByPredicate do the following:
```
	for _, deviceId := range group.deviceIds {
		ret = append(ret, &DeviceInfo{DeviceId: deviceId, GroupId: groupId})
	}
```

Is this what you mean by?

ikicha: nvm, I just thought we could get deviceIds from p.devices with predicates, but for group, the current code looks better.

frontend/src/operator/webui/src/app/device.service.ts
retroinspect: I guess "non-group scene" means the view with all devices. 
Refresh works for the case.

ikicha (approved):

PR #317, opened 2023-07-19
Allow to run host orchestrator locally

jemoreira:
by making https and cvd executor user optional.

### DIFF
--- a/frontend/src/host_orchestrator/main.go
+++ b/frontend/src/host_orchestrator/main.go
@@ -56,9 +56,6 @@ func startHttpServer(port string) error {
 }
 
 func startHttpsServer(port string, certPath string, keyPath string) error {
-	if port == "" {
-		return nil
-	}
 	log.Println(fmt.Sprint("Host Orchestrator is listening at https://localhost:", port))
 	return http.ListenAndServeTLS(fmt.Sprint(":", port),
 		certPath,
@@ -108,6 +105,7 @@ func main() {
 	webUIUrlStr := fromEnvOrDefault("ORCHESTRATOR_WEBUI_URL", DefaultWebUIUrl)
 	certPath := filepath.Join(tlsCertDir, "cert.pem")
 	keyPath := filepath.Join(tlsCertDir, "key.pem")
+	cvdUser := fromEnvOrDefault("ORCHESTRATOR_CVD_USER", "")
 
 	pool := operator.NewDevicePool()
 	polledSet := operator.NewPolledSet()
@@ -148,6 +146,7 @@ func main() {
 			return orchestrator.NewAndroidCIBuildAPI(http.DefaultClient, abURL, credentials)
 		},
 		UUIDGen: func() string { return uuid.New().String() },
+		CVDUser: cvdUser,
 	}
 	im := orchestrator.NewCVDToolInstanceManager(&opts)
 	debugStaticVars := debug.StaticVariables{
@@ -185,6 +184,9 @@ func main() {
 		func() error { return operator.SetupDeviceEndpoint(pool, config, socketPath)() },
 		func() error { return startHttpsServer(httpsPort, certPath, keyPath) },
 		func() error { return startHttpServer(httpPort) },
+	}
+	if httpsPort != "" {
+		starters = append(starters, func() error {return startHttpsServer(httpsPort, certPath, keyPath)})
 	}
 	start(starters)
 }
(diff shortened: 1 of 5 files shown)


frontend/src/host_orchestrator/main.go
```diff
@@ -57,6 +56,9 @@ func startHttpServer(port string) error {
 }
 
 func startHttpsServer(port string, certPath string, keyPath string) error {
+	if port == "" {
```
ser-io, line 59: When reading the code I think having a more upfront logic towards starting the https server is easier to follow, and let the function panic if `port` was empty.

```
starters := []func() error{
		func() error { return operator.SetupDeviceEndpoint(pool, config, socketPath)() },
		func() error { return startHttpServer(httpPort) },
}
if httpsPort != "" {
  starters = append(starters, func() error { return startHttpsServer(httpsPort, certPath, keyPath) })
}
```

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -140,6 +143,28 @@ func NewCVDToolInstanceManager(opts *CVDToolInstanceManagerOpts) *CVDToolInstanc
 	}
 }
 
+func newCVDExecContext(execContext ExecContext, user string) CVDExecContext {
+	if user != "" {
+		// Without a user, cvd is executed as the current user.
```
ser-io, line 148: let's add this comment as a general comment for the whole `newCVDExecContext` function.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -813,14 +837,12 @@ func (c *cvdCommand) startCVDServer() error {
 	return cmd.Run()
 }
 
-func buildCvdCommand(ctx context.Context, execContext ExecContext, androidHostOut string, home string, cvdBin string, args ...string) *exec.Cmd {
-	newArgs := []string{"-u", cvdUser, envVarHome + "=" + home}
+func buildCVDEnv(androidHostOut, home string) []string {
```
ser-io, line 840: nit: `cvdEnv` only as function name.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -140,6 +143,28 @@ func NewCVDToolInstanceManager(opts *CVDToolInstanceManagerOpts) *CVDToolInstanc
 	}
 }
 
+func newCVDExecContext(execContext ExecContext, user string) CVDExecContext {
```
ser-io, line 146: let's move `user` as the first parameter and and emphasizes the big difference between passing a value and not in the comments.

frontend/src/host_orchestrator/orchestrator/instancemanager_test.go
```diff
@@ -42,6 +42,7 @@ func (AlwaysSucceedsValidator) Validate() error {
 }
 
 const fakeLatesGreenBuildID = "9551522"
+const cvdUser = "_cvd-executor"
```
ser-io, line 45: `fakeCVDUser` for consistency with other names.

ser-io, line 45: nit: wondering whether having the same production user name here could cause confusion in the future while reading test failure messages.

ser-io (changes requested):

frontend/src/host_orchestrator/orchestrator/instancemanager.go
jemoreira: The execContext parameter is arguably more important since it's being wrapped by the result, while the user is just a customization parameter.

frontend/src/host_orchestrator/main.go
jemoreira: Done

frontend/src/host_orchestrator/orchestrator/instancemanager.go
jemoreira: Done

frontend/src/host_orchestrator/orchestrator/instancemanager.go
jemoreira: Done

frontend/src/host_orchestrator/orchestrator/instancemanager_test.go
jemoreira: Done

jemoreira: Changed

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: true.

ser-io (approved):

PR #327, opened 2023-07-27
Implement cvd env ls

0405ysj:


frontend/src/liboperator/operator/operator.go
```diff
@@ -232,6 +360,46 @@ func grpcCallUnaryMethod(w http.ResponseWriter, r *http.Request, pool *DevicePoo
...
+		http.Error(w, "Device not found", http.StatusNotFound)
+		return
+	}
+
+	request := gopb.TypeInformationRequest{
+		ServiceName: vars["serviceName"],
+		TypeName: vars["typeName"],
+	}
+
+	devInfo := dev.info.(map[string]interface{})
```
ikicha, line 377: I think the code retrieving `conn` can be extracted as a method because it's duplicated across call/list/type

ikicha (changes requested):

frontend/src/liboperator/operator/operator.go
0405ysj: QQ about golang style issue,
I reduced the duplication, but and the part of duplication remains.
Do you think the duplicated code below should be extracted as well or not?

```
	vars := mux.Vars(r)
	devId := vars["deviceId"]
	dev := pool.GetDevice(devId)
	if dev == nil {
		http.Error(w, "Device not found", http.StatusNotFound)
		return
	}
	conn, err := dev.ConnectControlEnvProxyServer()
	if err != nil {
		http.Error(w, err.Error(), http.StatusNotFound)
		return
	}
	defer conn.Close()
```

0405ysj: @ikicha Gentle ping?

ikicha: I think it'd better be extracted as well. For resource close parts, the function might return clean-up function as return value.

0405ysj: Done.

PR #329, opened 2023-07-28
[HO] Code health.

ser-io:
- Breaks down instance manager into smaller independent units catching up with the Single Responsability Principle.


frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -105,36 +116,71 @@ func replyJSON(w http.ResponseWriter, obj interface{}, statusCode int) error {
 	return encoder.Encode(obj)
 }
 
+func createCVD(c Config, om OperationManager, uadr UserArtifactsManager) *createCVDHandler {
```
jemoreira, line 119: `newCreateCVDHandler` is probably better. `createCVD` implies the CVD is created in this function, which it isn't.

jemoreira (approved):

PR #293, opened 2023-06-09
Allow webui to be served on a path other than root

jemoreira:
This allows showing a remote host's web UI through the cloud orchestrator proxying mechanism.

PR #330, opened 2023-07-28
Change iframe URL for the device UI to absolute path

k311093:


ikicha:
I think it causes the bug in #293

k311093:
Thank you for your comment.

I'll find another fix that does not affects #293

PR #331, opened 2023-08-01
Change device iframe URL to absolute path(starts with '/')

k311093:
Using window.location.pathname as prefix to make possible to serve normally under proxy.


ikicha:
@jemoreira , can you test if it meets your requirements?

frontend/src/operator/webui/src/app/safe-device-url.pipe.ts
```diff
@@ -10,7 +10,7 @@ export class SafeDeviceUrlPipe implements PipeTransform {
   transform(deviceId: string) {
     /* eslint-disable */
     // DO NOT FORMAT THIS LINE
-    return this.sanitizer.bypassSecurityTrustResourceUrl(`devices/${deviceId}/files/client.html`);
+    return this.sanitizer.bypassSecurityTrustResourceUrl(`/${window.location.pathname}devices/${deviceId}/files/client.html`);
```
jemoreira, line 13: I expected this to be `${window.location.pathname}/devices...`.
For every page I tested, the `pathname` value starts with a `/`, but not all end with one so I expected to see `/` between `${window.location.pathname}` and `devices`. Some pages do end in `/`, but the browser normally ignores `//` and treats it like a single `/`.

k311093: I tried to use this under proxy suburl. It seems like have a problem with window.location.pathname I'll check it further and will update the PR again.

ikicha (approved):

ikicha (changes requested):
As @k311093 said in offline, we need to find a way to support both internal and external.

PR #339, opened 2023-08-08
[HO] Download same artifacts bundle once.

ser-io:


frontend/src/host_orchestrator/orchestrator/artifacts/artifacts.go
```diff
@@ -115,31 +104,53 @@ func (h *Manager) GetBootloaderBundle(buildID, target string, fetcher Fetcher) (
...
+	OutDir string
+	Error  error
+}
+
+type downloadContext struct {
+	mutex  sync.Mutex
+	result *downloadResult
+}
+
+func (c *downloadContext) Run(f func() (string, error)) (string, error) {
```
jemoreira, line 120: The parameter and return value of `Run` could use `downloadResult` instead of `(string, error)`

frontend/src/host_orchestrator/orchestrator/artifacts/artifacts.go
```diff
@@ -115,31 +104,53 @@ func (h *Manager) GetBootloaderBundle(buildID, target string, fetcher Fetcher) (
...
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	if c.result != nil {
+		return c.result.OutDir, c.result.Error
```
jemoreira, line 124: Returning a previous result would be incorrect if the functor passed by parameter is not the same as the previous one, but the API doesn't make this clear.

A better api would be something like this:
```
func (r *registry) RetrieveOrDownload(dir string, downloadFn func() downloadResult) downloadResult {
    // Acquires r.mutex
    dc := r.get(dir)
    // Acquires dc.mutex
    return dc.RetrieveOrDownload(downloadFn)
}

// Then to use it:
registry.RetrieveOrDownload(outDir, downloadFn)
// instead of
registry.Get(outDir).Run(downloadFn)
```

"RetrieveOrDownload" is the best name I could think of, but it could definitely be improved.

frontend/src/host_orchestrator/orchestrator/artifacts/artifacts.go
```diff
@@ -101,12 +90,12 @@ func (h *Manager) GetKernelBundle(buildID, target string, fetcher Fetcher) (stri
...
 	}
-	return h.syncDownload(buildID+target+"kernel", f)
+	return h.registry.Get(outDir).Run(f)
 }
 
 func (h *Manager) GetBootloaderBundle(buildID, target string, fetcher Fetcher) (string, error) {
+	outDir := fmt.Sprintf("%s/%s_%s__bootloader", h.rootDir, buildID, target)
 	f := func() (string, error) {
-		outDir := fmt.Sprintf("%s/%s_%s__bootloader", h.rootDir, buildID, target)
 		if err := createDir(outDir); err != nil {
```
jemoreira, line 99: nit: I'd prefer `outDir` to be a parameter of the functor instead of a capture.

jemoreira, line 99: nit: Creating the directory is better suited for the registry than the download function.

jemoreira (changes requested):

frontend/src/host_orchestrator/orchestrator/artifacts/artifacts.go
ser-io: `downloadResult` is no longer needed.

frontend/src/host_orchestrator/orchestrator/artifacts/artifacts.go
ser-io: Modified API as suggested, used `GetOrDownload` name.

frontend/src/host_orchestrator/orchestrator/artifacts/artifacts.go
ser-io: done

ser-io: we've been doing it this way because `cvd fetch` needs to create the out directory to set relevant permissions.

jemoreira: I don't understand what `fetch_cvd` has to do with this. My request is just to move the call to `createDir(outDir)` in line 105 to when the `downloadContext` is created around line 151 (probably in a `newDownloadContext` function). Functionality won't change, only the organization of the code will.

I just noticed that all functors but one call `createDir`: Why does GetCVDBundle not call `createDir`?

ser-io: in the GetCVDBundle, the fetcher is the one that creates the directory, which the fetch implementation we use right now, ends up executing `fetch_cvd`. This is an implementation detail that could be handled in a follow up PR as it has been like this. Leaving the responsibility to the fetcher will be the way to go.

jemoreira: we can address in a follow up PR, but I don't like to leave it to the fetcher: When the registry "acquires" a directory it should expect that directory to not exist yet and use the opportunity to assert that and then create it.

jemoreira (approved):

frontend/src/host_orchestrator/orchestrator/artifacts/artifacts.go
ser-io: sounds good.

PR #344, opened 2023-08-14
[HO] Returns the AndroidCIBundle in fetch artifacts action response.

ser-io:
- The returned object will contain the downloaded build id in cases where it was not set in the original request.

### DIFF
--- a/frontend/src/host_orchestrator/orchestrator/fetchartifactsaction.go
+++ b/frontend/src/host_orchestrator/orchestrator/fetchartifactsaction.go
@@ -15,6 +15,7 @@
 package orchestrator
 
 import (
+	"encoding/json"
 	"fmt"
 	"log"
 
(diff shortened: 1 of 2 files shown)


frontend/src/host_orchestrator/orchestrator/fetchartifactsaction.go
```diff
@@ -110,7 +110,7 @@ func (a *FetchArtifactsAction) startDownload(op apiv1.Operation) OperationResult
 	if err != nil {
 		return OperationResult{Error: operator.NewInternalError(errMsgFailedFetchingArtifacts, err)}
 	}
-	return OperationResult{}
+	return OperationResult{Value: a.req.AndroidCIBundle}
```
jemoreira, line 113: nit: IIUC this could be `build` instead of `a.req.AndroidCIBundle`. Which would avoid the need to modify the request.

jemoreira (approved):

frontend/src/host_orchestrator/orchestrator/fetchartifactsaction.go
ser-io: This is a good point, the request shouldn't be modified, I'll make the proper changes.

About returning the broader bundle object, I think it's necessary. I couldn't move forward with the `system image` bundle as it also needs a `main` bundle info in the request, this has not been modeled yet in the API, but it might contain 2 build objects in the request having to return 2 build objects as well.

PR #346, opened 2023-08-18
POST /cvds to support specify group name & instance names

retroinspect:
### Changes
- `POST /cvds` now accept group name and instance names as body content, thus user can specify the names
- If `group_name` or `instance_names` are not provided, the default group name becomes `default` and instance names become `cvd-1`, `cvd-2`, etc.


- The test request below creates 2 cvds in a group "bar"
```
curl -d  '{ 
    "group_name": "bar",                                                                                                                                                                                                                  
    "cvd": {
        "name": "",
        "build_source": {
            "android_ci_build_source": {
                "branch": "aosp-main",
                "build_id": "10678986",
                "target": "aosp_cf_x86_64_phone-trunk_staging-userdebug"
            }
        },
        "status": "a",
        "displays": []
    },
   "instance_names": ["device1", "device2"]
}' -H "Content-Type: application/json" -k -X POST [http://localhost:1080/cvds](http://localhost:1081/cvds)
```

- The logic to generate instanceNumbers to decide instance names were removed as `instanceCounter` does not have to deal with concurrency as `instanceCounter` is a value on `CreateCVDAction` which is created per request.  


### Limitation
- Multiple `POST /cvds` calls are not possible so far
  - Maintaining `instanceCounter` from `instanceManager` instead of `CreateCVDAction` would make the host orchestrator be aware of the first available instance id
- Creating different specs(target, build id, branch) of cvds is not possible so far as cvd cli (`cvd start`) does not support it

### DIFF
--- a/frontend/src/host_orchestrator/orchestrator/createcvdaction.go
+++ b/frontend/src/host_orchestrator/orchestrator/createcvdaction.go
@@ -18,7 +18,6 @@ import (
 	"errors"
 	"fmt"
 	"log"
-	"strconv"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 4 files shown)


retroinspect:
@ikicha @k311093 @jemoreira @ser-io PTAL

retroinspect:
Will be rebased on #350

ser-io:
Let's split the [Use groupName & instanceNames](https://github.com/google/android-cuttlefish/pull/346/commits/007187a34940d7a475951f5a14e486abe3b21008) commit into two commits: one for instance names and one for group names this way we are able to validate/test these new features one a time by checking out into the specific commit.

retroinspect:
Since new POST /cvds API uses cvd load, this PR can be closed :\

PR #350, opened 2023-08-26
[HO] `cvd start` using canonical configs.

ser-io:


frontend/src/liboperator/api/v1/messages.go
```diff
@@ -154,6 +159,23 @@ type CVD struct {
 	Displays []string `json:"displays"`
 }
 
+type Instance struct {
```
jemoreira, line 162: Dont' try to parse the contents of the config file from the orchestrators (yet). Just pass them as is to the `cvd load/start` command.

The reason for this is that what goes in those files is still a bit unstable and this project doesn't really have a need to parse that yet. So treating those files as black boxes will save us a lot of time.

jemoreira (changes requested):

frontend/src/liboperator/api/v1/messages.go
ser-io: I understand. Done.

jemoreira: Looks like you could delete these structures now

ser-io: I thought I did it, good catch. Thanks.

jemoreira (approved):

PR #349, opened 2023-08-25
[HO] Fix aosp branch and target

retroinspect:
- Changed branch name from aosp-master to aosp-main
- Changed target from aosp_cf_x86_64_phone-userdebug to aosp_cf_x86_64_phone-trunk_staging-userdebug

PR #352, opened 2023-08-29
Fix target name according to current build

retroinspect:
- Changed arm target to `aosp_cf_arm64_only_phone-trunk_staging-userdebug`
- Changed x86_64 branch from aosp-main-throttled to aosp-main

This resolves [discussion on #349](https://github.com/google/android-cuttlefish/pull/349#discussion_r1305982993)


retroinspect:
@ikicha PTAL

ikicha (approved):

ikicha:
LGTM, @jemoreira, WDYT?

docker/download-aosp.sh
```diff
@@ -27,12 +27,15 @@ function parse_opts() {
 					ARCH=$2
 					case "${ARCH}" in
 						x86)
+							TARGET='aosp_x86-trunk_staging-userdebug'
```
jemoreira, line 30: Please double check that these are the right targets: The original ones had `cf` and `phone` in them, but these don't.

It's also strange that the target for arm64 is a cf phone, but those for intel are not even cuttlefish.

retroinspect: Thanks. Updated x86-64 target as cf target.
But I noticed that there is no cf target for x86 . 
What should be the proper target for x86?

retroinspect: @jemoreira PTAL

PR #354, opened 2023-08-31
Fix cvd name after `cvd` version update.

ser-io:
- After bumping `cvd` version in commit#497d7f09eb161cb095fb66b8be104f6bd75412de returned cvd names changed from `cvd-1` to `1`, which breaks compatibility with names used in `GET /devices` API.

### DIFF
--- a/frontend/src/host_orchestrator/orchestrator/createcvdaction.go
+++ b/frontend/src/host_orchestrator/orchestrator/createcvdaction.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
-	"strconv"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -156,7 +155,7 @@ func (a *CreateCVDAction) launchWithCanonicalConfig(op apiv1.Operation) (*apiv1.
 	if err := cmd.Run(); err != nil {
 		return nil, operator.NewInternalError(ErrMsgLaunchCVDFailed, err)
 	}
-	fleet, err := cvdFleet(a.execContext, a.paths.CVDBin())
+	fleet, err := getCVDFleet(a.execContext, a.paths.CVDBin())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


frontend/src/host_orchestrator/orchestrator/createcvdaction.go
```diff
@@ -156,7 +155,7 @@ func (a *CreateCVDAction) launchWithCanonicalConfig(op apiv1.Operation) (*apiv1.
 	if err := cmd.Run(); err != nil {
 		return nil, operator.NewInternalError(ErrMsgLaunchCVDFailed, err)
 	}
-	fleet, err := cvdFleet(a.execContext, a.paths.CVDBin())
+	fleet, err := getCVDFleet(a.execContext, a.paths.CVDBin())
```
jemoreira, line 158: `cvdFleet` was better: https://google.github.io/styleguide/go/decisions#getters

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -114,57 +115,73 @@ type cvdInstance struct {
...
+	return group + "-" + name
+}
+
+func decodeCVDName(name string) (string, string, error) {
+	parts := strings.SplitN(name, "-", 2)
+	if len(parts) != 2 {
+		return "", "", errors.New("invalid <GROUP>-<NAME> format")
+	}
+	return parts[0], parts[1], nil
+}
```
jemoreira, line 160: Don't do this, the latest version of `cvd` already prints instance name, group name and webrtc device id separately.

jemoreira (changes requested):

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: I see, I'll abandon this PR in favor of using `webrtc device id`: https://github.com/google/android-cuttlefish/pull/355

PR #356, opened 2023-09-13
Use credentials with canonical configs.

ser-io:


frontend/src/host_orchestrator/orchestrator/createcvdaction.go
```diff
@@ -337,3 +356,57 @@ func validateRequest(r *apiv1.CreateCVDRequest) error {
...
+	}
+	if err := cvd.Exec(ctx, "touch", name); err != nil {
+		return "", err
+	}
+	if err := cvd.Exec(ctx, "chgrp", "cvdnetwork", name); err != nil {
+		return "", err
+	}
+	if err := cvd.Exec(ctx, "chmod", "0620", name); err != nil {
+		return "", err
+	}
```
jemoreira, line 392: Why is this done with three commands instead of
```
os.WriteFile(name, credentials, 0620)
```
?

frontend/src/host_orchestrator/orchestrator/createcvdaction.go
```diff
@@ -337,3 +356,57 @@ func validateRequest(r *apiv1.CreateCVDRequest) error {
...
+
+func createCredsFile(ctx cvd.CVDExecContext) (string, error) {
+	name, err := tempFilename("cvdload*.creds")
+	if err != nil {
+		return "", err
+	}
+	if err := cvd.Exec(ctx, "touch", name); err != nil {
+		return "", err
+	}
+	if err := cvd.Exec(ctx, "chgrp", "cvdnetwork", name); err != nil {
```
jemoreira, line 387: why does `cvdnetwork` need write access to the credentials file?

frontend/src/host_orchestrator/orchestrator/createcvdaction.go
ser-io: Using os.WriteFile(), `_cutf-operator` will be the owner of the file. Created the file this way to have the cvd user: ` _cvd-executor` being the owner instead.

frontend/src/host_orchestrator/orchestrator/createcvdaction.go
ser-io: this way we can write in the credentials from the `host_orchestrator` process run by the `_cutf-operator`.

frontend/src/host_orchestrator/orchestrator/createcvdaction.go
jemoreira: A comment saying just that would be nice here.

frontend/src/host_orchestrator/orchestrator/createcvdaction.go
jemoreira: The HO may or may not run in a single user environment. Allowing any type of access to the `cvdnetwork` group is risky, even it's only for writing.
If possible, create the file with the operator owner, write it and then change the ownership. That would keep access only in the users that should have it.
Or at least remove the group access after writing if the above is not possible.

frontend/src/host_orchestrator/orchestrator/createcvdaction.go
ser-io: very true, done.

frontend/src/host_orchestrator/orchestrator/createcvdaction.go
ser-io: I see, I tried the `change ownership` approach but `chown` resulted in `operation not permitted` in this case. Implemented the temporary write permission approach.

jemoreira (approved):

PR #364, opened 2023-10-13
Fix errors in cuttlefish user package

jemoreira:
- Correct path in link to host orchestrator binary
- Correctly pass flags to command with start-stop-daemon


Mahmod (approved):
LGTM

PR #367, opened 2023-10-19
Fixing docker android 14

luke-ingle:
Android 14 behaves differently to Android 13 in that when cuttlefish runs, it needs to create a new super.img file over the original image. It does this by re-naming super.img.raw to super.img. However in docker if you use the --android parameter, this file is mounted directly rather than mounting a containing directory. This means that the inode of the file would change, and docker does not allow this. The end result is the following output after running cf_start_<name>:

```
assemble_cvd D 10-19 07:46:14   474   474 subprocess.cpp:335] --out
assemble_cvd D 10-19 07:46:14   474   474 subprocess.cpp:335] /tmp/nfrzwE
Failed to load library: libEGL.so
Failed to load library: libEGL.so.1
Failed to load library: libEGL.so
Failed to load library: libEGL.so.1
Failed to load vk
GPU auto mode: did not detect prerequisites for accelerated rendering support, enabling --gpu_mode=guest_swiftshader.
GPU auto mode: did not detect prerequisites for accelerated rendering support, enabling --gpu_mode=guest_swiftshader.
Requested resuming a previous session (the default behavior) but the base images have changed under the overlay, making the overlay incompatible. Wiping the overlay files.
Path for instance UDS: /tmp/cf_avd_1005
/bin/mv: cannot move '/home/vsoc-01/super.img.raw' to '/home/vsoc-01/super.img': Device or resource busy
Unable to replace original sparse image 1
launch_cvd E 10-19 07:46:15   448   448 subprocess.cpp:162] Subprocess 474 was interrupted by a signal: 6
launch_cvd E 10-19 07:46:15   448   448 main.cc:391] assemble_cvd returned -1
```

This commit changes the method of mounting the image files, so that they are copied to the same tmp directory as the cuttlefish files. Then the whole directory is mounted, so that the files inside can be tweaked at will. This allows the running of cuttlefish inside the docker container for android 14.

(NOTE: This is not an issue if you create the container without specifying cuttlefish/android files, and then manually ssh them inside once created. e.g.
```
cf_docker_create cf1
ssh vsoc-01@$(cf_get_ip cf1) -- 'tar xzvf -' < AOSP_IMAGE_FILES.tar.gz
ssh vsoc-01@$(cf_get_ip cf1) -- 'tar xzvf -' < cvd-host_package.tar.gz
```
works fine, but

```
[...]

PR #368, opened 2023-10-19
Removing base_instance_num from cf_start_

luke-ingle:
If you are attempting to do multi-tenancy cuttlefish, the docker script will currently automatically put `--base_instance_num=0` at the start of your start command. 

In the launch_cvd help, it states that this is for labelling the instances, and will begin labelling at this number.

```
-base_instance_num (The instance number of the device created. When
      `-num_instances N` is used, N instance numbers are claimed starting at
      this number.) type: int32 default: 1
```
If we set this value to 0, we are unable to create multi-tenancy cuttlefish instances, regardless of how many instances we pass in with `--num_instances=` parameter. We actually get an error informing us that cuttlefish was unable to interpret 0 as an id.

`run_cvd I 10-19 12:21:01  1549  1549 cuttlefish_config.cpp:55] Failed to interpret "0" as an id, using instance id 1`

This error can be seen in this line of code: https://github.com/luke-ingle/android-cuttlefish/blob/fixing-capability-to-multi-tenancy-cuttlefish/cvd/host/libs/config/cuttlefish_config.cpp#L53 which is doing a check to ensure the instance value is GREATER than 0.

```
  int instance = std::stoi(instance_str);
  if (instance <= 0) {
    LOG(INFO) << "Failed to interpret \"" << instance_str << "\" as an id, "
              << "using instance id " << kDefaultInstance;
    return kDefaultInstance;
  }
```

Doing the change in the PR remedies this problem, and then when you pass in `--num_instances=` parameter, you will create that many instances as expected.

### DIFF
--- a/docker/setup.sh
+++ b/docker/setup.sh
@@ -528,7 +528,7 @@ function __gen_gethome_func_name {
 function __gen_funcs {
 	local name=$1
 	local instance_id=$(cf_get_instance_id ${name})
-	local vcid_opt="--base_instance_num=${instance_id}"
+	local vcid_opt
 	local login_func
 	local start_func
 	local stop_func
--- a/docker/setup.sh
+++ b/docker/setup.sh
@@ -52,7 +52,7 @@ function cf_allocate_instance_id {
 	done
 	local sorted;
 	IFS=$'\n' sorted=($(sort -n <<<"${ids[*]}")); unset IFS
-	local prev=0
+	local prev=1
 	for id in ${sorted[@]}; do
 		if [[ "${prev}" -lt "${id}" ]]; then
 			break;
@@ -349,7 +349,7 @@ function cf_docker_create {
 	    echo "Container ${name} does not exist.";
 
         local cf_instance=$(cf_allocate_instance_id)
-        if [ "${cf_instance}" -gt 7 ]; then
+        if [ "${cf_instance}" -gt 8 ]; then
                 echo "Limit is maximum 8 Cuttlefish instances."
                 return
         fi


luke-ingle:
Hi @kwstephenkim

Have you had a chance to review this change? It turns out this is needed if you are using cuttlefish from tip, because of the issues mentioned in the merged PR  https://github.com/google/android-cuttlefish/pull/367.

If you do not make this change and you use cuttlefish from tip, rather than a branch, the cuttlefish instance fails to load with an error. (An error the same as the one in my initial comment)

If there's any more info I can give, please let me know :)

kwstephenkim:
Thank you! 

I haven't had time to thoroughly read it yet. I guess, however, the instance_id (given to the Cuttlefish via launch_cvd arguments) is not allowed to be 0. 

Once upon a time, there was no `vcid_opt` or `--vsock_cid`. At the time, the instance ID should be unique across the docker containers. Thus, I think we used to allocated each from [0, 7]. (Please see the docker run lines). 

I think, the line you are removing might have to add +1 to the instance_id to pass to the launch_cvd via --base_instance_num. Otherwise, manually adding vsock_cid is a must.

By the way, I'd say "build.sh" is maintained but setup.sh hasn't been. Personally, I am planning re-writing the script to reflect recent changes. Not sure when it is going to be, though.

luke-ingle:
> I guess, however, the instance_id (given to the Cuttlefish via launch_cvd arguments) is not allowed to be 0.

Yep, absolutely. If you run help for launch_cvd you can see 2 instances where this is mentioned.

```
An instance number of i'th instance is determined by --num_instances=N and --base_instance_num=B. 
The instance number of i'th instance is B + i where i in [0, N-1] and B >= 1.
```
and 
```
-base_instance_num (The instance number of the device created. When
      `-num_instances N` is used, N instance numbers are claimed starting at
      this number.) type: int32 default: 1
```
Admittedly not that clear, but heavily implied it must be 1 or greater. (It's much more obvious in the cpp code I mentioned above)

> I think, the line you are removing might have to add +1 to the instance_id to pass to the launch_cvd via --base_instance_num.
So I took some time this morning to try out a few things, and originally tried just changing the default instance id to 1 and kept passing it to launch_cvd, which worked, except that if you used vsock_cid option the cvd numbering would start at 2, instead of 1. e.g. for 8 instances...
```
cvd-4: Virtual device booted successfully
cvd-5: VIRTUAL_DEVICE_BOOT_COMPLETED
cvd-5: Virtual device booted successfully
cvd-2: VIRTUAL_DEVICE_BOOT_COMPLETED
cvd-2: Virtual device booted successfully
cvd-3: VIRTUAL_DEVICE_BOOT_COMPLETED
cvd-3: Virtual device booted successfully
cvd-9: VIRTUAL_DEVICE_BOOT_STARTED
cvd-7: VIRTUAL_DEVICE_BOOT_STARTED
cvd-6: VIRTUAL_DEVICE_BOOT_STARTED
cvd-9: VIRTUAL_DEVICE_BOOT_COMPLETED
cvd-9: Virtual device booted successfully
cvd-8: VIRTUAL_DEVICE_BOOT_STARTED
cvd-7: VIRTUAL_DEVICE_BOOT_COMPLETED
cvd-7: Virtual device booted successfully
cvd-6: VIRTUAL_DEVICE_BOOT_COMPLETED
cvd-6: Virtual device booted successfully
cvd-8: VIRTUAL_DEVICE_BOOT_COMPLETED
cvd-8: Virtual device booted successfully
```

Keeping the removal of the base_instance_num as per my initial commit, remedies this so I have left it in, and added the instance number correction, to match how cuttlefish behaves.

Let me know if there's anything else I should do, but from my testing it seems to now behave correctly for multiple instances and if you set --vsock-cid or not.

kwstephenkim (approved):

PR #370, opened 2023-10-22
Fix operator https not running

yongwoo42:
1. there is no where to specify operator_tls_cert_dir to create cert files
 *  put this variable on cuttlefish-operator.default

2. call startHttpsServer when httpPort is not available but httpPort has default port, no chance to run HttpsServer

### DIFF
--- a/frontend/debian/cuttlefish-user.cuttlefish-operator.default
+++ b/frontend/debian/cuttlefish-user.cuttlefish-operator.default
@@ -14,7 +14,8 @@ operator_https_port=1443
 #
 # The directory where the TLS certificate needed to establish the HTTPS
 # communication lives.
-operator_tls_cert_dir="/etc/cuttlefish-common/operator/cert"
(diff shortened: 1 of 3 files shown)


frontend/debian/cuttlefish-user.cuttlefish-operator.default
```diff
@@ -14,8 +14,7 @@ operator_https_port=1443
 #
 # The directory where the TLS certificate needed to establish the HTTPS
 # communication lives.
-# Defaults to "/etc/cuttlefish-common/operator/cert".
-# operator_tls_cert_dir=
```
jemoreira, line 18: This change is unnecessary, that's indeed the default. Instead insert this line at the beginning of the `gen_cert()` function in cuttlefihs-operator.init:
```
operator_tls_cert_dir=${operator_tls_cert_dir:-/etc/cuttlefish-common/operator/cert}
```

jemoreira (approved):

PR #374, opened 2023-10-26
Build(deps): Bump google.golang.org/grpc from 1.40.0 to 1.56.3 in /frontend/src/liboperator

jemoreira:
This fixes https://github.com/google/android-cuttlefish/pull/373


jemoreira:
Didn't actually fix it, the problems seems to be the update is not compatible with go 1.13

PR #376, opened 2023-10-31
Add API to provide hint to expect messages soon

jmacnak:
... so that the PollingConnector can poll more frequently after a device display power event occurs (where we expect a new display stream).

Bug: b/306499233
Test: lunch aosp_cf_x86_64_foldable-userdebug && m
Test: cvd start
Test: press UI fold button and notice smaller delay for display 2

### DIFF
--- a/frontend/src/operator/intercept/js/server_connector.js
+++ b/frontend/src/operator/intercept/js/server_connector.js
@@ -90,6 +90,13 @@ class Connector {
   async sendToDevice(msg) {
     throw 'Not implemented!';
   }
+
+  // Provides a hint to this controller that it should expect messages from the
+  // signaling server soon. This is useful for a connector which polls for
+  // example which might want to poll more quickly for a period of time.
+  expectMessagesSoon(durationMilliseconds) {
+    throw 'Not implemented!';
+  }
 }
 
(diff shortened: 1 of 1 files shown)


frontend/src/operator/intercept/js/server_connector.js
```diff
@@ -227,22 +235,39 @@ class PollingConnector extends Connector {
...
       let messages = await this.#pollMessages();
 
-      // Do exponential backoff on the polling up to 60 seconds
-      currentPollDelay = Math.min(60000, 2 * currentPollDelay);
+      if (Date.now() < this.#pollQuicklyUntil) {
+        currentPollDelay = shortPollDelay;
+      } else {
+        // Do exponential backoff on the polling up to 60 seconds
+        currentPollDelay = Math.min(60000, 2 * currentPollDelay);
+      }
```
jemoreira, line 249: nit: Extract this to a helper function named like `calcNextPollDelay()`.

jemoreira (approved):

frontend/src/operator/intercept/js/server_connector.js
jmacnak: Done. PTAL!

PR #382, opened 2023-11-16
Implement cvd stop action.

ser-io:


frontend/src/host_orchestrator/orchestrator/cvd/cvd.go
```diff
@@ -122,6 +124,50 @@ func (c *Command) Run() error {
 	return nil
 }
 
+// Same as `Run()`, but error won't be populated in case of failed executions, the caller gets the exit code value.
+func (c *Command) Execute() (int, error) {
+	// Makes sure cvd server daemon is running before executing the cvd command.
```
jemoreira, line 129: why? The server should start on its own automatically

frontend/src/host_orchestrator/orchestrator/cvd/cvd.go
```diff
@@ -122,6 +124,50 @@ func (c *Command) Run() error {
 	return nil
 }
 
+// Same as `Run()`, but error won't be populated in case of failed executions, the caller gets the exit code value.
```
jemoreira, line 127: Can `Run` be implemented by calling `Execute` then and processing the exit code it returns?

frontend/src/host_orchestrator/orchestrator/execcvdcommandaction.go
```diff
@@ -0,0 +1,114 @@
...
+		execContext:     newCVDExecContext(opts.ExecContext, opts.CVDUser),
+	}
+}
+
+func (a *ExecCVDCommandAction) Run() (apiv1.Operation, error) {
+	if err := validateExecCVDCommandRequest(a.req); err != nil {
+		return apiv1.Operation{}, operator.NewBadRequestError("invalid ExecCVDCommandRequest", err)
+	}
+	op := a.om.New()
+	go func(op apiv1.Operation) {
```
jemoreira, line 62: The response from this api is not an operation.

jemoreira (changes requested):

frontend/src/liboperator/api/v1/messages.go
```diff
@@ -177,3 +177,12 @@ type UploadDirectory struct {
...
 	Items []*UploadDirectory `json:"items"`
 }
+
+type ExecCVDCommandRequest struct {
+	Args []string `json:"args"`
+}
+
+type ExecCVDCommandResponse struct {
+	ExecID   string `json:"exec_id"`
+	ExitCode int    `json:"exit_code"`
```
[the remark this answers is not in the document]
ser-io: simplified the PR so it executes `cvd stop` only, this is no longer relevant.

frontend/src/host_orchestrator/main.go
```diff
@@ -123,10 +123,11 @@ func main() {
...
 	imPaths := orchestrator.IMPaths{
-		RootDir:          *imRootDir,
-		CVDToolsDir:      *imRootDir,
-		ArtifactsRootDir: filepath.Join(*imRootDir, "artifacts"),
-		RuntimesRootDir:  filepath.Join(*imRootDir, "runtimes"),
+		RootDir:              *imRootDir,
+		CVDToolsDir:          *imRootDir,
+		ArtifactsRootDir:     filepath.Join(*imRootDir, "artifacts"),
+		RuntimesRootDir:      filepath.Join(*imRootDir, "runtimes"),
+		CVDCMDOutputsRootDir: filepath.Join(*imRootDir, "cvdcmdout"),
```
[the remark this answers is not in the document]
ser-io: ditto.

frontend/src/host_orchestrator/orchestrator/cvd/cvd.go
ser-io: we should add the explanation in this line instead.

https://github.com/google/android-cuttlefish/blob/main/frontend/src/host_orchestrator/orchestrator/cvd/cvd.go#L127-L132

frontend/src/host_orchestrator/orchestrator/cvd/cvd.go
ser-io: simplified the PR so it executes cvd stop only, this is no longer relevant.

frontend/src/host_orchestrator/orchestrator/execcvdcommandaction.go
ser-io: simplified the PR so it executes cvd stop only, this is no longer relevant.

frontend/src/host_orchestrator/orchestrator/stopcvdaction.go
```diff
@@ -0,0 +1,76 @@
...
+	if err := a.cvdDownloader.Download(a.cvdToolsVersion, a.paths.CVDBin(), a.paths.FetchCVDBin()); err != nil {
+		return nil, operator.NewInternalError("failed downloading the cvd binary", err)
+	}
+	cmdOpts := cvd.CommandOpts{
+		Home: a.paths.RuntimesRootDir,
+	}
+	cmd := cvd.NewCommand(a.execContext, a.paths.CVDBin(), []string{"stop"}, cmdOpts)
+	if err := cmd.Run(); err != nil {
+		return nil, operator.NewInternalError("cvd stop failed", err)
+	}
```
jemoreira, line 74: How does this work when you have created two instances in the same host?

ser-io: `cvd stop` stops all the instances, tested this running `cvd directly`. as creating multiple instances in HO is currently broken.

jemoreira: This is a deficiency of `cvd` that we can't fix now, but we don't have to carry it over to the HO. Can we use selectors to tell `cvd` what instance we want to stop? It's ok if it stops everything in the group now, but when that is fixed we'll get the correct behavior without any future work needed on our part.

ser-io: I see, will update this PR with selectors, however we need to be able to create multiple groups/cvds  to test `cvd stop` with selectors, creating multiple groups/instances is broken. I'll update the PR after it's fixed.

ser-io: implemented cvd selectors for `cvd stop`. PTAL

jemoreira (approved):

PR #385, opened 2023-11-21
Set separated default cvd binary target between x86_64 and arm64

0405ysj:
An implementation consideration:
Should we keep the legacy argument `orchestrator_cvdbin_android_build_target` for the maintaining existing HO servers, or deprecate this?
I prefer to delete the old argument. On the other hand, I'm worried if there's any machine using that argument.


ikicha:
we can do that in go code with build tag(https://pkg.go.dev/go/build#hdr-Build_Constraints) I think it might be better in readability. @jemoreira , WDYT?

ikicha:
nits: how about constants_x86.go and constants_arm64.go instead?

ikicha (approved):
thanks!

0405ysj:
Could you merge this? I cannot merge this by myself.

ikicha:
merged.

frontend/src/host_orchestrator/constants_arm64.go
```diff
@@ -0,0 +1,23 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build arm64
+// +build arm64
+
+package main
+
+const (
+	defaultCVDBinAndroidBuildID     = "11125998"
```
jemoreira, line 21: Be careful when choosing a build id that is significantly different for both platforms:
There have been several changes introduced into `cvd` recently, we've tried to keep up from the HO side, but it's inevitable to fall behind sometimes. If you choose different versions of `cvd` for different architectures the code that calls it may be incompatible with at least one of them.

0405ysj: QQ: Does upgrading cvd binary build ID for x86 cause any problem? If there's no issue, I want to upgrade it as similar number which arm64 have, though it couldn't be exact same number.

ikicha: 10797123 is the most similar build id for arm64

0405ysj: Then matching or downgrading arm64 version similar to x86_64 looks better? No strong opinions from me.

ikicha: As Jorge mentioned, cvd has been changed a lot recently, so for the stability, using current version for both x86 and arm64 looks good to me, and then, we can update them soon after checking stability and compatibility with HO

0405ysj: That makes sense. I'll create a short patch that downgrading arm64 build ID.

0405ysj: https://github.com/google/android-cuttlefish/pull/386

0405ysj: Merged following PR, so set as resolved :)

PR #387, opened 2023-11-22
Mini-merge: Create fix-up commit and reset index

Databean:
Create a fix-up commit as a hack around issues in file merging to catch any missing changes. This final commit forces all files covered by mappings to have the same contents as the source files.

Reset the working tree and index at the end so there is nothing to commit.


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #389, opened 2023-11-24
_cutf-operator run operator in HO, not root

0405ysj:


0405ysj:
Without this change, I found operator process is not killed when removing cuttlefish-orchestration or stopping cuttlefish-host_orchestrator service.

0405ysj:
@ikicha Could you merge this change? This was the root cause the reason we can not see our instances in HO.

ikicha (approved):

PR #393, opened 2023-12-07
Warn on unused results

Databean:
Equivalent to https://android-review.git.corp.google.com/c/device/google/cuttlefish/+/2856428

Tested with `meson compile`


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #395, opened 2023-12-08
Start GitHub action to validate that cvd builds

Databean:
On the GitHub workflow runner this fails with errors that aren't on my machine. The meson build file is fixed in this commit, but c++ compilation errors remain.


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #397, opened 2023-12-12
Depend on openssl instead of boringssl

jemoreira:


cvd/debian/control
```diff
@@ -12,6 +12,7 @@ Build-Depends: config-package-dev,
                libjsoncpp-dev,
                libcurl4-openssl-dev,
                libgoogle-glog-dev,
+               libssl-dev,
```
Databean, line 15: For `libssl-dev` and some of these other packages that provide `.so` files, should they also be included in `Depends`? They're needed for both compilation and at runtime for dynamic linking.

On the other hand, if ` ${misc:Depends}` and `${shlibs:Depends` include these, should `clang` and `meson` etc. be filtered out?

Databean (approved):

PR #399, opened 2023-12-14
Adds group properties to api.CVD definition

ser-io:


frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -58,6 +58,7 @@ func (c *Controller) AddRoutes(router *mux.Router) {
 	router.Handle("/cvds",
 		httpHandler(newCreateCVDHandler(c.Config, c.OperationManager, c.UserArtifactsManager))).Methods("POST")
 	router.Handle("/cvds", httpHandler(&listCVDsHandler{Config: c.Config})).Methods("GET")
+	router.Handle("/cvdgroups", httpHandler(&listCVDGroupsHandler{Config: c.Config})).Methods("GET")
```
jemoreira, line 61: cvd groups have no significant meaning on their own. Either add an extra `group` field to the cvd descriptions we return or make the `cvds` resource a subresource of groups.

I think the extra fields in `CVD` makes the most sense for now: It's the simplest of the two while returning the same amount of information while not tying us to a specific solution prematurely (you can continue to return the group name as part of the CVD object even if you later decide to add more endpoints and functionality).

Similarly, you can always specify the group name when creating a new cvd instead of creating a group first and then adding cvds (not even sure this is supported yet).

jemoreira (changes requested):

frontend/src/host_orchestrator/orchestrator/controller.go
ser-io: added the extra fields to CVD.

jemoreira (changes requested):

frontend/src/liboperator/api/v1/messages.go
```diff
@@ -148,7 +148,11 @@ type Operation struct {
 }
 
 type CVD struct {
-	// [Output Only]
+	// [Output Only] Identifier within the whole fleet. Format: "{group}/{name}".
+	ID string `json:"id"`
```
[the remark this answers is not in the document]
ser-io: agree, done.

jemoreira (approved):

PR #400, opened 2023-12-21
Add script to facilitate import of AOSP code

jemoreira:


tools/minimerge/import.sh
```diff
@@ -0,0 +1,52 @@
...
+
+import || (git reset --hard "${current_commit}"; exit)
+
+#Find and update last imported commit
+pushd "${CUTTLEFISH_DIR}" >/dev/null
+current_import="$(git rev-parse HEAD)"
+popd >/dev/null
+
+pushd "${DIR}" >/dev/null
+sed -i -e "s|${last_imported_commit}|${current_import}|g" "${METADATA_FILE}"
```
Databean, line 55: Rather than risk a bad interaction with `sed`, can this overwrite the `tools/minimerge/cuttlefish_import_metadata.txt` entirely?

tools/minimerge/import.sh
```diff
@@ -0,0 +1,52 @@
...
+
+echo "Importing AOSP's device/google/cuttlefish project"
+
+# Do these early to avoid doing unnecessary work when these fail
+# Find last imported commit
+stat "${METADATA_FILE}" >/dev/null
+last_imported_commit="$(grep '^[[:space:]]*last_import[[:space:]]*:' "${METADATA_FILE}" | cut -d: -f2 | grep -o -E '[^[:space:]]+' )"
+# Find current commit
+pushd "${DIR}" >/dev/null
+[[ `git status --porcelain` ]] >/dev/null && (echo "Repository is not in clean state"; exit)
```
Databean: Is this trying to test if `git status --porcelain` returns non-empty output? On my machine `git status --porcelain` writes to `stderr` so I'm not how that is supposed to work.

It seems to work as intended in `bash`, but in `zsh` I need to use
```bash
[[ `git status --porcelain 2>&1` ]]
```
for this to work consistently. Not sure how important it is since the file explicitly uses `/bin/bash`.

Also `[[` is not POSIX, it's a [bash extension](https://stackoverflow.com/a/47576482).

tools/minimerge/import.sh
```diff
@@ -0,0 +1,52 @@
...
+
+function import() {
+  "${TOOL}" \
+    --source="${CUTTLEFISH_DIR}" \
+    --destination="${REPO_DIR}" \
+    --rev_range="${last_imported_commit}..HEAD" \
+    "@${DIR}/cuttlefish_args.txt"
+}
+
+import || (git reset --hard "${current_commit}"; exit)
```
Databean: Does `exit` here exit the script? It seems to me like it exits the subshell created by the parentheses.

For example with this `test.sh`:
```bash
#!/bin/bash
(echo "hello"; exit; echo "abc")
(echo "goodbye"; exit; echo "def")
```
Running it produces:
```shell
$ bash test.sh
hello
goodbye
```

tools/minimerge/import.sh
Databean: Is `exit` here intended to stop the script? It seems to me like it exits the subshell created by the parentheses.

For example with this `test.sh`:
```bash
#!/bin/bash
(echo "hello"; exit; echo "abc")
(echo "goodbye"; exit; echo "def")
```
Running it produces:
```shell
$ bash test.sh
hello
goodbye
```

Databean (approved):

tools/minimerge/import.sh
jemoreira: TBH I found this on stackoverflow.

It's ok if it only works on bash.

tools/minimerge/import.sh
jemoreira: Fixed it

tools/minimerge/import.sh
jemoreira: Fixed it

tools/minimerge/import.sh
jemoreira: Done

PR #403, opened 2024-01-08
Operator supports device pre-registration

jemoreira:
This gives the CVD server the chance to provide the group and instance names to the operator for displaying in the WebUI.

### DIFF
--- a/frontend/src/liboperator/api/v1/messages.go
+++ b/frontend/src/liboperator/api/v1/messages.go
@@ -192,8 +192,15 @@ type CVD struct {
 // Identifier within the whole fleet. Format: "{group}/{name}".
 func (c *CVD) ID() string { return c.Group + "/" + c.Name }
 
+type DeviceDescriptor struct {
+	DeviceId string `json:"device_id"`
+	GroupId  string `json:"group_id"`
+	Owner    string `json:"owner,omitempty"`
+	Name     string `json:"name,omitempty"`
+}
+
 type DeviceInfoReply struct {
-	DeviceId         string      `json:"device_id"`
+	DeviceDescriptor
 	RegistrationInfo interface{} `json:"registration_info"`
 }
 
--- a/frontend/src/liboperator/api/v1/messages.go
+++ b/frontend/src/liboperator/api/v1/messages.go
@@ -14,6 +14,33 @@
 
 package v1
 
+// Control messages
+
+type ControlMsgHeader struct {
+	Type string `json:"message_type"`
+}
+
+type PreRegisterMsg struct {
+	// Type must be set to "pre-register"
+	ControlMsgHeader
+	GroupName string `json:"group_name"`
+	Owner     string `json:"owner"`
+	Devices   []struct {
+		Id   string `json:"id"`
+		Name string `json:"name"`
+	} `json:"devices"`
+}
+
+type RegistrationStatusReport struct {
+	Id     string `json:"id"`
+	Status string `json:"status"`
+	Msg    string `json:"message"`
+}
+
+type PreRegistrationResponse []RegistrationStatusReport
+
+// Device and client messages
+
 type RegisterMsg struct {
 	DeviceId string      `json:"device_id"`
 	Port     int         `json:"device_port"`
--- a/frontend/src/liboperator/operator/devices.go
+++ b/frontend/src/liboperator/operator/devices.go
@@ -23,10 +23,19 @@ import (
 	"sync"
 )
 
+type DeviceDesc struct {
+	DeviceId string `json:"device_id"`
+	GroupId  string `json:"group_id"`
+	Owner    string `json:"owner"`
+	Name     string `json:"name"`
+}
+
 type Device struct {
-	info    interface{}
-	groupId string
-	conn    *JSONUnix
+	// Provided by the device at registration time
+	privateData interface{}
+	// Provided at pre-registration time
+	desc DeviceDesc
+	conn *JSONUnix
 	// Reverse proxy to the client files
 	Proxy *httputil.ReverseProxy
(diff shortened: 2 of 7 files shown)


frontend/src/liboperator/operator/devices.go
```diff
@@ -103,22 +117,33 @@ func (d *Device) ToClient(id int, msg interface{}) error {
...
 
+// preDevice holds information about a device between pre-registration and registration.
+type preDevice struct {
+	Desc  DeviceDesc
+	RegCh chan bool
+}
+
 // Keeps track of the registered devices.
 type DevicePool struct {
+	// preDevicesMtx must always be locked before devicesMtx
```
ser-io, line 123: rather than having two mutexes and two maps in the DevicePool data structure, would having a single mutex/map pair to handle pre registration and devices work here? Suggesting this as it would be easier to keep track of one mutex.


```
type preRegistration struct {
	desc  DeviceDesc
        done chan bool
}

type poolEntry {
  preReg *preRegistration
  device *Device
}

type DevicePool {
  mtx sync.Mutex
  entries  map[string]*poolEntry
}
```

jemoreira: I thought about this and ended up preferring the two mutexes approach by a very slight margin. These are the reasons:
* The `done` channel makes no sense after registration, that fact is easily conveyed by it going out of scope
* The DeviceDescriptor is an integral part of the Device, the fact that it was provided at pre-registration is just a detail
* [minor] Listing registered devices with two maps doesn't require filtering
* [minor] No need to have a status field (or equivalent) to know if a device is pre-registered or fully registered

An alternative would be to use the same mutex for both maps. The downside of that is that interactions with registered devices (signaling) can't happen concurrently with pre-registration operations.

I'm ok with using the same mutex for both structs, but I strongly prefer having two structs over just one. WDYT?

frontend/src/liboperator/operator/devices.go
```diff
@@ -36,18 +40,13 @@ type Device struct {
...
-}
-
 type Group struct {
 	deviceIds []string
 }
 
 const DEFAULT_GROUP_ID = "default"
 
-func NewDevice(conn *JSONUnix, port int, info interface{}) *Device {
+func newDevice(id string, conn *JSONUnix, port int, privateData interface{}) *Device {
```
ser-io, line 49: suggest to use a new type here, i.e `type devicePrivateData interface{}` here.

```
func newDevice(id string, conn *JSONUnix, port int, data devicePrivateData) *Device

func groupIdFromPrivateData(data devicePrivateData) string
```

frontend/src/liboperator/operator/devices_test.go
```diff
@@ -123,19 +125,57 @@ func TestPoolRegister(t *testing.T) {
...
 
+func TestPoolPreRegister(t *testing.T) {
+	p := NewDevicePool()
+	p.CancelPreRegistration("inexistent_device")
+	if err := p.PreRegister(&apiv1.DeviceDescriptor{
+		DeviceId: "d1",
+		GroupId:  "g1",
+		Owner:    "o1",
+		Name:     "n1",
+	}, make(chan bool, 1)); err != nil {
```
ser-io, line 136: nit: declaring a device descriptor variable in a single statement makes it easier to follow.

```
dd1 := &apiv1.DeviceDescriptor{
  DeviceId: "d1",
  GroupId:  "g1",
  Owner:    "o1",
  Name:     "n1",
}
if err := p.PreRegister(dd1, make(chan bool, 1)); err != nil {
  t.Fatal("Failed to pre-register: ", err)
}
```

frontend/src/liboperator/operator/operator.go
```diff
@@ -66,7 +74,28 @@ func SetupDeviceEndpoint(pool *DevicePool, config apiv1.InfraConfig, path string
 			}
 			go deviceEndpoint(NewJSONUnix(c), pool, config)
 		}
+	}, nil
+}
+
+// Sets up a unix socket for server control and returns a function that listens on the socket
+// until an error occurrs.
+func SetupPreregistrationEndpoint(pool *DevicePool, path string) (func() error, error) {
```
ser-io, line 82: nit: should this be `SetupControlEndpoint` instead?

frontend/src/liboperator/operator/devices.go
```diff
@@ -103,22 +112,31 @@ func (d *Device) ToClient(id int, msg interface{}) error {
 	return client.Send(msg)
 }
 
+// preDevice holds information about a device between pre-registration and registration.
```
ser-io, line 115: nit: would the name `preRegProcess` be more appropriate here? the `preDevice` name is descriptive by itself.

frontend/src/liboperator/operator/devices.go
jemoreira: ack, but will address it in a different PR since this one only changed the name of the field, the type was already there.

frontend/src/liboperator/operator/operator.go
jemoreira: done

frontend/src/liboperator/operator/devices_test.go
jemoreira: done

frontend/src/liboperator/api/v1/messages.go
```diff
@@ -14,6 +14,32 @@
 
 package v1
 
+// Control messages
+
+type ControlMsgHeader struct {
+	Type string `json:"message_type"`
+}
+
+type PreRegisterMsg struct {
```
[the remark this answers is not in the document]
jemoreira: done

ser-io (approved):

PR #407, opened 2024-01-12
Github action to publish artifacts

ser-io:


tools/buildimage/install_gce_guest_environment.sh
```diff
@@ -0,0 +1,107 @@
...
+done
+
+mount_point="/mnt/image"
+sudo mkdir ${mount_point}
+
+# offset value is 262144 * 512, the `262144`th is the sector where the `Linux filesystem` partition
+# starts and `512` bytes is the sectors size. See `sudo fdisk -l disk.raw`.
+sudo mount -o loop,offset=$((262144 * 512)) ${diskraw} ${mount_point}
+
+cat <<'EOF' >/tmp/rc.local
```
jemoreira, line 55: Could this be in a file by its own and copied to /tmp/rc.local here?

jemoreira (approved):

tools/buildimage/install_gce_guest_environment.sh
ser-io: yes it could, decided to keep it on the same file to keep it as closer as possible to the article https://cloud.google.com/compute/docs/images/install-guest-environment#debian_1.

PR #412, opened 2024-01-19
Tool to export host image to GCE.

ser-io:


tools/imagetogce/main.sh
```diff
@@ -0,0 +1,101 @@
...
+
+gcloud storage cp image.tar.gz  gs://${gce_bucket}/${name}.tar.gz
+
+echo "Creating image ..."
+gcloud compute images create ${name} \
+  --source-uri gs://${gce_bucket}/${name}.tar.gz \
+  --family ${image_family_name}
+
+# Cleanup
+rm image.zip
```
jemoreira, line 98: With `set -e` this line may not run if an error occurs in the image creation script. You can force the cleanup to run deferring it to run on exit using a trap right after the file is downloaded.
The same applies to image.tar.gz and the file in the bucket.

jemoreira (approved):

tools/imagetogce/main.sh
ser-io: Done. Good call, didn't know about `trap` in bash.

PR #415, opened 2024-01-22
Dockerfile for arm server

ikicha:


docker/arm-server/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+EXPOSE 6520-6620
+WORKDIR /root
+USER root
+
+RUN set -x
+
+RUN apt update
+RUN apt install -y --no-install-recommends git wget unzip
+RUN apt install -y --no-install-recommends ca-certificates && update-ca-certificates
+RUN git clone https://github.com/google/android-cuttlefish.git
```
k311093, line 14: How about moving this line with line 6 (WORKDIR /root) to before line 19 to separate stages for installing packages and building android-cuttlefish?

k311093 (changes requested):

k311093 (approved):

PR #425, opened 2024-01-31
Enable building cvd on github workflow

jemoreira:


cvd/meson.build
```diff
@@ -57,9 +57,9 @@ cvd_sources = [
   'cuttlefish/host/commands/cvd/lock_file.cpp',
   'cuttlefish/host/commands/cvd/logger.cpp',
   'cuttlefish/host/commands/cvd/main.cc',
-  'cuttlefish/host/commands/cvd/metrics/cvd_metrics_api.cpp',
+  # 'cuttlefish/host/commands/cvd/metrics/cvd_metrics_api.cpp',
```
ser-io, line 60: why not removing this?

ser-io (approved):

cvd/meson.build
jemoreira: done

PR #426, opened 2024-01-31
Move cvd to base/ in preparation to release it as part of cuttlefish-base

jemoreira:


.github/workflows/presubmit.yaml
```diff
@@ -55,6 +55,6 @@ jobs:
     - name: Install dependencies
       run: apt install -y clang pkg-config meson libfmt-dev libgflags-dev libjsoncpp-dev protobuf-compiler libgtest-dev libcurl4-openssl-dev libprotobuf-c-dev libgoogle-glog-dev libssl-dev libxml2-dev openssl uuid-dev
```
Databean, line 56: These need to be part of the `cuttlefish-base` workflow now as well.

Databean (approved):

.github/workflows/presubmit.yaml
jemoreira: Removed the last commit from this PR as it depends on a change submitted to AOSP.

PR #428, opened 2024-02-01
Dockerfile for ARM64 uses debian packages built with local code

0405ysj:
Also I did some refactoring around dockerfile for ARM64.

BTW, this PR weren't verified in ARM64 environment, just only in x86_64 env w/ my local machine.
I'll proceed to verify when review is approved.

### DIFF
--- a/docker/arm-server/Dockerfile
+++ b/docker/arm-server/Dockerfile
@@ -20,17 +20,8 @@ RUN set -x
 # Install dependants of CF debian packages.
 RUN apt update
 RUN apt install -y --no-install-recommends \
-    build-essential:native \
     ca-certificates \
-    config-package-dev \
     curl \
-    debhelper-compat \
-    dpkg-dev \
-    gcc-10 \
-    git \
-    golang-1.13 \
-    libprotobuf-dev \
-    protobuf-compiler \
     sudo
 RUN update-ca-certificates
 
--- a/docker/arm-server/Dockerfile
+++ b/docker/arm-server/Dockerfile
@@ -1,30 +1,53 @@
-FROM ubuntu:22.04
+# Docker image for running CF instances in ARM64 server.
+# Docker image includes HO(Host Orchestrator) inside,
+# so it could execute CF instance with API in HO.
+FROM ubuntu:22.04 AS cuttlefish-arm64
 
+# Expose Operator Port (HTTP:1080, HTTPS:1443)
+EXPOSE 1080 1443
+# Expose HO(Host Orchestrator) Port (HTTP:2080, HTTPS:2443)
+EXPOSE 2080 2443
+# Expose WebRTC Port
 EXPOSE 15550-15560
-EXPOSE 1443
+# Expose ADB Port
+# Corresponding ADB port for CF instance is, 6520+instance_num-1.
 EXPOSE 6520-6620
+
 USER root
 
 RUN set -x
 
+# Install dependants of CF debian packages.
 RUN apt update
-RUN apt install -y --no-install-recommends git
-RUN apt install -y --no-install-recommends ca-certificates && update-ca-certificates
-RUN apt install -y --no-install-recommends dpkg-dev build-essential:native config-package-dev debhelper-compat golang-1.13 libprotobuf-dev protobuf-compiler curl sudo
-RUN apt install -y --no-install-recommends gcc-10
+RUN apt install -y --no-install-recommends \
+    build-essential:native \
+    ca-certificates \
+    config-package-dev \
+    curl \
+    debhelper-compat \
+    dpkg-dev \
+    gcc-10 \
+    git \
+    golang-1.13 \
+    libprotobuf-dev \
+    protobuf-compiler \
+    sudo
+RUN update-ca-certificates
 
+# ADD CF debian packages built by
+# docker/debs-builder-docker/build-debs-with-docker.sh.
+COPY ./out/cuttlefish-base_*.deb /root/
+COPY ./out/cuttlefish-user_*.deb /root/
+COPY ./out/cuttlefish-orchestration_*.deb /root/
+
+# Install CF debian packages.
 WORKDIR /root
-RUN git clone https://github.com/google/android-cuttlefish.git
-WORKDIR /root/android-cuttlefish/base
-RUN dpkg-buildpackage
-WORKDIR /root/android-cuttlefish/frontend
-RUN dpkg-buildpackage
+RUN apt install -y -f \
+    /root/cuttlefish-base_*.deb \
+    /root/cuttlefish-user_*.deb \
+    /root/cuttlefish-orchestration_*.deb
 
-WORKDIR /root/android-cuttlefish
-RUN apt install -y -f ./cuttlefish-base_*.deb ./cuttlefish-user_*.deb ./cuttlefish-orchestration_*.deb
 RUN echo "num_cvd_accounts=100" >> /etc/default/cuttlefish-host-resources
-
-WORKDIR /root
 
 RUN usermod -aG kvm root
 RUN usermod -aG cvdnetwork root
--- a/docker/arm-server/build.sh
+++ b/docker/arm-server/build.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+# This shell script exists for building ARM64 docker image.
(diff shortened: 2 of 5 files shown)


docker/arm-server/Dockerfile
```diff
@@ -1,30 +1,53 @@
...
-
+# Docker image for running CF instances in ARM64 server.
+# Docker image includes HO(Host Orchestrator) inside,
+# so it could execute CF instance with API in HO.
+FROM ubuntu:22.04 AS cuttlefish-arm64
+
+# Expose Operator Port (HTTP:1080, HTTPS:1443)
+EXPOSE 1080 1443
+# Expose HO(Host Orchestrator) Port (HTTP:2080, HTTPS:2443)
+EXPOSE 2080 2443
```
ikicha, line 9: I think there are conflicts with https://cs.android.com/android/platform/superproject/main/+/main:device/google/cuttlefish/tools/launch_cvd_arm64_server_docker.sh;l=83?q=launch_cvd_arm
 because I uses 2443 for remote web ui, please modify the script as well.

docker/arm-server/run.sh
```diff
@@ -0,0 +1,2 @@
+# Shell script to run docker image.
+docker run --privileged -d -P -it cuttlefish-arm64
```
ikicha, line 2: does it work even without volume options? run_services.sh expects cf packages.

0405ysj: Yes, it works. Uploading & installing debian packages for CF (e.g. cuttlefish-base) is done in build phase. (Defined in Dockerfile) When starting `/run_services.sh`, the packages already lives in the docker image.

docker/arm-server/Dockerfile
0405ysj: I think we don't need to change. Regardless it's operator or HO, both supports WebUI with operator's API. Was ther e any error when using WebUI of HO when running the script?

docker/arm-server/build.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# This shell script exists for building ARM64 docker image.
+# Docker image includes HO(Host Orchestrator) inside,
+# so it could execute CF instance with API in HO.
+
+# Build debian packages(e.g. cuttlefish-base) with docker.
+pushd ..
+./debs-builder-docker/build-debs-with-docker.sh
```
ikicha, line 9: I thought we can still build debian package inside docker. maybe we can provide options: one for local source tree, the other for remote source tree.

docker/arm-server/Dockerfile
ikicha: i mean current logic set up port forward rule like 2443(remote) -> xxxxx(randomly assigned in host) -> 1443(docker), but if we expose 2443 as well, there will be two rule for 2443. so we need to modify 2443 in the script to the other port

ikicha (commented):
If we still want to build debian package inside docker(the source would be either local or remote) but we don't want build-only deps, we can use https://docs.docker.com/build/building/multi-stage/

docker/arm-server/run.sh
ikicha: [this line](https://github.com/google/android-cuttlefish/blob/main/docker/arm-server/run_services.sh#L6) will fail, because there is no /root/cvd_home/bin/cvd in this case, isn't it?

docker/arm-server/Dockerfile
0405ysj: AFAIU, `EXPOSE` in Dockerfile doesn't expose port via `-p` option, and it's exposed with randomly assigned port for `-P` option.

Just for my understanding on `tools/launch_cvd_arm64_server_docker.sh`, if we expose port 2443, does it try to make port forwarding via ssh with same port like below? Unless, I don't think we need to modify the script.
- 2443(remote) -> xxxxx(randomly assigned in host) -> 2443(docker)
- 2443(remote) -> xxxxx(randomly assigned in host) -> 1443(docker)

docker/arm-server/run.sh
0405ysj: Yes, you're right theoretically since it doesn't have cvd binary. But it worked well.

Or, do I remove this shell script? This script isn't used by any operation, just for reference to run docker image.

docker/arm-server/build.sh
0405ysj: Doesn't `build-debs-with-docker.sh` build debian packages like cuttlefish-base, cuttlefish-user, or cuttlefish-orchestration inside Docker?

- Docker image X(debs-builder-docker/build-debs-with-docker.sh) builds CF debian packages.
- Build docker image Y(arm-server/build.sh) with installing CF debian packages on top of linux.
- Run docker image Y to run CF instance inside.

I thought we already build debian packages inside docker, no?

docker/arm-server/Dockerfile
0405ysj: Thought again, I changed my mind. Regardless exposing 2443 or not, we should edit the script since we should consider machine using the script already have HO inside.

0405ysj: Done, and submitted into AOSP. Let this be resolved.

docker/arm-server/run.sh
0405ysj: Ah, maybe we could do like this. Running cvd start in /run_services.sh only when cvd binary exists as our intent. Then we could continuously run `build-debs-with-docker.sh` and CF with HO API both.

docker/arm-server/build.sh
ikicha: But there is separate dockerfile for X. can we use https://docs.docker.com/build/building/multi-stage/ to combine them?

docker/arm-server/Dockerfile
```diff
@@ -1,30 +1,53 @@
-FROM ubuntu:22.04
-
+# Docker image for running CF instances in ARM64 server.
+# Docker image includes HO(Host Orchestrator) inside,
+# so it could execute CF instance with API in HO.
+FROM ubuntu:22.04 AS cuttlefish-arm64
```
ikicha, line 4: why do we need 'as cuttlefish-arm64' here?

docker/arm-server/build.sh
ikicha: Ack

docker/arm-server/Dockerfile
```diff
@@ -1,30 +1,53 @@
...
 RUN set -x
 
+# Install dependants of CF debian packages.
 RUN apt update
-RUN apt install -y --no-install-recommends git
-RUN apt install -y --no-install-recommends ca-certificates && update-ca-certificates
-RUN apt install -y --no-install-recommends dpkg-dev build-essential:native config-package-dev debhelper-compat golang-1.13 libprotobuf-dev protobuf-compiler curl sudo
-RUN apt install -y --no-install-recommends gcc-10
-
+RUN apt install -y --no-install-recommends \
```
ikicha, line 22: if we decided not to build debian packages here, we need only a part of them.

docker/arm-server/Dockerfile
0405ysj: Just for setting target name & distinguishing for visibility. I don't have strong opinion on here.

docker/arm-server/run.sh
0405ysj: Edited run_services.sh, to be able to run regardless existence of `/root/cvd_home/bin/cvd`.

docker/arm-server/Dockerfile
0405ysj: OOC, when you made the initial script, would there be some packages installed for other purposes?

At least for me, `curl` and `sudo` would be useful. Do you have any packages want to make remained?

ikicha: no, i just installed them to build deb packages.

0405ysj: Okay, I got that. I'll test with clearing every dependencies except `curl` and `sudo` then.

docker/arm-server/Dockerfile
0405ysj: Close this conversation, since there looks no more discussion.

ikicha (approved):

PR #437, opened 2024-02-07
Let docker/arm-server/Dockerfile be from debian:stable-20211011

0405ysj:
After cvd became a part of cuttlefish-base, some packages weren't installed well. Synced host OS with `cuttlefish-hostpkg-builder`.

### DIFF
--- a/docker/arm-server/Dockerfile
+++ b/docker/arm-server/Dockerfile
@@ -1,7 +1,7 @@
 # Docker image for running CF instances in ARM64 server.
 # Docker image includes HO(Host Orchestrator) inside,
 # so it could execute CF instance with API in HO.
-FROM ubuntu:22.04 AS cuttlefish-arm64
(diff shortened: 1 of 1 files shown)


ikicha:
I chose ubuntu 22.04 because our servers are using it, does it work well in our servers? and can we use 22.04 in cuttlefish-hostpkg-builder instead?

0405ysj:
Yes, I'm currently checking if debian packages working well in ARM64 machines, and it works well.

Also, verified working well for POST /cvds w/ user_build_source, by checking with aosp & internal arm64 series.

0405ysj:
BTW, I think we can choose using 22.04 in `cuttlefish-hostpkg-builder` for arm64 purpose.
But the problem is, other component using that docker image might be broken.
For example, we might not be able to build cuttlefish debian packages via docker image & using it in x86_64 debian machines.

ikicha (approved):

PR #447, opened 2024-02-14
Build cvd as part of cuttlefish-base in ubuntu 20.04

jemoreira:
* Fixes the build
* Re-adds cvd to debian package
* configures presubmit build on 20.04
* No longer import single forked file from aosp


base/debian/control
```diff
@@ -4,15 +4,33 @@ Section: misc
 Priority: optional
 Build-Depends: config-package-dev,
                debhelper-compat (= 12),
+               clang,
+               pkg-config,
+               meson,
+               libfmt-dev,
```
Databean, line 10: If we're linking these statically it's fine to be in `Build-Depends`, but if we're linking dynamically should these be in `Depends` as well?

Databean (approved):

base/debian/control
jemoreira: Done.
Added some of the packages with the -dev suffix because those had the version in the package name and were different between ubuntu20.04, 22.04 and debian. The -dev package depends on the correct one for each platform.

PR #452, opened 2024-02-22
Create backward compatibility symlinks in case of failures

dimorinny:
Currently cuttlefish runtime symlinks are created only in case of succesfull boot. Acloud cannot pull them when cvd isn't booted, so we cannot see launcher/kernel logs on infra.

Fixing it by executing symlinks creating before cmd result check.

Test: locally emulate vmm issue verified symlinks are created
Test: launcher log is pulled in case of vmm failure
Bug: 325936323
Change-Id: I66d6aea50f011adde19fe175b460832ea281317c


rmuthiah (approved):

rmuthiah:
Closed in favor of https://github.com/google/android-cuttlefish/pull/454

PR #456, opened 2024-02-22
Allow using service account credentials for fetching artifacts.

ser-io:


frontend/src/liboperator/api/v1/messages.go
```diff
@@ -111,6 +111,8 @@ type AndroidCIBundle struct {
 }
 
 // Use `X-Cutf-Host-Orchestrator-BuildAPI-Creds` http header to pass the Build API credentials.
+//
+// - On GCE, use special value "gce" to use the instance service account credentials.
```
jemoreira, line 115: If the remote hosts are configured to use credentials from GCE metadata, it should be transparent to the user/client. Instead of expecting users to request the `gce` special value, make it a configuration of the host orchestrator read at start time and applied to all instances **unless** overridden by the user/client.

jemoreira (changes requested):

frontend/src/liboperator/api/v1/messages.go
ser-io: agree, this should be a host config parameter, sent out https://github.com/google/android-cuttlefish/pull/459 for review.

PR #458, opened 2024-02-23
Support getting download URLs from the LUCI API

Databean:
BuildBucket API:
https://cr-buildbucket.appspot.com/rpcexplorer/services/buildbucket.v2.Builds/SearchBuilds

BuildBucket API access:
https://pkg.go.dev/go.chromium.org/luci/auth/client/cmd/luci-auth

Cloud storage API:
https://cloud.google.com/storage/docs/json_api/v1/objects/get

Getting access to cloud storage:
https://www.chromium.org/chromium-os/developer-library/reference/tools/gsutil/#configure-authentication-boto

See b/322559348 for usage examples with Google credentials.

Bug: b/322559348

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd_parser.cc
+++ b/base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd_parser.cc
@@ -35,7 +35,8 @@ namespace cuttlefish {
 namespace {
 
 constexpr char kUsageMessage[] =
-    "*_build flags accept values in the following format:\n"
+    "*_build flags other than --chrome_os_build accept values in the following "
+    "format:\n"
     "{<branch> | <build_id>}[/<build_target>][{<filepath>}]\n"
     "For example: "
     "\"aosp-main/aosp_cf_x86_64_phone-trunk_staging-userdebug{file.txt}\""
@@ -130,6 +131,10 @@ std::vector<Flag> GetFlagsVector(FetchFlags& fetch_flags,
   flags.emplace_back(
       GflagsCompatFlag("otatools_build", vector_flags.otatools_build)
           .Help("source for the host ota tools"));
+  flags.emplace_back(
+      GflagsCompatFlag("chrome_os_build", vector_flags.chrome_os_build)
+          .Help("source for a ChromeOS build. Formatted as as a numeric build "
+                "id, or '<project>/<bucket>/<builder>'"));
 
   flags.emplace_back(
       GflagsCompatFlag("boot_artifact", vector_flags.boot_artifact)
@@ -162,9 +167,9 @@ Result<int> GetNumberOfBuilds(
        {flags.default_build.size(), flags.system_build.size(),
         flags.kernel_build.size(), flags.boot_build.size(),
         flags.bootloader_build.size(), flags.android_efi_loader_build.size(),
-        flags.otatools_build.size(), flags.boot_artifact.size(),
-        flags.download_img_zip.size(), flags.download_target_files_zip.size(),
-        subdirectory_flag.size()}) {
+        flags.otatools_build.size(), flags.chrome_os_build.size(),
+        flags.boot_artifact.size(), flags.download_img_zip.size(),
+        flags.download_target_files_zip.size(), subdirectory_flag.size()}) {
     if (flag_size == 0) {
       // a size zero flag vector means the flag was not given
       continue;
--- a/base/cvd/cuttlefish/host/libs/web/chrome_os_build_string.cpp
+++ b/base/cvd/cuttlefish/host/libs/web/chrome_os_build_string.cpp
@@ -1,0 +1,86 @@
+//
+// Copyright (C) 2024 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "host/libs/web/chrome_os_build_string.h"
+
+#include <optional>
+#include <string>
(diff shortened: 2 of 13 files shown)


jemoreira:
If you don't plan to merge this to AOSP too, you need to remove a few lines from cuttlefish_args.txt in tools/minimerge to prevent an import from overwriting your changes.

base/cvd/cuttlefish/host/libs/web/chrome_os_build_string.h
```diff
@@ -0,0 +1,44 @@
...
+  std::string builder;
+};
+
+using ChromeOsBuildString = std::variant<ChromeOsBuilder, std::string>;
+
+Result<ChromeOsBuildString> ParseChromeOsBuildString(
+    const std::string& build_string);
+
+Flag GflagsCompatFlag(const std::string& name,
+                      std::optional<ChromeOsBuildString>& value);
```
cjreynol, line 40: Do we need this overload?  `android_build_string.h` has a single value overload to support `--host_package_build`, which is mapped to the non-vector field `FetchFlags.host_package_build`.  I did not see where this is used, only the vector version.

base/cvd/cuttlefish/host/libs/web/credential_source.cc
```diff
@@ -73,13 +75,15 @@ Result<std::unique_ptr<CredentialSource>> GetCredentialSourceLegacy(
   } else if (credential_source == "") {
     LOG(VERBOSE) << "Probing acloud credentials at " << oauth_filepath;
     if (FileExists(oauth_filepath)) {
-      std::ifstream stream(oauth_filepath);
-      auto attempt_load =
-          RefreshCredentialSource::FromOauth2ClientFile(http_client, stream);
+      std::string oauth_contents;
+      CF_EXPECT(
+          android::base::ReadFileToString(oauth_filepath, &oauth_contents));
```
cjreynol, line 80: More of an inquiry than a comment.  Do you think we be using this android helper instead of `file.h:ReadFileContents` (which is used elsewhere in this file)?  That version has more verbose error messages, but the implementations are quite different so I am not sure about all of the tradeoffs between one or the other.  

Mostly I would vote for consistency or making the reason for the diverging choice obvious.  I'll try to dig a bit deeper once I finish reviewing the rest of the changes.

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd_parser.cc
```diff
@@ -130,6 +130,9 @@ std::vector<Flag> GetFlagsVector(FetchFlags& fetch_flags,
   flags.emplace_back(
       GflagsCompatFlag("otatools_build", vector_flags.otatools_build)
           .Help("source for the host ota tools"));
+  flags.emplace_back(
+      GflagsCompatFlag("chrome_os_build", vector_flags.chrome_os_build)
+          .Help("source for a ChromeOS build"));
```
cjreynol, line 135: We should give the user some hint of the format for this flag, that it can be a single value or a triple separated by slashes.

cjreynol: Also, `kUsageMessage` above is no longer accurate about the format of `*_build` flags with this addition.  Perhaps we update with the format of this flag in that message instead of adding it here?  Your call either way, but we should note the exception.

base/cvd/cuttlefish/host/libs/web/luci_build_api.cpp
```diff
@@ -0,0 +1,154 @@
...
+  headers.emplace_back("Accept: application/json");        // Output format
+  return headers;
+}
+
+Result<std::vector<std::string>> LuciBuildApi::CloudStorageHeaders() {
+  std::vector<std::string> headers;
+  if (storage_credential_source_) {
+    std::string credential =
+        CF_EXPECT(storage_credential_source_->Credential());
+    LOG(ERROR) << "storage credential: '" << credential << "'";
```
cjreynol, line 59: Isn't this leaking the credentials in logs?

base/cvd/cuttlefish/host/libs/web/chrome_os_build_string.h
Databean: Removed.

base/cvd/cuttlefish/host/libs/web/chrome_os_build_string.cpp
```diff
@@ -0,0 +1,79 @@
...
+        if (!match.value.empty()) {
+          value = CF_EXPECT(ParseChromeOsBuildString(match.value));
+        }
+        return {};
+      });
+}
+
+Flag GflagsCompatFlag(const std::string& name,
+                      std::vector<std::optional<ChromeOsBuildString>>& value) {
+  return GflagsCompatFlag(name).Setter(
```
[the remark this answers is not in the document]
Databean: I've implemented a `Getter`, but the "Current value" logic seems to be broken elsewhere, I'm not seeing the current value reflected in the output for e.g. `cvd fetch --help --credential_filepath=abc`.

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd_parser.cc
Databean: Done

Databean:
> If you don't plan to merge this to AOSP too, you need to remove a few lines from `cuttlefish_args.txt` in tools/minimerge to prevent an import from overwriting your changes.

Should I be merging it into AOSP?

Databean (commented):
Thanks for reviewing!

cjreynol (approved):

base/cvd/cuttlefish/host/libs/web/chrome_os_build_string.cpp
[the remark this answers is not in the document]
Databean: This ended up a bit more complicated to satisfy multiple versions of `fmtlib` from [7.1.3](https://fmt.dev/7.1.3/api.html#formatting-user-defined-types) through [10.2.0](https://fmt.dev/10.2.0/api.html#formatting-user-defined-types). 7.1.3 is the version I see being used on the GitHub Actions runner.

If you'd like to validate on [Compiler Explorer](https://godbolt.org/z/bo7fbsWfb), this strategy seems to work from 5.2.0 onwards.

PR #462, opened 2024-02-29
Use instance service account credentials if available.

ser-io:


frontend/src/host_orchestrator/orchestrator/createcvdaction.go
```diff
@@ -157,6 +158,12 @@ func (a *CreateCVDAction) launchWithCanonicalConfig(op apiv1.Operation) (*apiv1.
 			}
 		}()
 		args = append(args, "--credential_source="+filename)
+	} else if isRunningOnGCE() {
+		if ok, err := hasServiceAccountAccessToken(); err != nil {
+			return nil, fmt.Errorf("service account token check failed: %w", err)
```
jemoreira, line 163: would it be better to just try to run `cvd` without credentials in this case?

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -586,3 +594,23 @@ func contains(s []uint32, e uint32) bool {
...
+	return err == nil
+}
+
+// For instances running on GCE, checks whether the instance was created with a service account having an access token.
+func hasServiceAccountAccessToken() (bool, error) {
+	req, err := http.NewRequest("GET", "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token", nil)
+	if err != nil {
+		return false, err
+	}
+	req.Header.Set("Metadata-Flavor", "Google")
```
jemoreira, line 609: Is this required? I was under the impression that this is something the server sets in its responses, not something you had to set in the request.

frontend/src/host_orchestrator/orchestrator/createcvdaction.go
ser-io: true, added a log message though.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: yes, https://cloud.google.com/compute/docs/metadata/querying-metadata#parts-of-a-request

jemoreira (approved):

PR #465, opened 2024-03-05
Added bridge type support, and change iptables to iptables-legacy bec…

WeiChungHsu:
Added bridge type support because some system may not have bridge type loaded. Changed iptables to ptables-legacy because the new kernel may not work with the new iptables tool.


rmuthiah:
Question - have you tested this on corp devices as well?

jemoreira (approved):

WeiChungHsu:
> Question - have you tested this on corp devices as well?

Yes, tested on corp devices (cloudtop gLinux) and it is working fine.

rmuthiah (approved):

PR #474, opened 2024-03-13
Test cvd load on presubmit

jemoreira:


e2etests/environment_specs/aosp_main_x64_phone_x2.json
```diff
@@ -0,0 +1,21 @@
+{
+  "instances": [
+    {
+      "name": "ins-1",
+      "disk": {
+        "default_build": "@ab\/aosp-main\/aosp_cf_x86_64_phone-trunk_staging-userdebug"
+      }
+    }, {
```
ser-io, line 8: nit:  new line.

e2etests/boot_tests.bzl
```diff
@@ -10,3 +10,17 @@ def launch_cvd_boot_test(name, branch, target):
             "no-sandbox",
         ],
     )
+
+def cvd_load_boot_test(name, env_file, size = "medium"):
```
ser-io, line 14: do we need boot in the name? why not `cvd_load_test`?

e2etests/environment_specs/aosp_main_x64_phone_x2.json
jemoreira: done

e2etests/boot_tests.bzl
jemoreira: Probably not, but that naming scheme is in the other tests. If you want to see it changed I'll do it in a different PR.

The reason for the "boot" in the name is legacy: Cuttlefish is used to run CTS, so when you only want to make sure it boots it's usually called "boot test".

ser-io (approved):

PR #480, opened 2024-03-21
Use --instance_nums to pass instance number list

k311093:
To create an instance number list from orchestrator to cvd need to use the argument named instance_nums to pass intended numbers.


ikicha (approved):

ikicha:
please check failures in kokoro(might need to fix test as well?)

PR #482, opened 2024-03-23
Add .clang-format file to base/cvd

Databean:
This does not yet add any GitHub action validation that the formatting is actually followed.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,22 @@
+#
+# Copyright (C) 2020 The Android Open Source Project
+#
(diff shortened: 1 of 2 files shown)


base/cvd/.clang-format
jemoreira, line 1: This could be at the top of the repository, that way it would cover any new C++ code we add anywhere else.

Databean: Done.

jemoreira (approved):

Databean:
Thanks for reviewing!

PR #483, opened 2024-03-25
Implement multi-vd vsock connectivity configuration

sferrogoo:
This PR enables `cvd` to configure a `vsock_guest_group` launch parameter as part of a new `connectivity` section, which controls vsock communications isolation. The `vhost-device-vsock` service documents this feature with examples at: https://github.com/rust-vmm/vhost-device/blob/main/vhost-device-vsock/README.md , for reference the `groups` parameter used there is identical in format and function to the new `vsock_guest_group` argument that was added in this PR (and below CL)

Note that this PR is intended to be used with several other Cuttlefish changes, all of which are visible and documented at: [aosp/2995598](https://android-review.git.corp.google.com/c/device/google/cuttlefish/+/2995598)  There is an example config there as well for reference, and instructions on how this feature was tested.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/parser/instance/cf_connectivity_configs.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/parser/instance/cf_connectivity_configs.cpp
@@ -1,0 +1,43 @@
+/*
+ * Copyright (C) 2024 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <fmt/format.h>
+#include <json/json.h>
+
+#include "common/libs/utils/result.h"
+
+namespace cuttlefish {
+
+std::string GenerateGroupFlag(const Json::Value& instances_json) {
+  std::vector<std::string> vsock_groups;
+
+  for (const auto& instance_json : instances_json) {
+    vsock_groups.emplace_back(
+        instance_json["connectivity"]["vsock"]["guest_group"].asString());
+  }
+
+  std::string args = fmt::format("{}", fmt::join(vsock_groups, ","));
+  return "--vsock_guest_group=" + args;
+}
+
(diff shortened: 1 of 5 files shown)


base/cvd/meson.build
```diff
@@ -79,6 +79,7 @@ cvd_sources = [
   'cuttlefish/host/commands/cvd/parser/instance/cf_security_configs.cpp',
   'cuttlefish/host/commands/cvd/parser/instance/cf_streaming_configs.cpp',
   'cuttlefish/host/commands/cvd/parser/instance/cf_vm_configs.cpp',
+  'cuttlefish/host/commands/cvd/parser/instance/cf_connectivity_configs.cpp',
```
jemoreira, line 82: These are in alphabetical order

base/cvd/cuttlefish/host/commands/cvd/parser/instance/cf_connectivity_configs.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include "common/libs/utils/result.h"
+
+namespace cuttlefish {
+
+std::string GenerateGroupFlag(const Json::Value& instances_json) {
+  std::vector<std::string> vsock_groups;
+
+  for (const auto& instance_json : instances_json) {
+    vsock_groups.emplace_back(
+        instance_json["connectivity"]["vsock"]["guest_group"].asString());
```
jemoreira, line 29: Will this abort if 'vsock' or 'guest_group' are not in the config file?

jemoreira (changes requested):
Please add a description of what that vsock_guest_group flag does to the commit message and the PR description.

jemoreira:
Please squash the two commits together.

base/cvd/meson.build
sferrogoo: Done.

base/cvd/cuttlefish/host/commands/cvd/parser/instance/cf_connectivity_configs.cpp
sferrogoo: I suspected the same - to address did some manual testing with valid and blank values, assigned to arbitrary VMs in the cluster.  The result is actually desirable here, I get a blank return value in case of unspecified, which also is valid in context of the group argument expects (blank value when not used.).

sferrogoo:
> Please squash the two commits together.

Done.

> Please add a description of what that vsock_guest_group flag does to the commit message and the PR description.

Done.

jemoreira (approved):
Approved it, but I'd suggest you wait for the change in aosp to be approved before submitting this one, in cases there are changes requested there.

sferrogoo:
> Approved it, but I'd suggest you wait for the change in aosp to be approved before submitting this one, in cases there are changes requested there.

Will do.  Looks like some kokoro failures came in so I'll push a few tweaks here as well.

Databean (approved):

PR #486, opened 2024-03-26
Set umask to 0 while creating directory at untar

k311093:
While creating a directory at untarring artifacts, there is a possibility to create a directory not with permissions 0774.

This causes _cvd-executor failing to create a lock file for an image and then fails to execute cvd.

Set umask 0 while creating directory at untarring to create directory with proper permissions.

### DIFF
--- a/frontend/src/host_orchestrator/orchestrator/userartifacts.go
+++ b/frontend/src/host_orchestrator/orchestrator/userartifacts.go
@@ -21,6 +21,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"syscall"
 
 	apiv1 "github.com/google/android-cuttlefish/frontend/src/liboperator/api/v1"
 	"github.com/google/android-cuttlefish/frontend/src/liboperator/operator"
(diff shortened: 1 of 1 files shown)


frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -168,6 +169,11 @@ func Untar(dst string, src string) error {
 		switch header.Typeflag {
 		case tar.TypeDir:
 			if _, err := os.Stat(target); err != nil {
+				oldmask := syscall.Umask(0)
+				defer func() {
```
0405ysj, line 173: Seems defer keyword executes when right before function ends.

0405ysj (changes requested):

frontend/src/host_orchestrator/orchestrator/userartifacts.go
k311093: Removed defer block and added code for restoring old mask

ikicha (approved):

0405ysj (approved):

ser-io:
let's squash all these commits into one. 

this way we avoid having the `defer` issue introduced in https://github.com/google/android-cuttlefish/pull/486/commits/0d2aa01b98eb48cbf6c87088045b8a4d39a899d4 (fixed later at https://github.com/google/android-cuttlefish/pull/486/commits/d9bd9a74fec137605a96720b282e0de2c20d57d5) in our history.

ser-io (changes requested):
let's squash all these commits into one for this PR.

this way we avoid having the defer issue introduced in https://github.com/google/android-cuttlefish/commit/0d2aa01b98eb48cbf6c87088045b8a4d39a899d4 (fixed later at https://github.com/google/android-cuttlefish/commit/d9bd9a74fec137605a96720b282e0de2c20d57d5) in our history.

k311093:
Squashed all commits into one

ser-io (approved):

PR #487, opened 2024-03-27
Extract .tar.gz user artifacts.

ser-io:


frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -78,6 +78,8 @@ func (c *Controller) AddRoutes(router *mux.Router) {
 		httpHandler(&listUploadDirectoriesHandler{c.UserArtifactsManager})).Methods("GET")
 	router.Handle("/userartifacts/{name}",
 		httpHandler(&createUpdateUserArtifactHandler{c.UserArtifactsManager})).Methods("PUT")
+	router.Handle("/userartifacts/{dir}/{name}/:extract",
```
jemoreira, line 81: why an extra endpoint for this?

can't it be an extra parameter in the upload operation?

jemoreira:
The kokoro failure this PR is hitting is a known issue, you can ignore it and bypass presubmits

frontend/src/host_orchestrator/orchestrator/controller.go
ser-io: the upload endpoint is a blocking/sync endpoint based on chunks. The extraction operation is a lengthy one for which we needed a new async endpoint.

PR #488, opened 2024-03-28
Extract .zip user artifacts.

ser-io:


frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -219,3 +224,38 @@ func Untar(dst string, src string) error {
...
 	}
 }
+
+func Unzip(dstDir string, src string) error {
+	r, err := zip.OpenReader(src)
+	if err != nil {
+		return err
+	}
+	defer r.Close()
+	extractAndCopy := func(dst string, src *zip.File) error {
```
jemoreira, line 234: `extractTo` or `extractToDestination` are better names, the fact that it uses `Copy` is an implementation detail.

jemoreira (approved):

PR #491, opened 2024-03-29
Added env DOCKER_RUNNING=true in Dockerfile

WeiChungHsu:
To added env DOCKER_RUNNING=true in Dockerfile, and RUN commands could use this var to detect if it is inside the docker container.


0405ysj (approved):

ikicha:
Isn't checking if "/.dockerenv" exist enough?

0405ysj:
> Isn't checking if "/.dockerenv" exist enough?

In his experiment, `/.dockerenv` wasn't recognized in `cuttlefish-base.postinst`. That lead `kvm` group being missed.

WeiChungHsu:
> > Isn't checking if "/.dockerenv" exist enough?
> 
> In his experiment, `/.dockerenv` wasn't recognized in `cuttlefish-base.postinst`. That lead `kvm` group being missed.

Thanks for helping me answer the question. And yes, the `/.dockerenv` is missing when running Dockerfile RUN commands. I search google, stockoverflow and docker sites, this file `/.dockerenv` is not persistent and a good way to check if you are inside docker.

On the other hand, there is no good way and no official suggested solution to detect if you are inside a docker.

docker/arm-server/Dockerfile
```diff
@@ -3,6 +3,9 @@
 # so it could execute CF instance with API in HO.
 FROM debian:stable-20211011 AS cuttlefish-arm64
 
+# to detect if it is running inside a docker
+ENV DOCKER_RUNNING=true
```
jemoreira, line 7: There is code in other places (example: cvd) that checks for the existence of the `.dockerenv` file. It's probably best to make sure it exists here than to come up with another mechanism, something like `RUN touch /.dockerenv` should work.

jemoreira (changes requested):

jemoreira (approved):

PR #497, opened 2024-04-04
Restore luci downloading to fetch_cvd

Databean:
This was originally added in PR #458 but clobbered in an import at PR #478.

Bug: b/322559348
Test: b/322559348#comment4


jemoreira (changes requested):
Please add an entry to `base/debian/changelog` for this feature. It probably needs to be a new UNRELEASED version.

Databean:
Updated the `changelog` file.

jemoreira (approved):

PR #498, opened 2024-04-04
Generate c++ documentation on GitHub Pages with doxygen

Databean:
Sample view: https://databean.github.io/android-cuttlefish/


ser-io (approved):

.github/workflows/documentation.yaml
```diff
@@ -0,0 +1,55 @@
+name: Documentation 
+
+on:
+  push:
```
jemoreira, line 4: Does "push" cover PR merges too?

jemoreira (approved):

.github/workflows/documentation.yaml
Databean: I think so. [Events that trigger workflows](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push) is vague, but [Understand GitHub Actions: Understanding the workflow file](https://docs.github.com/en/actions/learn-github-actions/understanding-github-actions#understanding-the-workflow-file) writes

> a workflow run is triggered every time someone pushes a change to the repository or merges a pull request.

PR #501, opened 2024-04-10
Always attempt to download full set of system imgs

cjreynol:
The current logic might only download the `system` and `product` images from the system build's `img.zip`.  We want the full set of system images if they exist.  That means we always expect a `system.img`, the rest are optional.

Since we always download the target files if there is a `--system_build` present we might as well skip the `img.zip` and extract them from the target files.

Removed `init_boot` from the list of system images to fetch.  See bug for context.

This also fixes that files from this particular codepath were not added to the `fetcher_config.json` and not de-sparsed.

Bug: 333616172
Test: cvd fetch --target_directory=/tmp/cvd/chd_target_files --default_build=aosp-main --system_build=11692693
Test: # slightly older aosp-main build
Test: cd /tmp/cvd/fetch_test
Test: shasum -b vbmeta_system.zip
Test: bsdtar -xf system/<target_files_zip>
Test: shasum -b IMAGES/vbmeta_system.zip
Test: # the hashes should match

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
+++ b/base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
@@ -497,55 +497,44 @@ Result<void> FetchTarget(BuildApi& build_api, LuciBuildApi& luci_build_api,
     DeAndroidSparse({target_files});
 
     if (flags.download_img_zip) {
-      std::string system_img_zip_name = GetBuildZipName(*builds.system, "img");
-      Result<std::string> system_img_zip_result = build_api.DownloadFile(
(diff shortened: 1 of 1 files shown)


cjreynol:
@Databean would you please review when you have time.  Thanks!

cjreynol:
@jemoreira brought up the concern that perhaps `target_files/IMAGES/system.img` might not exist.  If we keep downloading it from the `img.zip` we are at least staying consistent with out current behavior and not potentially introducing a breaking change.

Databean (approved):

cjreynol:
Added fallback logic for the [case pointed out by @jemoreira](https://github.com/google/android-cuttlefish/pull/501#issuecomment-2046265386).

PR #508, opened 2024-04-17
Run host orchestrator service behind nginx.

ser-io:
- Start using a reverse proxy in the host as new services running along with the host orchestrator service will be added, i.e a new service for logs retrieval.


frontend/debian/cuttlefish-orchestration.postinst
```diff
@@ -11,6 +11,11 @@ case "$1" in
         # The cvdnetwork group is created by cuttlefish-base
         usermod -a -G cvdnetwork,kvm _cvd-executor
     fi
+
+    # Reload nginx with the orchestration configuration
+    rm /etc/nginx/nginx.conf
+    ln -s /etc/cuttlefish-orchestration/nginx.conf /etc/nginx/nginx.conf
+    sudo nginx -s reload
```
jemoreira, line 18: Could you put this configuration file in `/etc/nginx/conf.d/host_orchestrator.conf` instead? That avoids having to override the entire nginx config and plays well with other services that could be installed in the host.
That would also allow you to enable and disable that configuration on host_orchestrator service start/stop, simply by changing the file extension and reloading nginx.

frontend/src/host_orchestrator/main.go
```diff
@@ -56,39 +54,13 @@ func startHttpServer(addr string, port int) error {
 	return http.ListenAndServe(fmt.Sprintf("%s:%d", addr, port), nil)
 }
 
-func startHttpsServer(addr string, port int, certPath string, keyPath string) error {
```
jemoreira, line 59: Can we keep this functionality? Being able to run https is not a bad idea, even if we don't use it at the moment.

jemoreira (changes requested):

frontend/src/host_orchestrator/main.go
ser-io: Done, also kept the logic of starters, we we can start multiple https servers.

frontend/debian/cuttlefish-orchestration.install
```diff
@@ -1 +1,3 @@
 host/packages/cuttlefish-orchestration/* /
+.sslcert/cert.pem /etc/cuttlefish-orchestration/ssl/cert/
+.sslcert/key.pem /etc/cuttlefish-orchestration/ssl/cert/
```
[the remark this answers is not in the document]
ser-io: nginx use the config parameters `ssl_certificate` and `ssl_certificate_key` for which we can set different values pointing to different paths.

frontend/debian/cuttlefish-orchestration.postinst
jemoreira: > different nginx packages can potentially install different defaults we won't be having control over

This would still be problem if the other package is installed **after** `cuttlefish-orchestration`, as that package would still override our config.

I'm thinking other users of the orchestration package could decide to install something like a /health endpoint or some other service in the host for which they want to use nginx but we're taking over nginx config completely.

> one server and different location entries

I don't know what this means, I'm not familiar enough with nginx configuration. Based on my previous argument, I would still prefer the conf.d approach.

ser-io: Agree, added our own specific config under `/etc/nginx/conf.d`.

docker/orchestration/run_services.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+service nginx start
```
0405ysj, line 3: Out of curiosity, what's the precondition to run `nginx` when not using Docker? Do we need any extra edit on here?

https://github.com/google/android-cuttlefish/blob/main/README.md#debian-packages

frontend/debian/cuttlefish-orchestration.cuttlefish-host_orchestrator.default
```diff
@@ -1,22 +1,13 @@
 # defaults for cuttlefish-host_orchestrator
 
-# The network address to listen on for both HTTP and HTTPS.
+# The network address to listen on.
 # Defaults to localhost only.
 orchestrator_listen_address=0.0.0.0
 #
 # The port on which the orchestrator server should listen on for plain HTTP.
-# Defaults to 2080.
+# Defaults to 2081.
```
0405ysj, line 8: Do we need this configuration? IIUC, `nginx` would wrap Host Orchestrator, and the actual ports should be accessed by users would set in `/etc/nginx/conf.d/cuttlefish-orchestration.conf`.

frontend/debian/cuttlefish-orchestration.cuttlefish-host_orchestrator.default
```diff
@@ -1,22 +1,13 @@
...
 # orchestrator_http_port=
 #
-# The port on which the orchestrator server should listen on for HTTPS.
-# The orchestrator won't server over HTTPS if the port is not set.
-orchestrator_https_port=2443
```
0405ysj, line 13: Random idea: Couldn't we get rid of `operator_https_port` as well?

frontend/src/host_orchestrator/main.go
```diff
@@ -101,9 +100,7 @@ func newOperatorProxy(port int) *httputil.ReverseProxy {
 }
 
 func main() {
-	httpPort := flag.Int("http_port", 2080, "Port to listen on for HTTP requests.")
-	httpsPort := flag.Int("https_port", -1, "Port to listen on for HTTPS requests.")
-	tlsCertDir := flag.String("tls_cert_dir", defaultTLSCertDir, "Directory with the TLS certificate.")
+	httpPort := flag.Int("http_port", 2081, "Port to listen on for HTTP requests.")
```
0405ysj, line 103: Do we really need to expose this port as numeric one? Regarding nginx configuration, I could find setting `proxy_pass` with `.sock` file. (e.g. `proxy_pass http://unix:/path/to/socket`)

frontend/debian/cuttlefish-orchestration.cuttlefish-host_orchestrator.default
ser-io: yes, we should be able to continue accessing the HO service directly.

frontend/debian/cuttlefish-orchestration.cuttlefish-host_orchestrator.default
ser-io: yes, `ssl` would be now a responsibility of `nginx`, this would be done in future PRs.

frontend/src/host_orchestrator/main.go
ser-io: I see your point, however in this case this port number doesn't need to be exposed externally, in fact it shouldn't be exposed. Exposing the port number is deliberate action, so a I think it's ok to continue using a port number.

frontend/debian/cuttlefish-orchestration.cuttlefish-host_orchestrator.default
0405ysj: SG.

frontend/debian/cuttlefish-orchestration.cuttlefish-host_orchestrator.default
0405ysj: Could you elaborate on here? What cases do we need to access HO directly via `2081`?

frontend/src/host_orchestrator/main.go
0405ysj: Acknowledged.

jemoreira (approved):

docker/orchestration/run_services.sh
0405ysj: While on chat, doing `apt install` for existing 3 CF debian packages would be okay. Before that, `apt install nginx` seems to be needed.

ser-io: when installing the packages directly in the host, `systemd` starts the service, there's no need to start it manually as within a container.

frontend/debian/cuttlefish-orchestration.cuttlefish-host_orchestrator.default
0405ysj: While chatting with him, heard answer that `2081` would be continuously used for debugging purposes.

0405ysj (approved):

PR #511, opened 2024-04-19
Make the instance database multi-process safe instead of just thread safe

jemoreira:


base/cvd/cuttlefish/host/commands/cvd/selector/creation_analyzer.cpp
```diff
@@ -85,23 +73,11 @@ static std::unordered_map<unsigned, InstanceLockFile> ConstructIdLockFileMap(
...
-
 Result<std::vector<PerInstanceInfo>>
 CreationAnalyzer::AnalyzeInstanceIdsInternal(
     const std::vector<unsigned>& requested_instance_ids) {
   CF_EXPECT(!requested_instance_ids.empty(),
             "Instance IDs were specified, so should be one or more.");
-  for (const auto id : requested_instance_ids) {
-    CF_EXPECT(IsIdAvailable(instance_database_, id),
-              "instance ID #" << id << " is requested but not available.");
-  }
```
Databean, line 103: Is this validation still happening somewhere else?

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/selector/creation_analyzer.cpp
jemoreira: Yes, in instace_database.cpp line 118. Creation of the new group+instances will fail because the requested instance id is in use.

PR #503, opened 2024-04-11
Persist cvd's instance database to a file

jemoreira:

PR #515, opened 2024-04-24
Add ADB port into ListCVDsResponse

0405ysj:
The purpose of this PR is to show adb port of CF instance in `GET /cvds`. For instance:
```
{"cvds":[{"group":"cvd","name":"1","build_source":{},"status":"Running","displays":["720 x 1280 ( 320 )"],"webrtc_device_id":"cvd-1","adb_port":"6520"}]}
```

FYI, this PR doesn't work in latest code, since `GET /cvds` in HO on top of docker instance is already broken after https://github.com/google/android-cuttlefish/pull/503. About that, I created a bug internally.

### DIFF
--- a/frontend/src/host_orchestrator/orchestrator/instancemanager.go
+++ b/frontend/src/host_orchestrator/orchestrator/instancemanager.go
@@ -115,12 +115,16 @@ type cvdGroup struct {
 	Instances []*cvdInstance `json:"instances"`
 }
 
-func (g *cvdGroup) toAPIObject() []*apiv1.CVD {
+func (g *cvdGroup) toAPIObject() ([]*apiv1.CVD, error) {
 	result := make([]*apiv1.CVD, len(g.Instances))
 	for i, item := range g.Instances {
-		result[i] = item.toAPIObject(g.Name)
+		instance, err := item.toAPIObject(g.Name)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get cvdInstance API Object")
+		}
+		result[i] = instance
 	}
-	return result
+	return result, nil
 }
 
 type cvdInstance struct {
@@ -129,9 +133,18 @@ type cvdInstance struct {
 	Displays       []string `json:"displays"`
 	InstanceDir    string   `json:"instance_dir"`
 	WebRTCDeviceID string   `json:"webrtc_device_id"`
+	ADBSerial	   string   `json:"adb_serial"`
 }
 
-func (i *cvdInstance) toAPIObject(group string) *apiv1.CVD {
+func (i *cvdInstance) toAPIObject(group string) (*apiv1.CVD, error) {
+	_, port, err := net.SplitHostPort(i.ADBSerial)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse ADB serial: %w", err)
+	}
+	adbPort, err := strconv.Atoi(port)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert ADB port to integer: %w", err)
+	}
 	return &apiv1.CVD{
 		Group: group,
 		Name:  i.InstanceName,
@@ -140,7 +153,8 @@ func (i *cvdInstance) toAPIObject(group string) *apiv1.CVD {
 		Status:         i.Status,
 		Displays:       i.Displays,
 		WebRTCDeviceID: i.WebRTCDeviceID,
-	}
+		ADBPort:        adbPort,
+	}, nil
 }
 
 func cvdFleet(ctx cvd.CVDExecContext) (*cvdFleetOutput, error) {
--- a/frontend/src/liboperator/api/v1/messages.go
+++ b/frontend/src/liboperator/api/v1/messages.go
@@ -187,6 +187,8 @@ type CVD struct {
 	Displays []string `json:"displays"`
 	// [Output Only]
 	WebRTCDeviceID string `json:"webrtc_device_id"`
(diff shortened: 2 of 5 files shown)


frontend/src/liboperator/api/v1/messages.go
```diff
@@ -187,6 +187,8 @@ type CVD struct {
 	Displays []string `json:"displays"`
 	// [Output Only]
 	WebRTCDeviceID string `json:"webrtc_device_id"`
+	// [Output Only]
+	ADBPort string `json:"adb_port"`
```
jemoreira, line 191: this should be `int`

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -129,9 +133,17 @@ type cvdInstance struct {
...
 
-func (i *cvdInstance) toAPIObject(group string) *apiv1.CVD {
+func (i *cvdInstance) toAPIObject(group string) (*apiv1.CVD, error) {
+	host, port, err := net.SplitHostPort(i.ADBSerial)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse ADB serial: %w", err)
+	}
+	if host != "0.0.0.0" {
+		return nil, fmt.Errorf("invalid host address. (Expect: 0.0.0.0, Actual: %s)", host)
+	}
```
jemoreira, line 146: Don't assert on the IP. This code doesn't seem to use it and it could start failing in the future if it changes to the external IP address, IPv6 or something else we haven't thought about.

jemoreira (changes requested):

frontend/src/liboperator/api/v1/messages.go
0405ysj: Done.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
0405ysj: Removed the assertion. Done.

k311093 (approved):

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -129,9 +133,18 @@ type cvdInstance struct {
 	Displays       []string `json:"displays"`
 	InstanceDir    string   `json:"instance_dir"`
 	WebRTCDeviceID string   `json:"webrtc_device_id"`
+	ADBSerial	   string   `json:"adb_serial"`
 }
 
-func (i *cvdInstance) toAPIObject(group string) *apiv1.CVD {
+func (i *cvdInstance) toAPIObject(group string) (*apiv1.CVD, error) {
+	_, port, err := net.SplitHostPort(i.ADBSerial)
```
ser-io, line 140: let's put the `ADBSerial` parsing logic into its own function.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -129,9 +133,18 @@ type cvdInstance struct {
 	Displays       []string `json:"displays"`
 	InstanceDir    string   `json:"instance_dir"`
 	WebRTCDeviceID string   `json:"webrtc_device_id"`
+	ADBSerial	   string   `json:"adb_serial"`
```
ser-io, line 136: nit: format the code with `go fmt`

frontend/src/liboperator/api/v1/messages.go
```diff
@@ -187,6 +187,8 @@ type CVD struct {
 	Displays []string `json:"displays"`
 	// [Output Only]
 	WebRTCDeviceID string `json:"webrtc_device_id"`
+	// [Output Only]
+	ADBPort int `json:"adb_port"`
```
ser-io, line 191: why not return the original `ADB serial` and get rid of the parsing logic and error handling here?

0405ysj: What's the reason why we should move parsing code to the client side? Is making same with `cvd fleet` the purpose of `GET /cvds`?

ser-io: not exactly the same but as closer as possible. in this case, if `cvd fleet` returns empty or misspelled  `ADBSerial` values, the host orchestrator should provide the same values to clients rather than failing. clients can  then raise more relevant errors to the final users.

we should see the host orchestrator as an http wrapper of `cvd` given clients the same values presented in `cvd` outputs.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
0405ysj: Removed parsing logic regarding `ADBSerial` in https://github.com/google/android-cuttlefish/pull/522, since HO won't handle it.

frontend/src/host_orchestrator/orchestrator/instancemanager.go
0405ysj: Done in https://github.com/google/android-cuttlefish/pull/522.

frontend/src/liboperator/api/v1/messages.go
0405ysj: Okay. Then I'll just pass `ADBSerial` value in HO, and handle it in client side like `cvdr` binary. Done in https://github.com/google/android-cuttlefish/pull/522, by changing the field from `ADBPort` to `ADBSerial`.

PR #523, opened 2024-04-29
Change mode when creating a directory

0405ysj:
Recently I faced at the permission related issue when doing `POST /cvds` after `GET /cvds`.

And found the root cause `_cutf-operator` trying to change mode of `/tmp/cvd` directory, which is owned by `_cvd-executor`.

I can resolve this issue with this PR, but I still have doubts on:
- Why can't `mkdir` function set the mode?
- Why doesn't the group of `_cvd-executor` set as `cvdnetwork`?
- Why are the users different between `GET /cvds` and `POST /cvds`?

### DIFF
--- a/base/cvd/cuttlefish/common/libs/utils/files.cpp
+++ b/base/cvd/cuttlefish/common/libs/utils/files.cpp
@@ -129,8 +129,11 @@ Result<void> EnsureDirectoryExists(const std::string& directory_path,
                                                      << strerror(errno));
   }
 
+  CF_EXPECTF(chmod(directory_path.c_str(), mode) == 0,
+            "Failed to set permission on {}: {}", directory_path, strerror(errno));
+
   if (group_name != "") {
-    ChangeGroup(directory_path, group_name);
+    CF_EXPECT(ChangeGroup(directory_path, group_name));
   }
 
   return {};
--- a/base/cvd/cuttlefish/host/commands/cvd/main.cc
+++ b/base/cvd/cuttlefish/host/commands/cvd/main.cc
@@ -102,13 +102,6 @@ Result<void> KillOldServer() {
 Result<void> EnsureCvdDirectoriesExist() {
   // This is accessed by all users.
   CF_EXPECT(EnsureDirectoryExists(CvdDir(), 0777));
-  // The process' umask likely prevented the right permissions to be applied
-  // to the cvd directory. Changing the umask before this call would only set
-  // the permissions for a newly created directory, while this approach ensures
-  // the permissions are correct even if the directory already exists.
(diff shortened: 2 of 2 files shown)


ser-io:
faced the same issue, fix is in https://github.com/google/android-cuttlefish/pull/520/commits/80a612e065568fdeff99ccd6538a1c6dcf290d64, when using `fetch_cvd` we were not using the `_cvd-executor` as the effective user, hence the error.

jemoreira:
> Why can't mkdir function set the mode?

it does, but according to the process' `umask`, if you want something more permissive you have to use `chmod` or change the `umask`.

> Why doesn't the group of _cvd-executor set as cvdnetwork?

That's a good idea, no reason AFAIK

> Why are the users different between GET /cvds and POST /cvds?

@ser-io

base/cvd/cuttlefish/common/libs/utils/files.cpp
```diff
@@ -129,8 +129,11 @@ Result<void> EnsureDirectoryExists(const std::string& directory_path,
                                                      << strerror(errno));
   }
 
+  CF_EXPECTF(chmod(directory_path.c_str(), mode) == 0,
+            "Failed to set permission on {}: {}", directory_path, strerror(errno));
+
   if (group_name != "") {
-    ChangeGroup(directory_path, group_name);
+    CF_EXPECT(ChangeGroup(directory_path, group_name));
```
jemoreira, line 136: Please add these changes in aosp too.

jemoreira (approved):

ser-io:
> Why doesn't the group of _cvd-executor set as cvdnetwork?
_cvd-executor is part of the cvdnetwork group: https://github.com/ser-io/android-cuttlefish/blob/main/frontend/debian/cuttlefish-orchestration.postinst#L12

> Why are the users different between GET /cvds and POST /cvds?
we are not using the configured cvd user when calling `fetch_cvd` hence the difference, sent out a PR fixing this: https://github.com/google/android-cuttlefish/pull/524

base/cvd/cuttlefish/common/libs/utils/files.cpp
0405ysj: I'll do that. Thanks for the mentioning :)

0405ysj:
> _cvd-executor is part of the cvdnetwork group: https://github.com/ser-io/android-cuttlefish/blob/main/frontend/debian/cuttlefish-orchestration.postinst#L12

I think I missed to say a part of contents I found. I was suspicious about why the group of `/tmp/cvd` wasn't set as `cvdnetwork`, when that dir is created by `_cvd-executor`.

PR #526, opened 2024-04-30
Remove unused known_paths.h

cjreynol:
The matching .cpp file does not appear to be copied over, and it is unused.

Test: meson compile


cjreynol:
@jemoreira review when you have time, please :)

jemoreira (approved):

PR #529, opened 2024-04-30
Print new group's status on cvd start

jemoreira:


jemoreira:
Need to land something else first

PR #536, opened 2024-05-03
Build `cvd` with bazel

jemoreira:
And allocd too.

Dependencies are taken from bazel registry when possible.

Protobuf files are now in their original locations since bazel doesn't have the same restrictions as meson.


ser-io (approved):

base/cvd/.gitignore
```diff
@@ -1,2 +1 @@
-html
-latex
```
Databean, line 2: These are from the doxygen output

Databean (approved):

PR #538, opened 2024-05-03
Factor out a helper to make download steps clearer

cjreynol:
The initial step of needing to retrieve the download URL from AB and then using that to do the actual download has left me briefly confused multiple times.

Test: cvd fetch --target_directory=/tmp/cvd/fetch_test --default_build=aosp-main
Test: # verify all typical artifacts are still downloaded

### DIFF
--- a/base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
+++ b/base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
@@ -336,9 +336,8 @@ Result<std::unordered_set<std::string>> BuildApi::Artifacts(
   return CF_EXPECT(std::move(res));
 }
 
-Result<void> BuildApi::ArtifactToFile(const DeviceBuild& build,
-                                      const std::string& artifact,
-                                      const std::string& path) {
+Result<std::string> BuildApi::GetArtifactDownloadUrl(
+    const DeviceBuild& build, const std::string& artifact) {
   std::string download_url_endpoint =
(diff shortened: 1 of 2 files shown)


base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
```diff
@@ -359,7 +358,13 @@ Result<void> BuildApi::ArtifactToFile(const DeviceBuild& build,
...
             "URL endpoint did not have json path: " << json);
-  std::string url = json["signedUrl"].asString();
+  return json["signedUrl"].asString();
+}
+
+Result<void> BuildApi::ArtifactToFile(const DeviceBuild& build,
+                                      const std::string& artifact,
+                                      const std::string& path) {
+  const auto url = CF_EXPECT(GetArtifactDownloadUrl(build, artifact));
   CF_EXPECT(CF_EXPECT(http_client->DownloadToFile(url, path)).HttpSuccess());
```
Databean, line 368: not part of this CL, but nested `CF_EXPECT` calls will generate messy errors as the macro expansion of the inner `CF_EXPECT` will become part of the "failing expression" text of the outer `CF_EXPECT`.

Databean (approved):

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
cjreynol: Created #539 to address this

PR #540, opened 2024-05-03
Fallback to $HOME and current directory

jemoreira:
when $ANDROID_HOST_OUT is not defined

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/common_utils.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/common_utils.cpp
@@ -30,6 +30,31 @@
 
 namespace cuttlefish {
 
+namespace {
+/*
+ * Most branches read the kAndroidHostOut environment variable, but a few read
+ * kAndroidSoongHostOut instead. Cvd will set both variables for the subtools
+ * to the first value it finds:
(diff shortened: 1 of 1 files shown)


base/cvd/cuttlefish/host/commands/cvd/common_utils.cpp
```diff
@@ -30,6 +30,32 @@
...
+    return envs.at(kAndroidHostOut);
+  }
+  if (Contains(envs, kAndroidSoongHostOut)) {
+    return envs.at(kAndroidSoongHostOut);
+  }
+  if (Contains(envs, "HOME") &&
+      FileExists(envs.at("HOME") + "/bin/cvd_internal_start")) {
+    return envs.at("HOME");
+  }
+  std::unique_ptr<char, void (*)(void*)> cwd(getcwd(nullptr, 0), &free);
```
cjreynol, line 54: `files.h::CurrentDirectory()` already implements this the same way (with a little logging if there is a failure)

cjreynol (approved):

base/cvd/cuttlefish/host/commands/cvd/common_utils.cpp
jemoreira: Done

PR #542, opened 2024-05-04
Add local fetch caching to `cvd fetch`

cjreynol:
IFF the `--target_directory` and `/tmp/cvd/<user_id>/cache` can be
hardlinked, repeated `cvd fetch` commands on the same build+target will
reference cached files and hardlink the files to the target directory
for use.

The extract and renaming steps still occur every time, this is only
caching for downloads against AB.

This new behavior is on by default (`--enable_caching=true`).

Test: cvd fetch --target_directory=/tmp/cvd/fetch_test
--default_build=11801630
Test: ls
/tmp/cvd/<user_id>/cache/11801630/aosp_cf_x86_64_phone-trunk_staging-userdebug/
Test: # repeat fetch command and review `fetch.log` to see the archive
downloads do not happen


cjreynol:
The first commit with `files.h/cpp` helper functions will be committed in aosp/3077287 and copied over to keep things in sync.

cjreynol:
@jemoreira please review when you have a chance :)

base/cvd/cuttlefish/host/libs/web/android_build_api.h
```diff
@@ -71,7 +71,7 @@ class BuildApi {
  public:
   BuildApi() = delete;
   BuildApi(BuildApi&&) = default;
```
Databean, line 73: FYI this can now have some weird behaviors, e.g. creating a `BuildApi` from a `CachingBuildApi`

[Compiler explorer](https://godbolt.org/z/asbY83e61)

base/cvd/cuttlefish/host/libs/web/caching_build_api.cpp
```diff
@@ -0,0 +1,141 @@
...
+    std::unique_ptr<CredentialSource> credential_source, std::string api_key,
+    const std::chrono::seconds retry_period, std::string api_base_url,
+    const bool enable_caching, const std::string cache_base_path)
+    : BuildApi(std::move(http_client), std::move(inner_http_client),
+               std::move(credential_source), std::move(api_key), retry_period,
+               std::move(api_base_url)),
+      enable_caching_(enable_caching),
+      cache_base_path_(std::move(cache_base_path)) {
+  // the cache directory must exist for the CanCache check to be true
+  enable_caching_ = enable_caching_ && EnsureCacheDirectory(cache_base_path_);
```
Databean, line 88: An alternative to accepting `enable_caching` as an argument and member variable here is to create a public function
```
std::unique_ptr<BuildApi> CreateBuildApi(...);
```
that internally creates either kind.

base/cvd/cuttlefish/common/libs/utils/files.cpp
```diff
@@ -81,6 +81,30 @@ bool FileExists(const std::string& path, bool follow_symlinks) {
...
+Result<dev_t> FileDeviceId(const std::string& path) {
+  struct stat out;
+  CF_EXPECTF(stat(path.c_str(), &out) == 0,
+             "stat() failed trying to retrive device ID information for \"{}\" "
+             "with error: {}",
+             path, strerror(errno));
+  return out.st_dev;
+}
+
+Result<bool> CanHardLink(const std::string& path1, const std::string& path2) {
```
jemoreira, line 93: Rename the parameters to make it clear one is the target and the other one the parent directory of the link that can be created.

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
```diff
@@ -314,7 +316,7 @@ Result<std::vector<std::string>> FetchSystemImgZipImages(
       system_img_zip_name);
 }
 
-Result<BuildApi> GetBuildApi(const BuildApiFlags& flags) {
+Result<CachingBuildApi> GetBuildApi(const BuildApiFlags& flags) {
```
jemoreira, line 319: nit: It seems you made `CachingBuildApi` inherit from `BuildApi`, so here you could return `std::unique_ptr<BuildApi>` and use the high level interface everywhere else.

base/cvd/cuttlefish/host/libs/web/caching_build_api.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+bool CachingBuildApi::CanCache(const std::string& target_directory) {
+  return enable_caching_ &&
+         CanHardLink(target_directory, cache_base_path_).ok();
+}
+
+Result<std::string> CachingBuildApi::DownloadFile(
+    const Build& build, const std::string& target_directory,
+    const std::string& artifact_name) {
+  if (!CanCache(target_directory)) {
```
jemoreira, line 99: Why not download to cache and copy to `target_directory` when it can't hardlink? In fact, the artifact may already be present in the cache, copying is probably faster than downloading again.

The tradeoff here is using {filesize} more space to gain speed in future downloads.

base/cvd/cuttlefish/host/libs/web/caching_build_api.cpp
```diff
@@ -0,0 +1,141 @@
...
+               std::move(api_base_url)),
+      enable_caching_(enable_caching),
+      cache_base_path_(std::move(cache_base_path)) {
+  // the cache directory must exist for the CanCache check to be true
+  enable_caching_ = enable_caching_ && EnsureCacheDirectory(cache_base_path_);
+};
+
+bool CachingBuildApi::CanCache(const std::string& target_directory) {
+  return enable_caching_ &&
+         CanHardLink(target_directory, cache_base_path_).ok();
```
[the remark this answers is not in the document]
cjreynol: Fixed, went with the `Result<bool>` return and `CF_EXPECT(CanCache(` everywhere

base/cvd/cuttlefish/host/libs/web/android_build_api.h
cjreynol: Thanks for the catch!  Deleted the move constructors for both.  They were unused anyways.

base/cvd/cuttlefish/host/libs/web/caching_build_api.cpp
cjreynol: Good idea, I like that better than the internal member variable.  Done.

base/cvd/cuttlefish/common/libs/utils/files.cpp
```diff
@@ -81,6 +81,30 @@ bool FileExists(const std::string& path, bool follow_symlinks) {
...
+  return CF_EXPECT(FileDeviceId(path1)) == CF_EXPECT(FileDeviceId(path2));
+}
+
+Result<std::string> CreateHardLink(const std::string& target,
+                                   const std::string& hardlink) {
+  if (!FileExists(hardlink)) {
```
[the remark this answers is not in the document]
cjreynol: Done, and added https://android-review.git.corp.google.com/c/device/google/cuttlefish/+/3077287/comment/5a551381_a4058ef4/
to the AOSP CL.

base/cvd/cuttlefish/host/libs/web/caching_build_api.cpp
```diff
@@ -0,0 +1,141 @@
...
+  std::string build_cache;
+  std::string target_artifact;
+  std::string cache_artifact;
+  std::string target_backup_artifact;
+  std::string cache_backup_artifact;
+};
+
+bool EnsureCacheDirectory(const std::string& cache_base_path) {
+  Result<void> result = EnsureDirectoryExists(cache_base_path);
+  if (!result.ok()) {
```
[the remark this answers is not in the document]
cjreynol: Done

base/cvd/cuttlefish/common/libs/utils/files.cpp
cjreynol: This helper is only checking if the two paths can be hardlinked, there are no particular semantics for which is which.  The parameters are distinct for `CreateHardLink`, but here we are just calling `stat` and checking both have the same device ids.  They could both be any type of file.

base/cvd/cuttlefish/common/libs/utils/files.cpp
```diff
@@ -81,6 +81,30 @@ bool FileExists(const std::string& path, bool follow_symlinks) {
...
+    CF_EXPECTF(link(target.c_str(), hardlink.c_str()) == 0,
+               "link() failed trying to create hardlink from \"{}\" to \"{}\" "
+               "with error: {}",
+               target, hardlink, strerror(errno));
+  }
+  return hardlink;
```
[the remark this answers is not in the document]
cjreynol: Done.

I added logic to check if the files are already hardlinked.  I also added the ability to force an overwriting the existing file.

We cannot simply error if the file already exists if we want to match existing `cvd fetch` behavior, because right now it overwrites existing files.

We also cannot rely on surfacing an error if the file already exists but is not a hardlink of the file in the cache because some files (like [`misc_info.txt`](https://github.com/google/android-cuttlefish/blob/61c7bc37409dbbbf196599c8901d27483fe1f8df/base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc#L463)) have cases where the file is allowed to be missing and anywhere in the download is allowed to fail.  I discovered while testing the logic that verified the existing file was hardlinked that the error would be silently fail and then not get an entry in the `fetcher_config.json`. The overwriting matches the current behavior.

base/cvd/cuttlefish/host/libs/web/caching_build_api.cpp
cjreynol: I am not necessarily against it, but I would want to have something concrete in place around fetched artifact cleanup before we increase the size impact beyond what we already are with this hardlinking version.  I will add a comment with this idea to b/296620322 with this idea.  

For now I think this version works and handles the typical developer `cvd load` case, so it is still an improvement.

base/cvd/cuttlefish/common/libs/utils/files.cpp
```diff
@@ -81,6 +81,60 @@ bool FileExists(const std::string& path, bool follow_symlinks) {
...
+  CF_EXPECTF(
+      stat(path.c_str(), &out) == 0,
+      "stat() failed trying to retrieve inode num information for \"{}\" "
+      "with error: {}",
+      path, strerror(errno));
+  return out.st_ino;
+}
+
+Result<bool> AreHardLinked(const std::string& path1, const std::string& path2) {
+  return CF_EXPECT(FileInodeNumber(path1)) == CF_EXPECT(FileInodeNumber(path2));
```
Databean, line 109: inode numbers are only guaranteed to be unique within a single filesystem, so it's the same file if both `st_dev` and `st_ino` match.

(there might also be other conditions for it being the same file like bind mounting, not sure)

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
```diff
@@ -332,9 +334,11 @@ Result<BuildApi> GetBuildApi(const BuildApiFlags& flags) {
...
           flags.credential_flags.service_account_filepath));
 
-  return BuildApi(std::move(retrying_http_client), std::move(curl),
-                  std::move(credential_source), flags.api_key,
-                  flags.wait_retry_period, flags.api_base_url);
+  const auto cache_base_path = PerUserDir() + "/cache";
+  return CreateBuildApi(std::move(retrying_http_client), std::move(curl),
+                        std::move(credential_source), std::move(flags.api_key),
+                        flags.wait_retry_period, std::move(flags.api_base_url),
+                        flags.enable_caching, cache_base_path);
```
Databean, line 341: nit: `std::move(cache_base_path)`

Databean (approved):

base/cvd/cuttlefish/common/libs/utils/files.cpp
cjreynol: You are right, we need the combination of both.  Expanded the conditional

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
cjreynol: Done.

PR #549, opened 2024-05-08
Add documentation about building docker image

0405ysj:


README.md
```diff
@@ -50,3 +50,32 @@ The following script can be used to build a host image for Google Compute Engine
...
+
+## Docker
+
+We also provide the docker image which installed cuttlefish debian packages
+inside; including `cuttlefish-base`, `cuttlefish-user`, and
+`cuttlefish-orchestration`.
+
+### Build docker image manually
+
+Please run below command to build manually.
```
ikicha, line 62: is it worth mentioning that it should run in arm machine?

0405ysj: Yeah, I think so. Users can know with reading path of shell script, but not directly.

Additionally, what about this?
- Mentioning it should work in arm64.
- Changing the dir name from `docker/arm-server` to `docker/linux-env-builder-docker` and mentioning it should work in arm64 and x86_64 both. (PLMK if you have better naming suggestion.)

README.md
```diff
@@ -56,24 +56,24 @@ to obtain the script.
...
+Currently it's available for x86_64 and ARM64 architecture.
 
 ### Build docker image manually
 
 Please run below command to build manually.
 
 ```bash
 cd /path/to/android-cuttlefish
-cd docker/arm-server
+cd docker/linux-infrastructure-docker
```
ser-io, line 67: the `docker` suffix here is redundant as we are within the docker directory.

why the name `linux-infrastructure`? from my understanding, this directory seems related to build a docker image ready to manage cuttlefish instances through out the host orchestrator, wondering whether the full path name: `android-cuttlefish/docker/orchestration` makes sense here?

0405ysj: I actually waited for the better naming, as I wrote in the other comment. There's no huge meaning for setting `linux-infrastructure`, the primary purpose is to avoid dir name relying on specific architecture.(e.g. arm)

0405ysj: Just for clarification, what's your suggestion exactly? Is it changing directory name only, or the docker image name as well? For dir name, I 100% agree. For the image name, I have minor concern that it may cause little confusion to developers to distinguish between host orchestrator, cloud orchestrator, and the docker named with `cuttlefish-orchestration` or similar.

OTOH, I don't hate your suggestion since it sounds better than my naming at least I think. I'll change that in advance.

0405ysj: Changed to the `orchestration`.

ikicha (approved):
Please update our internal guide accordingly because it changes docker image name

Nirzak:
The is not issue reporing section, so I am reporting in here. The docker build option is giving error during building docker images on arm64 instance. here what I did:

ran ./build.sh under arm-server section.

Here's the errors what came

mk-build-deps: Unable to install cuttlefish-common-build-deps-depends at /usr/bin/mk-build-deps line 459.
mk-build-deps: Unable to install all build-dep package

=> ERROR [11/17] RUN apt install -y -f     /root/cuttlefish-base_*.deb     /root/cuttlefish-user_*.deb     /root/cuttlefish-orch  1.0s

0405ysj:
Hello Nirzak,

Sorry, but I need more information to help the issue you've faced at.

- Machine information ran `./build.sh` in `docker/orchestration`(`docker/arm-server` in the past) : Does the OS equal to Linux? Is the arch in x86_64 or arm64?
- Which version or commit of this github repo you've used?
- What was the result for below? Would debian packages including `cuttlefish-base` be made well?
```bash
cd docker/debs-builder-docker
./main.sh
```

Nirzak:
May be I found the error. My arch is arm64 linux. so it actually giving the following error
```
Correcting dependencies...Starting pkgProblemResolver with broken count: 1
Starting 2 pkgProblemResolver with broken count: 1
Investigating (0) cuttlefish-common-build-deps-depends:arm64 < 0.9.29 @iU K Nb Ib >
Broken cuttlefish-common-build-deps-depends:arm64 Depends on bazel:arm64 < none @un H >
  Removing cuttlefish-common-build-deps-depends:arm64 because I can't find bazel:arm64
Done
```
May be this one is leading to the other errors. I have checked the codes and here is what I have found

`deb [arch=amd64 signed-by=/usr/share/keyrings/bazel-archive-keyring.gpg] https://storage.googleapis.com/bazel-apt stable jdk1.8"` on build-hostpkg.sh file

isn't here the arch should be aarch64 for arm64 architecture?

Nirzak:
Tried arm64 anf aarch64 on the arch section of bazel repo. but seems like it couldn't find bazel arm64 deb package from repo.

but they actually supports arm64 linux here https://github.com/bazelbuild/bazel/releases/download/7.1.2/bazel-7.1.2-linux-arm64

Nirzak:
I have tried to include the following lines in the build-hostpkg.sh file to make it compatible with arm64. but looks like they are not executing or something. I can't see any info about the in the log.

```
    curl -o bazel https://releases.bazel.build/7.1.2/release/bazel-7.1.2-linux-arm64
    chmod +x bazel
    sudo mv bazel /bin
```

0405ysj:
Thanks for the investigation! We're currently looking at the same page, and checked bazel is not installable by apt in aarch64/arm64 architecture.

@jemoreira Could you look at here? Currently `cuttlefish-base` is not built in arm64, because of lack of `bazel`.

ikicha:
@ser-io can we use bazelisk? https://bazel.build/install/bazelisk

Nirzak:
The problem is bazelisk is also not available as a APT package. whenever we are trying to install cuttlefish-common-base deb it by default is checking the bazel in the installed packages. so installing by other means will not reflect if we don't ignore the check from dependency checks or cuttlefish.

Nirzak:
Please allow us to create issues on this repository. Thanks

I have set up cuttlefish successfully using docker on my amd machine after getting unsuccessful attempts on arm. the docker image was built without any errors. and also the container started without any error. but the problem is the web GUI isn't showing any devices. but almost 40-50 ports are open for devices also I can see network interfaces for them inside the container.. then what is this issue? isn't there should be a device running already inside the container? or do we have to manually download the Android image and cvd package and then run it inside the docker? @0405ysj @ikicha 
Thanks

image

PR #525, opened 2024-04-30
Validate cvd configs with a proto definition

Databean:
`cf_flags_validator.cpp` included functionality to validate field types inside a json structure based on a rigid schema. This commit introduces the capability to pull the schema from a protocol buffer definition.

Proto3 includes a canonical json translation:

https://protobuf.dev/programming-guides/proto3/#json

This is validated by replacing a single nested json structure with proto validation. It produces errors when the field types are wrong.

Bug: b/337089452

PR #550, opened 2024-05-08
Use protocol buffers to interpret `cvd start` / `cvd load` config files

Databean:
This is a step towards accepting encoded protocol buffers in addition to json configuration files for `cvd start` / `cvd load`.

Outstanding issues are implementing merge logic and override logic matching what we have with json. These will likely require proto reflection to implement.

Supersedes PR #525.

Bug: b/337089452


base/cvd/cuttlefish/host/commands/cvd/parser/instance/cf_boot_configs.cpp
```diff
@@ -18,33 +18,45 @@
...
+static bool EnableBootAnimation(const Instance& instance) {
+  const auto& boot = instance.boot();
+  if (boot.has_enable_bootanimation()) {
+    return instance.boot().enable_bootanimation();
+  } else {
+    return CF_DEFAULTS_ENABLE_BOOTANIMATION;
+  }
+}
+
+static Result<std::string> BtCfg(const Instance& instance) {
```
jemoreira, line 41: nit: `BtCfg` made think of bluetooth instead of boot. Rename as `BootCfg` or just `Cfg`

jemoreira (approved):

base/cvd/cuttlefish/host/commands/cvd/parser/instance/cf_boot_configs.cpp
Databean: Done.

Databean:
Thanks for reviewing!

PR #551, opened 2024-05-08
Add explicit sign conversions to many integer comparisons

Databean:
This fixes some of the "signed compared to unsigned" errors that scroll by when building with bazel.


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #552, opened 2024-05-09
Include `compile_commands.json` generator

Databean:
Running `bazel run @hedron_compile_commands//:refresh_all` will generate a `compile_commands.json` at the root of the `WORKSPACE` that can be used by text editors for autocompletion. This has to be re-generated manually by re-running the command.

`compile_commands.json` is used by text editors / LSP servers for autocompletion.

Bug: b/339691377


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #553, opened 2024-05-09
Install bazel with bazelisk in arm64

jemoreira:


jemoreira:
So far I was only able to test that bazel is installed properly, but couldn't build because [bazel can't connect to the network over IPv6](https://github.com/bazelbuild/bazel/issues/2486#issuecomment-1870698756)

ikicha (approved):

ser-io (approved):

tools/installbazel/main.sh
```diff
@@ -18,7 +18,7 @@
 
 set -e
 
-function install_bazel() {
+function install_bazel_x86_64() {
```
ikicha, line 21: Can we use bazelisk for every arch?

tools/installbazel/main.sh
```diff
@@ -28,4 +28,16 @@ function install_bazel() {
...
 }
 
-install_bazel
+function install_bazel_aarch64() {
+  BAZELISK_VERSION=v1.19.0
+  apt install wget
+  tmpdir="$(mktemp -t -d bazel_installer_XXXXXX)"
+  trap "rm -rf $tmpdir" EXIT
+  pushd "${tmpdir}"
+  wget "https://github.com/bazelbuild/bazelisk/releases/download/${BAZELISK_VERSION}/bazelisk-linux-arm64"
```
ikicha, line 37: we may want to check checksum for the binary

ikicha, line 37: or we may install it via npm because npm is anyway used for web ui build.

jemoreira: I couldn't find a checksum in https://github.com/bazelbuild/bazelisk/releases/tag/v1.19.0.

tools/installbazel/main.sh
0405ysj: I think we can use bazelisk for x86_64 as well. You could see `bazelisk-linux-amd64` [here](https://github.com/bazelbuild/bazelisk/releases).


Should the version of `bazel` sync with any other places beyond this github repo? \
Unless, using `bazelisk` for `arm64` only would diverge the version used for `bazel`.

But I prefer to do this in the following PR, not this one. Since the primary goal is fixing build issue for ARM64 first.

PR #558, opened 2024-05-13
Rename `super` and `retry_period` config proto fields, keeping the json name

Databean:
The json name is part of our public interface, but the proto name conflicts with a language keyword in one of the supported proto languages.


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #559, opened 2024-05-14
Deserialize the flag values validated in `graphics_config_test.cc`

Databean:
The protobuf binary format is not guaranteed to be stable, only the deserialization is guaranteed to be stable.

In practice this caused a presubmit failure: http://shortn/_WOVqWc1jIc

Test: bazel run '...'


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #564, opened 2024-05-17
installbazel.sh Install symlink to /usr/bin/bazel in arm64

awheeler-mmi:
We are seeing build errors on our ARM64 Ubuntu 24.04 servers, if they have either the default bazel installed, or no bazel installed.    This commit resolves the issue.  I believe adding `--prepend-path /usr/local/bin` to the `debuild` commands would also work.

### DIFF
--- a/base/debian/rules
+++ b/base/debian/rules
@@ -1,7 +1,7 @@
 #!/usr/bin/make -f
 
 # Start build by executing:
-# $ debuild --no-tgz-check -us -uc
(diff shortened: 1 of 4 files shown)


jemoreira:
>  adding --prepend-path /usr/local/bin to the debuild commands would also work.

Could you do this instead? It's cleaner than changing the file under `/usr/bin`. This script may run in many environments including workstations where it's not ok to just replace the bazel binary.

PR #568, opened 2024-05-23
Change base stable version to 0.9.28

WeiChungHsu:
Change base stable version to 0.9.28


Databean:
There are 41 commits here but only the last one seems to contribute to the `Files changed`, can this be changed to just the last commit?

PR #576, opened 2024-06-03
Start systemd-journal-gatewayd.

ser-io:


frontend/debian/cuttlefish-orchestration.cuttlefish-host_orchestrator.init
```diff
@@ -90,6 +90,9 @@ start_orchestrator() {
 }
 
 start() {
+  # Make sure systemd-journal-gatewayd is running
+  service systemd-journal-gatewayd start
```
Databean, line 94: `cuttlefish-host_orchestrator` is a systemd service itself, right?

Should this [package in](https://wiki.debian.org/Teams/pkg-systemd/Packaging) a [systemd service file](https://wiki.debian.org/systemd/Services)?

Based on [unit file rules](https://manpages.debian.org/bookworm/systemd/systemd.unit.5.en.html) IIUC it would need `After=` and then either `Wants=` or `Requires=` on this other service.

ser-io: the `cuttlefish-host_orchestrator` is defined with the traditional `init` script rather than with systemd unit file. Attempted the migration to `systemd` unit files (https://github.com/ser-io/android-cuttlefish/commit/4f62fcfe48039576f399201f85c9bbd371a78615) in which the suggested line `Wants=systemd-journal-gatewayd.service` does the work. The problem with the migration is that `systemd` isn't installed in Docker failing to start the  host orchestrator service there if using systemd unit files with current scripts, so we need to stick with the init script.

Databean: In chat @ser-io mentioned [`systemd-sysv-generator`](https://man7.org/linux/man-pages/man8/systemd-sysv-generator.8.html) is used on this file for systemd. Would it be possible to list the dependency in the `INIT INFO` block at the top of this file?

It looks like `systemd-sysv-generator` can read `INIT INFO` blocks, [extract Required-Start blocks](https://github.com/systemd/systemd/blob/828e52bf8fae8778cc784d207252a6db82e4bee2/src/sysv-generator/sysv-generator.c#L558), and build systemd [wants / before / after](https://github.com/systemd/systemd/blob/828e52bf8fae8778cc784d207252a6db82e4bee2/src/sysv-generator/sysv-generator.c#L414) state that is [written into the service file](https://github.com/systemd/systemd/blob/828e52bf8fae8778cc784d207252a6db82e4bee2/src/sysv-generator/sysv-generator.c#L154).

Databean: It's unfortunate for the `systemd` case where i assume it wants visibility into cross-service dependencies, rather than having services internally start other services.

Can you please leave a comment about this limitation?

Databean (approved):

frontend/debian/cuttlefish-orchestration.cuttlefish-host_orchestrator.init
ser-io: Done.

PR #579, opened 2024-06-07
Update cuttlefish-common release version

cjreynol:
Bump v0.9.29 to unstable

### DIFF
--- a/base/debian/changelog
+++ b/base/debian/changelog
@@ -1,4 +1,4 @@
-cuttlefish-common (0.9.29) UNRELEASED; urgency=medium
+cuttlefish-common (0.9.29) unstable; urgency=medium
 
   [ A. Cody Schuffelen ]
   * Add capability to `cvd fetch` to download ChromeOS builds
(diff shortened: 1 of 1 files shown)


base/debian/changelog
```diff
@@ -1,4 +1,4 @@
-cuttlefish-common (0.9.29) UNRELEASED; urgency=medium
+cuttlefish-common (0.9.29) unstable; urgency=medium
```
Databean, line 1: Should this become `stable` now?

There is some [automation](https://github.com/google/android-cuttlefish/blob/1cd99c5c5c9dc48327ad0dda56c6c84e5ef8efd7/tools/cuttlefish-host-image-installer/update_github_releases.sh#L12) around finding the latest `stable` version.

jemoreira (approved):

base/debian/changelog
cjreynol: I thought we need to teamfood the unstable release for a bit before releasing a stable version?

The linked code should find `v0.9.28`, which I think is the outcome we want until we verify this release does not need any obvious bugfixes?

Databean: Ah I see, sorry for the trouble.

PR #581, opened 2024-06-07
Use `tools/buildutils/build_packages.sh`

ser-io:
- Use `tools/buildutils/build_packages.sh` in artifacts workflow.


.github/workflows/artifacts.yaml
```diff
@@ -20,7 +20,9 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@a81bbbf8298c0fa03ea29cdc473d45769f953675 # aka v2
     - name: Build debs
-      run: bash build_debs.sh
+      run: |
+        bash tools/buildutils/build_packages.sh
+        ls -la
```
jemoreira, line 25: nit: `ls -la *.dev` will produce a more concise output

jemoreira (approved):

.github/workflows/artifacts.yaml
ser-io: good point. done.

PR #582, opened 2024-06-07
Update U-boot repository URL and sync with codelinaro

WeiChungHsu:
Update U-boot repository URL and sync with codelinaro to add the recent changes from codelinaro side.


.github/workflows/reusable-build.yaml
```diff
@@ -14,6 +14,8 @@ jobs:
       run: pbuilder-dist stable arm64 create
     - name: Pbuilder arm64 update
       run: pbuilder-dist stable arm64 update
+    - name: Install bazel
+      run: cd tools/cuttlefish-host-image-installer && ./utils/install_bazel_package.sh
```
jemoreira, line 18: Use `https://github.com/google/android-cuttlefish/blob/main/tools/buildutils/installbazel.sh` instead.

tools/cuttlefish-host-image-installer/utils/install_bazel_package.sh
jemoreira, line 1: This file duplicates `https://github.com/google/android-cuttlefish/blob/main/tools/buildutils/installbazel.sh`

jemoreira (changes requested):

.github/workflows/reusable-build.yaml
WeiChungHsu: Since it is cross-platform compilation (we don't have ARM64 github runner yet), we can't directly execute installbazel.sh or install_bazel_package.sh. Instead, I have to hardcode function install_bazel_aarch64() part of codes into build_cf_packages.sh.
When ARM64 github runner is available, I will change it to execute installbazel.sh directly.

tools/cuttlefish-host-image-installer/utils/install_bazel_package.sh
WeiChungHsu: Since it is cross-platform compilation (we don't have ARM64 github runner yet), we can't directly execute installbazel.sh or install_bazel_package.sh. Instead, I have to hardcode function install_bazel_aarch64() part of codes into build_cf_packages.sh.
When ARM64 github runner is available, I will change it to execute installbazel.sh directly.

WeiChungHsu: And I removed this file tools/cuttlefish-host-image-installer/utils/install_bazel_package.sh, because cross-platform can't directly execute it.
This script doesn't work in cross-platform compilation.

jemoreira (approved):

PR #589, opened 2024-06-19
[CVD] Introduce instance states to the database

jemoreira:
and other small changes needed to keep records in database before and after the time the instances are running.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/selector/instance_group_record.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/selector/instance_group_record.cpp
@@ -93,8 +93,18 @@ void LocalInstanceGroup::SetProductOutPath(
   group_proto_.set_product_out_path(product_out_path);
 }
 
+void LocalInstanceGroup::SetAllStates(cvd::InstanceState state) {
+  for (auto& instance: Instances()) {
+    instance.SetState(state);
+  }
+}
+
 TimeStamp LocalInstanceGroup::StartTime() const {
   return CvdServerClock::from_time_t(group_proto_.start_time_sec());
+}
+
+void LocalInstanceGroup::SetStartTime(TimeStamp time) {
+  group_proto_.set_start_time_sec(CvdServerClock::to_time_t(time));
 }
 
 LocalInstanceGroup::LocalInstanceGroup(
--- a/base/cvd/cuttlefish/host/commands/cvd/selector/instance_group_record.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/selector/instance_group_record.cpp
@@ -59,14 +59,38 @@ Result<LocalInstanceGroup> LocalInstanceGroup::Create(
 
   for (const auto& instance_proto : group_proto.instances()) {
     instances.emplace_back(group_proto, instance_proto);
-    ids.insert(instance_proto.id());
+    auto id = instance_proto.id();
+    if (id != 0) {
+      // Only non-zero ids are checked, zero means no id has been assigned yet.
+      CF_EXPECTF(ids.find(id) == ids.end(),
+                 "Instances must have unique ids, found '{}' repeated", id);
+      ids.insert(id);
+    }
     names.insert(instance_proto.name());
   }
-  CF_EXPECT(ids.size() == (size_t)group_proto.instances_size(),
-            "Instances must have unique ids");
   CF_EXPECT(names.size() == (size_t)group_proto.instances_size(),
             "Instances must have unique names");
   return LocalInstanceGroup(group_proto, instances);
+}
+
+void LocalInstanceGroup::SetHomeDir(const std::string& home_dir) {
+  CHECK(group_proto_.home_directory().empty())
+      << "Home directory can't be changed once set";
+  group_proto_.set_home_directory(home_dir);
+}
+
+void LocalInstanceGroup::SetHostArtifactsPath(
+    const std::string& host_artifacts_path) {
+  CHECK(group_proto_.host_artifacts_path().empty())
+      << "Host artifacts path can't be changed once set";
+  group_proto_.set_host_artifacts_path(host_artifacts_path);
+}
+
+void LocalInstanceGroup::SetProductOutPath(
+    const std::string& product_out_path) {
+  CHECK(group_proto_.product_out_path().empty())
+      << "Product out path can't be changed once set";
+  group_proto_.set_product_out_path(product_out_path);
 }
 
 TimeStamp LocalInstanceGroup::StartTime() const {
--- a/base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
@@ -224,7 +224,7 @@ Result<StatusFetcherOutput> StatusFetcher::FetchStatus(
 }
 
 Result<Json::Value> StatusFetcher::FetchGroupStatus(
-    const selector::LocalInstanceGroup& group,
+    selector::LocalInstanceGroup& group,
     const RequestWithStdio& original_request) {
   Json::Value group_json(Json::objectValue);
   group_json["group_name"] = group.GroupName();
@@ -241,10 +241,20 @@ Result<Json::Value> StatusFetcher::FetchGroupStatus(
                                  original_request.FileDescriptors()};
   auto [_, instances_json, group_response] =
       CF_EXPECT(FetchStatus(group_request));
-  CF_EXPECT(
-      group_response.status().code() == cvd::Status::OK,
-      fmt::format("Running `cvd status --all_instances` for group \"{}\" failed",
-                  group.GroupName()));
+  if (group_response.status().code() != cvd::Status::OK) {
+    group.SetAllStates(cvd::INSTANCE_STATE_UNREACHABLE);
+    CF_EXPECT(instance_manager_.UpdateInstanceGroup(group));
+    instances_json = Json::Value(Json::arrayValue);
+    for (const auto& instance: group.Instances()) {
+      Json::Value instance_json;
+      instance_json["instance_name"] = instance.PerInstanceName();
+      instances_json.append(instance_json);
+    }
+  }
+  for (size_t i = 0; i < group.Instances().size(); ++i) {
+    instances_json[(int)i]["status"] =
+        cvd::InstanceState_Name(group.Instances()[i].State());
+  }
   group_json["instances"] = instances_json;
   return group_json;
 }
--- a/base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
@@ -16,11 +16,14 @@
 
 #include "host/commands/cvd/server_command/status_fetcher.h"
 
+#include <cctype>
 #include <map>
 #include <mutex>
 #include <sstream>
+#include <string>
 #include <vector>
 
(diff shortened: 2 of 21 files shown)


base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
```diff
@@ -224,10 +224,28 @@ Result<StatusFetcherOutput> StatusFetcher::FetchStatus(
...
-    const selector::LocalInstanceGroup& group,
+    selector::LocalInstanceGroup& group,
     const RequestWithStdio& original_request) {
   Json::Value group_json(Json::objectValue);
   group_json["group_name"] = group.GroupName();
+  std::set<cvd::InstanceState> reachable_states{
+      cvd::InstanceState::RUNNING, cvd::InstanceState::STARTING,
+      cvd::InstanceState::UNREACHABLE};
+  CF_EXPECT(!group.Instances().empty(), "Group can't be empty");
+  if (!Contains(reachable_states, group.Instances()[0].State())) {
```
Databean, line 235: Should this be checked for every instance rather than `Instances()[0]`?

base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
```diff
@@ -224,10 +224,28 @@ Result<StatusFetcherOutput> StatusFetcher::FetchStatus(
...
 
 Result<Json::Value> StatusFetcher::FetchGroupStatus(
-    const selector::LocalInstanceGroup& group,
+    selector::LocalInstanceGroup& group,
     const RequestWithStdio& original_request) {
   Json::Value group_json(Json::objectValue);
   group_json["group_name"] = group.GroupName();
+  std::set<cvd::InstanceState> reachable_states{
+      cvd::InstanceState::RUNNING, cvd::InstanceState::STARTING,
+      cvd::InstanceState::UNREACHABLE};
```
Databean, line 233: It's odd naming to have `UNREACHABLE` as a member of `reachable_states`.

Maybe `states_to_message`?

base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
```diff
@@ -224,10 +224,28 @@ Result<StatusFetcherOutput> StatusFetcher::FetchStatus(
 }
 
 Result<Json::Value> StatusFetcher::FetchGroupStatus(
-    const selector::LocalInstanceGroup& group,
+    selector::LocalInstanceGroup& group,
```
Databean, line 227: nit: now that this is an [input/output parameter](https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs), it should be after the input parameters.

base/cvd/cuttlefish/host/commands/cvd/selector/instance_group_record.cpp
```diff
@@ -59,20 +59,54 @@ Result<LocalInstanceGroup> LocalInstanceGroup::Create(
...
   }
-  CF_EXPECT(ids.size() == (size_t)group_proto.instances_size(),
-            "Instances must have unique ids");
   CF_EXPECT(names.size() == (size_t)group_proto.instances_size(),
             "Instances must have unique names");
   return LocalInstanceGroup(group_proto, instances);
 }
 
+void LocalInstanceGroup::SetHomeDir(const std::string& home_dir) {
+  CHECK(group_proto_.home_directory().empty())
```
Databean, line 77: Should these functions return a `Result`?

base/cvd/cuttlefish/host/commands/cvd/unittests/selector/instance_database_test.cpp
```diff
@@ -476,5 +476,50 @@ TEST_F(CvdInstanceDatabaseJsonTest, DumpLoadDumpCompare) {
...
+
+  auto find_res = db.FindGroup(Query("group_name", "grp1"));
+  ASSERT_TRUE(find_res.ok()) << find_res.error().Message();
+
+  ASSERT_EQ(find_res->HomeDir(), Workspace() + "/grp1_home");
+  ASSERT_EQ(find_res->ProductOutPath(), "/path/to/product");
+  ASSERT_EQ(find_res->Instances()[0].InstanceId(), 1);
+  ASSERT_EQ(find_res->Instances()[1].InstanceId(), 2);
+  ASSERT_EQ(find_res->Instances()[0].State(), cvd::InstanceState::STARTING);
+  ASSERT_EQ(find_res->Instances()[1].State(), cvd::InstanceState::STARTING);
```
Databean, line 521: This final `ASSERT` block could use `EXPECT` which is [non-fatal](https://google.github.io/googletest/primer.html#assertions). The test still fails but the other `EXPECT` calls will still be run and can provide additional information in the test results.

base/cvd/cuttlefish/host/commands/cvd/selector/instance_group_record.cpp
```diff
@@ -59,20 +59,54 @@ Result<LocalInstanceGroup> LocalInstanceGroup::Create(
 
   for (const auto& instance_proto : group_proto.instances()) {
     instances.emplace_back(group_proto, instance_proto);
-    ids.insert(instance_proto.id());
+    auto id = instance_proto.id();
+    if (id != 0) {
+      // Only non-zero ids are checked, zero means no id has been assigned yet.
+      CF_EXPECTF(ids.find(id) == ids.end(),
+                 "Instances must have unique ids, foud {} repeated", id);
```
Databean, line 66: nit: `found`

nit: use quotes or a separator around `{}` to distinguish the user-provided input from the error message.

base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
```diff
@@ -241,10 +259,20 @@ Result<Json::Value> StatusFetcher::FetchGroupStatus(
...
+    for (const auto& instance: group.Instances()) {
+      Json::Value instance_json;
+      instance_json["instance_name"] = instance.PerInstanceName();
+      instances_json.append(instance_json);
+    }
+  }
+  for (size_t i = 0; i < group.Instances().size(); ++i) {
+    instances_json[(int)i]["status"] =
+        cvd::InstanceState_Name(group.Instances()[i].State());
+  }
```
Databean, line 275: Would it be possible to break up this method into two methods based on the two large behaviors it has?

1. Update `selector::LocalInstanceGroup` state based on `cvd status`
2. Produce a JSON report representing the `selector::LocalInstanceGroupState`

If there's a lot of `cvd status` output not captured in `selector::LocalInstanceGroupState`, is it possible to add more fields to capture that additional state?

base/cvd/cuttlefish/host/commands/cvd/selector/instance_group_record.cpp
jemoreira: A fail of this check represents a coding error. Normally I would try to make the compiler detect these, but in this case I can't, so a runtime abort is the next best thing.

base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp

I was under the (wrong) impression that only one control socket was contacted for the entire group. Ended up rewriting most of my changes in this file and adding a `UpdateInstance` function to isntance manager and database.

base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
jemoreira: Doesn't apply after addressing the previous comment. Move this to `FetchOneInstanceStatus` which almost everything else calls here. Updating the instance state there is two lines, so probably doesn't make much sense to extract.

base/cvd/cuttlefish/host/commands/cvd/unittests/selector/instance_database_test.cpp
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
```diff
@@ -43,6 +46,33 @@ struct IdAndPerInstanceName {
...
+  for (size_t i = 0; i < name.size(); ++i) {
+    // Replace underscores with spaces
+    if (name[i] == '_') {
+      name[i] = ' ';
+      continue;
+    }
+    // All characters but the first of each word should be lowercase
+    bool first = (i == 0 || name[i-1] == ' ');
+    if (!first) {
+      name[i] = std::tolower(name[i]);
```
Databean, line 69: Because `std::string` is made of `char`s, note that for `std::tolower`, [cppreference writes](https://en.cppreference.com/w/cpp/string/byte/tolower)

> Like all other functions from [`<cctype>`](https://en.cppreference.com/w/cpp/header/cctype), the behavior of `std::tolower` is undefined if the argument's value is neither representable as `unsigned char` nor equal to [`EOF`](https://en.cppreference.com/w/cpp/io/c). To use these functions safely with plain `char`s (or `signed char`s), the argument should first be converted to `unsigned char`:
```c++
static_cast<char>(std::tolower(static_cast<unsigned char>(ch)))
```

base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
```diff
@@ -241,10 +284,20 @@ Result<Json::Value> StatusFetcher::FetchGroupStatus(
                                  original_request.FileDescriptors()};
   auto [_, instances_json, group_response] =
       CF_EXPECT(FetchStatus(group_request));
-  CF_EXPECT(
-      group_response.status().code() == cvd::Status::OK,
-      fmt::format("Running `cvd status --all_instances` for group \"{}\" failed",
-                  group.GroupName()));
+  if (group_response.status().code() != cvd::Status::OK) {
```
Databean, line 287: Kind of an ambiguous question, but should this case be represented by a failed `Result` rather than a successful `Result` with a non-`OK` status? What distinguishes these conditions?

Probably out of the scope of this change though.

base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
Databean: It still feels somewhat unrelated to the rest of the behavior of this function and I'm generally not a fan of output parameters, but it also looks inconvenient to extract.

Also, `FetchStatus` calls `FetchOneInstanceStatus` which calls `UpdateInstance` in the same unreachable state, is it necessary to do the same here?

base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
```diff
@@ -43,6 +46,33 @@ struct IdAndPerInstanceName {
   unsigned id;
 };
 
+// The most important thing this function does is turn "INSTANCE_STATE_RUNNING"
+// into "Running". Some external tools (like the host orchestrator) already
+// depend on this string.
+std::string HumanFriendlyStateName(cvd::InstanceState state) {
+  std::string name = cvd::InstanceState_Name(state);
+  // Drop the enum name prefix
+  std::string prefix = "INSTANCE_STATE_";
```
Databean, line 55: nit: `std::string_view`

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
jemoreira: Fixed

PR #592, opened 2024-06-24
Removing ext/stdio_filebuf.h include

cjreynol:
Fails to build in google3.  

Since this is now a `SharedFD`-based operation, I moved the logic there instead of `files.h/cpp`

Bug: 348040850
Test: cvd fetch --target_directory=/tmp/cvd/fetch-test --default_build=aosp-main

### DIFF
--- a/base/cvd/cuttlefish/common/libs/fs/shared_fd.cpp
+++ b/base/cvd/cuttlefish/common/libs/fs/shared_fd.cpp
@@ -30,11 +30,16 @@
 #include <cstddef>
 
 #include <algorithm>
+#include <cstdlib>
 #include <sstream>
+#include <string>
+#include <string_view>
+#include <utility>
 #include <vector>
 
 #include <android-base/file.h>
 #include <android-base/logging.h>
+#include <fmt/format.h>
 
 #include "common/libs/fs/shared_buf.h"
 #include "common/libs/fs/shared_select.h"
@@ -522,6 +527,18 @@ SharedFD SharedFD::Mkstemp(std::string* path) {
   } else {
     return SharedFD(std::shared_ptr<FileInstance>(new FileInstance(fd, 0)));
   }
+}
+
+Result<std::pair<SharedFD, std::string>> SharedFD::Mkostemp(
+    const std::string_view path, const int flags) {
+  // mkostemp replaces the Xs with random selections to make a unique filename
+  auto temp_path = fmt::format("{}XXXXXX", path);
+  const int fd = mkostemp(temp_path.data(), flags);
+  CF_EXPECTF(fd != -1, "Error creating temporary file: {}", strerror(errno));
+  auto shared_fd =
+      SharedFD(std::shared_ptr<FileInstance>(new FileInstance(fd, 0)));
+  return std::make_pair<SharedFD, std::string>(std::move(shared_fd),
+                                               std::move(temp_path));
 }
 
 SharedFD SharedFD::ErrorFD(int error) {
--- a/base/cvd/cuttlefish/BUILD.bazel
+++ b/base/cvd/cuttlefish/BUILD.bazel
@@ -4,6 +4,7 @@ cc_library(
         "common/libs/fs/epoll.cpp",
         "common/libs/fs/shared_buf.cc",
         "common/libs/fs/shared_fd.cpp",
+        "common/libs/fs/shared_fd_stream.cpp",
         "common/libs/utils/archive.cpp",
         "common/libs/utils/base64.cpp",
         "common/libs/utils/environment.cpp",
--- a/base/cvd/cuttlefish/common/libs/fs/shared_fd.h
+++ b/base/cvd/cuttlefish/common/libs/fs/shared_fd.h
@@ -38,6 +38,7 @@
 #include <memory>
 #include <sstream>
 #include <string>
+#include <string_view>
 #include <utility>
 #include <vector>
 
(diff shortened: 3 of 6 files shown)


base/cvd/cuttlefish/common/libs/fs/shared_fd.cpp
```diff
@@ -524,6 +527,24 @@ SharedFD SharedFD::Mkstemp(std::string* path) {
   }
 }
 
+Result<std::pair<SharedFD, std::string>> SharedFD::Mkostemp(
+    const std::string& path) {
+  // mkostemp replaces the Xs with random selections to make a unique filename
+  auto temp_path = path + "XXXXXX";
+  const int fd = mkostemp(temp_path.data(), O_CLOEXEC);
+  CF_EXPECTF(fd != -1, "Error creating temporary file: {}", strerror(errno));
+  auto shared_fd = SharedFD::Dup(fd);
```
jemoreira, line 536: `Dup` is unnecessary here, from inside the `SharedFD` class the private constructor should be used:
```
SharedFD(std::shared_ptr<FileInstance>(new FileInstance(fd, fd<0?errno:0)));
```

jemoreira (changes requested):

base/cvd/cuttlefish/common/libs/fs/shared_fd.cpp
```diff
@@ -524,6 +527,24 @@ SharedFD SharedFD::Mkstemp(std::string* path) {
   }
 }
 
+Result<std::pair<SharedFD, std::string>> SharedFD::Mkostemp(
```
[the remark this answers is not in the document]
Databean: https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs writes

>
> Prefer using return values over output parameters: they improve readability, and often provide the same or better performance.

Is matching the type signature of `stdlib.h` more important than following this part of the style guide?

base/cvd/cuttlefish/common/libs/fs/shared_fd.cpp
cjreynol: Done

base/cvd/cuttlefish/common/libs/fs/shared_fd.cpp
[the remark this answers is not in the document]
cjreynol: I also prefer the return value and taking care of the `path` copy inside `Mkostemp` instead of leaving it to the caller, enforcing the caller has both values.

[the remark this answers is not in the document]
jemoreira: In my opinion, yes. Especially because it doesn't forbid the use of in/out parameters, it just says "prefer" the return approach.
In any case I don't have a strong opinion here, that's what I marked it as a "nit", either way is fine by me.

base/cvd/cuttlefish/common/libs/fs/shared_fd.cpp
```diff
@@ -524,6 +527,25 @@ SharedFD SharedFD::Mkstemp(std::string* path) {
...
+    const std::string& path) {
+  // mkostemp replaces the Xs with random selections to make a unique filename
+  auto temp_path = path + "XXXXXX";
+  const int fd = mkostemp(temp_path.data(), O_CLOEXEC);
+  CF_EXPECTF(fd != -1, "Error creating temporary file: {}", strerror(errno));
+  auto shared_fd =
+      SharedFD(std::shared_ptr<FileInstance>(new FileInstance(fd, 0)));
+  CF_EXPECTF(shared_fd->IsOpen(),
+             "Error creating SharedFD from existing file descriptor for \"{}\"",
+             path);
```
jemoreira, line 540: This isn't necessary anymore.

base/cvd/cuttlefish/common/libs/fs/shared_fd.cpp
```diff
@@ -524,6 +527,25 @@ SharedFD SharedFD::Mkstemp(std::string* path) {
...
+  CF_EXPECTF(close(fd) != -1,
+             "Error closing original file descriptor after creating SharedFD "
+             "for \"{}\": ",
+             path, strerror(errno));
```
jemoreira, line 544: This will actually close the file descriptor before returning it, it was only needed because of the `Dup`. By passing the fd to the SharedFD constructor the newly created object takes ownership of that fd and you don't need to worry about it anymore.

base/cvd/cuttlefish/common/libs/fs/shared_fd.cpp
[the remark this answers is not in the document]
jemoreira: > taking care of the path copy inside Mkostemp

No copy would be needed with the in/out parameter approach. I also don't see the value of having both the template and the path after the file has been created, although it doesn't hurt either.

jemoreira (changes requested):

base/cvd/cuttlefish/common/libs/fs/shared_fd.cpp
cjreynol: Thanks for the catch!

base/cvd/cuttlefish/common/libs/fs/shared_fd.cpp
cjreynol: Removed

Databean (approved):

jemoreira (approved):

PR #602, opened 2024-06-28
Support take_snapshot command.

ser-io:


frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -64,6 +64,8 @@ func (c *Controller) AddRoutes(router *mux.Router) {
 		httpHandler(newExecCVDCommandHandler(c.Config, c.OperationManager, "stop"))).Methods("DELETE")
 	router.Handle("/cvds/{group}/{name}/:powerwash",
 		httpHandler(newExecCVDCommandHandler(c.Config, c.OperationManager, "powerwash"))).Methods("POST")
+	router.Handle("/cvds/{group}/{name}/:snapshot",
```
jemoreira, line 67: This API might be more useful like this:
```
GET /cvds/{group}/{name}/snapshots -> list snapshots for that device
POST /cvds/{group}/{name}/snapshots -> create a new snapshot
GET /cvds/{group}/{name}/snapshots/{name} -> get information or download snapshot
PUT /cvds/{group}/{name}/snapshots/{name} -> upload a snaphot
```

That last one is useful to start a device from a snapshot created elsewhere. It would need the create cvd endpoint to support fetching the artifacts without starting until the snapshot is uploaded though, but it's probably a good idea to start thinking about that.

jemoreira (changes requested):

PR #606, opened 2024-07-03
Add WebSocket endpoint for adb proxy

k311093:
Add url /devices/dev_id/adb as WebSocket endpoint that proxies WebSocket payloads to the tcp port for the adb of device for dev_id.


frontend/src/liboperator/operator/operator.go
```diff
@@ -480,6 +484,111 @@ func openwrt(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
...
+	adbPort := 0
+	if adb_port, ok := devInfo["adb_port"]; ok {
+		adbPort = int(adb_port.(float64))
+	} else {
+		for i, id := range pool.DeviceIds() {
+			if id == devId {
+				adbPort = 6520 + i
+				break
+			}
+		}
```
0405ysj, line 512: I think this fallback mechanism is not reliable. Especially when the base instance number is not 1. Shouldn't we make backporting changes for previous CF releases instead?

frontend/src/liboperator/operator/operator.go
```diff
@@ -480,6 +484,111 @@ func openwrt(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
...
+		log.Print("Error while connect to ADB: ", err)
+		return
+	}
+	defer tcpConn.Close()
+
+	// Redirect WebSocket to ADB tcp socket
+	redirectWsToTcp(wsConn, tcpConn)
+}
+
+func redirectWsToTcp(wsConn *websocket.Conn, tcpConn net.Conn) {
```
0405ysj, line 539: Couldn't bi-directional data transmission be happened with `io.copy`? Just out of curiosity, I think it would make code much simpler if it's available.

0405ysj (changes requested):

frontend/src/liboperator/operator/operator.go
k311093: Changed to use io.Copy.

frontend/src/liboperator/operator/operator.go
k311093: @ser-io @jemoreira I have some question about this.

Does it okay to move this logic to orchestrator? or it needs to be in operator. Since operator does not have adb connection info for the device I added adb port info to device registration info, but it might be a problem when device is launched with old host packages.

I added the fallback logic, but there are some situations this might not work.

But if this logic is in the orchestrator, it might be possible to get adb_serial info from orchestrator.

Do you have any suggestions for that?

frontend/src/liboperator/operator/operator.go
```diff
@@ -480,6 +485,100 @@ func openwrt(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
...
+type wsIoWrapper struct {
+	wsConn *websocket.Conn
+	pos int
+	buf []byte
+}
+
+var _ io.ReadWriteCloser = (*wsIoWrapper)(nil)
+
+func (w *wsIoWrapper) Read(buf []byte) (int, error) {
+	if w.buf == nil || w.pos == len(w.buf) {
```
0405ysj, line 557: nit: `w.pos >= len(w.buf)`?

frontend/src/liboperator/operator/operator.go
```diff
@@ -480,6 +485,100 @@ func openwrt(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
...
+
+func (w *wsIoWrapper) Write(buf []byte) (int, error) {
+	err := w.wsConn.WriteMessage(websocket.BinaryMessage, buf)
+	if err != nil {
+		return 0, err
+	}
+	return len(buf), nil
+}
+
+func (w *wsIoWrapper) Close() error {
```
0405ysj, line 578: When is it used?

frontend/src/liboperator/operator/operator.go
```diff
@@ -480,6 +485,100 @@ func openwrt(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
...
+	wsConn *websocket.Conn
+	pos int
+	buf []byte
+}
+
+var _ io.ReadWriteCloser = (*wsIoWrapper)(nil)
+
+func (w *wsIoWrapper) Read(buf []byte) (int, error) {
+	if w.buf == nil || w.pos == len(w.buf) {
+		_, buf, err := w.wsConn.ReadMessage()
```
0405ysj, line 558: As we talked, the variable name `buf` seems duplicated.

frontend/src/liboperator/operator/operator.go
k311093: Changed to use the expression.

frontend/src/liboperator/operator/operator.go
k311093: Changed buf name that is duplicated.

frontend/src/liboperator/operator/operator.go
k311093: Close() is not used, removed the function and changed implementing interface from io.ReadWriteCloser to io.ReadWriter.

frontend/src/liboperator/operator/operator.go
```diff
@@ -480,6 +485,104 @@ func openwrt(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
...
+	// Redirect WebSocket to ADB tcp socket
+	wsWrapper := &wsIoWrapper{
+		wsConn: wsConn,
+		pos: 0,
+		buf: nil,
+	}
+
+	go func() {
+		io.Copy(wsWrapper, tcpConn)
+		wsWrapper.Close()
```
0405ysj, line 545: As we talked before, please ensure both connection being closed when it ends.

frontend/src/liboperator/operator/operator.go
```diff
@@ -480,6 +485,104 @@ func openwrt(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
...
+	}()
+	io.Copy(tcpConn, wsWrapper)
+	tcpConn.Close()
```
0405ysj, line 548: ditto.

frontend/src/liboperator/operator/operator.go
```diff
@@ -480,6 +485,104 @@ func openwrt(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
...
+}
+
+// Wrapper for implementing io.ReadWriteCloser of websocket.Conn
+type wsIoWrapper struct {
+	wsConn *websocket.Conn
+	pos int
+	buf []byte
```
0405ysj, line 555: nit: Could you leave a short comment that those fields are for `Read` function only?

frontend/src/liboperator/operator/operator.go
0405ysj: AFAIK, we would unite Operator and Host Orchestrator into one. (Correct me if I'm wrong.)

Then, it's better to be in Host Orchestrator, since it holds information about adb_serial.
TBH, we don't need ADB port but ADB serial instead, isn't it?

0405ysj (changes requested):

frontend/src/liboperator/operator/operator.go
k311093: This code had already patched for connection closing.

frontend/src/liboperator/operator/operator.go
k311093: This code had also patched for connection closing.

frontend/src/liboperator/operator/operator.go
k311093: Added comment for the field that is used for Read

frontend/src/liboperator/operator/operator.go
k311093: If an instance is started by `cvd start` it would be fine. adb_serial for host orchestrator is got by `cvd fleet` which shows instances created by `cvd start` (of course created by host orchestrator).

If supporting ADB connection proxy from host orchestrator it might need to rename `/devices/{deviceId}/adb` endpoint to `cvds/{group}/{name}/adb`.

Or another option might be just return an error `Device is not supported for ADB` when deviceInfo does not have ADB port for operator, and backport ADB port registration change(http://aosp/3154596) for the device when it is needed.

@ser-io @jemoreira That do you think?

k311093: Decided not to use fallback logic but just return errors when adb_port is not registered.

0405ysj (approved):

ser-io:
Team remember to use `git commit --amend` and `git push -f` when addressing comments, it makes the history of the repository cleaner and the PRs easier to review. See [PR#307](https://github.com/google/cloud-android-orchestration/pull/307) (having 50 threads and only 1 commit) as a reference. Thanks!

frontend/src/liboperator/operator/operator.go
```diff
@@ -533,60 +534,49 @@ func adbProxy(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
...
+}
 
-			wsData := make([]byte, len(buf))
-			copy(wsData, buf)
+var _ io.ReadWriteCloser = (*wsIoWrapper)(nil)
```
ser-io, line 548: why is this needed?

frontend/src/liboperator/operator/operator.go
```diff
@@ -533,60 +534,49 @@ func adbProxy(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
...
+	if w.buf == nil || w.pos == len(w.buf) {
+		_, buf, err := w.wsConn.ReadMessage()
+		if err != nil {
+			return 0, err
 		}
-	}(wsConn, wsChan)
+		w.buf = buf
+		w.pos = 0
+	}
+	nRead := copy(buf[:], w.buf[w.pos:])
```
ser-io, line 559: we can use `buf` directly here as it's already an slice.

frontend/src/liboperator/operator/operator.go
```diff
@@ -533,60 +534,49 @@ func adbProxy(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
...
+	pos int
+	buf []byte
+}
 
-			wsData := make([]byte, len(buf))
-			copy(wsData, buf)
+var _ io.ReadWriteCloser = (*wsIoWrapper)(nil)
 
-			wsChan <- wsData
+func (w *wsIoWrapper) Read(buf []byte) (int, error) {
```
ser-io, line 550: nit: use `p` argument name rather than `buf` to avoid confusion with `w.buf` using the name in the original signature: `Read(p []byte) (n int, err error)` https://pkg.go.dev/io#Reader

ser-io (commented):
We should use `git commit --amend` and `git push -f `when addressing comments and avoid new commits, it makes the history of the repository cleaner and the PRs easier to review. See https://github.com/google/cloud-android-orchestration/pull/307 (having 50 threads and only 1 commit) as a reference.

frontend/src/liboperator/operator/operator.go
k311093: This code is for confirming that `wsIoWrapper` implements `io.ReadWriteCloser` interface

frontend/src/liboperator/operator/operator.go
k311093: I'll upload separated PR for this.

frontend/src/liboperator/operator/operator.go
k311093: I'll upload separated PR for this.

k311093:
> We should use `git commit --amend` and `git push -f `when addressing comments and avoid new commits, it makes the history of the repository cleaner and the PRs easier to review. See [google/cloud-android-orchestration#307](https://github.com/google/cloud-android-orchestration/pull/307) (having 50 threads and only 1 commit) as a reference.

Acknowledged, thanks!

frontend/src/liboperator/operator/operator.go
ser-io: I see, from https://go.dev/doc/effective_go#interfaces_and_types this is needed when interface checks happens at runtime which is not the case here. 

```
The appearance of the blank identifier in this construct indicates that the declaration exists only for the type checking, not to create a variable. Don't do this for every type that satisfies an interface, though. By convention, such declarations are only used when there are no static conversions already present in the code, which is a rare event.
```

frontend/src/liboperator/operator/operator.go
```diff
@@ -480,6 +485,94 @@ func openwrt(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
...
+		wsConn: wsConn,
+		pos:    0,
+		buf:    nil,
+	}
+	defer wsWrapper.Close()
+
+	// Redirect WebSocket to ADB tcp socket
+	go func() {
+		io.Copy(wsWrapper, tcpConn)
+		wsWrapper.Close()
```
jemoreira, line 534: Should this be `tcpConn.Close()`? The `wsWrapper` is already closed on line 529.

ikicha:
If we have websocket endpoints for adb, can we simplify cvdr impl for adb connect? WDYT @jemoreira @0405ysj

jemoreira:
> If we have websocket endpoints for adb, can we simplify cvdr impl for adb connect? WDYT @jemoreira @0405ysj

Websocket doesn't work in AppEngine, so we need both implementations in cvdr unfortunately.

0405ysj:
> If we have websocket endpoints for adb, can we simplify cvdr impl for adb connect?

What I heard from Jaeman was, there are 2 types of AppEngine environments; standard vs flexible.
Standard is what we're currently using, but it doesn't support websocket.
Flexible supports websocket, but this type is not available for the internal uses.

AFAIK Jaeman is currently investigating around AppEngine deeply.

frontend/src/liboperator/operator/operator.go
k311093: When I tested it, deferred wsWrapper.Close() was not called with spawned go routine, so it need to close explicitly io.Copy() call at line 536 to be exited when tcpConn is closed.

PR #607, opened 2024-07-05
Add bazel detection when installation.

spiritedRunning:
In my case, can't install bazel by installbazel.sh successfully, it keep return `curl: (7) Failed to connect to bazel.build port 443: Connection refused` error, so I tried to install bazel manually.
After successful installation of bazel, the `install_bazel_x86_64` function still tries to install it. This would be causing failure when local environment can't access bazel.build server.

### DIFF
--- a/tools/buildutils/build_packages.sh
+++ b/tools/buildutils/build_packages.sh
@@ -25,7 +25,7 @@ function build_package() {
 REPO_DIR="$(realpath "$(dirname "$0")/../..")"
 INSTALL_BAZEL="$(dirname $0)/installbazel.sh"
 
-sudo "${INSTALL_BAZEL}"
+command -v bazel &> /dev/null || sudo "${INSTALL_BAZEL}"
 install_debuild_dependencies
 
 build_package "${REPO_DIR}/base"


tools/buildutils/installbazel.sh
```diff
@@ -19,13 +19,17 @@
...
 
 function install_bazel_x86_64() {
-  echo "Installing bazel"
-  apt install apt-transport-https curl gnupg -y
-  curl -fsSL https://bazel.build/bazel-release.pub.gpg | gpg --dearmor >bazel-archive-keyring.gpg
-  mv bazel-archive-keyring.gpg /usr/share/keyrings
-  echo "deb [arch=amd64 signed-by=/usr/share/keyrings/bazel-archive-keyring.gpg] https://storage.googleapis.com/bazel-apt stable jdk1.8" | tee /etc/apt/sources.list.d/bazel.list
-  # bazel needs the zip command to gather test outputs but doesn't depend on it
-  apt-get update && apt-get install -y bazel zip unzip
+  if command -v bazel &> /dev/null; then
```
ser-io, line 22: let's put this condition around `sudo "${INSTALL_BAZEL}"` in https://github.com/google/android-cuttlefish/blob/main/tools/buildutils/build_packages.sh#L28

spiritedRunning: done. re-commit.

tools/buildutils/build_packages.sh
```diff
@@ -25,7 +25,11 @@ function build_package() {
 REPO_DIR="$(realpath "$(dirname "$0")/../..")"
 INSTALL_BAZEL="$(dirname $0)/installbazel.sh"
 
-sudo "${INSTALL_BAZEL}"
+if command -v bazel &> /dev/null; then
```
ser-io, line 28: let's rather use the OR (||) having a single self explanatory line: `command -v bazel &> /dev/null || sudo "${INSTALL_BAZEL}"`

ser-io (changes requested):

ser-io:
Zach, this PR should contain only one commit, please use `git rebase` rather than `git merge` when bringing changes from upstream and use `git commit --amend` and `git push -f`  when addressing comments.

tools/buildutils/build_packages.sh
spiritedRunning: sure.

spiritedRunning: I have do rebase and push again. Please help to check if meets the submission requirement.

ser-io (approved):

spiritedRunning:
Hi, does any problem with this kokoro task?

ser-io:
> Hi, does any problem with this kokoro task?

let's try updating the branch, a new forced push might re-trigger the workflow.

Databean:
@ser-io a Google engineer has to apply the `kokoro:run` label. I've triggered the run now.

jemoreira (approved):

ser-io:
> @ser-io a Google engineer has to apply the `kokoro:run` label. I've triggered the run now.

Thanks Cody! I wasn't aware of this.

PR #608, opened 2024-07-09
Suggest an up-to-date value for `cvd fetch --default_build`.

Databean:
Bug: b/351915411
Test: bazel run cuttlefish:cvd -- fetch --help


ser-io (approved):

Databean:
Thanks for reviewing!

PR #611, opened 2024-07-09
Add menu for copying adb connect command that uses cvdr

k311093:
Add sub-menu for device that copies adb connect command to the clipboard.

the command would be like

`cvdr connect --connect_agent=websocket_agent --service_url=https://127.0.0.1:1443/ --host=websocket cvd-1`

### DIFF
--- a/frontend/src/operator/webui/src/app/app.module.ts
+++ b/frontend/src/operator/webui/src/app/app.module.ts
@@ -10,8 +10,10 @@ import {MatButtonModule} from '@angular/material/button';
 import {MatCardModule} from '@angular/material/card';
 import {MatCheckboxModule} from '@angular/material/checkbox';
 import {MatIconModule} from '@angular/material/icon';
+import {MatMenuModule} from '@angular/material/menu';
 import {MatSidenavModule} from '@angular/material/sidenav';
 import {MatSlideToggleModule} from '@angular/material/slide-toggle';
+import {MatSnackBarModule} from '@angular/material/snack-bar';
 import {MatToolbarModule} from '@angular/material/toolbar';
 import {ViewPaneComponent} from './view-pane/view-pane.component';
 import {SafeDeviceUrlPipe} from './safe-device-url.pipe';
@@ -32,8 +34,10 @@ import {RouterModule} from '@angular/router';
     MatCardModule,
     MatCheckboxModule,
     MatIconModule,
+    MatMenuModule,
     MatSidenavModule,
     MatSlideToggleModule,
+    MatSnackBarModule,
     MatToolbarModule,
     FormsModule,
     HttpClientModule,
--- a/frontend/src/operator/webui/src/app/device-pane/device-pane.component.html
+++ b/frontend/src/operator/webui/src/app/device-pane/device-pane.component.html
@@ -17,7 +17,11 @@
           (change)="displaysService.toggleVisibility(device.device_id)"
           >{{ device.device_id }}</mat-slide-toggle
         >
-        <button mat-icon-button class="menu" title="menu">
+        <button
+          mat-icon-button class="menu" title="menu"
+          [matMenuTriggerFor]="deviceMenu"
+          [matMenuTriggerData]="{device_id: device.device_id}"
+        >
           <mat-icon>more_vert</mat-icon>
         </button>
       </li>
@@ -29,3 +33,9 @@
     Show All
   </button>
 </div>
+
+<mat-menu #deviceMenu="matMenu">
+  <ng-template matMenuContent let-device_id="device_id">
+    <button mat-menu-item (click)="copyAdbCommand(device_id)">Copy ADB connect command</button>
+  </ng-template>
+</mat-menu>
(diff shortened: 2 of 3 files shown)


0405ysj (approved):

ikicha (approved):

ser-io:
this needs further discussion, the command `cvdr connect --connect_agent=websocket_agent --service_url=https://127.0.0.1:1443/ --host=websocket cvd-1` is not going to work for instances created in GCP for example.

ser-io (changes requested):
this needs further discussion, the command `cvdr connect --connect_agent=websocket_agent --service_url=https://127.0.0.1:1443/ --host=websocket cvd-1` is not going to work for instances created in GCP for example.

k311093:
> this needs further discussion, the command `cvdr connect --connect_agent=websocket_agent --service_url=https://127.0.0.1:1443/ --host=websocket cvd-1` is not going to work for instances created in GCP for example.

I checked with GCP instance and found that handshake is succeeded but it looks like failing to sending data, it might be because it uses WebSocket.

It might be two options to solve this
  1. Proxy WebSocket at CO
  2. Under GCP environment, copies normal cvdr command like `cvdr connect service_url=https://.../ --zone=us-central1-c --host=cf-00000-0000 cvd-1`

Looks like Option 2 can be simpler.

What do you think?

0405ysj:
Random idea: Mixed approach seems better for me, to make it clear for any types of our environment(Operator, HO, and CO).

- Define a new subpath that providing cvdr command text.
- WebUI reads response from that subpath, for providing command text via clicking button.
- Override that subpath by CO.

ser-io:
> > this needs further discussion, the command `cvdr connect --connect_agent=websocket_agent --service_url=https://127.0.0.1:1443/ --host=websocket cvd-1` is not going to work for instances created in GCP for example.
> 
> I checked with GCP instance and found that handshake is succeeded but it looks like failing to sending data, it might be because it uses WebSocket.
> 
> It might be two options to solve this
> 
> 1. Proxy WebSocket at CO
> 2. Under GCP environment, copies normal cvdr command like `cvdr connect service_url=https://.../ --zone=us-central1-c --host=cf-00000-0000 cvd-1`
> 
> Looks like Option 2 can be simpler.
> 
> What do you think?

I think this should be a placeholder to be filled out only by the `CO`, there should not be a `cvdr` command in this UI if there's not a `CO`  in the middle. If I only use Cuttlefish on my machine I'm still able to connect to devices via adb and I don't need `cvdr`.

Using `cvdr` with port forwarding connecting  to either the `HO` or the `Operator` instead of the `CO` is confusing to me as it's not being used as intended from the original design,  `service_url` is meant to be `CO` API only, not the `Operator` or the `HO` API. As  some functionalities of the `CO` API is just a proxy we can afford to omit the `CO` when developing, but we should not confuse the original concepts.

By implementing  **Option.1 `Proxy WebSocket at CO`** See b/352377928, we can continue to be aligned with `cvdr` original design of connecting only with the `CO`.

If we do need to connect directly to `HO` or  `Operator` APIs running on a remote machine via port forwarding, let's have a meeting to discuss it.

k311093:
> > > this needs further discussion, the command `cvdr connect --connect_agent=websocket_agent --service_url=https://127.0.0.1:1443/ --host=websocket cvd-1` is not going to work for instances created in GCP for example.
> > 
> > 
> > I checked with GCP instance and found that handshake is succeeded but it looks like failing to sending data, it might be because it uses WebSocket.
> > It might be two options to solve this
> > 
> > 1. Proxy WebSocket at CO
> > 2. Under GCP environment, copies normal cvdr command like `cvdr connect service_url=https://.../ --zone=us-central1-c --host=cf-00000-0000 cvd-1`
> > 
> > Looks like Option 2 can be simpler.
> > What do you think?
> 
> I think this should be a placeholder to be filled out only by the `CO`, there should not be a `cvdr` command in the UI if there's not a `CO` in the middle. If using cuttlefish on local workflows devices are already connected via adb and there's no need for `cvdr`.
> 
> Using `cvdr` with port forwarding connecting to either the `HO` or the `Operator` instead of the `CO` is confusing as it's not being used as intended from the original design, `service_url` is meant to be `CO` API only, not the `Operator` or the `HO` API. As some functionalities of the `CO` API is just a proxy we can afford to omit the `CO` when developing, but we should not promote these shortcut to final users.
> 
> By implementing **Option.1 `Proxy WebSocket at CO`** See b/352377928, we can continue to be aligned with `cvdr` original design of connecting only with the `CO`.
> 
> If we do need to connect directly to `HO` or `Operator` APIs running on a remote machine via port forwarding, let's have a meeting to discuss it.

Oh, I see the concepts between CO and cvdr.

Let me first try to proxy WebSocket at CO. And then fix cvdr to follow the concept `cvdr connect --connect_agent=websocket_agent --service_url=https://address.to.co/ --host=cf-aaaaaa cvd-1` (I used `service_url` as endpoint URL for Operator, will change to CO's url).

And then I'll revisit this PR later.

Thanks!

PR #612, opened 2024-07-10
Address comments for previous PR

k311093:
https://github.com/google/android-cuttlefish/pull/606

Modify `buf[:]` to `buf` since it is already slice.
Modify parameter name `buf` to `p` to avoid confusion.

### DIFF
--- a/frontend/src/liboperator/operator/operator.go
+++ b/frontend/src/liboperator/operator/operator.go
@@ -547,7 +547,7 @@ type wsIoWrapper struct {
 
 var _ io.ReadWriteCloser = (*wsIoWrapper)(nil)
 
-func (w *wsIoWrapper) Read(buf []byte) (int, error) {
+func (w *wsIoWrapper) Read(p []byte) (int, error) {
 	if w.buf == nil || w.pos >= len(w.buf) {
 		_, readBuf, err := w.wsConn.ReadMessage()
 		if err != nil {
@@ -556,7 +556,7 @@ func (w *wsIoWrapper) Read(buf []byte) (int, error) {
 		w.buf = readBuf
 		w.pos = 0
 	}
-	nRead := copy(buf[:], w.buf[w.pos:])
+	nRead := copy(p, w.buf[w.pos:])
 	w.pos += nRead
 	return nRead, nil
 }


frontend/src/liboperator/operator/operator.go
```diff
@@ -556,17 +556,17 @@ func (w *wsIoWrapper) Read(buf []byte) (int, error) {
...
 	}
-	nRead := copy(buf[:], w.buf[w.pos:])
+	nRead := copy(buf, w.buf[w.pos:])
 	w.pos += nRead
 	return nRead, nil
 }
 
-func (w *wsIoWrapper) Write(buf []byte) (int, error) {
-	err := w.wsConn.WriteMessage(websocket.BinaryMessage, buf)
+func (w *wsIoWrapper) Write(p []byte) (int, error) {
```
ser-io, line 564: nit:  this name change is more relevant to the Read method: `func (w *wsIoWrapper) Read(buf []byte) (int, error) {`

ser-io (changes requested):

ser-io:
Let's make the lines in the first comment https://github.com/google/android-cuttlefish/pull/612#issue-2399696488 part of the commit message so these can be seen with `git log`

```
Modify buf[:] to buf since it is already slice.
Modify parameter name buf to p to avoid confusion.
```

frontend/src/liboperator/operator/operator.go
k311093: Changed argument name of the Read() function instead of this.

ser-io (approved):
nit: we can add the following to the commit description as `Address comments for previous PR` only does not describe what the changes are about in a direct way.

Modify buf[:] to buf since it is already slice.
Modify parameter name buf to p to avoid confusion.

k311093:
> Let's make the lines in the first comment [#612 (comment)](https://github.com/google/android-cuttlefish/pull/612#issue-2399696488) part of the commit message so these can be seen with `git log`
> 
> ```
> Modify buf[:] to buf since it is already slice.
> Modify parameter name buf to p to avoid confusion.
> ```

Modified commit message. Thanks!

PR #614, opened 2024-07-11
cvd bug report API

ser-io:


frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -76,6 +76,9 @@ func (c *Controller) AddRoutes(router *mux.Router) {
 	// reached. Be prepared to retry if the deadline was reached.
 	router.Handle("/operations/{name}/:wait",
 		httpHandler(&waitOperationHandler{c.OperationManager, c.WaitOperationDuration})).Methods("POST")
+	router.Handle("/cvdbugreports",
+		httpHandler(newCreateCVDBugReportHandler(c.Config, c.OperationManager))).Methods("POST")
+	router.Handle("/cvdbugreports/{uuid}/:download", &downloadCVDBugReportHandler{c.Config}).Methods("POST")
```
jemoreira, line 81: why POST? Does it delete it from the remote host?

frontend/src/host_orchestrator/orchestrator/createcvdbugreportaction.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+const BugReportZipFileName = "cvd_bugreport.zip"
+
+func (a *CreateCVDBugReportAction) createBugReport(uuid string, op apiv1.Operation) {
+	result := &OperationResult{}
+	filename := filepath.Join(a.paths.CVDBugReportsDir, uuid, "cvd_bugreport.zip")
+	cmd := cvd.NewCommand(
+		a.execContext,
+		[]string{"host_bugreport", "--output=" + filename},
```
jemoreira, line 69: FYI: This command will block reading from `stdin` if there are more than one group running.

frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -76,6 +76,9 @@ func (c *Controller) AddRoutes(router *mux.Router) {
 	// reached. Be prepared to retry if the deadline was reached.
 	router.Handle("/operations/{name}/:wait",
 		httpHandler(&waitOperationHandler{c.OperationManager, c.WaitOperationDuration})).Methods("POST")
+	router.Handle("/cvdbugreports",
```
jemoreira, line 79: Despite the (misleading) command name, host bug reports in `cvd` are taken from a group, not the entire host. There should be a way to associate the bugreport to the group it was taken from, both at creation time and later. The best way to do this is to make the bugreports a subresource of `/cvds`

frontend/src/host_orchestrator/orchestrator/controller.go
jemoreira, line 81: For completion, add a `DELETE /{bugreportspath}/{uuid}` endpoint.

frontend/src/liboperator/api/v1/messages.go
```diff
@@ -227,3 +227,7 @@ type EmptyResponse struct{}
 
 type StopCVDResponse = EmptyResponse
 
+type UUID struct {
```
jemoreira, line 230: A full struct to represent a uuid is overkill, just a string would be enough.

jemoreira (changes requested):

frontend/src/host_orchestrator/orchestrator/controller.go
ser-io: good catch, just carried the POST, changed it to GET.

frontend/src/host_orchestrator/orchestrator/createcvdbugreportaction.go
ser-io: good point, made the group name required.

frontend/src/host_orchestrator/orchestrator/controller.go
ser-io: Good point,

The new API looks like:

```
#  `:bugreport` is an operation of the cvd resource which creates the `cvdbugreport` resource.
POST /cvds/{group}/:bugreport

#  then we add new endpoints handling `cvdbugreport` as a main resource.
GET /cvdbugreport/{uuid}
DELETE /cvdbugreport/{uuid}
```

frontend/src/liboperator/api/v1/messages.go
ser-io: True, was under the impression a single string was not a valid JSON value.

frontend/src/host_orchestrator/orchestrator/controller.go
ser-io: Done

jemoreira (approved):

PR #615, opened 2024-07-16
Add grub-efi-amd64-bin into control

410063005:
Make the generated *.base.deb can be installed on an AMD64 Ubuntu 20.04 machine with grub-efi-amd64-bin (not grub-efi-ia32-bin)


adelva1984:
I think this change is wrong because our current bootloader structure uses U-Boot to chain into GRUB, and U-Boot runs in 32-bit mode on x86 and doesn't support 64-bit EFI apps. Therefore replacing this dependency will cause only the 64-bit binary to be installed on some systems and break otheros support. Why is this ia32 package not available in Ubuntu 20.04?

PR #617, opened 2024-07-16
Use iptables when iptables-legacy isn't available

jemoreira:
and ensure /run and /var/run exist before using them

### DIFF
--- a/base/debian/cuttlefish-base.cuttlefish-host-resources.init
+++ b/base/debian/cuttlefish-base.cuttlefish-host-resources.init
@@ -59,6 +59,8 @@ ebtables=${ebtables:-ebtables}
 iptables=$(which iptables-legacy)
 iptables=${iptables:-iptables}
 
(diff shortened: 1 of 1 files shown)


base/debian/cuttlefish-base.cuttlefish-host-resources.init
```diff
@@ -55,6 +55,12 @@ ebtables=$(which ebtables-legacy)
 # If there's no ebtables-legacy, it's probably an older distro,
 # and ebtables should support broute directly
 ebtables=${ebtables:-ebtables}
+# Similarly for iptables
+iptables=$(which iptables-legacy)
+iptables=${iptables:-iptables}
+
+# Make sure these directories exist, since they are used multiple times
+mkdir -p /run /var/run
```
Databean, line 63: Should we have a `set -e` somewhere at the top in case this fails?

Databean (approved):

WeiChungHsu (approved):

PR #618, opened 2024-07-17
Fix double closure of sockets while proxying ADB

k311093:
While proxying ADB, wsWrapper.Close() is executed twice, at deferred function and inside go routine. Call wsWrapper.Close() only at go routine that writes data to the wsWrapper.

Also move tcpConn.Close() after io.Copy(...) to avoid confusion.

### DIFF
--- a/frontend/src/liboperator/operator/operator.go
+++ b/frontend/src/liboperator/operator/operator.go
@@ -27,6 +27,7 @@ import (
 	"net/url"
 	"os"
 	"strconv"
+	"strings"
 
 	apiv1 "github.com/google/android-cuttlefish/frontend/src/liboperator/api/v1"
 	"github.com/gorilla/mux"
@@ -540,10 +541,17 @@ func adbProxy(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
 
 	// Redirect WebSocket to ADB tcp socket
 	go func() {
-		io.Copy(wsWrapper, tcpConn)
-		wsWrapper.Close()
+		// TODO: Replace with checking net.ErrClosed after Go 1.16
+		if _, err := io.Copy(wsWrapper, tcpConn); err != nil && !strings.Contains(err.Error(), "use of closed network connection") {
+			log.Print("Error while io.Copy from ADB to WebSocket: ", err)
+		}
+		if err = wsWrapper.Close(); err != nil {
+			log.Print("Error while closing WebSocket: ", err)
+		}
 	}()
-	io.Copy(tcpConn, wsWrapper)
+	if _, err = io.Copy(tcpConn, wsWrapper); err != nil {
+		log.Print("Error while io.Copy from WebSocket to ADB: ", err)
+	}
 }
 
 // Wrapper for implementing io.ReadWriteCloser of websocket.Conn
@@ -561,6 +569,9 @@ func (w *wsIoWrapper) Read(p []byte) (int, error) {
 	if w.buf == nil || w.pos >= len(w.buf) {
 		_, readBuf, err := w.wsConn.ReadMessage()
 		if err != nil {
+			if websocket.IsUnexpectedCloseError(err, websocket.CloseGoingAway) {
+				return 0, io.EOF
+			}
 			return 0, err
 		}
 		w.buf = readBuf
--- a/frontend/src/liboperator/operator/operator.go
+++ b/frontend/src/liboperator/operator/operator.go
@@ -537,7 +537,6 @@ func adbProxy(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
 		pos:    0,
 		buf:    nil,
 	}
-	defer wsWrapper.Close()
 
 	// Redirect WebSocket to ADB tcp socket
 	go func() {


frontend/src/liboperator/operator/operator.go
```diff
@@ -513,7 +513,6 @@ func adbProxy(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
 		log.Print("Error while connect to ADB: ", err)
 		return
 	}
-	defer tcpConn.Close()
```
ser-io, line 516: Suggest to rather keep the two defer calls: `defer tcpConn.Close()` and `defer wsWrapper.Close()` as they are and remove `wsWrapper.Close()` within the goroutine. I prefer this approach as it's a common Golang idiom when working with `Closeable` instances.

ser-io (changes requested):
As a follow up, we should log the errors of  `wsWrapper.Close()`, `io.Copy(wsWrapper, tcpConn)` and `tcpConn.Close()` calls.

frontend/src/liboperator/operator/operator.go
jemoreira: The `defer tcpConn.Close()` should stay, but the one for `wsWrapper` needs to go in the goroutine. The reason is that the copy from `wsWrapper` to `tcpConn` can finish before the goroutine completes writing to `wsWrapper`, which will lead to `wsWrapper` being closed in the middle of a write operation. To avoid this, sockets should be closed in the goroutine they're written to. Alternatively, you can keep the defer for `wsWrapper` but must ensure the function doesn't return until the goroutine finishes.

k311093: Changed to keep tcpConn.Close() and close wsWrapper inside goroutine.

frontend/src/liboperator/operator/operator.go
```diff
@@ -537,14 +538,24 @@ func adbProxy(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
...
-		io.Copy(wsWrapper, tcpConn)
-		wsWrapper.Close()
+		_, err := io.Copy(wsWrapper, tcpConn)
+		if err != nil {
+			if !strings.Contains(err.Error(), "use of closed network connection") {
+				log.Print("Error while io.Copy from ADB to WebSocket: ", err)
+			}
+		}
+		err = wsWrapper.Close()
+		if err != nil {
```
jemoreira, line 551: nit:
```
if err := wsWrapper.Close(); err != nil {
...
}
```

frontend/src/liboperator/operator/operator.go
```diff
@@ -537,14 +538,24 @@ func adbProxy(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
...
 		buf:    nil,
 	}
-	defer wsWrapper.Close()
 
 	// Redirect WebSocket to ADB tcp socket
 	go func() {
-		io.Copy(wsWrapper, tcpConn)
-		wsWrapper.Close()
+		_, err := io.Copy(wsWrapper, tcpConn)
+		if err != nil {
```
jemoreira, line 545: nit:
```
if _, err := io.Copy(...); err != nil {
...
}
```

frontend/src/liboperator/operator/operator.go
```diff
@@ -537,14 +538,24 @@ func adbProxy(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
...
+				log.Print("Error while io.Copy from ADB to WebSocket: ", err)
+			}
+		}
+		err = wsWrapper.Close()
+		if err != nil {
+			log.Print("Error while closing WebSocket: ", err)
+		}
 	}()
-	io.Copy(tcpConn, wsWrapper)
+	_, err = io.Copy(tcpConn, wsWrapper)
```
jemoreira, line 555: nit:
```
if _, err := io.Copy(...); err != nil {
...
}
```

jemoreira (approved):

frontend/src/liboperator/operator/operator.go
k311093: Done

frontend/src/liboperator/operator/operator.go
k311093: Done

frontend/src/liboperator/operator/operator.go
k311093: Done

k311093:
Hi @ser-io , can you please review this PR?

frontend/src/liboperator/operator/operator.go
```diff
@@ -540,10 +541,18 @@ func adbProxy(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
 
 	// Redirect WebSocket to ADB tcp socket
 	go func() {
-		io.Copy(wsWrapper, tcpConn)
-		wsWrapper.Close()
+		if _, err := io.Copy(wsWrapper, tcpConn); err != nil {
+			if !strings.Contains(err.Error(), "use of closed network connection") {
```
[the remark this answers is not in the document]
k311093: I tried to use net.ErrClosed, but it failed to build at golang 1.15 because net.ErrClosed was introduced later.

I added a TODO comment about that.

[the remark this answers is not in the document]
ser-io: sounds good.

nit: let's move the condition to the above if though, e.g ` err != nil &&  !strings.Contains(err.Error(), "use of closed network connection") {` and have one `if` only.

ser-io (approved):

frontend/src/liboperator/operator/operator.go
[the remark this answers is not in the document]
k311093: Changed condition statements to one line.

PR #623, opened 2024-07-19
base: debian: control: grub-efi-ia32-bin should also be !arm64

grandpaul:
When adding grub-efi-amd64-bin to cuttlefish-base, it accidentally made grub-efi-ia32-bin to be installed to all architectures.

Mark grub-efi-ia32-bin to !arm64 again so that it can be installed on arm64 platform.


base/debian/control
```diff
@@ -35,7 +35,7 @@ Depends: adduser,
          ebtables-legacy | ebtables,
          f2fs-tools,
          grub-efi-arm64-bin [arm64],
-         grub-efi-ia32-bin | grub-efi-amd64-bin [!arm64],
+         grub-efi-ia32-bin [!arm64] | grub-efi-amd64-bin [!arm64],
```
jemoreira, line 38: A comment on the other PR mentioned that depending on the amd64 version was wrong: https://github.com/google/android-cuttlefish/pull/615#issuecomment-2233833959

We should just revert it to depending on the `ia32` package only.

jemoreira (changes requested):

base/debian/control
WeiChungHsu: Please revert it, because this blocks the ARM64 image/packages build. Thanks!

grandpaul:
The latest commit builds on arm64 again. I'd like to close this pull request because it is not needed anymore. Thanks a lot.

PR #628, opened 2024-07-23
Use lesser id first while allocating id for CVD

k311093:
While launching device using automatic device id allocator, device id is not predictable because it allocates id from unordered_set. Change it to use ordered set to allocate lesser id first.

### DIFF
--- a/base/cvd/cuttlefish/common/libs/utils/unique_resource_allocator.h
+++ b/base/cvd/cuttlefish/common/libs/utils/unique_resource_allocator.h
@@ -20,6 +20,7 @@
 #include <memory>
 #include <mutex>
 #include <optional>
+#include <set>
 #include <type_traits>
 #include <unordered_set>
 #include <utility>
@@ -281,13 +282,13 @@ class UniqueResourceAllocator {
    *
    * The itr must belong to available_resources_.
    */
-  const T* RemoveFromPool(const typename std::unordered_set<T>::iterator itr) {
+  const T* RemoveFromPool(const typename std::set<T>::iterator itr) {
     T tmp = std::move(*itr);
     available_resources_.erase(itr);
     const auto [new_itr, _] = allocated_resources_.insert(std::move(tmp));
     return std::addressof(*new_itr);
   }
-  std::unordered_set<T> available_resources_;
+  std::set<T> available_resources_;
   std::unordered_set<T> allocated_resources_;
   std::mutex mutex_;
 };


0405ysj (approved):

ikicha (commented):
it looks like it's shared code between this repo and AOSP, should we update both?

https://cs.android.com/android/platform/superproject/main/+/main:device/google/cuttlefish/common/libs/utils/unique_resource_allocator.h;l=1?q=device%2Fgoogle%2Fcuttlefish%2Fcommon%2Flibs%2Futils%2Funique_resource_allocator.h&sq=

and also can we make a logic to find proper adb dynamic(it means we need an API to query assigned instance num) instead of assuming it's starting with instance_num 1...

k311093:
> it looks like it's shared code between this repo and AOSP, should we update both?
> 
> https://cs.android.com/android/platform/superproject/main/+/main:device/google/cuttlefish/common/libs/utils/unique_resource_allocator.h;l=1?q=device%2Fgoogle%2Fcuttlefish%2Fcommon%2Flibs%2Futils%2Funique_resource_allocator.h&sq=
> 

It looks like we need to update both, but need to check it need to be manually merged. @Databean 

> and also can we make a logic to find proper adb dynamic(it means we need an API to query assigned instance num) instead of assuming it's starting with instance_num 1...

It would be possible to make a logic to find ports by using `/cvds` HO API, let me update the logic as follow-ups after competing 
 support canonical config for the script.

Databean:
It's fine to update common files either here or in AOSP, but note that the next merge from AOSP will overwrite any changes made here. If you make changes to common files here I recommend sending a corresponding CL to AOSP soon afterwards.

Merges are done with the [`minimerge` script](https://github.com/google/android-cuttlefish/blob/e123d86c45a6f03252301917973466b1572dad7a/tools/minimerge/import.sh) based on a [mapping of shared files](https://github.com/google/android-cuttlefish/blob/main/tools/minimerge/cuttlefish_args.txt). The script will add commits to your current `git` branch based on AOSP commits since the last merge.

k311093:
> It's fine to update common files either here or in AOSP, but note that the next merge from AOSP will overwrite any changes made here. If you make changes to common files here I recommend sending a corresponding CL to AOSP soon afterwards.
> 
> Merges are done with the [`minimerge` script](https://github.com/google/android-cuttlefish/blob/e123d86c45a6f03252301917973466b1572dad7a/tools/minimerge/import.sh) based on a [mapping of shared files](https://github.com/google/android-cuttlefish/blob/main/tools/minimerge/cuttlefish_args.txt). The script will add commits to your current `git` branch based on AOSP commits since the last merge.

Thanks for the answer :)

I'll try to upload this change to the AOSP after this PR is merged.

jemoreira:
> it looks like it's shared code between this repo and AOSP, should we update both?
> 
> https://cs.android.com/android/platform/superproject/main/+/main:device/google/cuttlefish/common/libs/utils/unique_resource_allocator.h;l=1?q=device%2Fgoogle%2Fcuttlefish%2Fcommon%2Flibs%2Futils%2Funique_resource_allocator.h&sq=
> 
> and also can we make a logic to find proper adb dynamic(it means we need an API to query assigned instance num) instead of assuming it's starting with instance_num 1...

yes, please update the aosp code too

PR #631, opened 2024-07-24
Remove unique_resource_allocator and its tests from minimerge import list

k311093:
unique_resource_allocator is only used by cvd at GitHub side, not for AOSP. So remove unique_resource_allocator related code from AOSP and also remove from import list.


Databean (approved):

Databean:
Thanks!

PR #632, opened 2024-07-24
Support user artifact path for CVD creation with canonical config

k311093:
Add support for the notation `@user_artifact` to specify user artifact path at canonical config.

Example config for using user artifact while launching CVD with canonical config.

```
{
  "env_config": {
    "common": {
      "host_package": "@user_artifact/$user_artifact_id"
    },
    "instances": [
      {
        "vm": {
          "memory_mb": 8192,
          "setupwizard_mode": "OPTIONAL",
          "cpus": 8
        },
        "disk": {
          "default_build": "@user_artifact/$user_artifact_id"
        }
      }
    ]
  }
}
```

### DIFF
--- a/frontend/src/host_orchestrator/orchestrator/createcvdaction.go
+++ b/frontend/src/host_orchestrator/orchestrator/createcvdaction.go
@@ -15,6 +15,7 @@
 package orchestrator
 
 import (
+	"bytes"
 	"encoding/json"
 	"io/ioutil"
 	"log"
@@ -123,6 +124,9 @@ func (a *CreateCVDAction) launchWithCanonicalConfig(op apiv1.Operation) (*apiv1.
 	if err != nil {
 		return nil, err
 	}
+	data = bytes.ReplaceAll(data,
+		[]byte(apiv1.EnvConfigUserArtifactsVar + "/"),
+		[]byte(a.userArtifactsDirResolver.GetDirPath("")))
 	configFile, err := createTempFile("cvdload*.json", data, 0640)
 	if err != nil {
 		return nil, err
--- a/frontend/src/liboperator/api/v1/messages.go
+++ b/frontend/src/liboperator/api/v1/messages.go
@@ -110,6 +110,9 @@ type AndroidCIBundle struct {
 	Type ArtifactsBundleType `json:"type"`
 }
 
+// Prefix for specifying user artifact path while creating CVD with CreateCVDRequest.EnvConfig.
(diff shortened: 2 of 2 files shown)


ikicha (approved):

0405ysj (approved):

frontend/src/host_orchestrator/orchestrator/createcvdaction.go
```diff
@@ -118,8 +119,13 @@ func (a *CreateCVDAction) launchCVD(op apiv1.Operation) {
 	}
 }
 
+const userArtifactPrefix = "@user_artifact/"
```
ser-io, line 122: Let's move this constant to our public API and add the relevant documentation to it. Let's put it closer to the `CreateCVDRequest` struct https://github.com/google/android-cuttlefish/blob/main/frontend/src/liboperator/api/v1/messages.go#L114.

In the context of the public API, I'd recommend constant name: `EnvConfigUserArtifactsVar`, using only the notation literal `@user_artifacts` as the value and leaving the `/` character to the internal replace logic.

ser-io (changes requested):

frontend/src/host_orchestrator/orchestrator/createcvdaction.go
```diff
@@ -118,8 +119,13 @@ func (a *CreateCVDAction) launchCVD(op apiv1.Operation) {
...
+
 func (a *CreateCVDAction) launchWithCanonicalConfig(op apiv1.Operation) (*apiv1.CreateCVDResponse, error) {
 	data, err := json.MarshalIndent(a.req.EnvConfig, "", " ")
+	data = bytes.ReplaceAll(data,
```
ser-io, line 126: let's move this after the error handling: `if err != nil { ...` lines

ser-io (changes requested):

frontend/src/host_orchestrator/orchestrator/createcvdaction.go
k311093: Moved `userArtifactPrefix` to `messages.go` and renamed it to `EnvConfigUserArtifactsVar`

frontend/src/host_orchestrator/orchestrator/createcvdaction.go
k311093: Moved ReplaceAll(...) code after error handling.

ser-io (approved):

PR #636, opened 2024-07-27
cvd fleet prints full info for non-running instances

jemoreira:
By not relying only on the output of cvd_internal_status, but instead keeping the relevant information in the database.
Some info is still parsed from the status command output the first time it's run (right after a successful boot) and then stored in the database: webrtc_device_id and adb_port.
Additionally, cvd fleet prints the adb port on its own, which is more useful than the serial for remote use cases.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
@@ -23,6 +23,7 @@
 #include <string>
 #include <vector>
 
+#include <android-base/parseint.h>
 #include <android-base/strings.h>
 #include <fmt/core.h>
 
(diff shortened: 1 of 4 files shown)


base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
```diff
@@ -75,6 +76,49 @@ std::string HumanFriendlyStateName(cvd::InstanceState state) {
...
 
+// Adds more information to the json object returned by cvd_internal_status,
+// including some that cvd_internal_status normally returns but doesn't when the
+// instance is not running.
+void EnrichInstanceJson(const selector::LocalInstanceGroup& group,
+                        const cvd::Instance& instance,
+                        Json::Value& instance_json) {
+  CHECK(instance_json.isMember("instance_name") &&
+        instance_json["instance_name"].isString())
+      << "JSON returned from status fetcher has unexpected format";
```
Databean, line 87: Should this function return a `Result` instead rather than `abort`ing based on invalid output from a different executable?

base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
```diff
@@ -188,11 +235,12 @@ Result<StatusFetcherOutput> StatusFetcher::FetchOneInstanceStatus(
   auto response = ResponseFromSiginfo(infop);
   if (response.status().code() != cvd::Status::OK) {
     instance.set_state(cvd::INSTANCE_STATE_UNREACHABLE);
-    instance_manager_.UpdateInstance(group, instance);
     instance_status_json["warning"] = "cvd status failed";
-    instance_status_json["instance_name"] = instance.name();
-    instance_status_json["status"] = HumanFriendlyStateName(instance.state());
+  } else {
+    CF_EXPECT(UpdateInstanceWithStatusResult(instance, instance_status_json));
```
Databean, line 240: nit: can the `if` be the positive conditional, i.e.
```
if (v1 == v2) {
  ...
} else {
  ...
}
```
rather than
```
if (v1 != v2) {
  ...
} else {
  ...
}
```
See [go/tott/685](http://go/tott/685)

base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
```diff
@@ -75,6 +76,49 @@ std::string HumanFriendlyStateName(cvd::InstanceState state) {
...
+  instance_json["instance_name"] = instance.name();
+  instance_json["status"] = HumanFriendlyStateName(instance.state());
+  instance_json["adb_port"] = instance.adb_port();
+  instance_json["assembly_dir"] = group.AssemblyDir();
+  instance_json["instance_dir"] = group.InstanceDir(instance);
+  instance_json["instance_name"] = instance.name();
+  instance_json["webrtc_device_id"] = instance.webrtc_device_id();
+  instance_json["web_access"] =
+      fmt::format("https://localhost:1443/devices/{}/files/client.html",
+                  instance.webrtc_device_id());
```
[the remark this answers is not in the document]
jemoreira: Nope, fixed to only include this for running instances.

base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
```diff
@@ -75,6 +76,49 @@ std::string HumanFriendlyStateName(cvd::InstanceState state) {
...
+  CHECK(instance_json.isMember("instance_name") &&
+        instance_json["instance_name"].isString())
+      << "JSON returned from status fetcher has unexpected format";
+  instance_json["instance_name"] = instance.name();
+  instance_json["status"] = HumanFriendlyStateName(instance.state());
+  instance_json["adb_port"] = instance.adb_port();
+  instance_json["assembly_dir"] = group.AssemblyDir();
+  instance_json["instance_dir"] = group.InstanceDir(instance);
+  instance_json["instance_name"] = instance.name();
+  instance_json["webrtc_device_id"] = instance.webrtc_device_id();
```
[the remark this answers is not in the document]
jemoreira: I don't like "Enrich" that much either, any suggestion is welcome.

This code intentionally overrides fields from `cvd_internal_status` for two reasons:
1. `cvd` knows more about the host than the cuttlefish devices does. For example, `cvd` knows where the host orchestrator is, if that ever changes cvd can easily change at the same time because they are released together, from the same source; the AOSP code can't react to those changes easily.
2. This JSON is `cvd`'s output, it's better to have `cvd` be aware of and enforce its own interface instead of relying on `cvd_internal_status` from who-knows-what-branch. My intention is evolve this to only rely on `cvd_internal_status`'s exit code for the instance and other information that can change while it's running (displays), but for everything that stays constant after boot (ADB, webrtc_id, etc) rely on the instance database, but it'll require more work to get there.

base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp

base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
jemoreira: This check shouldn't be here, the code below doesn't even access that member from the JSON object.

base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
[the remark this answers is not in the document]
Databean: `OverrideInstanceJson`?

Databean (approved):

PR #640, opened 2024-07-31
Release instance id lock on interrupt

jemoreira:


base/cvd/cuttlefish/host/commands/cvd/server_command/start.cpp
```diff
@@ -755,6 +755,14 @@ Result<cvd::Response> CvdStartCommandHandler::Handle(
                         "`cvd reset` to ensure a clean state";
         }
 
+        for (const auto& instance: group.Instances()) {
+          InstanceLockFileManager ilfm;
+          if (instance.id() > 0) {
+            // Do this before the id is reset in the next block
+            ilfm.RemoveLockFile(instance.id());
+          }
+        }
```
Databean, line 764: If this is deleting the locks indicating some instance ids are in use, should those instances be shut down too?

Databean (changes requested):

base/cvd/cuttlefish/host/commands/cvd/server_command/start.cpp
jemoreira: that happens on line 750: `subprocess_waiter_.Interrupt()`

Databean: That `Interrupt()` will `SIGKILL` the process, does that guarantee all the instance subprocesses are shut down in `--daemon` mode?

jemoreira: In most cases it does. There is a small window in which the subprocess gets killed and the instances remain, that's why there is warning below.

Databean (approved):

PR #642, opened 2024-07-31
Fix shortcomings in cvd fleet

jemoreira:
Prints more information for cancelled instances.

Don't print adb port and webrtc id when the instance isn't running.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
@@ -83,17 +83,17 @@ void OverrideInstanceJson(const selector::LocalInstanceGroup& group,
                         Json::Value& instance_json) {
   instance_json["instance_name"] = instance.name();
   instance_json["status"] = HumanFriendlyStateName(instance.state());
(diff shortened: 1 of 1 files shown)


base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
```diff
@@ -83,11 +83,14 @@ void OverrideInstanceJson(const selector::LocalInstanceGroup& group,
                         Json::Value& instance_json) {
   instance_json["instance_name"] = instance.name();
   instance_json["status"] = HumanFriendlyStateName(instance.state());
-  instance_json["adb_port"] = instance.adb_port();
+  if (instance.id() > 0) {
+    // These only make sense for active instances
+    instance_json["webrtc_device_id"] = instance.webrtc_device_id();
+    instance_json["adb_port"] = instance.adb_port();
+  }
```
Databean, line 90: Can this be combined with the other `if (instance.id() > 0) {` block?

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp

PR #646, opened 2024-08-01
Always output when oauth file is checked in `cvd fetch`

cjreynol:
When the `.acloud_oauth2.dat` file is implicitly used there is no message to the user on default verbosity settings.  If that file is outdated users get a credential failure even though no credential flags were provided.  That makes it difficult to debug without code familiarity.

Bug: 356892183
Test: cvd fetch --target_directory=/tmp/cvd/fetch_test --default_build=aosp-main

### DIFF
--- a/base/cvd/cuttlefish/host/libs/web/credential_source.cc
+++ b/base/cvd/cuttlefish/host/libs/web/credential_source.cc
@@ -73,20 +73,22 @@ Result<std::unique_ptr<CredentialSource>> GetCredentialSourceLegacy(
   if (credential_source == "gce") {
     result = GceMetadataCredentialSource::Make(http_client);
   } else if (credential_source == "") {
-    LOG(VERBOSE) << "Probing oauth credentials at " << oauth_filepath;
     if (FileExists(oauth_filepath)) {
       std::string oauth_contents = CF_EXPECT(ReadFileContents(oauth_filepath));
       auto attempt_load = RefreshCredentialSource::FromOauth2ClientFile(
           http_client, oauth_contents);
       if (attempt_load.ok()) {
         result.reset(new RefreshCredentialSource(std::move(*attempt_load)));
+        LOG(INFO) << "\"" << oauth_filepath
+                  << "\" was found, using that as credentials";
       } else {
-        LOG(ERROR) << "Failed to load oauth2 credentials: "
-                   << attempt_load.error().FormatForEnv();
+        LOG(ERROR) << "Failed to load oauth credentials from \""
+                   << oauth_filepath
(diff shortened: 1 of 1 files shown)


base/cvd/cuttlefish/host/libs/web/credential_source.cc
```diff
@@ -73,20 +73,22 @@ Result<std::unique_ptr<CredentialSource>> GetCredentialSourceLegacy(
...
-        LOG(ERROR) << "Failed to load oauth2 credentials: "
-                   << attempt_load.error().FormatForEnv();
+        LOG(ERROR) << "Failed to load oauth credentials from \""
+                   << oauth_filepath
+                   << "\":" << attempt_load.error().FormatForEnv();
       }
     } else {
       LOG(INFO) << "\"" << oauth_filepath
-                << "\" missing, running without credentials";
+                << "\" file is missing, running without credentials";
```
jemoreira, line 91: nit: I find `"/path/to/oauth/file/oauth.dat" file is missing` redundant, just `"/path....dat" is missing` sounds better IMHO.

base/cvd/cuttlefish/host/libs/web/credential_source.cc
```diff
@@ -73,20 +73,22 @@ Result<std::unique_ptr<CredentialSource>> GetCredentialSourceLegacy(
...
   } else if (credential_source == "") {
-    LOG(VERBOSE) << "Probing oauth credentials at " << oauth_filepath;
     if (FileExists(oauth_filepath)) {
       std::string oauth_contents = CF_EXPECT(ReadFileContents(oauth_filepath));
       auto attempt_load = RefreshCredentialSource::FromOauth2ClientFile(
           http_client, oauth_contents);
       if (attempt_load.ok()) {
         result.reset(new RefreshCredentialSource(std::move(*attempt_load)));
+        LOG(INFO) << "\"" << oauth_filepath
+                  << "\" file found, using that as credentials";
```
jemoreira, line 83: nit: same comment as below about dropping the word `file`.

jemoreira (approved):

base/cvd/cuttlefish/host/libs/web/credential_source.cc
cjreynol: Done.  Also removed "file" file from found case.

base/cvd/cuttlefish/host/libs/web/credential_source.cc
cjreynol: Done.

Databean (approved):

PR #654, opened 2024-08-05
Separate stop and generic command handlers

jemoreira:
While there was some code in common the generic.cpp file had many special cases just to handle stop. Getting rid of those improves readability enough that it's OK to tolerate some repeated code between the two.


base/cvd/cuttlefish/host/commands/cvd/server_command/stop.cpp
```diff
@@ -0,0 +1,304 @@
...
+#include "host/commands/cvd/server_command/utils.h"
+#include "host/commands/cvd/types.h"
+
+namespace cuttlefish {
+namespace {
+
+constexpr char kSummaryHelpText[] =
+    "Run cvd stop --help for command description";
+
+}  // namespace
```
Databean, line 51: `CvdStopCommandHandler` could be in the unnamed namespace as well, only `NewCvdStopCommandHandler` needs to be outside.

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/server_command/stop.cpp
jemoreira: ACK.

I just copied everything from generic.cpp and deleted the things not relevant to the stop command. I think moving things to the empty namespace should include all other commands and be done in a separate PR.

PR #655, opened 2024-08-05
Fix host_bugreport when ANDROID_HOST_OUT is not set.

ser-io:
- Command `host_bugreport` uses the ANDROID_HOST_OUT env variable to locate the `adb`, when the variable is not set, `cvd` uses the HOME value instead.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/server_command/generic.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/server_command/generic.cpp
@@ -380,6 +380,7 @@ CvdGenericCommandHandler::ExtractInfo(const RequestWithStdio& request) {
                                   .args = cmd_args,
                                   .envs = envs};
   result.envs["HOME"] = home;
+  result.envs[kAndroidHostOut] = android_host_out;
   extracted_info.invocation_info = result;
   extracted_info.group = instance_group;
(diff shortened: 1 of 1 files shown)


base/cvd/cuttlefish/host/commands/cvd/server_command/generic.cpp
```diff
@@ -174,7 +174,10 @@ Result<cvd::Response> CvdGenericCommandHandler::Handle(
   if (is_non_help_cvd && ui_response_type == UiResponseType::kNoTTY) {
     return CF_EXPECT(NoTTYResponse(request));
   }
-
+  if (group_opt) {
+    // Update `ANDROID_HOST_OUT` with selected group's host artifacts path.
+    invocation_info.envs[kAndroidHostOut] = group_opt->HostArtifactsPath();
```
jemoreira, line 179: nit: `ExtractInfo` is a better place for this, around line 398.

jemoreira (approved):

base/cvd/cuttlefish/host/commands/cvd/server_command/generic.cpp
ser-io: good point, done.

PR #656, opened 2024-08-06
Let `CvdServerHandler` instances create their own `SubprocessWaiter`s

Databean:
These don't need to be shared between instances, and each instance can only handle one process at a time anyway.


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #659, opened 2024-08-07
Don't use `SubprocessWaiter` in uninterruptible contexts

Databean:
Since PR #517 removed interrupt handling, a lot of classes have been unnecessarily using `SubprocessWaiter`, which provides additional functionality around being interrupted.


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #661, opened 2024-08-08
Split cvd start into create and start

jemoreira:
Split cvd start functionality into two commands

The commands are
- create: Creates the new instance groups, allocates instance ids and its
locks under /tmp/acloud_cvd_temp. There is an option to start groups
immediately after creation, enabled by default.
- start:  Starts existing instance groups that are not currently running.


base/cvd/cuttlefish/host/commands/cvd/server_command/create.cpp
```diff
@@ -0,0 +1,339 @@
...
+  for (const auto& arg : args) {
+    load_command.add_args(arg);
+  }
+  return RequestWithStdio(request_proto, request.FileDescriptors());
+}
+
+Result<cvd_common::Envs> GetEnvs(const RequestWithStdio& request) {
+  cvd_common::Envs envs =
+      cvd_common::ConvertToEnvs(request.Message().command_request().env());
+  if (envs.at("HOME").empty()) {
```
Databean, line 164: `at` throws a `std::out_of_range` exception on access failure and this happens before the `Contains` check, so it could crash in an unexpected way.

base/cvd/cuttlefish/host/commands/cvd/server_command/create.cpp
```diff
@@ -0,0 +1,339 @@
...
+
+constexpr char kDetailedHelpText[] =
+    R"""(
+Usage:
+cvd create [--product_path=PATH] [--host_path=PATH] [--[no]start] [START_ARGS]
+cvd create --config_file=PATH [--[no]start]
+
+Creates and starts a new cuttlefish instance group.
+
+--host_path=PATH     The path to the directory containing the Cuttlefish Host
```
ser-io, line 60: what do you think of `host_out=PATH` and `product_out=PATH`? the terms `host_out` and `product_out` might be self explanatory if you are familiar with Android development.

base/cvd/cuttlefish/host/commands/cvd/server_command/create.cpp
```diff
@@ -0,0 +1,339 @@
...
+RequestWithStdio CreateStartCommand(const RequestWithStdio& request,
+                                    const selector::LocalInstanceGroup& group,
+                                    const cvd_common::Args& args,
+                                    const cvd_common::Envs& envs) {
+  cvd::Request request_proto;
+  auto& load_command = *request_proto.mutable_command_request();
+  for (const auto& [key, value] : envs) {
+    (*load_command.mutable_env())[key] = value;
+  }
+  load_command.set_working_directory(
```
ser-io, line 148: nit: `start_command`.

I'd recommend  `command` only as there's no other command instance in this function, same in `CreateLoadCommand`

ser-io (changes requested):

frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -346,8 +346,8 @@ type startCVDParams struct {
 	BootloaderDir string
 }
 
-func (h *startCVDHandler) Start(p startCVDParams) error {
-	args := []string{groupNameArg, "start", daemonArg, reportAnonymousUsageStatsArg}
+func (h *startCVDHandler) Create(p startCVDParams) error {
```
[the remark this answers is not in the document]
jemoreira: yes, but that causes a name conflict with the request handler somewhere else. I'll do the rename in another PR as it could get big and involve other things not related to this change.

base/cvd/cuttlefish/host/commands/cvd/server_command/create.cpp
```diff
@@ -0,0 +1,337 @@
...
+    groups.push_back(
+        CF_EXPECT(instance_manager_.CreateInstanceGroup(group_creation_info)));
+  } else {
+    auto& group = groups[0];
+    CF_EXPECTF(group.Instances().size() == group_creation_info.instances.size(),
+               "Mismatch in number of instances from analisys: {} vs {}",
+               group.Instances().size(), group_creation_info.instances.size());
+    // The instances don't have an id yet
+    for (size_t i = 0; i < group.Instances().size(); ++i) {
+      auto& instance = group.Instances()[i];
```
ser-io, line 268: out of curiosity, do we have any high level overview comment around the relationship between `instances`, instances_info` and `instance_file_lock` concepts, they seem to map 1:1, but still I cannot get the full picture from the code only, wondering whether we could add some comment about it if it does not exist. thanks.

jemoreira: Those need a refactor, not a comment.

- The cvd::Instance is the representation of an instance for cvd, it's the object stored in the database and what should be used almost everywhere.
- An instance_info was initial information about an instance before it was added to the database, the class doesn't exist anymore, but there is InstanceGroupInfo which is the same thing for groups. A LocalInstanceGroup is created from a InstanceGroupInfo.
- instance_file_lock is missnamed. It represents a lock on the instance id (not the instance itself) so that different instances don't use the same id.

I have done some refactors around this, but there is still some more to do.

base/cvd/cuttlefish/host/commands/cvd/server_command/create.cpp
```diff
@@ -0,0 +1,337 @@
...
+  return CurrentDirectory();
+}
+
+struct CreateFlags {
+  std::string host_path;
+  std::string product_path;
+  bool start;
+  std::string config_file;
+};
+Result<CreateFlags> ParseCommandFlags(const cvd_common::Envs& envs,
```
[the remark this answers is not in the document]
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/server_command/create.cpp
ser-io: I see, appreciate the context, thanks!

PR #662, opened 2024-08-08
Remove stdio file descriptors from RequestWithStdio

Databean:
In practice, these file descriptors were usually the same as the regular stdio file descriptors, but were sometimes `/dev/null`. This commit regresses on the `/dev/null` cases and collapses everything into the "inherit stdio" case, sometimes introducing extra output.


Databean:
Superseded by #664.

PR #663, opened 2024-08-08
Fixes create cvd with local image and add e2e test.

ser-io:


frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -76,11 +77,11 @@ func (s *CVDSelector) ToCVDCLI() []string {
 }
 
 // Creates a CVD execution context from a regular execution context.
-// If a non-empty user name is provided the returned execution context executes commands as that user.
-func newCVDExecContext(execContext ExecContext, user string) cvd.CVDExecContext {
-	if user != "" {
+// If a non-nil user is provided the returned execution context executes commands as that user.
+func newCVDExecContext(execContext ExecContext, usr *user.User) cvd.CVDExecContext {
```
jemoreira, line 81: nit: It's better to have two functions instead of accepting a possibly `nil` (or empty) parameter in a single function. Something like:
```
newCVDExecContext(execContext ExecContext)
newCVDExecContextWithUser(execContext ExecContext, user user.User)
```
The second one could add the args and the call the first one.

jemoreira (approved):

frontend/src/host_orchestrator/orchestrator/instancemanager.go
ser-io: Agree, I'll handle that in a follow up PR.

ser-io: https://github.com/google/android-cuttlefish/pull/665

PR #664, opened 2024-08-09
Replace `SharedFD`s in `RequestWithStdio` with `std::[io]stream&`s.

Databean:
These are more convenient to write to than `SharedFD`s. Since `cvd` has stopped using a background server, the only remaining cases were that the regular stdin, stdout, stderr streams were used, or no stream was being used at all.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/server_client.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/server_client.cpp
@@ -23,6 +23,7 @@
 #include "common/libs/fs/shared_fd.h"
 #include "common/libs/utils/result.h"
 #include "common/libs/utils/unix_sockets.h"
+#include "cuttlefish/host/commands/cvd/server_client.h"
 
 namespace cuttlefish {
 
@@ -64,7 +65,7 @@ Result<std::optional<RequestWithStdio>> GetRequest(const SharedFD& client) {
     LOG(DEBUG) << "Has credentials, uid=" << creds->uid;
   }
 
-  return RequestWithStdio(std::move(request), std::move(fds));
+  return RequestWithStdio::StdIo(std::move(request));
 }
 
 Result<void> SendResponse(const SharedFD& client,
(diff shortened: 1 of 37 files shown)


base/cvd/cuttlefish/host/commands/cvd/server_client.cpp
```diff
@@ -81,30 +82,44 @@ Result<void> SendResponse(const SharedFD& client,
...
 
-SharedFD RequestWithStdio::Out() const {
-  return fds_.size() > 1 ? fds_[1] : SharedFD();
+RequestWithStdio RequestWithStdio::NullIo(cvd::Request message) {
+  return RequestWithStdio(std::move(message), NullIn(), NullOut(), NullOut());
 }
 
-SharedFD RequestWithStdio::Err() const {
-  return fds_.size() > 2 ? fds_[2] : SharedFD();
+RequestWithStdio RequestWithStdio::InheritIo(cvd::Request message,
```
jemoreira, line 103: nit: `InheritIO` to match the style of `SharedFD` (instead of `InheritIo` and `SharedFd`)

base/cvd/cuttlefish/host/commands/cvd/server_command/stop.cpp
```diff
@@ -133,31 +131,28 @@ Result<selector::LocalInstanceGroup> CvdStopCommandHandler::PromptUserForGroup(
...
+  request.Out() << selector.Menu() << "\n";
   auto menu = selector.Menu();
 
-  CF_EXPECT_EQ(WriteAll(request.Out(), menu + "\n"), (ssize_t)menu.size() + 1);
   std::unique_ptr<InterruptibleTerminal> terminal_ =
-      std::make_unique<InterruptibleTerminal>(request.In());
+      std::make_unique<InterruptibleTerminal>();
 
-  const bool is_tty = request.Err()->IsOpen() && request.Err()->IsATTY();
+  const bool is_tty = isatty(2);
```
jemoreira, line 140: This should check `0` and `1`, since it uses `request.Out()` and `request.In()` below.

Given that the check is performed on those file descriptors, the code below should use `std::cin` and `std::cout` instead of `request.In()` and `request.Out()`.

jemoreira (approved):

base/cvd/cuttlefish/host/commands/cvd/server_command/stop.cpp
jemoreira: Never mind, this check is on `2` because it prints with colors later on `cerr`

cjreynol (approved):

base/cvd/cuttlefish/host/commands/cvd/server_client.cpp
Databean: The [style guide](https://google.github.io/styleguide/cppguide.html#General_Naming_Rules) writes

> For names written in mixed case (also sometimes referred to as "[camel case](https://en.wikipedia.org/wiki/Camel_case)" or "[Pascal case](https://en.wiktionary.org/wiki/Pascal_case)"), in which the first letter of each word is capitalized, prefer to capitalize abbreviations as single words, e.g., `StartRpc()` rather than `StartRPC()`

Databean:
Thanks for reviewing!

PR #665, opened 2024-08-09
Code health: avoid requiring to pass a nil value.

ser-io:
- Context: https://github.com/google/android-cuttlefish/pull/663#discussion_r1711971144
- New function `newCVDExecContext` avoid calls with explicit nil user values

### DIFF
--- a/frontend/src/host_orchestrator/orchestrator/instancemanager.go
+++ b/frontend/src/host_orchestrator/orchestrator/instancemanager.go
@@ -77,25 +77,30 @@ func (s *CVDSelector) ToCVDCLI() []string {
 }
 
 // Creates a CVD execution context from a regular execution context.
-// If a non-nil user is provided the returned execution context executes commands as that user.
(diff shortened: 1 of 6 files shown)


frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -99,6 +87,23 @@ func newCVDExecContext(execContext ExecContext, usr *user.User) cvd.CVDExecConte
 	}
 }
 
+// Creates a CVD execution context from a regular execution context
+// for a given user. If user is nil, the running process's user will be used.
+func newCVDExecContextWithUser(execContext ExecContext, usr *user.User) cvd.CVDExecContext {
+	if usr == nil {
+		return newCVDExecContext(execContext)
+	}
```
jemoreira, line 95: This defeats the purpose of the refactor. Either return an error on `nil` or make the parameter not a pointer.

jemoreira (changes requested):

PR #667, opened 2024-08-09
Make cvdStartHandler a function named CreateCVD

jemoreira:


frontend/src/host_orchestrator/orchestrator/instancemanager.go
```diff
@@ -347,7 +343,7 @@ type startCVDParams struct {
 	BootloaderDir string
 }
 
-func (h *startCVDHandler) Create(p startCVDParams) error {
+func CreateCVD(execCtx cvd.CVDExecContext, p startCVDParams) error {
```
ser-io, line 346: nit: `ctx`. 

`ctx` is pretty common in Golang for contexts instances, and in this function we don't have any other context variable to differentiate from.

ser-io (approved):

frontend/src/host_orchestrator/orchestrator/instancemanager.go
jemoreira: Done

PR #668, opened 2024-08-09
Fix resizing in webui when there is more than one display for a single device

matvore:


frontend/src/operator/webui/src/app/view-pane/view-pane.component.ts
```diff
@@ -110,6 +141,7 @@ export class ViewPaneComponent implements OnInit, OnDestroy, AfterViewInit {
     this.displaysService.getDeviceVisibilities().pipe(
       map(deviceVisibilityInfos =>
         deviceVisibilityInfos.map(info => {
+          console.log('new info:', info);
```
k311093, line 144: Can you please remove this message?

k311093 (changes requested):
Thanks for the change!

frontend/src/operator/webui/src/app/view-pane/view-pane.component.ts
matvore: Done.

I've also reverted the change in behavior re: custom icon panel assumed presence, and updated the comment. The justification I was using to change the behavior didn't make any sense, and it seems like it would be easy to make it work properly in a separate patch.

matvore:
I changed this again to reference the DisplayInfo type from the server_connector module, rather than add a new type. I think this is better.

k311093 (approved):

PR #672, opened 2024-08-12
Host Orchestrator supports new cvd flows

jemoreira:


frontend/src/host_orchestrator/orchestrator/cvd/cvd.go
```diff
@@ -87,11 +88,10 @@ func (c *Command) Run() error {
...
 
 func cvdEnv(androidHostOut string) []string {
-	env := []string{}
-	if androidHostOut != "" {
-		env = append(env, envVarAndroidHostOut+"="+androidHostOut)
+	if androidHostOut == "" {
+		return nil
 	}
-	return env
+	return append(os.Environ(), envVarAndroidHostOut+"="+androidHostOut)
```
ser-io, line 94: nit: I think it would be useful having the explanation written in the commit message as a note here.

ser-io (approved):

frontend/src/host_orchestrator/orchestrator/cvd/cvd.go
jemoreira: Done

PR #673, opened 2024-08-12
Don't require ANDROID_HOST_OUT on cvd start

jemoreira:
cvd start uses the host artifacts path from the group to be started, which is stored in the instance database. It's not necessary then to required ANDROID_HOST_OUT to be defined and point to a valid location.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/server_command/start.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/server_command/start.cpp
@@ -612,18 +612,17 @@ Result<cvd::Response> CvdStartCommandHandler::Handle(
                                .follow_symlink = false}));
     }
   }
-  CF_EXPECT(Contains(envs, kAndroidHostOut));
-  const auto bin = CF_EXPECT(FindStartBin(envs.at(kAndroidHostOut)));
-
   // update DB if not help
   // collect group creation infos
   CF_EXPECT(Contains(supported_commands_, subcmd),
             "subcmd should be start but is " << subcmd);
   const bool is_help = CF_EXPECT(IsHelpSubcmd(subcmd_args));
-  CF_EXPECT(ConsumeDaemonModeFlag(subcmd_args));
-  subcmd_args.push_back("--daemon=true");
 
   if (is_help) {
+    auto it = envs.find(kAndroidHostOut);
+    CF_EXPECT(it != envs.end());
+    const auto bin = CF_EXPECT(FindStartBin(it->second));
+
     Command command =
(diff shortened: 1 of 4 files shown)


base/cvd/cuttlefish/host/commands/cvd/server_command/start.cpp
```diff
@@ -643,6 +644,7 @@ Result<cvd::Response> CvdStartCommandHandler::Handle(
 
   CF_EXPECT(UpdateArgs(subcmd_args, group));
   CF_EXPECT(UpdateEnvs(envs, group));
+  const auto bin = CF_EXPECT(FindStartBin(group.HostArtifactsPath()));
```
Databean, line 647: [`FindStartBin`](https://github.com/google/android-cuttlefish/blob/19709a6a16df349034cc4426f2e8081a2386b66c/base/cvd/cuttlefish/host/commands/cvd/server_command/start.cpp#L483) suggests running `lunch` if it can't find the executable, does that still help?

base/cvd/cuttlefish/host/commands/cvd/server_command/start.cpp
```diff
@@ -612,18 +612,16 @@ Result<cvd::Response> CvdStartCommandHandler::Handle(
...
   // collect group creation infos
   CF_EXPECT(Contains(supported_commands_, subcmd),
             "subcmd should be start but is " << subcmd);
   const bool is_help = CF_EXPECT(IsHelpSubcmd(subcmd_args));
-  CF_EXPECT(ConsumeDaemonModeFlag(subcmd_args));
-  subcmd_args.push_back("--daemon=true");
 
   if (is_help) {
+    CF_EXPECT(Contains(envs, kAndroidHostOut));
+    const auto bin = CF_EXPECT(FindStartBin(envs.at(kAndroidHostOut)));
```
Databean, line 623: `at` can throw an exception and `Contains` + `at` is a double lookup. Can this use `find`?

See [go/totw/132#safe-lookup](https://go.corp.google.com/totw/132#safe-lookup)

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/server_command/start.cpp
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/server_command/start.cpp
jemoreira: Done

PR #676, opened 2024-08-13
Move tag stable commit id to the new tag commit id with stable status.

WeiChungHsu:
Move tag stable commit id to the new tag commit id with stable status.


tools/cuttlefish-host-image-installer/update_github_releases.sh
```diff
@@ -20,6 +20,19 @@ echo "$changes" > changelog
 
 STABLE_VERSION="v$version"
 
+echo "step 2: move stable release to new commit id"
+commit_id=$(git  rev-list -n 1 $STABLE_VERSION )
+stable_commit_id=$(git  rev-list -n 1 stable )
```
Databean, line 25: nit: extra space in `git  rev-list`

tools/cuttlefish-host-image-installer/update_github_releases.sh
```diff
@@ -20,6 +20,19 @@ echo "$changes" > changelog
...
+if [[ "$commit_id" == "$stable_commit_id" ]]; then
+  echo "same commit, no need to change code base"
+else
+  git tag -d stable
+  git push -d origin stable
+  git push origin :refs/tags/stable
+  git tag stable $commit_id
```
Databean, line 32: `git tag -f` and `git push -f` allow overriding tags, then it doesn't need two steps to delete and recreate.

Databean (approved):

tools/cuttlefish-host-image-installer/update_github_releases.sh
```diff
@@ -20,6 +20,19 @@ echo "$changes" > changelog
 
 STABLE_VERSION="v$version"
 
+echo "step 2: move stable release to new commit id"
```
Databean, line 23: What's the significance of having both `stable` and `latest` github releases if they're updated at the same time?

WeiChungHsu: Currently, they are the same. Originally, we want stable release stores stable version, and latest release stores latest version. But we want users to use stable release instead of ToT, therefore, the latest release still stores stable version.

tools/cuttlefish-host-image-installer/update_github_releases.sh
WeiChungHsu: Done.

tools/cuttlefish-host-image-installer/update_github_releases.sh
WeiChungHsu: Done.

Databean (approved):

PR #688, opened 2024-08-21
Make acloud delete run cvd rm

jemoreira:
By replacing stop_cvd with script that calls `cvd rm`

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/server_command/acloud_command.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/server_command/acloud_command.cpp
@@ -24,6 +24,7 @@
 #include "common/libs/fs/shared_buf.h"
 #include "common/libs/fs/shared_fd.h"
 #include "common/libs/utils/environment.h"
+#include "common/libs/utils/files.h"
 #include "common/libs/utils/result.h"
 #include "cuttlefish/host/commands/cvd/cvd_server.pb.h"
 #include "host/commands/cvd/acloud/converter.h"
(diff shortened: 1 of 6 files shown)


base/cvd/cuttlefish/host/commands/cvd/server_command/acloud_command.cpp
```diff
@@ -187,25 +189,63 @@ Result<cvd::Response> AcloudCommand::HandleLocal(
...
+  std::string cvd_internal_stop_path =
+      fmt::format("{}/bin/cvd_internal_stop", host_path);
+  if (FileExists(cvd_internal_stop_path)) {
+    // cvd_internal_stop exists, stop_cvd is just a symlink to it
+    CF_EXPECT(RemoveFile(stop_cvd_path), "Failed to remove stop_cvd file");
+  } else {
+    // cvd_internal_stop doesn't exist, stop_cvd is the actual executable file
+    CF_EXPECT(RenameFile(stop_cvd_path, cvd_internal_stop_path),
+              "Failed to rename stop_cvd as cvd_internal_stop");
+  }
```
Databean, line 234: If someone runs:
```
$ m
$ acloud create
$ acloud delete
$ m hosttar
```
Does `m hosttar` restore the build-system-intended `cvd_internal_stop` implementation or does it package the altered implementation?

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/server_command/acloud_command.cpp
jemoreira: I ran this:
```
acloud create --local-instance --local-image
cat `which stop_cvd`
m hosttar
cat `which stop_cvd`
```

The `stop_cvd` in my `$PATH` is the script, but the `stop_cvd` in the tar file is the link to `cvd_internal_stop`. I don't know why it works like that, but that's the best outcome we could hope for in this situation.

PR #700, opened 2024-08-28
Verify the content of the bugreport zip file.

ser-io:


e2etests/orchestration/common.go
```diff
@@ -170,6 +172,9 @@ func DownloadHostBugReport(srv orchclient.HostOrchestratorService, group string)
 	if err := f.Close(); err != nil {
 		return err
 	}
+	if err := RunCmd("unzip", "-qt", filename); err != nil {
```
jemoreira, line 175: Could this be done with the `archive/zip` go package? Probably by reading the file without writing anywhere.

I'd like to avoid calling external commands like this when a go library can be used as commands are implicit external dependencies that may fail depending on the environment the code is run in, while a library is packaged within the binary and guaranteed to work.

jemoreira (approved):

e2etests/orchestration/common.go
ser-io: Done! yes, reading the file spotted the error, e.g: `flate: corrupt input before offset 1`

e2etests/orchestration/common.go
```diff
@@ -186,3 +191,23 @@ func UploadAndExtract(srv orchclient.HostOrchestratorService, remoteDir, src str
 	}
 	return nil
 }
+
```
jemoreira, line 194: A comment describing how the file is verified would be nice.

jemoreira (approved):

e2etests/orchestration/common.go
ser-io: Done.

PR #701, opened 2024-08-28
Generate a default instance name on load

jemoreira:


base/cvd/cuttlefish/host/commands/cvd/parser/load_configs_parser.cpp
```diff
@@ -271,6 +273,27 @@ Result<LoadDirectories> GenerateLoadDirectories(
   return result;
 }
 
+void GenerateMissingInstanceNames(std::vector<std::string>& instance_names) {
```
Databean, line 276: On the premise that "[Prefer using return values over output parameters: they improve readability, and often provide the same or better performance.](https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs)", can this be returned rather than an output parameter?

It should be possible to implement the same behavior with a return value and without copying the input using a signature like
```
std::vector<std::string> FillInMissingInstanceNames(std::vector<std::string> instance_names) { ... }
```
and invoking it like
```
flags.instance_names = FillInMissingInstanceNames(std::move(flags.instance_names));
```

base/cvd/cuttlefish/host/commands/cvd/parser/load_configs_parser.cpp
```diff
@@ -286,6 +309,7 @@ Result<CvdFlags> ParseCvdConfigs(const EnvironmentSpecification& launch,
   for (const auto& instance: launch.instances()) {
     flags.instance_names.push_back(instance.name());
   }
+  GenerateMissingInstanceNames(flags.instance_names);
```
Databean, line 312: Can the behavior in lines 309-312 be part of `ParseLaunchCvdConfigs`?

base/cvd/cuttlefish/host/commands/cvd/parser/load_configs_parser.cpp
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/parser/load_configs_parser.cpp
jemoreira: `ParseLaunchCvdConfigs` seems to be used to generate the flags for `cvd_internal_start` as a list of strings. The instance names are not passed to `cvd_internal_start` so I'd rather not mix those.

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/parser/load_configs_parser.cpp
```diff
@@ -271,6 +273,29 @@ Result<LoadDirectories> GenerateLoadDirectories(
   return result;
 }
 
+std::vector<std::string> GenerateMissingInstanceNames(
```
ser-io, line 276: WDYT of `void FillEmptyInstanceNames(std::vector<std::string>& names)`?

With `GenerateMissingInstanceNames`, at a first reading I thought we were generating only the missing names, although we return the existing ones as well. If keeping this name, let's add a brief comment explaining this to the function.

ser-io (approved):

base/cvd/cuttlefish/host/commands/cvd/parser/load_configs_parser.cpp
jemoreira: Done

PR #713, opened 2024-09-04
Split generic command handler into individual commands and utility functions

jemoreira:
The generic command handler was handling `mkdir`, `ln`, `clear` and `host_bugreport`, which had very little in common, resulting in many (hard to keep track of) conditional code to handle the different cases.

`mkdir` and `ln` don't need to be `cvd` commands, these are utility functions now.

`clear` and `host_bugreport` have their own command handlers now.


base/cvd/cuttlefish/host/commands/cvd/server_command/serial_launch.cpp
```diff
@@ -264,11 +264,7 @@ class SerialLaunchCommand : public CvdServerHandler {
 
     size_t index = 0;
     for (const auto& device : devices) {
-      auto& mkdir_cmd = *req_protos.emplace_back().mutable_command_request();
-      *mkdir_cmd.mutable_env() = client_env;
-      mkdir_cmd.add_args("cvd");
-      mkdir_cmd.add_args("mkdir");
-      mkdir_cmd.add_args(device.home_dir);
+      CF_EXPECT(EnsureDirectoryExists(device.home_dir, 0775, ""));
```
Databean, line 267: Not sure how relevant this is given it's probably safe to delete `serial_launch.cpp` entirely, but this is a behavioral change. It used to be deferred execution that `mkdir` was added to the sequence of commands, but now this happens immediately.

base/cvd/cuttlefish/host/commands/cvd/acloud/converter.cpp
```diff
@@ -354,17 +354,8 @@ Result<ConvertedAcloudCreateCommand> ConvertAcloudCreate(
...
-      cvd::Request& mkdir_request = request_protos.emplace_back();
-      auto& mkdir_command = *mkdir_request.mutable_command_request();
-      mkdir_command.add_args("cvd");
-      mkdir_command.add_args("mkdir");
-      mkdir_command.add_args("-p");
-      mkdir_command.add_args(host_dir);
-      auto& mkdir_env = *mkdir_command.mutable_env();
-      mkdir_env[kAndroidHostOut] = host_artifacts_path->second;
-    }
+    CF_EXPECT(EnsureDirectoryExists(host_dir, 0775, /* group_name */ ""));
```
Databean, line 357: This is a behavioral change: [`try-acloud` invokes `ConvertAcloudCreate`](https://github.com/google/android-cuttlefish/blob/f1e6468df30d3b594776c27a042c452e2d915193/base/cvd/cuttlefish/host/commands/cvd/server_command/try_acloud.cpp#L102), but doesn't actually invoke the commands.

The deferred execution here allows callers to choose later whether to make stateful changes to the filesystem or not.

base/cvd/cuttlefish/host/commands/cvd/server_command/clear.cpp
```diff
@@ -0,0 +1,114 @@
...
+}
+
+Result<cvd::Response> CvdClearCommandHandler::Handle(
+    const RequestWithStdio& request) {
+  CF_EXPECT(CanHandle(request));
+
+  cvd::Response response;
+  response.mutable_command_response();
+
+  auto precondition_verified = VerifyPrecondition(request);
```
Databean, line 71: Something for another time, but it would nice if [`VerifyPrecondition`](https://github.com/google/android-cuttlefish/blob/f1e6468df30d3b594776c27a042c452e2d915193/base/cvd/cuttlefish/host/commands/cvd/server_command/utils.cpp#L55) had a more descriptive name.

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/server_command/clear.cpp
jemoreira: Agreed.
Most likely it should be deleted altogether since all it does is check that ANDROID_HOST_OUT is defined in the environment IIRC

base/cvd/cuttlefish/host/commands/cvd/acloud/converter.cpp
jemoreira: true, but creating that directory doesn't have bad side effects.

The directory in question is `/tmp/acloud_image_artifacts`, which acloud uses to store the downloaded images (in subdirectories). Once an acloud command creates that directory it stays created, so the acloud command (and the translator) are built to work well whether the directory already exists or not.

base/cvd/cuttlefish/host/commands/cvd/server_command/serial_launch.cpp
jemoreira: I also think this should be deleted, will do it in the next PR.

PR #716, opened 2024-09-06
Get rid of VerifyPrecondition

jemoreira:
It was only checking that ANDROID_HOST_OUT was always present in the environment, which is not always true. To make that check pass a default path was calculated when the request was created and added to the environment, only to be replaced with the value from an existing group most of the time.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/server_command/display.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/server_command/display.cpp
@@ -65,7 +65,6 @@ class CvdDisplayCommandHandler : public CvdServerHandler {
 
   Result<cvd::Response> Handle(const RequestWithStdio& request) override {
     CF_EXPECT(CanHandle(request));
-    CF_EXPECT(VerifyPrecondition(request));
     cvd_common::Envs envs = request.Envs();
 
     auto [_, subcmd_args] = ParseInvocation(request.Message());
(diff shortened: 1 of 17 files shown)


base/cvd/cuttlefish/host/commands/cvd/server_command/display.cpp
```diff
@@ -99,14 +98,13 @@ class CvdDisplayCommandHandler : public CvdServerHandler {
...
-        ConcatToString(envs.at(kAndroidHostOut), "/bin/", kDisplayBin);
-    std::string home = Contains(envs, "HOME")
-                           ? envs.at("HOME")
-                           : CF_EXPECT(SystemWideUserHome());
+        ConcatToString(android_host_out, "/bin/", kDisplayBin);
+    std::string home = Contains(envs, "HOME") ? envs.at("HOME")
+                                              : CF_EXPECT(SystemWideUserHome());
     envs["HOME"] = home;
-    envs[kAndroidSoongHostOut] = envs.at(kAndroidHostOut);
+    envs[kAndroidSoongHostOut] = android_host_out;
```
Databean, line 107: Here and in other commands:

Should this also set `envs["ANDROID_HOST_OUT"]`? Since `AndroidHostPath` can sometimes return a value when `$ANDROID_HOST_OUT` is not set, this part could work only for something later to fail.

jemoreira: Done

Databean (approved):

PR #717, opened 2024-09-06
To create another iso image file which use aosp kernel.

WeiChungHsu:
To create another iso image file which use aosp kernel.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/flags_defaults.h
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/flags_defaults.h
@@ -206,7 +206,7 @@
 // Streaming default parameters
 #define CF_DEFAULTS_START_WEBRTC false
 #define CF_DEFAULTS_START_WEBRTC_SIG_SERVER true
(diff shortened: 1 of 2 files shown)


tools/cuttlefish-host-image-installer/addpreseed_aosp.sh
```diff
@@ -0,0 +1,88 @@
...
+# add preseed to console based installer
+chmod ug+w initrd.gz
+gzip -d -f initrd.gz
+echo preseed.cfg | cpio -H newc -o -A -F initrd
+echo after_install_1.sh | cpio -H newc -o -A -F initrd
+gzip -9 initrd
+chmod a-w initrd.gz
+# add preseed to GTK based installer
+chmod ug+w gtk
+cd gtk
```
jemoreira, line 48: nit:
```
cd somedir
# do stuff in somedir
cd ..
```

is better handled using `pushd`:
```
pushd somedir
# do stuff in somedir
popd
```
If `somedir` is ever moved to location deeper in the directory tree, just modifying the `pushd` line is enough as `popd` will work fine, but `cd ..` would need to modified as well with the original approach.
```

jemoreira (approved):

PR #720, opened 2024-09-10
Use /usr/bin/env bash instead of /bin/bash

xdevs23:
It's more universally supported and makes the script run on NixOS

### DIFF
--- a/.kokoro/presubmit.sh
+++ b/.kokoro/presubmit.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -e -x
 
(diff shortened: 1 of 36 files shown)


0405ysj (changes requested):
I agree `/usr/bin/env bash` is better than `/bin/bash`, but I prefer to have united shebang between all script files under `docker/`. Could you modify other files as well, such as `image-build-bazel-wrapper.sh` or `guest/run_services.sh`?

Also after your modification, I would recommend you to amend commit with forced push due to our branch maintenance strategy. Thanks for the contribution!

xdevs23:
Sure. I force-pushed and also rebased on top of main. There is one change in a C++ file that appears to be generating a script, so make sure that is correct, as I don't have sufficient knowledge about that specific instance to know whether it's correct.

0405ysj (approved):

Databean (approved):

ser-io (approved):

PR #722, opened 2024-09-11
Set start_time value at creating an instance

oribe1115:
Previously InstanceGroup was set `start_time` information at creating, however, the operation was lost when splitting `cvd start` into `create` and `start` (#661).
Then, currently `cvd create` outputs the zero value of unix time as the `start_time`.
```
        "start_time" : "1970-01-01 00:00:00"
```

It brought back the operation and confirmed it works well locally.
```
        "start_time" : "2024-09-11 08:01:17"
```


Databean (approved):

Databean:
Thanks!

PR #726, opened 2024-09-20
Disable acloud-translator so that acloud would always execute the Pyt…

WeiChungHsu:
Disable acloud-translator so that acloud would always execute the Python codes.


jemoreira:
This isn't the best place to disable that functionality. You may want to do it in AOSP and completely bypass the call to `cvd`.

rmuthiah (approved):

PR #729, opened 2024-09-23
Create compatibility link from cvd create

jemoreira:
This links are meant to provide backwards compatibility with acloud and launch_cvd, which only makes sense when starting the device at creation time. It's also only at creation time that the check for default group can be made as by the time start executes a default group has been chosen that can't be differentiated from a custom one.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/server_command/create.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/server_command/create.cpp
@@ -27,6 +27,7 @@
 #include <android-base/strings.h>
 
 #include "common/libs/utils/contains.h"
+#include "common/libs/utils/environment.h"
 #include "common/libs/utils/files.h"
 #include "common/libs/utils/flag_parser.h"
 #include "common/libs/utils/result.h"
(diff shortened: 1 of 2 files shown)


base/cvd/cuttlefish/host/commands/cvd/server_command/create.cpp
```diff
@@ -296,6 +315,43 @@ Result<selector::LocalInstanceGroup> CvdCreateCommandHandler::GetOrCreateGroup(
...
+    CF_EXPECT(EnsureSymlink(group.HomeDir() + "/cuttlefish",
+                            system_wide_home + "/cuttlefish"));
+  }
+  // The config file needs to be copied instead of symlinked because when the
+  // group is removed the original file will be deleted leaving the symlink
+  // dangling. The config file in the home directory is used by
+  // cvd_internal_start to persist the user's choice for
+  // -report_anonymous_usage_stats.
+  CF_EXPECT(
+      Copy(group.InstanceDir(group.Instances()[0]) + "/cuttlefish_config.json",
```
Databean, line 341: The `group.Instances()[0]` access should be after the `CF_EXPECT_NE` check, as the `CF_EXPECT_NE` check proves that `group.Instances()` is not empty.

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/server_command/create.cpp
jemoreira: Done

PR #734, opened 2024-09-26
Add more log output to `cvd fetch` stderr

Databean:
... to debug an issue around build downloading getting stuck.

As written this change increases the stderr output from a regular `cvd fetch` from about 300 lines / 28KiB to 217K lines / 19 MB.

Bug: b/364058922
Test: /usr/bin/bazel run commands/cvd:cvd -- fetch --target_directory=/tmp/a --default_build=aosp-main/aosp_cf_x86_64_phone-trunk_staging-userdebug


ser-io:
@Databean  what do you think of printing the libcurl version to the logs as well? https://curl.se/libcurl/c/curl_version.html

jemoreira:
`cvd fetch` output is already quite spamy, this makes it even worst.

We need to separate what goes to stderr from what goes to fetch.log and keep the stderr output reasonable.

jemoreira (approved):

jemoreira:
> `cvd fetch` output is already quite spamy, this makes it even worst.
> 
> We need to separate what goes to stderr from what goes to fetch.log and keep the stderr output reasonable.

This is addressed in https://github.com/google/android-cuttlefish/pull/743, once that lands this can be merged too.

jemoreira (approved):

ser-io (approved):

PR #740, opened 2024-10-01
Enhance HO logs.

ser-io:


e2etests/orchestration/common.go
```diff
@@ -131,6 +131,33 @@ func (h *DockerHelper) RunContainer(img string, hostPort int) (string, error) {
 	return createRes.ID, nil
 }
 
+func (h *DockerHelper) PullLogs(id string) error {
```
jemoreira, line 134: what logs are being pulled here?

ser-io: these are the containers logs which contain the host orchestrator logs.

jemoreira (approved):

PR #749, opened 2024-10-13
Introduce isolation in artifacts.yaml

c0rv4x:
Add isolation in Github CI/CD to avoid unsafe code being executed in the self-hosted runner. This could otherwise lead to the runner takeover


jemoreira (approved):

.github/workflows/artifacts.yaml
```diff
@@ -75,6 +75,9 @@ jobs:
         path: ${{ env.x86_64_image_path }}
   export-docker-image-arm64:
```
0405ysj, line 76: We have another job `export-docker-image-x86_64` above. If it does for arm64, why don't we do it for x86_64 as well?

.github/workflows/artifacts.yaml
```diff
@@ -75,6 +75,9 @@ jobs:
...
     runs-on: arm-ubuntu-arm-22.04-4core
+    container:
+      image: arm64v8/ubuntu:22.04  # Specify an arm64 Docker image here
+      options: --security-opt=no-new-privileges  # Optional: Add additional Docker security options
```
0405ysj, line 80: Looks like it fails the presubmit check, could you check it?

0405ysj (changes requested):

.github/workflows/artifacts.yaml
c0rv4x: I skipped this on purpose since the job is run in `runs-on: ubuntu-22.04`, which is a Github-hosted runner. It will have it's own isolation so this is not really necessary.

0405ysj: Regardless of the purpose of this PR, it shouldn't break any presubmit checks. This PR is currently breaking `export-docker-image-arm64`.
https://github.com/google/android-cuttlefish/actions/runs/11336098940/job/31525430431?pr=749

c0rv4x: Sorry, it takes some time to see if it possible to do docker-in-docker for bazel to work on building docker images

c0rv4x:
Hey, I have updated the code, but can we launch it on your setup?

.github/workflows/artifacts.yaml
c0rv4x: x86_64 is run on Github-hosted runner, so it would be unnecessary

0405ysj:
> Hey, I have updated the code, but can we launch it on your setup?

Ah, sure. I wasn't able to visit here for a couple of days due to my personal reasons. You wanted to trigger the presubmit, right?

c0rv4x:
Hey!The whole github actions pipeline would be great. I need to see if i have any discrepancies with your self hosted setupBest,Anatoly IvanovOn 25 Oct 2024, at 05:29, seungjaeyoo ***@***.***> wrote:﻿

Hey, I have updated the code, but can we launch it on your setup?

Ah, sure. I wasn't able to visit here for a couple of days due to my personal reasons. You wanted to trigger the presubmit, right?

0405ysj:
Sorry, but I don't have enough bandwidth to try that today. Perhaps I can try with your pending PR next Monday.

Out of curiosity, what's the purpose for checking the discrepancies? It seems it still fails the presubmit check.

.github/workflows/artifacts.yaml
```diff
@@ -87,7 +90,7 @@ jobs:
     - name: bazel version
       run: bazel version
     - name: Build docker image
-      run: bazel build --sandbox_writable_path=$HOME //docker:orchestration_image_tar
+      run: bazel build --sandbox_writable_path=$HOME //docker:orchestration_image_tar --action_env=PLATFORM=linux/arm64
```
0405ysj, line 93: Here, future users need to specify the arch when building docker image, though they may not be interested.

0405ysj (changes requested):

adelva1984:
Please open this PR later if applicable/possible.

PR #758, opened 2024-10-16
Fix race in some `CredentialSource` implementations

Databean:
There is concurrent access to the `BuildApi` class which uses `CredentialSource`. Accessing `Credential()` in some implementations of `CredentialSource` mutates internal state.

I don't believe this will affect the flakes we're currently seeing in production, as that code path uses `FixedCredentialSource` which does not mutate internal state.

PR also includes some refactoring of the `credential_source.*` files in earlier commits.

### DIFF
--- a/base/cvd/cuttlefish/host/libs/web/credential_source.cc
+++ b/base/cvd/cuttlefish/host/libs/web/credential_source.cc
@@ -112,16 +112,16 @@ Result<std::unique_ptr<CredentialSource>> GetCredentialSourceLegacy(
 
 GceMetadataCredentialSource::GceMetadataCredentialSource(
     HttpClient& http_client)
-    : http_client(http_client) {
-  latest_credential = "";
-  expiration = std::chrono::steady_clock::now();
+    : http_client_(http_client) {
+  latest_credential_ = "";
+  expiration_ = std::chrono::steady_clock::now();
 }
 
 Result<std::string> GceMetadataCredentialSource::Credential() {
-  if (expiration - std::chrono::steady_clock::now() < kRefreshWindow) {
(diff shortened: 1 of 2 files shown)


base/cvd/cuttlefish/host/libs/web/credential_source.cc
```diff
@@ -108,28 +189,27 @@ Result<std::unique_ptr<CredentialSource>> GetCredentialSourceLegacy(
...
 }
 
 Result<std::string> GceMetadataCredentialSource::Credential() {
-  if (expiration - std::chrono::steady_clock::now() < kRefreshWindow) {
+  std::lock_guard lock(latest_credential_mutex_);
+  if (expiration_ - std::chrono::steady_clock::now() < kRefreshWindow) {
     CF_EXPECT(RefreshCredential());
   }
-  return latest_credential;
+  return latest_credential_;
```
jemoreira, line 204: nit: This code is repeated among all 3 implementations. `Credential()`, `latest_credential_` and `latest_credential_mutex_` could be in the base class, where `RefreshCredential()` is declared as pure virtual and implemented by the subclasses.

jemoreira (approved):

base/cvd/cuttlefish/host/libs/web/credential_source.cc
Databean: Technically there are 4 implementations, `FixedCredentialSource` doesn't have this logic. You are correct that it's repeated and there could be some code reuse.

Databean: Added an intermediate `class RefreshingCredentialSource` that holds the state and refresh logic.

ser-io (approved):

Databean:
Thanks for reviewing!

PR #759, opened 2024-10-17
Isolate old server code and remove some duplication

jemoreira:


base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp
```diff
@@ -213,7 +214,8 @@ Result<StatusFetcherOutput> StatusFetcher::FetchStatus(
     instances.emplace_back(instance_record_result->first);
   } else {
     if (status_the_group_flag) {
-      instances = instance_group.Instances();
+      instances = std::vector<cvd::Instance>(instance_group.Instances().begin(),
+                                             instance_group.Instances().end());
```
Databean, line 218: nit: this assumes `Instances()` returns a reference to the same object between both calls. Can the returned value/reference from `Instances()` be stored in a local variable?

base/cvd/cuttlefish/host/commands/cvd/selector/instance_group_record.h
```diff
@@ -51,8 +51,12 @@ class LocalInstanceGroup {
...
   TimeStamp StartTime() const;
   void SetStartTime(TimeStamp time);
-  const std::vector<cvd::Instance>& Instances() const { return instances_; }
-  std::vector<cvd::Instance>& Instances() { return instances_; }
+  const google::protobuf::RepeatedPtrField<cvd::Instance>& Instances() const {
+    return group_proto_.instances();
+  };
+  google::protobuf::RepeatedPtrField<cvd::Instance>& Instances() {
+    return *group_proto_.mutable_instances();
+  };
```
Databean, line 59: In the future, do you think we could head in a direction of avoiding exposing the inner proto objects? From [go/UnhealthyIdioms#missing-abstractions-sea-of-details](https://go.corp.google.com/UnhealthyIdioms#missing-abstractions-sea-of-details):

> Systems sometimes lack key abstractions that would allow for conceptual chunking and ease of understanding. This may be seen as a sea of details, where you have to read and understand large swathes of the code and infer what it collectively means.
> - ...
> - Data Class
>   - Dumb data objects that only have getters and setters, with no behavior.
> - Unencapsulated Proto Buffers
>   - This is a special case of Data Class that is often seen in Google code.

The downside of exposing proto objects is we give up control over the API of this type, and can't introduce more behavior like validation in setters or member functions that calculate a response based on other fields.

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/run_server.cpp
```diff
@@ -139,16 +139,39 @@ Result<void> ImportResourcesImpl(const ParseResult& param) {
...
 }
 
+static Result<UnixMessageSocket> GetClient(const SharedFD& client) {
+  UnixMessageSocket result(client);
+  CF_EXPECT(result.EnableCredentials(true),
+            "Unable to enable UnixMessageSocket credentials.");
+  return result;
+}
+
+Result<void> SendResponse(const SharedFD& client,
```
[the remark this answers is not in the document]
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/server_command/status_fetcher.cpp

base/cvd/cuttlefish/host/commands/cvd/selector/instance_group_record.h
jemoreira: Agreed

PR #766, opened 2024-10-22
Move webrtcclient

ser-io:


jemoreira:
Those files seem to be at the top of the repo, not in the right subdirectory

ser-io:
> Those files seem to be at the top of the repo, not in the right subdirectory

Good catch, forgot to push the commit moving the files. Moved them as package under `libhoclient`.

jemoreira (approved):

PR #767, opened 2024-10-22
Wrap cvd::Instance in a LocalInstance class

jemoreira:
In preparation to moving business logic to the instance, group and manager classes.


base/cvd/cuttlefish/host/commands/cvd/selector/instance_record.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+std::string LocalInstance::instance_dir() const {
+  return fmt::format("{}/cuttlefish/instances/cvd-{}",
+                     group_proto_->home_directory(), id());
+}
+
+int LocalInstance::adb_port() const {
+  // run_cvd picks this port from the instance id and doesn't provide a flag
+  // to change in cvd_internal_flag
+  return id() > 0 ? id() + BASE_ADB_PORT - BASE_INSTANCE_ID : 0;
```
ser-io, line 46: nit: I think handling the zero  scenario as an early return makes it an easier reading in this case.

```
// TODO: This should never happen.
if id() == 0 {
  return 0;
}
```

also, the following order makes easier to read as well:  ` BASE_ADB_PORT + (id() - BASE_INSTANCE_ID)`

ser-io (approved):

PR #768, opened 2024-10-23
Fix forward use of `X-Cutf-Cloud-Orchestrator-Inject-BuildAPI-Creds`.

ser-io:
- Use of the header got broken in https://github.com/google/android-cuttlefish/commit/aa520271e19342cdda6abcd6e5e78da7c91635c4

### DIFF
--- a/frontend/src/libhoclient/host_orchestrator_client.go
+++ b/frontend/src/libhoclient/host_orchestrator_client.go
@@ -93,8 +93,13 @@ type HostOrchestratorService interface {
 	CreateBugreport(group string, dst io.Writer) error
 }
 
-const DefaultHostOrchestratorCredentialsHeader = "X-Cutf-Host-Orchestrator-BuildAPI-Creds"
-const BuildAPICredsUserProjectIDHeader = "X-Cutf-Host-Orchestrator-BuildAPI-Creds-User-Project-ID"
+const (
+	HTTPHeaderBuildAPICreds              = "X-Cutf-Host-Orchestrator-BuildAPI-Creds"
+	HTTPHeaderBuildAPICredsUserProjectID = "X-Cutf-Host-Orchestrator-BuildAPI-Creds-User-Project-ID"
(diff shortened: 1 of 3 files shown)


frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -465,10 +468,16 @@ func asWebRTCICEServers(in []opapi.IceServer) []webrtc.ICEServer {
...
 func addBuildAPICredsHeaders(rb *HTTPRequestBuilder, creds BuildAPICredential) {
-	if creds.AccessToken != "" {
-		rb.AddHeader(DefaultHostOrchestratorCredentialsHeader, creds.AccessToken)
-		if creds.UserProjectID != "" {
-			rb.AddHeader(BuildAPICredsUserProjectIDHeader, creds.UserProjectID)
-		}
+	if creds.AccessToken == "" {
+		return
+	}
+	if creds.AccessToken == "inject" {
```
jemoreira, line 474: nit: Instead of having 3 different behaviors based on the value it's better to have two functions:
```
func AddBuildCredsHeaders(rb *HTTPRequestBuilder, creds BuildAPICredential) {
  // Add the access token and project id (if not empty)
}

func requestBuildAPICredsInjection(rb *HTTPRequestBuilder) {
    // Adds the inject header
}
```

The caller probably knows which behavior it wants as it needs to handle refreshing the tokens when not injecting, so this approach is less error prone.

ser-io: agree that having two helpers is less error prone, however I had to add a new helper `handleBuildAPICredsHeaders` to avoid repeating the same code in the callers.

jemoreira: I see, that's still better though, as the function has a clear purpose easy to understand from the code.

jemoreira (approved):

PR #770, opened 2024-10-24
Let the caller to set the build api credentials.

ser-io:


frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -48,11 +48,26 @@ type ConnectWebRTCOpts struct {
...
 	AccessToken   string
 	UserProjectID string
 }
 
+func (c *AccessTokenBuildAPICreds) ApplyToHTTPRequest(rb *HTTPRequestBuilder) {
+	const HTTPHeaderBuildAPICreds = "X-Cutf-Host-Orchestrator-BuildAPI-Creds"
+	const HTTPHeaderBuildAPICredsUserProjectID = "X-Cutf-Host-Orchestrator-BuildAPI-Creds-User-Project-ID"
+	if c.AccessToken != "" {
+		rb.AddHeader(HTTPHeaderBuildAPICreds, c.AccessToken)
+		if c.UserProjectID != "" {
```
jemoreira, line 65: what happens if `AccessToken` is empty and `UserProjectID` is not?

frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -48,11 +48,26 @@ type ConnectWebRTCOpts struct {
 	LocalICEConfig *wclient.ICEConfig
 }
 
-type BuildAPICredential struct {
+type BuildAPICreds interface {
+	ApplyToHTTPRequest(rb *HTTPRequestBuilder)
+}
+
+type AccessTokenBuildAPICreds struct {
```
jemoreira, line 55: What about the one that asks the Cloud Orchestrator to inject? Are you going to implement it in the cloud orchestrator project?

jemoreira (approved):

frontend/src/libhoclient/host_orchestrator_client.go
ser-io: the `UserProjectID` value is only relevant if there's an access token.

frontend/src/libhoclient/host_orchestrator_client.go
ser-io: yes.

PR #777, opened 2024-10-30
Work with local libhoclient library and local host orchestrator api/v1 package in e2etests.

ser-io:


frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -386,6 +398,55 @@ func (c *HostOrchestratorServiceImpl) ListCVDs() ([]*hoapi.CVD, error) {
 	return res.CVDs, nil
 }
 
+func (c *HostOrchestratorServiceImpl) Powerwash(group, name string) error {
```
jemoreira, line 401: nit: Rename `group, name` as `group, instance` or `groupName, instanceName`.

frontend/src/libhoclient/host_orchestrator_client.go
jemoreira, line 1: These changes look like they should be in a different PR

jemoreira (approved):

frontend/src/libhoclient/host_orchestrator_client.go
ser-io: done.

frontend/src/libhoclient/host_orchestrator_client.go
ser-io: this code was removed from this PR following the comment below. I'll address the comment in the following PR containing this change.

PR #778, opened 2024-11-01
Support apt install in docker image

0405ysj:
With this PR, docker image could be built within various way.
- `docker/image-builder.sh cuttlefish-orchestration # For downloading packages from remote repository`
- `docker/image-builder.sh cuttlefish-orchestration -d # For building packages from local source`

### DIFF
--- a/.github/workflows/artifacts.yaml
+++ b/.github/workflows/artifacts.yaml
@@ -4,6 +4,7 @@ on:
     paths:
       - '.github/workflows/artifacts.yaml'
       - 'tools/buildimage/**'
+      - 'docker/**'
   push:
     branches:
       - main
@@ -60,14 +61,11 @@ jobs:
       run: |
         short_sha=$(echo ${{ github.sha }} | cut -c1-8)
         echo "x86_64_image_path=orchestration-image-x86_64-${short_sha}.tar" >> $GITHUB_ENV
-    - name: Install bazel
-      run: sudo bash tools/buildutils/installbazel.sh
-    - name: bazel version
-      run: bazel version
+        echo "image_name=cuttlefish-orchestration:${short_sha}" >> $GITHUB_ENV
     - name: Build docker image
-      run: bazel build --sandbox_writable_path=$HOME //docker:orchestration_image_tar
-    - name: Set filename of docker image
-      run: mv bazel-bin/docker/orchestration-image.tar ${{ env.x86_64_image_path }}
+      run: docker/image-builder.sh -t ${{ env.image_name }}
+    - name: Save docker image
+      run: docker save --output ${{ env.x86_64_image_path }} ${{ env.image_name }}
     - name: Publish docker image
       uses: actions/upload-artifact@c7d193f32edcb7bfad88892161225aeda64e9392 # aka v4.0.0
       with:
@@ -82,14 +80,11 @@ jobs:
       run: |
         short_sha=$(echo ${{ github.sha }} | cut -c1-8)
         echo "arm64_image_path=orchestration-image-arm64-${short_sha}.tar" >> $GITHUB_ENV
-    - name: Install bazel
-      run: sudo bash tools/buildutils/installbazel.sh
-    - name: bazel version
-      run: bazel version
+        echo "image_name=cuttlefish-orchestration:${short_sha}" >> $GITHUB_ENV
     - name: Build docker image
-      run: bazel build --sandbox_writable_path=$HOME //docker:orchestration_image_tar
-    - name: Set filename of docker image
-      run: mv bazel-bin/docker/orchestration-image.tar ${{ env.arm64_image_path }}
+      run: docker/image-builder.sh -t ${{ env.image_name }}
+    - name: Save docker image
+      run: docker save --output ${{ env.arm64_image_path }} ${{ env.image_name }}
     - name: Publish docker image
       uses: actions/upload-artifact@c7d193f32edcb7bfad88892161225aeda64e9392 # aka v4.0.0
       with:
--- a/docker/BUILD.bazel
+++ b/docker/BUILD.bazel
@@ -13,15 +13,15 @@
 # limitations under the License.
 
 genrule(
-    name = "orchestration_image_tar",
-    outs = ["orchestration-image.tar"],
-    cmd = "$(location :orchestration_image_builder) -o $@",
-    tools = [":orchestration_image_builder"],
+    name = "orchestration_image_dev_tar",
+    outs = ["orchestration-image-dev.tar"],
+    cmd = "$(location :orchestration_image_dev_builder) -o $@",
+    tools = [":orchestration_image_dev_builder"],
     visibility = ["//visibility:public"],
 )
 
 sh_binary(
-    name = "orchestration_image_builder",
+    name = "orchestration_image_dev_builder",
     srcs = ["image-build-bazel-wrapper.sh"],
     deps = ["@bazel_tools//tools/bash/runfiles"],
 )
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,7 +2,10 @@
 # Docker image includes HO(Host Orchestrator) inside,
 # so it could execute CF instance with API in HO.
 
-FROM debian:12 AS cuttlefish-hostpkg-builder
+ARG BUILD_OPTION=prod
+
+
+FROM debian:12 AS host-package-builder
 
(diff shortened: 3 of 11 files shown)


0405ysj:
Especially I have no strong opinion about naming, so please feel free to suggest any naming :)

0405ysj:
Perhaps performing `apt install` in the runtime is better than in the build time(current impl)?

0405ysj:
@jemoreira @ser-io @ikicha Gentle ping. Could I get review for this PR? This is for building ToT docker image which performs `apt install` inside for installing cuttlefish host packages like `cuttlefish-base`.

docker/image-builder.sh
```diff
@@ -7,18 +7,48 @@
...
 # Build docker image
 pushd $android_cuttlefish_root_dir
-docker build \
+DOCKER_BUILDKIT=1 docker build \
     --force-rm \
     --no-cache \
     -f docker/Dockerfile \
-    -t $tag \
+    -t $name \
+    --target $target \
```
ikicha, line 52: how about using --build-arg instead

docker/Dockerfile
```diff
@@ -44,19 +45,38 @@ RUN apt install -y --no-install-recommends \
...
+    wget
+RUN wget -qO- https://artifacts.codelinaro.org/artifactory/linaro-372-googlelt-gigabyte-ampere-cuttlefish-installer/gigabyte-ampere-cuttlefish-installer/latest/debian/linaro-glt-gig-archive-bookworm.asc \
+    | tee /etc/apt/trusted.gpg.d/linaro-glt-gig-archive-bookworm.asc
+RUN echo "deb https://artifacts.codelinaro.org/linaro-372-googlelt-gigabyte-ampere-cuttlefish-installer/gigabyte-ampere-cuttlefish-installer/latest/debian bookworm main" \
+    | tee /etc/apt/sources.list.d/linaro-glt-gig-archive-bookworm.list
+RUN apt update
+RUN apt install -y --no-install-recommends \
+    cuttlefish-base cuttlefish-user cuttlefish-orchestration
+
+RUN echo "num_cvd_accounts=100" >> /etc/default/cuttlefish-host-resources
```
ikicha, line 79: these block is duplicated, how about adding another stage for that?

```
FROM runner-base AS runner-with-local-source
// install debs from host-package-builder
FROM runner-base AS runner-with-apt-install
// install via apt

FROM runner-with-${build_option} AS runner
RUN echo "num_cvd_accounts=100" >> /etc/default/cuttlefish-host-resources
RUN usermod -aG kvm root
RUN usermod -aG cvdnetwork root
ENTRYPOINT ["/root/run_services.sh"]
```

docker/BUILD.bazel
```diff
@@ -13,9 +13,17 @@
 # limitations under the License.
 
 genrule(
-    name = "orchestration_image_tar",
-    outs = ["orchestration-image.tar"],
-    cmd = "$(location :orchestration_image_builder) -o $@",
+    name = "orchestration_image_with_local_source_tar",
```
ser-io, line 16: suffixes `local-source` and `apt-install` are implementation details, I suggest to use name reflecting how the images are meant to be used. Suggest names `orchestration_image` used for building the regular image using the published packages and using `orchestration_image_localdev` for the image using the packages built from local source.

Also, `apt-install` is misleading as both flows uses `apt install`.

docker/image-builder.sh
```diff
@@ -7,18 +7,48 @@
 script_location=`realpath -s $(dirname ${BASH_SOURCE[0]})`
 android_cuttlefish_root_dir=$(realpath -s $script_location/..)
 
-if [[ "$1" == "" ]]; then
-    tag=cuttlefish-orchestration
-else
-    tag=$1
-fi
+usage() {
+  echo "usage: $0 [-t <tag>] [-b <build-option>]"
```
ser-io, line 11: I suggest to make this script defaults to building the regular image, using the published packages. Then if we pass flag `-d` meant for `local development` we then use the `packages` built locally.

Let's avoid the use of `apt-install` as an option name as `apt install` is used both ways.

ser-io (changes requested):

.github/workflows/artifacts.yaml
```diff
@@ -51,47 +52,91 @@ jobs:
       with:
         name: image_gce_debian11_amd64
         path: image.tar.gz
-  export-docker-image-x86_64:
+  export-docker-image-x86_64-local-source:
```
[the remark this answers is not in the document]
0405ysj: Done.

docker/BUILD.bazel
0405ysj: Done.

docker/image-builder.sh
0405ysj: Done.

docker/Dockerfile
0405ysj: Thanks for the suggestion. It's nice! Done :)

docker/image-builder.sh
0405ysj: Done.

ikicha (approved):

docker/BUILD.bazel
```diff
@@ -12,6 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+genrule(
+    name = "orchestration_image_dev_tar",
+    outs = ["orchestration-image-dev.tar"],
+    cmd = "$(location :orchestration_image_builder) -o $@ -d",
```
ser-io, line 18: The difference between generating a `dev` image and a `prod` image is by appending `-d` in the command line which is then handled in two different bash scripts. I find this error prone, if we remove `-d` by mistake or introduce a bug in the scripts, it will create major problems down the line, for example not testing against latest changes allowing new PRs to be merged in.

I'd like to have a less error prone alternative here. My suggestion is  to only build `dev` image with `bazel`. Building the image with `bazel` was created only from the e2etests perspective as it makes running the e2e tests very simple using only one command. Now, the way the image is built with `bazel` is still a hacky way, using a wrapper bash script. Hence I prefer to avoid extending/accommodating  this solution to build the prod image and use the bash script directly.

ser-io (changes requested):

docker/BUILD.bazel
0405ysj: I agree with `prod` image `bazel` rule wouldn't be needed for performing e2e testing of cuttlefish host packages, but I think probably `prod` image `bazel` rule is needed for performing e2e testing of `cvdr`. Because versioning won't be same between the docker image and cuttlefish host packages. In case of modifying Dockerfile, performing `cvdr` e2e testing may not catch the error specified at `prod` image, and it can be mattered when releasing new version of `prod` image.

If you're still worried about error prone, what about explicitly giving parameter that this is for `dev` or `prod` in the script?

0405ysj: @ser-io Gentle ping.

ser-io: Ideally we should be using bazel rule https://github.com/bazelbuild/rules_docker/blob/master/docs/container.md#container_image to create Docker Image artifacts with `bazel`. I created the bash wrapper script hack to enable the e2etests using bazel, let's not extend this solution to build prod images as it's not ideal.

I understand the case of `cvdr`, however it's not clear to me how we will be using the rule in the e2e tests for `cvdr` as `cvdr` is in a different repository, we don't even use bazel in the `cvdr` repo yet. We can revisit this idea when implementing `cvdr` e2e tests, but let's wait until then and keep it simple for now.

0405ysj: Understood. I think not using bazel wrapper script for `prod` image is fine for now, but I'm still afraid of docker images between `prod` and `dev` image are diverged with unexpected behavior. So please let me know if there's a future plan for modifying bazel wrapper script.

OTOH, if we use `container_image` bazel rule someday, I think `prod` image should be built with that as well so far due to the above reason.

I'll be changing PR without using bazel wrapper for `prod` image.

0405ysj: Done.

ser-io: Yes, if there's a use case to build the `prod` image with `bazel` we should aim to build it with `container_image`.

ser-io (approved):

0405ysj:
@jemoreira @Databean Could I get overall code owner review? I think current approval is not enough for `e2etests/`.

Databean (approved):

PR #788, opened 2024-11-07
Add missing bazel dependencies for host/libs/directories

Databean:


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #791, opened 2024-11-08
Fix non-`cvd login` credentials

Databean:
Credential selection was interpreting missing `cvd login` crdentials as valid, and would only fall back when `cvd login` credentials were present but invalid.

Bug: b/378069857
Test: cvd fetch --credential_source=gce

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
+++ b/base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
@@ -359,9 +359,10 @@ Result<std::unique_ptr<BuildApi>> GetBuildApi(const BuildApiFlags& flags) {
       StringFromEnv("HOME", ".") + "/.acloud_oauth2.dat";
 
   std::unique_ptr<CredentialSource> credential_source =
-      cvd_creds.ok() ? std::move(*cvd_creds)
-                     : CF_EXPECT(GetCredentialSourceFromFlags(
-                           *retrying_http_client, flags, oauth_filepath));
+      cvd_creds.ok() && cvd_creds->get()
+          ? std::move(*cvd_creds)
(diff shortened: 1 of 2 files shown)


base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
```diff
@@ -359,9 +359,10 @@ Result<std::unique_ptr<BuildApi>> GetBuildApi(const BuildApiFlags& flags) {
       StringFromEnv("HOME", ".") + "/.acloud_oauth2.dat";
 
   std::unique_ptr<CredentialSource> credential_source =
-      cvd_creds.ok() ? std::move(*cvd_creds)
-                     : CF_EXPECT(GetCredentialSourceFromFlags(
-                           *retrying_http_client, flags, oauth_filepath));
+      cvd_creds.ok() && cvd_creds->get()
```
jemoreira, line 362: Given that `Err` and `Ok(nullptr)` are treated in the same way, could `CredentialForScopes` just return `Result<CredentialSource>`, returning an error when it would otherwise return `Ok(nullptr)`?

jemoreira (approved):

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
Databean: Unfortunately `Result<CredentialSource>` isn't valid because `CredentialSource` is a pure virtual class and therefore cannot be passed by value.

Databean: Changed `CredentialForScopes` so it doesn't ever return `nullptr`, reducing the space of return values.

Databean:
Thanks for reviewing!

PR #794, opened 2024-11-11
clang-tidy in `.bazelrc` and a github workflow

Databean:
References:
- https://github.com/erenon/bazel_clang_tidy
- https://clang.llvm.org/extra/clang-tidy/

Run locally with
```
bazel build //cuttlefish/... --config clang-tidy
```

The `clang-diagnostic-pragma-once-outside-header` check is disabled as bazel's `parse_headers` feature tries to compile every header as a source file.

We could re-enable that check if we switch to an include guard system, which would also better satisfy the style guide:

https://google.github.io/styleguide/cppguide.html#The__define_Guard

### DIFF
--- a/.github/workflows/presubmit.yaml
+++ b/.github/workflows/presubmit.yaml
@@ -57,7 +57,7 @@ jobs:
     - name: Setup apt
       run: apt update -y && apt upgrade -y
     - name: Install dependencies
-      run: apt install -y git clang libcurl4-openssl-dev
+      run: apt install -y git clang libcurl4-openssl-dev clang-tidy
     - name: Install bazel
       run: bash tools/buildutils/installbazel.sh
     - name: Build cvd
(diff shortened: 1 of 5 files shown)


ser-io (approved):

Databean:
> Is it possible to enable `--config clang-tidy` by default? So that I don't have to include it when I build manually.

Two issues:

1. `clang-tidy` is slow. After `bazel clean`, `bazel build `//...` is 24 seconds on my machine, while `bazel build //... --config clang-tidy` is 45 seconds.
2. It seems like this config completely replaces the original build commands, so it doesn't actually create the executables that a regular `bazel build ...` does. I don't have a `bazel-bin/cuttlefish/host/commands/cvd` after building with `--config=clang-tidy`.

Databean:
FYI: we were failing more of the default checks than initially clear. It seems like bazel consumes all the warning output in this configuration. I've upgraded all the warnings of the default checks to errors, and explicitly ignored the checks we are currently failing.

PR #800, opened 2024-11-15
Remove headers from `srcs` in cuttlefish/host/commands/cvd:libcvd

Databean:
Non-public headers can be in `srcs` and are not shared with other bazel targets, while public headers in `hdrs` are shared. In this case, mentioning headers in both places is redundant, so this commit first removes all headers from `srcs`, and leaves it to a future change to divide the headers between the two targets.


ser-io (approved):

jemoreira (approved):

Databean:
Thanks for reviewing!

PR #808, opened 2024-11-22
Update ExtractArchiveContents to support uncompressed archives

frankfeng579:


base/cvd/cuttlefish/common/libs/utils/archive.cpp
```diff
@@ -115,9 +117,72 @@ Result<std::string> ExtractImage(const std::string& archive_filepath,
   return {result.front()};
 }
 
+namespace {
+
+Result<std::string> hardlinkRecursively(const std::string& src_dir,
```
jemoreira, line 122: Should return `Result<std::vector<std::string>>` and avoid the out parameter since the string result is ignored anyways.

base/cvd/cuttlefish/common/libs/utils/archive.cpp
```diff
@@ -115,9 +117,72 @@ Result<std::string> ExtractImage(const std::string& archive_filepath,
...
+}
+
+}  // namespace
+
 Result<std::vector<std::string>> ExtractArchiveContents(
     const std::string& archive_filepath, const std::string& target_directory,
     const bool keep_archive) {
+  struct stat path_stat;
+  if (stat(archive_filepath.c_str(), &path_stat) == 0 &&
+      S_ISDIR(path_stat.st_mode)) {
```
jemoreira, line 176: `ExtractArchiveContents` should return an error if the source is a directory (not an archive). This check should be done in the caller and then decide whether to call `ExtractArchiveContents` or `HardlinkRecursively`.

base/cvd/cuttlefish/common/libs/utils/archive.cpp
jemoreira, line 122: rename as `HardlinkRecursively`

base/cvd/cuttlefish/common/libs/utils/archive.cpp
```diff
@@ -115,9 +117,72 @@ Result<std::string> ExtractImage(const std::string& archive_filepath,
   return {result.front()};
 }
 
+namespace {
+
+Result<std::string> hardlinkRecursively(const std::string& src_dir,
+                                        const std::string& dst_dir,
+                                        std::vector<std::string>& files) {
+  DIR* dir = opendir(src_dir.c_str());
```
jemoreira, line 125: use the `WalkDirectory` function in `base/cvd/cuttlefish/common/libs/util/files.h`. If you need more control on the order in which nodes are visited you can use `DirectoryContents` and `DirectoryExists` from the same header file.

base/cvd/cuttlefish/common/libs/utils/archive.cpp
```diff
@@ -115,9 +117,72 @@ Result<std::string> ExtractImage(const std::string& archive_filepath,
...
+}  // namespace
+
 Result<std::vector<std::string>> ExtractArchiveContents(
     const std::string& archive_filepath, const std::string& target_directory,
     const bool keep_archive) {
+  struct stat path_stat;
+  if (stat(archive_filepath.c_str(), &path_stat) == 0 &&
+      S_ISDIR(path_stat.st_mode)) {
+    std::vector<std::string> files;
+    CF_EXPECTF(hardlinkRecursively(archive_filepath, target_directory, files),
```
jemoreira, line 178: If `keep_archive` is `false` this should just move the directory instead of recursively hard linking and then recursively removing.

If `keep_archive` is `true` would it make sense to just create a symbolic link instead?

base/cvd/cuttlefish/common/libs/utils/archive_test.cpp
```diff
@@ -0,0 +1,90 @@
...
+  std::ofstream file(filepath);
+  file << text;
+  file.close();
+}
+
+// Test fixture for common setup and teardown
+class ExtractArchiveContentsTests : public ::testing::Test {
+ protected:
+  void SetUp() override {
+    temp_dir_ = TemporaryDir().path;
```
jemoreira, line 42: This creates the `TemporaryDir` object and destroys it immediately after, causing the temporary directory to be deleted. `temp_dir_` should hold the entire `TemporaryDir` object and `temp_dir_.path` used where necessary.

base/cvd/cuttlefish/common/libs/utils/archive_test.cpp
```diff
@@ -0,0 +1,90 @@
...
+    mkdir(temp_dir_.c_str(), 0755);
```
jemoreira, line 43: `TemporaryDir` creates the directory, it's not necessary to create it later as long the object is kept during the test duration.

base/cvd/cuttlefish/common/libs/utils/archive_test.cpp
```diff
@@ -0,0 +1,90 @@
...
+    archive_dir_ = temp_dir_ + "/device-imge-123.zip";
+    mkdir(archive_dir_.c_str(), 0755);
```
jemoreira, line 45: Use `EnsureDirectoryExists` from `files.h` to create directories and don't ignore the result of calling the function.

base/cvd/cuttlefish/common/libs/utils/archive_test.cpp
```diff
@@ -0,0 +1,90 @@
...
+    CreateTempFileWithText(archive_dir_ + "/file1.txt", "file1");
+    std::string sub_dir = archive_dir_ + "/sub_dir";
+    mkdir(sub_dir.c_str(), 0755);
+    CreateTempFileWithText(sub_dir + "/file2.txt", "file2");
+    extract_dir_ = temp_dir_ + "/target_dir";
+    mkdir(extract_dir_.c_str(), 0755);
+  }
+
+  void TearDown() override { RecursivelyRemoveDirectory(temp_dir_); }
```
jemoreira, line 54: Not needed if using `TemporaryDir`

jemoreira (changes requested):

frankfeng579:
Per review, archive.cpp should not be touched. Instead, HardLinkRecursively should be added to files.h+files.cpp and the remaining logic will go to the caller of ExtractArchiveContents(), fetch_cvd.cc in this case.

Close this pull request.

Databean:
FYI @frankfeng579 you can reuse the same pull request and `git push --force` to the development branch. That way it's easier to track your responses to the comments here.

PR #815, opened 2024-12-06
Use libcurl from bazel modules rather than with `-lcurl`

Databean:
This fixes the "clang: warning: -lcurl: 'linker' input unused [-Wunused-command-line-argument]" error, and no longer links with `-lcurl`.

Before https://github.com/bazelbuild/bazel-central-registry/pull/2398 , libcurl from the bazel registry did not link in any SSL version so HTTPS downloading was broken, but as of bazel libcurl 8.8.0.bcr.1 there is HTTPS support.

Bug: b/382734583
Test: /usr/bin/bazel run '//cuttlefish/host/commands/cvd' -- fetch --target_directory=/tmp/a --default_build=aosp-main/aosp_cf_x86_64_phone-trunk_staging-userdebug

### DIFF
--- a/base/cvd/MODULE.bazel
+++ b/base/cvd/MODULE.bazel
@@ -6,7 +6,7 @@
 ###############################################################################
 
 bazel_dep(name = "boringssl", version = "0.0.0-20240126-22d349c")
-bazel_dep(name = "curl", version = "8.7.1")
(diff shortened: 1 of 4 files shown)


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #816, opened 2024-12-07
Fold `CommandInvocation` into `CommandRequest`

Databean:
There was overlap between these types in the public interface.

Bug: b/382745645

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/command_request.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/command_request.cpp
@@ -25,53 +25,25 @@
 #include "host/commands/cvd/cli/selector/selector_common_parser.h"
 
 namespace cuttlefish {
-namespace {
-
-struct CommandInvocation {
-  std::string command;
-  std::vector<std::string> arguments;
-};
-
-CommandInvocation ParseInvocation(const CommandRequest& request) {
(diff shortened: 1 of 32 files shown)


base/cvd/cuttlefish/host/commands/cvd/cli/command_request.cpp
```diff
@@ -39,6 +40,14 @@ const selector::SelectorOptions& CommandRequest::Selectors() const {
 
 const cvd_common::Envs& CommandRequest::Env() const { return env_; }
 
+std::string CommandRequest::Subcommand() const {
+  return ParseInvocation(*this).command;
```
jemoreira, line 70: This has the potential of parsing multiple times in case of multiple calls to either of these functions. Can the call to `ParseInvocation` happen in the constructor?

jemoreira (approved):

jemoreira (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/command_request.cpp
Databean: Implemented in [48aa26a](https://github.com/google/android-cuttlefish/pull/816/commits/48aa26a5582041dbb51c1036d566c1a0c0990dbc).

Databean:
Thanks for reviewing!

PR #817, opened 2024-12-09
Default `CanHandle` implementation

Databean:
Deduplicates many identical `CanHandle` implementations, and also cleans up some local variables that were used only to return data from a function.

Bug: b/382794638


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #821, opened 2024-12-10
Add nginx configuration for WebSocket proxy

k311093:
While connecting to ADB via WebSocket, it also need to set additional header for nginx proxy. Change nginx config for that.

### DIFF
--- a/frontend/host/packages/cuttlefish-orchestration/etc/nginx/conf.d/cuttlefish-orchestration.conf
+++ b/frontend/host/packages/cuttlefish-orchestration/etc/nginx/conf.d/cuttlefish-orchestration.conf
@@ -36,6 +36,15 @@ server {
 
     # Host Orchestrator
     location / {
+        # ADB websocket backend
+        location ~* ^/.*/adb$ {
+            proxy_pass http://127.0.0.1:2081;
+            proxy_http_version 1.1;
+            proxy_set_header Upgrade $http_upgrade;
+            proxy_set_header Connection "Upgrade";
+            proxy_set_header Host $host;
+        }
+
         proxy_pass http://127.0.0.1:2081;
 
         # This is required for uploading huge artifacts.


0405ysj (approved):

jemoreira (approved):

frontend/host/packages/cuttlefish-orchestration/etc/nginx/conf.d/cuttlefish-orchestration.conf
```diff
@@ -36,6 +36,15 @@ server {
 
     # Host Orchestrator
     location / {
+        # ADB websocket backend
+        location ~* ^/.*/adb$ {
+            proxy_pass http://127.0.0.1:2081;
```
ser-io, line 41: can we avoid duplicating the `proxy_pass http://127.0.0.1:2081;` line using the following structure?

```
    location / {
        proxy_pass http://127.0.0.1:2081;

        # This is required for uploading huge artifacts.
        client_max_body_size 20G;

        # ADB websocket backend
        location ~* ^/.*/adb$ {
            proxy_http_version 1.1;
            proxy_set_header Upgrade $http_upgrade;
            proxy_set_header Connection "Upgrade";
            proxy_set_header Host $host;
        }
    }
```

k311093: I tried to remove proxy_pass http://127.0.0.1:2081; at the inner location match, but it seems like outer proxy_pass does not works for inner match. When I tried move the configuration below and use proxy_http_version, adb connection did not worked with handshake failure.

ser-io: Sounds good, thanks for confirming.

PR #826, opened 2024-12-13
Run clang-tidy in github actions on debian testing

Databean:
This has more parity with our developer workflow, and later clang-tidy supports better checks and fewer false positive errors.

The clang-tidy run is separated into a separate workflow so we can keep validation on the older debian environment, for parity with what we support for users.

Bug: b/384076029


Databean:
Superseded by https://github.com/google/android-cuttlefish/pull/827

PR #827, opened 2024-12-13
Use a prebuilt clang/LLVM to make the build hermetic

Databean:
LLVM version 16.0.4 was chosen for parity with our current developer desktop setup.

This makes the behavior of the compiler and clang-tidy more comparable between developer workstations and CI.

### DIFF
--- a/base/cvd/MODULE.bazel
+++ b/base/cvd/MODULE.bazel
@@ -33,3 +33,15 @@ git_override(
     commit = "f23d924918c581c68cd5cda5f12b4f8198ac0c35",
     remote = "https://github.com/erenon/bazel_clang_tidy.git",
 )
+
+bazel_dep(name = "toolchains_llvm", version = "1.2.0")
+
+# Configure and register the toolchain.
+llvm = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm")
+llvm.toolchain(
+   llvm_version = "16.0.0",
+)
+
+use_repo(llvm, "llvm_toolchain")
+
+register_toolchains("@llvm_toolchain//:all")
--- a/base/debian/control
+++ b/base/debian/control
@@ -2,26 +2,27 @@ Source: cuttlefish-common
 Maintainer: Cuttlefish Team
 Section: misc
 Priority: optional
-Build-Depends: config-package-dev,
-               debhelper-compat (= 12),
+Build-Depends: bazel [amd64],
                clang,
-               pkg-config,
+               config-package-dev,
+               debhelper-compat (= 12),
+               git,
+               libcurl4-openssl-dev,
                libfmt-dev,
                libgflags-dev,
-               libjsoncpp-dev,
-               libcurl4-openssl-dev,
                libgoogle-glog-dev,
                libgtest-dev,
+               libjsoncpp-dev,
+               libtinfo5,
+               libprotobuf-c-dev,
+               libprotobuf-dev,
                libssl-dev,
                libxml2-dev,
-               uuid-dev,
-               libprotobuf-dev,
-               libprotobuf-c-dev,
-               protobuf-compiler,
                libz3-dev,
+               pkg-config,
+               protobuf-compiler,
                sed,
-               git,
-               bazel [amd64]
+               uuid-dev
 Standards-Version: 4.5.0
 
 Package: cuttlefish-base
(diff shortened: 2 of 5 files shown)


base/cvd/MODULE.bazel
```diff
@@ -33,3 +33,15 @@ git_override(
     commit = "f23d924918c581c68cd5cda5f12b4f8198ac0c35",
     remote = "https://github.com/erenon/bazel_clang_tidy.git",
 )
+
+bazel_dep(name = "toolchains_llvm", version = "1.2.0")
+
+# Configure and register the toolchain.
+llvm = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm")
+llvm.toolchain(
+   llvm_version = "16.0.0",
```
cjreynol, line 42: NIT:  `clang --version` shows `16.0.6` for me.

It is only bugfixes, but I did not too deeply into the commit lists.
https://releases.llvm.org/

cjreynol (approved):

base/cvd/MODULE.bazel
Databean: `toolchains_llvm` has its own [hardcoded logic](https://github.com/bazel-contrib/toolchains_llvm/blob/0aa1ffdc304b02a1cf7b2556346f896de6982ec8/toolchain/internal/release_name.bzl#L82) to choose filenames for particular LLVM patch releases, and it hasn't been updated for 16.0.6. The closest we can get here is 16.0.4.

Databean: Actually `16.0.4` gives these mysterious errors on CI:
```
external/toolchains_llvm++llvm+llvm_toolchain_llvm/bin/clang: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by external/toolchains_llvm++llvm+llvm_toolchain_llvm/bin/clang)
external/toolchains_llvm++llvm+llvm_toolchain_llvm/bin/clang: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by external/toolchains_llvm++llvm+llvm_toolchain_llvm/bin/clang)
external/toolchains_llvm++llvm+llvm_toolchain_llvm/bin/clang: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by external/toolchains_llvm++llvm+llvm_toolchain_llvm/bin/clang)
external/toolchains_llvm++llvm+llvm_toolchain_llvm/bin/clang: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.30' not found (required by external/toolchains_llvm++llvm+llvm_toolchain_llvm/bin/clang)
external/toolchains_llvm++llvm+llvm_toolchain_llvm/bin/clang: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by external/toolchains_llvm++llvm+llvm_toolchain_llvm/bin/clang)
external/toolchains_llvm++llvm+llvm_toolchain_llvm/bin/clang: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `CXXABI_1.3.13' not found (required by external/toolchains_llvm++llvm+llvm_toolchain_llvm/bin/clang)
```

base/debian/control
```diff
@@ -2,26 +2,27 @@ Source: cuttlefish-common
 Maintainer: Cuttlefish Team
 Section: misc
 Priority: optional
-Build-Depends: config-package-dev,
-               debhelper-compat (= 12),
+Build-Depends: bazel [amd64],
                clang,
```
ser-io, line 6: curious about still needing `clang` as a dependency now that we will be using a pre-built clang binary?

ser-io (approved):

ikicha:
If we can use https://github.com/bazel-contrib/setup-bazel to set up bazel env in github action, it looks like we can also enable 'github action cached' disk cache for bazel.

base/debian/control
Databean: Removed.

Databean:
> If we can use https://github.com/bazel-contrib/setup-bazel to set up bazel env in github action, it looks like we can also enable 'github action cached' disk cache for bazel.

I added this as a dependency, but we might have to sacrifice parallelism between our workflows to get more value out of this cache as the cache can't be reused when the other workflows haven't run yet.

jemoreira (approved):

PR #834, opened 2024-12-21
cvd cache

cjreynol:
New command used to manage the local fetch cache.  Will eventually be
automatically invoked during every `cvd fetch` call to keep the cache
from growing without end.

There are a couple of preliminary commits adding some functionality for use by the new command.


base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+namespace {
+constexpr int kDefaultCacheSizeGigabytes = 25;
+
+constexpr char kDetailedHelpText[] =
+    R"(usage: cvd cache <action> [<flag>...]
+
+Example usage:
+    cvd cache where - display the filepath of the cache
+    cvd cache size - display the approximate size of the cache
```
jemoreira, line 48: nit: These two can be merged into one, either just `cvd cache` or something like `cvd cache info`. It can format its output in a way that's easy to extract the information in a script:
```
$ cvd cache info
path: /tmp/cvd/123456/cache
size: 320G
```
so you can run `cvd cache info | grep ^path: | cut -d\  -f2` to obtain the path.
```

File: base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp [hunk 2]
```diff
@@ -0,0 +1,207 @@
...
+    cvd cache cleanup - caps the cache at the default size
+    cvd cache cleanup --allowed_size_GB=<n> - caps the cache at the given size
+    cvd cache empty - wipes out all files in the cache directory
+
+**Notes**:
+    - size and cleanup round the cache size up to the nearest gigabyte for calculating
+    - cleanup uses last modification time to remove oldest files first
+)";
+
+constexpr char kSummaryHelpText[] = "Used to manage the files cached by cvd";
```
jemoreira, line 58: nit: "Manage the...", drop the "Used to"

base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+Example usage:
+    cvd cache where - display the filepath of the cache
+    cvd cache size - display the approximate size of the cache
+    cvd cache cleanup - caps the cache at the default size
+    cvd cache cleanup --allowed_size_GB=<n> - caps the cache at the given size
+    cvd cache empty - wipes out all files in the cache directory
+
+**Notes**:
+    - size and cleanup round the cache size up to the nearest gigabyte for calculating
```
jemoreira, line 54: nit: Drop "for calculating"

base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
```diff
@@ -0,0 +1,207 @@
...
+namespace {
+constexpr int kDefaultCacheSizeGigabytes = 25;
+
+constexpr char kDetailedHelpText[] =
+    R"(usage: cvd cache <action> [<flag>...]
+
+Example usage:
+    cvd cache where - display the filepath of the cache
+    cvd cache size - display the approximate size of the cache
+    cvd cache cleanup - caps the cache at the default size
```
jemoreira, line 49: Print the default size too.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
```diff
@@ -0,0 +1,207 @@
...
+                        const std::size_t allowed_size_GB) {
+  std::size_t cache_size = CF_EXPECT(GetDiskUsageGigabytes(cache_directory));
+  // Descending because elements are removed from the back
+  std::vector<std::string> cache_files =
+      CF_EXPECT(DirectoryContentsByModTimeDesc(cache_directory));
+  while (cache_size > allowed_size_GB) {
+    CF_EXPECTF(!cache_files.empty(),
+               "No more files found for deletion, but approximate cache size "
+               "of {}GB is still larger than allowed {}GB",
+               cache_size, allowed_size_GB);
```
jemoreira, line 122: How would this happen? Could you use `CHECK` instead to make it an assert instead of an error condition?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
```diff
@@ -0,0 +1,207 @@
...
+Result<void> RunEmpty(const std::string& cache_directory) {
+  CF_EXPECT(RecursivelyRemoveDirectory(cache_directory));
+  CF_EXPECT(EnsureDirectoryExists(cache_directory));
+  LOG(INFO) << fmt::format("Cache at \"{}\" has been emptied", cache_directory);
+  return {};
+}
+
+Result<void> RunSize(const std::string& cache_directory) {
+  std::size_t cache_size = CF_EXPECT(GetDiskUsageGigabytes(cache_directory));
+  LOG(INFO) << fmt::format("Cache at \"{}\": ~{}GB", cache_directory,
```
jemoreira, line 144: Print this to `std::cout`, same for `where`

base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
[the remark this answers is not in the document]
cjreynol: Agreed, done.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
cjreynol: Done.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
cjreynol: Done.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
cjreynol: Done.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
cjreynol: Barring a programming error it should never be able to happen within a single `cvd cache prune` execution, but since we are dealing with the filesystem it seemed prudent to check.  If two prunes were run simultaneously, or somebody was deleting files at the same time, it could theoretically happen.

I agree it should be an assert, done.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
cjreynol: Done.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
```diff
@@ -0,0 +1,206 @@
...
+Result<Action> ToAction(std::string_view key) {
+  if (key == "empty") {
+    return Action::Empty;
+  } else if (key == "info") {
+    return Action::Info;
+  } else if (key == "prune") {
+    return Action::Prune;
+  }
+  return CF_ERRF("Unable to determine action \"{}\"", key);
+}
```
Databean, line 67: Is there a way that the `cvd cache` subcommands can be different top-level `CvdCommandHandler`s? `CvdCommandHandler::CanHandle` has full access to the command line.

base/cvd/cuttlefish/common/libs/utils/files.cpp
```diff
@@ -229,6 +229,52 @@ Result<std::vector<std::string>> DirectoryContents(const std::string& path) {
...
+      CF_EXPECTF(DirectoryContentsPaths(path),
+                 "Failure retrieving contents of directory at \"{}\"", path);
+
+  auto not_self_or_parent_directory = [](std::string filepath) {
+    return !android::base::EndsWith(filepath, ".") &&
+           !android::base::EndsWith(filepath, "..");
+  };
+  std::vector<std::string> filtered;
+  std::copy_if(contents.begin(), contents.end(), std::back_inserter(filtered),
+               not_self_or_parent_directory);
```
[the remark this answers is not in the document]
cjreynol: We could, but we would need to update the `-std=c++17` `copt` we have set in 20 different places in `BUILD.bazel` files.  I'll do a follow-up PR for that.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
jemoreira: I wouldn't mind if we have something like `cvd clear_cache` instead of `cvd cache clear`. I'm thinking on the output of `cvd` or `cvd --help`, that now will show one line for `cache` and then you'll have to run `cvd cache --help` to what sub-sub-commands are there.

cjreynol: I think a handler for `cvd cache info`, `cvd cache prune`, etc would also each show a different top-level help line in `cvd help`.  Since it accumulates from all of its handlers.

I am not personally a fan of long, underscored command names.  Especially if there will be multiple with `_cache` in the name.

base/cvd/cuttlefish/common/libs/utils/files.cpp
```diff
@@ -229,6 +229,52 @@ Result<std::vector<std::string>> DirectoryContents(const std::string& path) {
...
+
+  // std::sort cannot provide a `Result`, but we want to capture the error
+  // message if something goes wrong
+  Result<std::chrono::system_clock::time_point> sort_error;
+  auto greater_mod_time_result = [&sort_error](const std::string a,
```
jemoreira, line 258: Shouldn't `a` be a reference too?

jemoreira (approved):

base/cvd/cuttlefish/common/libs/utils/files.cpp
cjreynol: Yes.  But even fixing that `clang-tidy` is still complaining about a moved value being used in the sort helper calls to `FileModificationTime`.  Will fix it once I figure that out.

base/cvd/cuttlefish/common/libs/utils/files.cpp
```diff
@@ -229,6 +229,52 @@ Result<std::vector<std::string>> DirectoryContents(const std::string& path) {
...
+  return result;
+}
+
+Result<std::vector<std::string>> DirectoryContentsByModTimeDesc(
+    const std::string& path) {
+  std::vector<std::string> contents =
+      CF_EXPECTF(DirectoryContentsPaths(path),
+                 "Failure retrieving contents of directory at \"{}\"", path);
+
+  auto not_self_or_parent_directory = [](std::string filepath) {
```
jemoreira, line 247: Should `filepath` be a `const std::string&` or `std::string_view`?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
Databean: Separate `CvdCommandHandler` implementations does not imply that the names have to be underscored, it's practically free-form on what any particular implementation handles based on the result of `CvdCommandHandler::CanHandle`. The only constraint enforced is that exactly one implementation handles any particular CLI command.

base/cvd/cuttlefish/common/libs/utils/files.cpp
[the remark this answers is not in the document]
cjreynol: Created a new bug to track that, closing this.

base/cvd/cuttlefish/common/libs/utils/files.cpp
cjreynol: Swapped to a reference, still looking into the `clang-tidy` error.

base/cvd/cuttlefish/common/libs/utils/files.cpp
```diff
@@ -229,6 +229,52 @@ Result<std::vector<std::string>> DirectoryContents(const std::string& path) {
   return ret;
 }
 
+Result<std::vector<std::string>> DirectoryContentsPaths(
+    const std::string& path) {
+  std::vector<std::string> result = CF_EXPECT(DirectoryContents(path));
+  std::for_each(result.begin(), result.end(), [&path](std::string& filename) {
+    filename = fmt::format("{}/{}", path, filename);
+  });
```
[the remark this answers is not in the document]
cjreynol: Done.

base/cvd/cuttlefish/common/libs/utils/files.h
```diff
@@ -51,6 +51,11 @@ Result<void> MoveDirectoryContents(const std::string& source,
                                    const std::string& destination);
 bool FileHasContent(const std::string& path);
 Result<std::vector<std::string>> DirectoryContents(const std::string& path);
+Result<std::vector<std::string>> DirectoryContentsPaths(
+    const std::string& path);
+// does not include self-reference or parent reference ("." or "..")
+Result<std::vector<std::string>> DirectoryContentsByModTimeDesc(
```
ser-io, line 57: Does `DirectoryContentsByModTimeDesc` return the absolute paths similar to `DirectoryContentsPaths` or just the base name similar to `DirectoryContents`? If returning the paths, should we name it `DirectoryContentsPathsByModTimeDesc` for consistency?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
```diff
@@ -0,0 +1,206 @@
...
+#include "common/libs/utils/result.h"
+#include "common/libs/utils/tee_logging.h"
+#include "host/commands/cvd/cli/commands/command_handler.h"
+#include "host/commands/cvd/cli/types.h"
+#include "host/commands/cvd/utils/common.h"
+
+namespace cuttlefish {
+
+namespace {
+constexpr int kDefaultCacheSizeGigabytes = 25;
```
ser-io, line 42: nit: use "GB" for a shorter name and to be consistent with variable names.

base/cvd/cuttlefish/common/libs/utils/files.h
```diff
@@ -51,6 +51,11 @@ Result<void> MoveDirectoryContents(const std::string& source,
                                    const std::string& destination);
 bool FileHasContent(const std::string& path);
 Result<std::vector<std::string>> DirectoryContents(const std::string& path);
+Result<std::vector<std::string>> DirectoryContentsPaths(
```
ser-io, line 54: What do you think of having only one helper function returning a vector of `stat` structs and let the caller handle the naming (filepath vs filename) and the sorting depending on their needs? I'm concerned about creating a precedent that could lead to adding other helper functions like `DirectoryContentsByModTimeAsc`, DirectoryContentsByCreateTimeDesc` and so on.

ser-io (dismissed):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
cjreynol: I have a version where I split the handlers, but ran into some issues.  Wrote about them in b/389147618 to tackle this later.  In short, the `help` logic is currently deficient for subcommands with their own sub-actions and the `CanHandle` implementation is not as straightforward as I would like.

Given that the current implementation is functioning and there are no particular plans to update this in the future, I am going to hold off on this for now to move forward to other tasks.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
cjreynol: Agreed, done.

base/cvd/cuttlefish/common/libs/utils/files.cpp
```diff
@@ -229,6 +229,52 @@ Result<std::vector<std::string>> DirectoryContents(const std::string& path) {
...
+  std::vector<std::string> filtered;
+  std::copy_if(contents.begin(), contents.end(), std::back_inserter(filtered),
+               not_self_or_parent_directory);
+
+  // std::sort cannot provide a `Result`, but we want to capture the error
+  // message if something goes wrong
+  Result<std::chrono::system_clock::time_point> sort_error;
+  auto greater_mod_time_result = [&sort_error](const std::string& a,
+                                               const std::string& b) {
+    auto a_time = FileModificationTime(a);
```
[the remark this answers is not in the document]
cjreynol: That is a good catch.  Theoretically the modification times could be getting modified by another process during the sort.  I also like this version better than the one with the complicated sort helper.

base/cvd/cuttlefish/common/libs/utils/files.h
cjreynol: I think the point of not starting down the path of any vague file-handling code becoming a helper is valid.  I moved the path sorting helper into the cache-handling logic.  

I do think `DirectoryContentPaths` is beneficial.  We have some [existing cases](https://cs.android.com/android/platform/superproject/main/+/main:device/google/cuttlefish/common/libs/utils/network.cpp;l=129;drc=5491a3733ee88c940533f7e2b21194a233c599c4) on the AOSP-side where returning full filepaths would be a better helper than just the filenames.  We have uses for both of them.

PR #835, opened 2024-12-21
Fix `GetDiskUsage` and add flag parsing overload

cjreynol:
`ParseInt` was failing on the output of `du` because it is in a format like: "<total num> <filepath>".  This version splits that and takes the first portion.  We should update this later to not rely on `du`, b/385384942 should encompass that.

Returning to the `atoi` implementation was not a good idea for the bytes version, because it was still susceptible to overflow issues.

Changed the return value to be a `size_t` instead of a signed number, because the sizes will not (should not?) ever be negative.

Added a `GflagsCompatFlag` overload for reading in flags that can be compared to the output of this helper.

The `GetDiskUsageGigabytes` will still be used in the upcoming `cvd cache`, while the `*Bytes` version will be used when we sync these changes back to AOSP.

Test: use the `cvd_cache` branch from #834 and run `cvd cache size`

### DIFF
--- a/base/cvd/cuttlefish/common/libs/utils/files.cpp
+++ b/base/cvd/cuttlefish/common/libs/utils/files.cpp
@@ -40,6 +40,7 @@
 #include <array>
 #include <cerrno>
 #include <chrono>
+#include <cstddef>
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
@@ -75,6 +76,8 @@
 #endif
 
 namespace cuttlefish {
+
+static constexpr char kWhitespaceCharacters[] = " \n\t\r\v\f";
 
 bool FileExists(const std::string& path, bool follow_symlinks) {
   struct stat st {};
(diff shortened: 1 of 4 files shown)


base/cvd/cuttlefish/common/libs/utils/files.cpp
```diff
@@ -613,35 +616,40 @@ bool FileIsSocket(const std::string& path) {
...
-  std::array<char, 1024> text_output{};
-  const auto bytes_read = read_fd->Read(text_output.data(), text_output.size());
-  CF_EXPECTF(bytes_read > 0, "Failed to read from pipe: {}", strerror(errno));
-  std::move(subprocess).Wait();
-  int result;
-  CF_EXPECTF(android::base::ParseInt(text_output.data(), &result),
-             "Failure parsing \"{}\" to integer.", text_output.data());
+
+  std::string out;
+  std::string err;
```
cjreynol, line 629: Include this in the error message outputs.  Was accidentally left out.

Databean (approved):

base/cvd/cuttlefish/common/libs/utils/files.cpp
```diff
@@ -613,35 +616,40 @@ bool FileIsSocket(const std::string& path) {
...
+  int return_code = RunWithManagedStdio(std::move(du_cmd), nullptr, &out, &err);
+  CF_EXPECT(return_code == 0, "Failed to run `du` command");
+  CF_EXPECT(out.empty() == false, "No output read from `du` command");
```
jemoreira, line 632: nit: `!out.empty()`?

jemoreira (approved):

base/cvd/cuttlefish/common/libs/utils/files.cpp
cjreynol: done.

base/cvd/cuttlefish/common/libs/utils/files.cpp
cjreynol: disagree but done :)

PR #836, opened 2024-12-21
Add scoped logger to avoid manual logger managing

cjreynol:
The manual managing leaves more opportunity for a mistake or error to cause the previous logger to never be reset.

Also, for tee logging we always wanted to set the Android LogSeverity to `VERBOSE` because the levels are managed internal to those loggers. This is automatically handled as well, which is why the `ScopedLogSeverity` is a member instead of being instantiated alongside the logger.

The new `LogToStderr` (and `ScopedTeeLogger`) will be used in the upcoming `cvd cache`, because it will eventually be called from `cvd fetch` invocations.

Test: TBD, I think a `cvd load` call will have successive logger setups?

### DIFF
--- a/base/cvd/cuttlefish/common/libs/utils/tee_logging.cpp
+++ b/base/cvd/cuttlefish/common/libs/utils/tee_logging.cpp
@@ -133,6 +133,18 @@ TeeLogger::TeeLogger(const std::vector<SeverityTarget>& destinations,
                      const std::string& prefix)
     : destinations_(destinations), prefix_(prefix) {}
 
+ScopedTeeLogger::ScopedTeeLogger(TeeLogger tee_logger)
+    // Set the android logger to full verbosity, the tee_logger will choose
+    // whether to write each line.
+    : scoped_severity_(android::base::VERBOSE) {
+  old_logger_ = android::base::SetLogger(tee_logger);
+}
+
+ScopedTeeLogger::~ScopedTeeLogger() {
+  // restore the previous logger
+  android::base::SetLogger(std::move(old_logger_));
+}
+
 // Copied from system/libbase/logging_splitters.h
 static std::pair<int, int> CountSizeAndNewLines(const char* message) {
   int size = 0;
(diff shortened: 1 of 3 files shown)


base/cvd/cuttlefish/common/libs/utils/tee_logging.cpp
```diff
@@ -133,6 +133,18 @@ TeeLogger::TeeLogger(const std::vector<SeverityTarget>& destinations,
                      const std::string& prefix)
     : destinations_(destinations), prefix_(prefix) {}
 
+ScopedTeeLogger::ScopedTeeLogger(TeeLogger tee_logger)
```
Databean, line 136: Can this accept a `android::base::LogFunction` for parity with `android::base::SetLogger`?

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
```diff
@@ -981,22 +981,10 @@ Result<void> FetchCvdMain(int argc, char** argv) {
   MetadataLevel metadata_level =
       isatty(0) ? MetadataLevel::ONLY_MESSAGE : MetadataLevel::FULL;
 
-  auto old_logger = android::base::SetLogger(
+  auto logger = ScopedTeeLogger(
```
Databean, line 984: nit: could be

```c++
ScopedTeeLogger logger(LogToStderrAndFiles({log_file}, "", metadata_level, flags.verbosity));
```

Databean (approved):

base/cvd/cuttlefish/common/libs/utils/tee_logging.cpp
cjreynol: From chat conversation:

The purpose of using a `TeeLogger` parameter is to imply different behavior from typical android logging.  The constructor forcibly sets the `LogSeverity` to `VERBOSE` for its lifespan, which is specific to the `tee_logging` implementation.  

Will keep it as is, at least for now.

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
cjreynol: Done, thanks for the suggestion

PR #839, opened 2024-12-31
Move power commands logic to LocalInstance class

jemoreira:
Restart commands talk directly to launcher monitor.


Databean:
For commit [5225f40](https://github.com/google/android-cuttlefish/pull/839/commits/5225f40e19ff38d76d92266b799af36c34e6eb5a), is it worth importing the commit history on these files into this git project?

Databean:
> For commit [5225f40](https://github.com/google/android-cuttlefish/pull/839/commits/5225f40e19ff38d76d92266b799af36c34e6eb5a), is it worth importing the commit history on these files into this git project?

Nevermind, the commit message answers my question.

>This is a stripped down version of the original as some dependencies are not yet in this repository. Only entire functions were removed from this version of the library so the interface and implementation of the imported bits are the same as the original; therefore when the full version is imported the code that uses this version will continue working correctly.

So the decision is deferred until the full import is done.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/host_tool_target.cpp
```diff
@@ -89,14 +88,29 @@ Result<std::string> HostToolTarget::GetRestartBinName() const {
...
+
 Result<std::string> HostToolTarget::GetPowerBtnBinName() const {
   return CF_EXPECT(GetBinName({"powerbtn_cvd"}));
 }
 
+Result<std::string> HostToolTarget::GetPowerBtnBinPath() const {
+  return fmt::format("{}/bin/{}", artifacts_path_,
+                     CF_EXPECT(GetPowerBtnBinName()));
+}
+
```
Databean, line 113: What is the distinction between functions that use `artifacts_path_` and functions that use `GetBinName()`?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/power_btn.cpp
```diff
@@ -0,0 +1,86 @@
...
+constexpr char kPowerBtnCmd[] = "powerbtn";
+
+class CvdDevicePowerBtnCommandHandler : public CvdCommandHandler {
+ public:
+  CvdDevicePowerBtnCommandHandler(InstanceManager& instance_manager)
+      : instance_manager_{instance_manager} {}
+
+  Result<bool> CanHandle(const CommandRequest& request) const override {
+    return request.Subcommand() == kPowerBtnCmd;
+  }
```
Databean, line 46: Can this and other new `CvdCommandHandler`s use the [default implementation of `CanHandle`](https://github.com/google/android-cuttlefish/blob/d4b8be15fc0e581e4e87c44fd6121f2b723af34a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/command_handler.cpp#L25)?

base/cvd/cuttlefish/host/commands/cvd/instances/instance_record.cpp
```diff
@@ -83,4 +96,94 @@ Result<Json::Value> LocalInstance::FetchStatus(std::chrono::seconds timeout) {
...
+  CF_EXPECT(WaitForRead(monitor_socket, boot_timeout.count()));
+  RunnerExitCodes boot_exit_code = CF_EXPECT(ReadExitCode(monitor_socket));
+  CF_EXPECT(boot_exit_code != RunnerExitCodes::kVirtualDeviceBootFailed,
+            "Boot failed");
+  CF_EXPECT(boot_exit_code == RunnerExitCodes::kSuccess,
+            "Unknown response" << static_cast<int>(boot_exit_code));
+
+  LOG(INFO) << "Restart successful";
+  return {};
+}
```
Databean, line 134: This makes the launcher control socket the API boundary between `cuttlefish-base` and `device/google/cuttlefish`, rather than executables.

The immediate problem I see there is that this sets the split between `LauncherAction` and `ExtendedLauncherAction` in stone. Does it seem valuable to finish the migration to `ExtendedLauncherAction` first to make it possible to give fields to the existing actions, or is it fine to keep supporting both kinds of `LauncherAction` indefinitely?

If we go through with this, it seems worth documenting the dependency on the AOSP side.

base/cvd/cuttlefish/host/commands/cvd/instances/instance_record.cpp
```diff
@@ -83,4 +96,94 @@ Result<Json::Value> LocalInstance::FetchStatus(std::chrono::seconds timeout) {
...
+  Json::CharReaderBuilder builder;
+  std::string config_file = instance_dir() + "/cuttlefish_config.json";
+  std::ifstream ifs(config_file);
+  std::string errorMessage;
+  Json::Value config;
+  bool parsed = Json::parseFromStream(builder, ifs, &config, &errorMessage);
+  CF_EXPECTF(std::move(parsed), "Could not read config file {}: {}",
+             config_file, errorMessage);
+  return config;
+}
```
Databean, line 167: `cuttlefish_config.json` also becomes part of the public API, with the caveats mentioned above.

base/cvd/cuttlefish/host/commands/cvd/instances/instance_record.cpp
```diff
@@ -19,14 +19,27 @@
...
+#include "host/libs/command_util/runner/defs.h"
+#include "host/libs/command_util/util.h"
 
 namespace cuttlefish {
 
 namespace {
 constexpr int BASE_ADB_PORT = 6520;
 constexpr int BASE_INSTANCE_ID = 1;
+
+void AddEnvironment(Command& cmd, const LocalInstance& instance) {
```
Databean, line 36: nit: `AddEnvironment` is not descriptive. Maybe `AddEnvironmentForInstance`?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/powerwash.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+  cvd_common::Args CmdList() const override { return {kPowerwashCmd}; }
+
+  Result<std::string> SummaryHelp() const override { return kSummaryHelpText; }
+
+  bool ShouldInterceptHelp() const override { return true; }
+
+  Result<std::string> DetailedHelp(
+      std::vector<std::string>& arguments) const override {
+    CHECK(!arguments.empty()) << "Missing subcommand";
```
Databean, line 111: nit: `CF_EXPECT`?

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/host_tool_target.cpp
jemoreira: With `BinName` the caller needs to know that the path to the binary withing the directory is `bin/{binary}` which is an internal detail of the host tools that could be encapsulated by returning the full path.

Deleted the `BinName` versions of these functions since those are no longer used. I intend to do the same for all other binaries here, but not in this PR.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/powerwash.cpp
jemoreira: Removed the check instead. The arguments are not used anymore.

base/cvd/cuttlefish/host/commands/cvd/instances/instance_record.cpp
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/instances/instance_record.cpp
jemoreira: With the host tools still in AOSP, `cvd` would always have to use `LauncherAction` to communicate with the launcher because there are some versions that don't support the extended ones.

I would argue that we finish the migration to `ExtendedLauncherAction` after we move the launcher to cuttlefish-base.

PR #844, opened 2025-01-09
add more logging for the requests that's processed by controler

viewshan:
The logging is used for debugging and triage the requests

### DIFF
--- a/frontend/src/host_orchestrator/orchestrator/controller.go
+++ b/frontend/src/host_orchestrator/orchestrator/controller.go
@@ -117,7 +117,9 @@ type handler interface {
 
 func httpHandler(h handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		log.Println("request:", r.Method, r.URL.Path)
+		log.Printf("handling request(%p) started %s %s", r, r.Method, r.URL.Path)
+		defer log.Printf("handling request(%p) ended", r)
+
 		res, err := h.Handle(r)
 		if err != nil {
 			log.Printf("request %q failed with error: %v", r.Method+" "+r.URL.Path, err)
@@ -473,7 +475,9 @@ type downloadCVDBugReportHandler struct {
 }
 
 func (h *downloadCVDBugReportHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	log.Println("request:", r.Method, r.URL.Path)
+	log.Printf("handling request(%p) started %s %s", r, r.Method, r.URL.Path)
+	defer log.Printf("handling request(%p) ended", r)
+
 	vars := mux.Vars(r)
 	uuid := vars["uuid"]
 	filename := filepath.Join(h.Config.Paths.CVDBugReportsDir, uuid, BugReportZipFileName)


frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -121,9 +121,11 @@ func httpHandler(h handler) http.Handler {
 			return
 		}
 		if res == nil {
+			defer log.Println("request has no response:", r.Method, r.URL.Path)
```
jemoreira, line 124: Don't defer, just log before the return statement

frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -121,9 +121,11 @@ func httpHandler(h handler) http.Handler {
...
 			w.WriteHeader(http.StatusOK)
 			return
 		}
+		defer log.Println("request process finished:", r.Method, r.URL.Path)
```
jemoreira, line 128: Don't defer, log after calling `replyJSONOK` instead

jemoreira (changes requested):
These are going to make the logs very verbose. I think this could only be included after we have support for log verbosity levels as provided by `log/slog`.

ser-io (changes requested):
Let's make the log message focused on the handler function rather on the request, we want to know the handler function started and the handler function exited, we cannot focused on the response cause the current handler cannot know whether the response was fully flushed to the final client, hence suggest adding the following lines at the beginning of handler functions modified in this PR.

```
log.Printf("handling request(%p) started %s %s", r, r.Method, r.URL.Path)                                                                                                            
defer log.Printf("handling request(%p) ended", r)  
```

frontend/src/host_orchestrator/orchestrator/controller.go
viewshan: have remove this particular log line as asked by another comment

ser-io:
Let's have a single commit in this PR.

To squash the current 5 commits into 1, you can use interactive rebase: https://git-scm.com/docs/git-rebase#_interactive_mode. When addressing comments, you can use `git commit --amend` and `git push -f`.

ser-io (changes requested):
Let's have one commit in this PR: https://github.com/google/android-cuttlefish/pull/844#issuecomment-2587409104

frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -125,6 +125,7 @@ func httpHandler(h handler) http.Handler {
 			return
 		}
 		replyJSONOK(w, res)
+		log.Println("handling request(%p) ended %s %s", r, r.Method, r.URL.Path)
```
ser-io, line 128: move this a `defer` statement below the `log.Println("handling request(%p) started %s %s", r, r.Method, r.URL.Path)`

frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -113,7 +113,7 @@ type handler interface {
 
 func httpHandler(h handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		log.Println("request:", r.Method, r.URL.Path)
+		log.Println("handling request(%p) started %s %s", r, r.Method, r.URL.Path)
```
ser-io, line 116: use `Printf`

frontend/src/host_orchestrator/orchestrator/controller.go
ser-io, line 128: remove r.Method and r.URL.Path from the ended log message, these are redundant in that statement.

ser-io (changes requested):

ser-io (approved):

frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -113,7 +113,9 @@ type handler interface {
 
 func httpHandler(h handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		log.Println("request:", r.Method, r.URL.Path)
+		log.Printf("handling request(%p) started %s %s", r, r.Method, r.URL.Path)
+		defer log.Printf("handling request(%p) ended", r)
+
```
ser-io, line 118: nit: remove empty line

ser-io (approved):

viewshan:
> These are going to make the logs very verbose. I think this could only be included after we have support for log verbosity levels as provided by `log/slog`.

Per comments from Sergio, have refactored the log msg and I am now only adding the start/end of the handlers

viewshan:
> These are going to make the logs very verbose. I think this could only be included after we have support for log verbosity levels as provided by `log/slog`.

refactored the log statements.

jemoreira (approved):

PR #848, opened 2025-01-14
Refactors for automatic cache pruning

cjreynol:
Clean up the boundary between `CvdFetchCommandHandler`/`FetchCvdMain` and make cache logic available for use in the `fetch` handler.

The main changes are around the `FetchCvdMain` parameters.  Everything else should be a move or rename.

Test: cvd fetch --target_directory=/tmp/cvd
--default_build=aosp-main/aosp_cf_x86_64_phone-trunk_staging-userdebug
Test: cvd fetch --help
Test: cvd help fetch
Test: cvd cache info
Test: cvd cache prune
Test: cvd cache empty


base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
```diff
@@ -968,9 +968,9 @@ std::string GetFetchLogsFileName(const std::string& target_directory) {
   return target_directory + "/fetch.log";
 }
 
-Result<void> FetchCvdMain(int argc, char** argv) {
-  android::base::InitLogging(argv, android::base::StderrLogger);
-  const FetchFlags flags = CF_EXPECT(GetFlagValues(argc, argv));
+Result<void> FetchCvdMain(std::vector<std::string>& args) {
+  android::base::InitLogging(nullptr, android::base::StderrLogger);
```
Databean: `InitLogging` [uses `argv[0]`](https://github.com/google/android-cuttlefish/blob/43ccb5b093de656aa5e48b474381af172227dbdb/base/cvd/libbase/logging.cpp#L347) for command line output, does this still output logs with the correct name?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/fetch.cpp
```diff
@@ -54,10 +72,12 @@ Result<std::string> CvdFetchCommandHandler::DetailedHelp(
     std::vector<std::string>&) const {
   std::vector<std::string> args = {"--help"};
   // TODO: b/389119573 - Should return the help text instead of printing it
-  CF_EXPECT(FetchCvdMain(args));
+  CF_EXPECT(GetFlagValues(args));
```
Databean, line 75: `GetFlagValues` seems like a somewhat ambiguous name if `fetch_cvd_parser.h` gets a wider reach.

Can this be a static function on `FetchFlags`, like `FetchFlags::Parse`?

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
cjreynol: Thanks for this catch!  I did not realize that connection.  

Define "correct name".  In the `fetch.log` the log entries now look like
```shell
cvd D 01-15 10:34:36 346421 346421 fetch_cvd.cc:955]
cvd I 01-15 10:34:36 346421 346421 fetch_cvd.cc:957] Completed all fetches
```
where `cvd` is from the `InitLogging` call in `cuttlefish/host/commands/cvd/main.cc`).  

I had not given it any thought until now, but `cvd` is technically more correct as the actual executable being run.  We still have the filenames+line numbers, and all of these log messages are still confined to the `fetch.log` which makes the context clear in my opinion.  If the output was intermingled like in the `launcher.log` I could see an argument for changing it, but I believe this is fine as is.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/fetch.cpp
cjreynol: Agreed, done.

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
cjreynol: Going to resolve this for now, but happy to discuss further and make changes in a follow-up PR.

PR #849, opened 2025-01-14
Simplify selectors

Databean:
Pushes down some hardcoded constants and replaces some object-oriented interfaces with imperative interfaces.

Bug: b/389804379
Test: /usr/bin/bazel run :cvd -- --instance_name=1 status

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/frontline_parser.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/frontline_parser.cpp
@@ -30,7 +30,8 @@
 #include "host/commands/cvd/cli/types.h"
 
 namespace cuttlefish {
-namespace {
+
+using selector::ArgumentsSeparator;
 
 /* the very first command line parser
  *
@@ -43,62 +44,10 @@ namespace {
  *     b) non-selector flags
  *  3. subcommand
  *  4. subcommand arguments
- *
- * This is currently on the client side but will be moved to the server
- * side.
  */
-class FrontlineParser {
-  using ArgumentsSeparator = selector::ArgumentsSeparator;
+Result<cvd_common::Args> ExtractCvdArgs(cvd_common::Args& args) {
+  CF_EXPECT(!args.empty());
 
- public:
-  // This call must guarantee all public methods will be valid
-  static Result<std::unique_ptr<FrontlineParser>> Parse(
-      const cvd_common::Args& all_args);
-
-  const std::string& ProgPath() const;
-  std::optional<std::string> SubCmd() const;
-  const cvd_common::Args& SubCmdArgs() const;
-  const cvd_common::Args& CvdArgs() const;
-
- private:
-  FrontlineParser(const cvd_common::Args& all_args);
-
-  // internal workers in order
-  Result<void> Separate();
-  Result<std::unique_ptr<ArgumentsSeparator>> CallSeparator();
-  struct FilterOutput {
-    bool clean;
-    bool help;
-    cvd_common::Args selector_args;
-  };
-  Result<FilterOutput> FilterNonSelectorArgs();
-
-  const cvd_common::Args all_args_;
-  const std::vector<std::string> internal_cmds_;
-  std::unique_ptr<ArgumentsSeparator> arguments_separator_;
-};
-
-Result<std::unique_ptr<FrontlineParser>> FrontlineParser::Parse(
-    const cvd_common::Args& all_args) {
-  CF_EXPECT(!all_args.empty());
-  std::unique_ptr<FrontlineParser> frontline_parser(
-      new FrontlineParser(all_args));
-  CF_EXPECT(frontline_parser != nullptr,
-            "Memory allocation for FrontlineParser failed.");
-  CF_EXPECT(frontline_parser->Separate());
-  return frontline_parser;
-}
-
-FrontlineParser::FrontlineParser(const cvd_common::Args& all_args)
-    : all_args_(all_args) {}
-
-Result<void> FrontlineParser::Separate() {
-  arguments_separator_ = CF_EXPECT(CallSeparator());
-  return {};
-}
-
-Result<std::unique_ptr<selector::ArgumentsSeparator>>
-FrontlineParser::CallSeparator() {
   ArgumentsSeparator::FlagsRegistration flag_registration{
       .known_boolean_flags = {},
       .known_value_flags = {selector::SelectorFlags::kGroupName,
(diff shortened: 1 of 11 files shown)


base/cvd/cuttlefish/host/commands/cvd/cli/frontline_parser.h
```diff
@@ -39,46 +32,7 @@ Result<cvd_common::Args> ExtractCvdArgs(cvd_common::Args& args);
...
-    cvd_common::Args selector_args;
-  };
-  Result<FilterOutput> FilterNonSelectorArgs();
-
-  cvd_common::Args server_supported_subcmds_;
-  const cvd_common::Args all_args_;
-  const std::vector<std::string> internal_cmds_;
-  std::unique_ptr<ArgumentsSeparator> arguments_separator_;
-};
+Result<cvd_common::Args> ExtractCvdArgs(cvd_common::Args& args);
```
jemoreira, line 36: Please update the comment over this to explicitly mention that it returns the cvd arguments after removing them from the input parameter.
That comment seems to have been originally for the FrontlineParser class, which did more than this function.

base/cvd/cuttlefish/host/commands/cvd/cli/frontline_parser.cpp
```diff
@@ -17,119 +17,40 @@
...
-}
-
-std::optional<std::string> FrontlineParser::SubCmd() const {
-  return arguments_separator_->SubCmd();
-}
-
-const cvd_common::Args& FrontlineParser::SubCmdArgs() const {
-  return arguments_separator_->SubCmdArgs();
-}
+  cvd_common::Args cmd_args{separated_arguments.sub_cmd_args};
```
jemoreira, line 49: Why copy the arguments? can't `cmd_args` be a reference?

base/cvd/cuttlefish/host/commands/cvd/cli/frontline_parser.cpp
```diff
@@ -17,119 +17,40 @@
...
+
+  SeparatedArguments separated_arguments = CF_EXPECT(SeparateArguments(args));
 
-  const auto prog_path = frontline_parser->ProgPath();
-  const auto new_sub_cmd = frontline_parser->SubCmd();
-  cvd_common::Args cmd_args{frontline_parser->SubCmdArgs()};
+  cvd_common::Args new_exec_args{separated_arguments.prog_path};
 
-  cvd_common::Args new_exec_args{prog_path};
+  const auto new_sub_cmd = separated_arguments.sub_cmd;
```
jemoreira, line 44: nit: `s/new_sub_cmd/sub_cmd`. I wasted some time searching for the `old_sub_cmd` here.

base/cvd/cuttlefish/host/commands/cvd/cli/frontline_parser.cpp
jemoreira, line 49: nit: `s/cmd_args/sub_cmd_args/`

base/cvd/cuttlefish/host/commands/cvd/cli/selector/arguments_lexer.cpp
```diff
@@ -33,6 +34,86 @@ bool Included(const std::string& item, Sets&&... containers) {
...
+ * A positional token starts with any character other than "-".
+ *
+ * Between flag tokens, there are "known" and "unknown" flag tokens.
+ *
+ */
+class ArgumentsLexerBuilder {
+  using FlagPatterns = ArgumentsLexer::FlagPatterns;
+
+ public:
+  static Result<std::unique_ptr<ArgumentsLexer>> Build();
```
jemoreira, line 110: Why can't this be a static function of `ArgumentsLexer` itself? Probably renamed as `Create`.

base/cvd/cuttlefish/host/commands/cvd/cli/selector/arguments_lexer.cpp
```diff
@@ -79,8 +160,12 @@ ArgumentsLexerBuilder::GenerateFlagPatterns(
   return flag_patterns;
 }
 
-Result<std::unique_ptr<ArgumentsLexer>> ArgumentsLexerBuilder::Build(
-    const LexerFlagsSpecification& known_flags) {
+Result<std::unique_ptr<ArgumentsLexer>> ArgumentsLexerBuilder::Build() {
+  LexerFlagsSpecification known_flags{
+      .known_boolean_flags = {SelectorFlags::kVerbosity},
+      .known_value_flags = {SelectorFlags::kGroupName,
+                            SelectorFlags::kInstanceName},
```
jemoreira, line 167: Probably not for this PR, but knowledge of what these flags are meant to be is duplicated between here and `ParseCommonSelectorArguments` in `selector_common_parser.cpp`. One way to fix that would be for this parser to return the cvd flags already parsed instead of a list of strings that need to be parsed again.

base/cvd/cuttlefish/host/commands/cvd/cli/selector/arguments_lexer.cpp
jemoreira, line 110: Why return a `Result`? There is no configuration or randomization for this builder, if it builds once it should always build, right?
My guess is that the `CF_EXPECT`s inside this function should probably be `CHECK` since those would indicate errors to be catch on the first run.

base/cvd/cuttlefish/host/commands/cvd/cli/selector/arguments_lexer.cpp
```diff
@@ -206,5 +291,16 @@ Result<std::vector<std::string>> ArgumentsLexer::Preprocess(
   return new_args;
 }
 
+Result<std::deque<ArgToken>> TokenizeArguments(
+    const std::vector<std::string>& args) {
+  ArgumentsLexerBuilder builder;
+
+  std::unique_ptr<ArgumentsLexer> lexer = CF_EXPECT(builder.Build());
+  CF_EXPECT(lexer.get());
```
jemoreira, line 299: `Build()` should return `Result<ArgumentsLexer>` instead of the pointer, so this line wouldn't be needed.

base/cvd/cuttlefish/host/commands/cvd/cli/selector/arguments_lexer.cpp
```diff
@@ -206,5 +291,16 @@ Result<std::vector<std::string>> ArgumentsLexer::Preprocess(
...
+
+  std::vector<ArgToken> as_vector = CF_EXPECT(lexer->Tokenize(args));
+  return std::deque<ArgToken>(as_vector.begin(), as_vector.end());
```
jemoreira, line 302: nit: I actually prefer this function to return a `vector` instead of a `deque`. There is no difference performance-wise, it's just that `deque` is used for a very specific purpose (efficient insertion and deletion at the front) which is not needed here. I prefer to introduce the `deque` at the caller site where the need for it is evident.

cjreynol (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/frontline_parser.cpp
Databean: Made it a `const` reference.

Databean: Done.

base/cvd/cuttlefish/host/commands/cvd/cli/frontline_parser.cpp
Databean: Done.

base/cvd/cuttlefish/host/commands/cvd/cli/selector/arguments_lexer.cpp
Databean: Done. It's still called `Build`.

base/cvd/cuttlefish/host/commands/cvd/cli/selector/arguments_lexer.cpp
Databean: Added a comment in both locations referencing the other place to update.

This is a downside of pushing down hardcoded values. This might be possible to address later in deduplicating `CvdFlag` and `Flag.

base/cvd/cuttlefish/host/commands/cvd/cli/selector/arguments_lexer.cpp
Databean: Done.

base/cvd/cuttlefish/host/commands/cvd/cli/selector/arguments_lexer.cpp
Databean: Done.

base/cvd/cuttlefish/host/commands/cvd/cli/selector/arguments_lexer.cpp
Databean: I don't see the harm in using `Result`s throughout considering that the public interface `TokenizeArguments` can fail based on user input anyway.

In general, I prefer to leave it up to the caller to decide whether to abort, whatever the condition is.

Databean:
Additional changes in the most recent push not related to comments.

- Corrects division of flags: there are actually three "value" flags, and no boolean flags.
- Deleted parsing support for boolean flags.
- Fixed tests.

base/cvd/cuttlefish/host/commands/cvd/cli/frontline_parser.cpp
```diff
@@ -17,119 +17,53 @@
...
+ * Being aware of valid subcommands and cvd-specific commands, it will
+ * separate the command line arguments into:
+ *
+ *  1. program path/name
+ *  2. cvd-specific arguments
+ *     a) selector flags
+ *     b) non-selector flags
+ *  3. subcommand
+ *  4. subcommand arguments
+ */
```
jemoreira, line 48: I think this applies to `SeparateArguments` instead.

jemoreira (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/frontline_parser.cpp
Databean: Fixed.

Databean:
Thanks for reviewing!

PR #852, opened 2025-01-16
Automatically prune the fetch cache after fetches

cjreynol:
Keeps the fetch cache from growing unbounded.

Test: # populate cache if necessary
Test: cvd fetch --target_directory=/tmp/cvd/fetch_test
--default_build=12924704,12924909,12924919
Test: cvd cache info
Test: cvd fetch --target_directory=/tmp/cvd/fetch_test2
--default_build=12924921 --max_cache_size_GB=2
Test: # only prunes when caching is enabled
Test: cvd fetch --target_directory=/tmp/cvd/fetch_test3
--default_build=12924578 --max_cache_size_GB=0 --enable_caching=false
Test: cvd cache info

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cache/cache.h
+++ b/base/cvd/cuttlefish/host/commands/cvd/cache/cache.h
@@ -23,6 +23,8 @@
 
 namespace cuttlefish {
 
+inline constexpr std::size_t kDefaultCacheSizeGb = 25;
+
 Result<std::string> EmptyCache(const std::string& cache_directory);
 Result<std::string> GetCacheInfo(const std::string& cache_directory,
                                  bool json_formatted);
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/fetch.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/fetch.cpp
@@ -29,10 +29,12 @@
 #include "common/libs/utils/result.h"
 #include "common/libs/utils/subprocess.h"
 #include "common/libs/utils/tee_logging.h"
+#include "host/commands/cvd/cache/cache.h"
 #include "host/commands/cvd/cli/commands/command_handler.h"
 #include "host/commands/cvd/cli/types.h"
 #include "host/commands/cvd/fetch/fetch_cvd.h"
 #include "host/commands/cvd/fetch/fetch_cvd_parser.h"
+#include "host/commands/cvd/utils/common.h"
 
 namespace cuttlefish {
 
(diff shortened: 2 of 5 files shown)


base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd_parser.h
```diff
@@ -69,6 +71,7 @@ struct BuildApiFlags {
   bool external_dns_resolver = kDefaultExternalDnsResolver;
   std::string api_base_url = kAndroidBuildServiceUrl;
   bool enable_caching = kDefaultEnableCaching;
+  std::size_t max_cache_size_GB = kDefaultCacheSizeGB;
```
Databean, line 74: nit: `max_cache_size_gb`

base/cvd/cuttlefish/host/commands/cvd/cache/cache.h
```diff
@@ -23,6 +23,8 @@
 
 namespace cuttlefish {
 
+inline constexpr std::size_t kDefaultCacheSizeGB = 25;
```
Databean, line 26: nit: `kDefaultCacheSizeGb`

As far as camel casing goes, the [style guide](https://google.github.io/styleguide/cppguide.html#General_Naming_Rules) says to consider an abbreviation a word.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/fetch.cpp
```diff
@@ -60,7 +62,16 @@ Result<void> CvdFetchCommandHandler::Handle(const CommandRequest& request) {
   ScopedTeeLogger logger(
       LogToStderrAndFiles({log_file}, "", metadata_level, flags.verbosity));
 
-  CF_EXPECT(FetchCvdMain(flags));
+  Result<void> result = FetchCvdMain(flags);
+  if (flags.build_api_flags.enable_caching) {
+    LOG(INFO) << "Checking if fetch cache needs to be pruned";
```
Databean, line 67: Can this only log in the case that files were actually deleted?

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/fetch.cpp
cjreynol: We would either need to rewrite `PruneCache` to return that information, or add another helper to give us sizes to compare.

I could raise the `LOG` level for these two to `DEBUG`, so it only ends up in the logfile?

My thinking was that it might actually raise awareness that the cache exists.  None of the logs ever mention it.  If a fetch completes quickly because of cache hits the user is given no indication why.

Databean: Can `PruneCache` return a structure with `cache_directory`, `cache_size`, and `allowed_size_gb`, allowing the caller to choose how to format the values? That makes `CvdFetchCommandHandler` here more responsible for the user-visible output, rather than the current division between the two.

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd_parser.h
cjreynol: Done.

base/cvd/cuttlefish/host/commands/cvd/cache/cache.h
cjreynol: Done.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/fetch.cpp
cjreynol: From our discussion:  Deleting the not-so-informative "Checking..." `LOG` line and will make another pass at the `cache` implementations to have them return information and delegate the output formatting to the callsites.

PR #854, opened 2025-01-16
Improve error message when `cvd` does not find a command handler

cjreynol:
Trying to provide the user enough help to unstick themselves instead of a vague error.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/request_context.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/request_context.cpp
@@ -17,9 +17,11 @@
 #include "host/commands/cvd/cli/request_context.h"
 
 #include <memory>
+#include <string>
 #include <vector>
 
 #include <android-base/logging.h>
+#include <android-base/strings.h>
 
 #include "common/libs/utils/result.h"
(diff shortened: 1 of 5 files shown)


jemoreira (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/request_context.cpp
```diff
@@ -104,7 +101,7 @@ RequestContext::RequestContext(
 
 Result<CvdCommandHandler*> RequestContext::Handler(
     const CommandRequest& request) {
-  return RequestHandler(request, request_handlers_);
+  return CF_EXPECT(RequestHandler(request, request_handlers_));
```
Databean, line 130: The `CF_EXPECT` here is load-bearing, it includes `RequestContext::Handler` in the output stack trace.

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/request_context.cpp
Databean: Sorry I got the old / new versions mixed up.

PR #857, opened 2025-01-16
Separate the cache logic from message formatting

cjreynol:
Allows the callsites to decide what to output.  Specifically, the auto prune in fetching can now omit the message if no files are removed.

Test: # populate cache if necessary
Test: cvd fetch --target_directory=/tmp/cvd/fetch_test --default_build=12924704,12924909,12924919
Test: cvd cache info
Test: cvd fetch --target_directory=/tmp/cvd/fetch_test2 --default_build=12924921 --max_cache_size_gb=2
Test: # only prunes when caching is enabled
Test: cvd fetch --target_directory=/tmp/cvd/fetch_test3 --default_build=12924578 --max_cache_size_gb=0 --enable_caching=false Test: cvd cache info
Test: cvd cache empty

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
@@ -18,6 +18,7 @@
 
 #include <cstddef>
 #include <memory>
+#include <sstream>
 #include <string>
 #include <string_view>
 #include <unordered_map>
@@ -25,6 +26,7 @@
 
 #include <android-base/logging.h>
 #include <fmt/format.h>
+#include <json/json.h>
 
 #include "common/libs/utils/files.h"
 #include "common/libs/utils/flag_parser.h"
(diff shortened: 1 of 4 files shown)


base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
```diff
@@ -93,6 +95,31 @@ Result<CacheArguments> ProcessArguments(
...
+  json_output["size_in_GB"] = std::to_string(cache_size);
+  return json_output.toStyledString();
+}
+
+std::string FormatPruneMessage(const PruneResult& result,
+                               const std::string& cache_directory,
+                               const std::size_t allowed_size) {
+  std::stringstream message;
+  if (result.before > result.after) {
+    message << fmt::format("Cache pruned from {}GB down to {}GB\n",
```
Databean, line 115: Rather than `message << fmt::format(...);`, this could be `fmt::print(message, ...);`

base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
```diff
@@ -109,21 +136,30 @@ Result<void> CvdCacheCommandHandler::Handle(const CommandRequest& request) {
       CF_EXPECT(ProcessArguments(request.SubcommandArguments()));
   std::string cache_directory = PerUserCacheDir();
   switch (arguments.action) {
-    case Action::Empty:
-      std::cout << CF_EXPECTF(EmptyCache(cache_directory),
-                              "Error emptying cache at {}", cache_directory);
+    case Action::Empty: {
+      CF_EXPECTF(EmptyCache(cache_directory), "Error emptying cache at {}",
+                 cache_directory);
+      std::cout << fmt::format("Cache at \"{}\" has been emptied\n",
```
Databean, line 142: I believe this can be `fmt::print(...)` without even mentioning `std::cout`.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/fetch.cpp
```diff
@@ -65,10 +65,16 @@ Result<void> CvdFetchCommandHandler::Handle(const CommandRequest& request) {
...
-    LOG(INFO) << CF_EXPECTF(
+    const PruneResult prune_result = CF_EXPECTF(
         PruneCache(cache_directory, flags.build_api_flags.max_cache_size_gb),
         "Error pruning cache at {} to {}GB", cache_directory,
         flags.build_api_flags.max_cache_size_gb);
+    if (prune_result.before > prune_result.after) {
+      LOG(INFO) << fmt::format(
+          "Cache at \"{}\" pruned from ~{}GB to ~{}GB of {}GB max size",
+          cache_directory, prune_result.before, prune_result.after,
+          flags.build_api_flags.max_cache_size_gb);
```
Databean, line 76: What's the distinction that causes `CvdCacheCommandHandler::Handle` to write directly to `std::cout` while `CvdFetchCommandHandler::Handle` writes to `LOG(INFO)` (iirc `stderr` with a prefix)?

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/fetch.cpp
cjreynol: In short, I do not have a strict distinction.  

Originally in the `cache` implementation it DID use logging with the "no-prefix" metadata setting to output to the user, but then there was a request to change it.  I agreed under the thinking that `cache` is outputting results to the user only for consumption via CLI output and fetch is more logfile-focused with keeping records of actions where the CLI output is simply keeping the user informed while they wait.  To muddy things even further, I have a `LOG(DEBUG)` message in the middle of `PruneCache` for some debugging information if the user wants to see which directories actually get deleted.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
cjreynol: Done.  Since the message generating was not a separate step, I inlined the logic instead of using the helper.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/cache.cpp
cjreynol: Done.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/fetch.cpp
cjreynol: From the group chat with @jemoreira we did not exactly come to a definitive path forward.

There was a general idea that `cvd` commands should output something to `stdout` and not everything to `stderr`, but I think I will leave it as is for now.  We can revisit at a later date and make adjustments to the handlers more broadly for this later on.

PR #853, opened 2025-01-16
Consolidate argument parsers

Databean:
Replaces `ArgumentLexer` with the `Flag` argument handling.

Bug: b/389804379
Test:
```
/usr/bin/bazel run :cvd
/usr/bin/bazel run :cvd -- create
/usr/bin/bazel run :cvd -- create
/usr/bin/bazel run :cvd -- help
/usr/bin/bazel run :cvd -- --instance_name 1 status
/usr/bin/bazel run :cvd -- --instance_name=1 status
/usr/bin/bazel run :cvd -- status
```

PR #860, opened 2025-01-21
Re-enable XML tests in `flag_parser_test.cpp`

Databean:
Re-enable XML tests in `flag_parser_test.cpp`

Introduces a dependency on libxml2.

Restores parts of tests deleted in https://github.com/google/android-cuttlefish/pull/853

Bug: b/389804379
Test: bazel test '//...'


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #868, opened 2025-01-28
Move `instances/{instance_lock,lock_file}.{cpp,h}

Databean:
... to a new subdirectory.

This is one step to making the filesystem organization match the build system.

Bug: b/392949844
Test: bazel test '//...'


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #874, opened 2025-01-31
Fix version tagging and version helper type

cjreynol:
[Update the return type of the version helper](https://github.com/google/android-cuttlefish/commit/4242056b1d5a56741a4953fb8197c539ef8c0f59) 

Also update the callsite, so there are no problems with scope of
objects.

Bug: 393455992

[Fix the genrule for cvd version output](https://github.com/google/android-cuttlefish/commit/877c318b4f166ff420b8b442329ae3dbc923b382) 

The build runs in a sandbox and not in the working directory the `bazel`
command is executed in, so the previous version never had an output to
populate the `cvd version` field.  Also, `git describe` was returning an
old tag.  I updated to return the latest HEAD commit hash instead.

If the
`--working_status_command=path/to/tools/buildutils/stamp_helper.sh` flag
is added to a build, the output goes into a file intended for embedding
outside information for use inside the build sandbox.

The commit hash should not change "often" (like a timestamp would be
different every single build), so we mark it as a stable key.  Then, if
the value changes, every affected action (marked `stamp = 1`) is
triggered.  Volatile keys do not trigger actions when they change.

### DIFF
--- a/tools/buildutils/stamp_helper.sh
+++ b/tools/buildutils/stamp_helper.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+#
+# https://bazel.build/docs/user-manual#workspace-status
+#
+# append `--workspace_status_command=path/to/stamp_helper.sh` to build command
+# to "stamp" this output to `bazel-out/stable_status.txt``
+# 
+# NOTES: 
+# - output must be a "key value" format on a single line
+# - output key must begin with "STABLE_" or the key and value end up in
+#   `volatile_status.txt`
+# - changes to `stable_status.txt` values triggers a re-run of the affected
+#   actions
+
+echo STABLE_HEAD_COMMIT `git rev-parse HEAD`
+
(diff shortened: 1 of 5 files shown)


tools/buildutils/stamp_helper.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+#
+# https://bazel.build/docs/user-manual#workspace-status
```
Databean, line 3: Maybe later we can do something here with [`dpkg-parsechangelog`](https://man7.org/linux/man-pages/man1/dpkg-parsechangelog.1.html) to populate the major and minor versions in the `cvd version` output. No need to look into this now though.

Databean (approved):

tools/buildutils/stamp_helper.sh
cjreynol: I also saw [a suggestion](https://stackoverflow.com/questions/58980342/with-bazel-how-can-you-retrieve-stamped-information-from-a-binary) to use a `cc_library` target with [the `linkstamp` attribute](https://bazel.build/reference/be/c-cpp#cc_library) to get these values instead?  The documentation is sparse.  

I'll make a bug and link to this comment for investigation later.

PR #875, opened 2025-01-31
Expose a function rather than a class for `creation_analyzer.h`

Databean:
The class exists only to call one function, so the public API is changed to something that takes the class creation inputs and returns the function output.

Bug: b/392949844
Test: bazel test '//...'


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #882, opened 2025-02-07
Introduce more HO APIs for caching user artifacts

0405ysj:
About the proposal of this design, please refer to go/cvdr-shared-ua.

Corresponding PR for CO and cvdr is at https://github.com/google/cloud-android-orchestration/pull/408.

### DIFF
--- a/frontend/host/packages/cuttlefish-orchestration/etc/nginx/conf.d/cuttlefish-orchestration.conf
+++ b/frontend/host/packages/cuttlefish-orchestration/etc/nginx/conf.d/cuttlefish-orchestration.conf
@@ -45,6 +45,15 @@ server {
             proxy_set_header Host $host;
         }
 
+        # acquiring lock for user_artifacts
+        location ~* ^/.*/:lock$ {
+            proxy_pass http://127.0.0.1:2081;
+            proxy_connect_timeout 600s;
+            proxy_send_timeout 600s;
+            proxy_read_timeout 600s;
+            send_timeout 600s;
+        }
+
         proxy_pass http://127.0.0.1:2081;
 
         # This is required for uploading huge artifacts.
--- a/frontend/src/host_orchestrator/go.mod
+++ b/frontend/src/host_orchestrator/go.mod
@@ -11,6 +11,7 @@ require (
 )
 
 require (
+	github.com/gofrs/flock v0.8.1 // indirect
 	github.com/golang/protobuf v1.4.3 // indirect
 	github.com/gorilla/websocket v1.5.3 // indirect
 	github.com/hashicorp/errwrap v1.0.0 // indirect
--- a/frontend/src/host_orchestrator/orchestrator/userartifacts.go
+++ b/frontend/src/host_orchestrator/orchestrator/userartifacts.go
@@ -16,6 +16,7 @@ package orchestrator
 
 import (
 	"archive/zip"
+	"context"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -25,7 +26,9 @@ import (
 	"os/user"
 	"path/filepath"
 	"strings"
+	"time"
 
+	"github.com/gofrs/flock"
 	apiv1 "github.com/google/android-cuttlefish/frontend/src/host_orchestrator/api/v1"
 	hoexec "github.com/google/android-cuttlefish/frontend/src/host_orchestrator/orchestrator/exec"
 	"github.com/google/android-cuttlefish/frontend/src/liboperator/operator"
@@ -50,6 +53,10 @@ type UserArtifactsManager interface {
 	UserArtifactsDirResolver
 	// Creates a new directory for uploading user artifacts in the future.
 	NewDir(dir string) (*apiv1.UploadDirectory, error)
+	// Acquire lock on the directory for preventing races.
+	LockFile(dir, name string) (*apiv1.LockFileResponse, error)
+	// Release lock on the directory locked for preventing races.
+	UnlockFile(dir, name string) (*apiv1.LockFileResponse, error)
 	// List existing directories
 	ListDirs() (*apiv1.ListUploadDirectoriesResponse, error)
 	// Update artifact with the passed chunk.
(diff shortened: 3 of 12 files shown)


0405ysj:
@ser-io @ikicha Could I get review for this PR? Thanks!

ikicha (commented):
Instead of providing Lock/Unlock API to a user, What do you think about about using it internally around `UploadFile`? managing lock/unlock from outside might be fragile. (e.g. CO forgot to call Unlock for some error case..)

0405ysj:
> Instead of providing Lock/Unlock API to a user, What do you think about about using it internally around UploadFile? managing lock/unlock from outside might be fragile. (e.g. CO forgot to call Unlock for some error case..)

I think that approach should be better, since it doesn't introduce extra communication between HO and CO. Unless SHA1 conflict, it should be safe. But for that, I should work on lock/unlock mechanism for `ExtractFile` as well.

frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -79,6 +80,7 @@ type HostOrchestratorService interface {
 
 	// Creates a directory in the host where user artifacts can be uploaded to.
 	CreateUploadDir() (string, error)
+	CreateUploadDirWithName(uploadDir string) (string, error)
```
ikicha, line 83: We want to specify `Name` to check if the file is identical during retrieving from the cache. If so, how about changing this method name to CreateUploadDirWithHash or something similar, and move hash calculation logic into here, instead of cvdr/CO to avoid unintentional usage of this API

0405ysj: I'm not certain about just introducing `CreateUploadDirWithHash` can avoid unintentional API usage. Which kinds of usage would you be worried at?

ikicha: Using the same directory name implies the hash of files are identical, doesn't it? I was worried that someone uses `CreateUploadDirWithName` API with their own name even though the contents aren't identical. If we have `CreateUploadDirWithHash(filePaths []string)` and it generate UploadDir by calculating hash inside the method, we can avoid that misuse.

0405ysj: Understood. Your concern is about function naming `CreateUploadDirWithName` can lead any misunderstanding? But I feel `CreateUploadDirWithHash(filePaths []string)` is not still a good way, since naming with `CreateUploadDirWithHash` implies desiring to call `UploadFile` or `ExtractFile` properly afterwards. Instead, what about having a new client api like `CreateUploadDirWithFiles`, which performs all tasks(creating dir, uploading files, extracting files)?

ikicha: SGTM, so will hash/lock/cache/upload happen under the hood with `CreateUploadDirWithFiles`?

0405ysj: Yes, and also including extracting task. But I want to hear @ser-io's opinion briefly before implementing with that way. Since perhaps there are some reasons for designing HO Client APIs for each task and use those in CO/cvdr.

ser-io (changes requested):
Can you add more context to the PR on the problems yo're trying to solve that requires adding the new flock based locking mechanism?

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -152,30 +230,36 @@ func writeChunk(filename string, chunk UserArtifactChunk) error {
 }
 
 func (m *UserArtifactsManagerImpl) ExtractArtifact(dir, name string) error {
-	dir = filepath.Join(m.RootDir, dir)
-	if ok, err := fileExist(dir); err != nil {
+	dirPath := filepath.Join(m.RootDir, dir)
```
ser-io, line 233: Check if there's an ongoing operation extracting the file here: https://github.com/google/android-cuttlefish/blob/main/frontend/src/host_orchestrator/orchestrator/controller.go#L573 and return the same operation rather than error. This might need modifying the operations API, if so do it in a different commit. 

There's no need to use the flock based concurrency mechanism here.

ser-io (changes requested):

frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -504,13 +505,23 @@ func (c *HostOrchestratorServiceImpl) UploadFileWithOptions(uploadDir string, fi
...
-	uploader := &FilesUploader{
-		HTTPHelper:    c.HTTPHelper,
-		DumpOut:       c.HTTPHelper.Dumpster,
-		UploadDir:     uploadDir,
-		UploadOptions: uploadOpts,
+	lockRes := &hoapi.LockFileResponse{}
+	if err := c.HTTPHelper.NewPostRequest(fmt.Sprintf("/userartifacts/%s/%s/:lock", uploadDir, filepath.Base(filename)), nil).JSONResDo(lockRes); err != nil {
+		return err
+	}
+	if !lockRes.UploadCompleted {
```
ser-io, line 512: what happens if clients don't call the :lock endpoints and rather start uploading files directly? Are we expecting all clients to be good citizens? we cannot rely on that. Let's elaborate on what's the problem you're trying to solve first, hence we can discuss on the potential API changes.

ser-io (changes requested):

ser-io (changes requested):

frontend/src/host_orchestrator/go.mod
```diff
@@ -11,6 +11,7 @@ require (
 )
 
 require (
+	github.com/gofrs/flock v0.8.1 // indirect
```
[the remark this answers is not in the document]
0405ysj: Multiple HOs should try to acquire and release on the same lock, since docker volume(representing user artifact dir) is being shared across multiple docker instances(representing each CF host).

When applying memory solution(I understood as using mutex), I should consider about shared memory across processes. Potentially semaphore could be the solution, but it seems each docker instance has isolated ipc namespace so that semaphore couldn't be shared across docker instance as default.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
0405ysj: As I wrote at another comment, it seems there was no consideration about multiple HOs/processes for the design of operation, I saw sync.rwmutex is used. I think using flock is still needed, to consider multiple HOs?

0405ysj:
> Can you add more context to the PR on the problems yo're trying to solve that requires adding the new flock based locking mechanism?

I forgot to add corresponding design document about this PR. I added at the top.

frontend/host/packages/cuttlefish-orchestration/etc/nginx/conf.d/cuttlefish-orchestration.conf
```diff
@@ -45,6 +45,15 @@ server {
             proxy_set_header Host $host;
         }
 
+        # acquiring lock for user_artifacts
```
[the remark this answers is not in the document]
0405ysj: AFAIR, there was default timeout period in nginx configuration. Since HO is wrapped with nginx, HO is already affected by nginx timeout configuration. Those lines are for increasing timeout, not to give timeout restriction.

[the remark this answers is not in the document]
ser-io: Yes, all of the HO should work under the timeout configs set in nginx for the HO.  if a new endpoint takes more than that it should handled  through out the operations api, but not by changing the nginx config.

frontend/src/host_orchestrator/go.mod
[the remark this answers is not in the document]
ser-io: I see, I wasn't aware of this. Added comments to the design doc.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
ser-io: Yes, this seems to be specific to the single server deployment. Added comments to the design doc.

frontend/src/libhoclient/host_orchestrator_client.go
ser-io: It's not about security, the comment is related to the API design itself. With these new methods, we cannot force every client implementation to call :lock before upload. We have control over the  HO behavior,  but every person can implement its own HO http client.

ser-io (changes requested):

PR #883, opened 2025-02-08
Simplify base64.cpp and fix 0 byte ending issue

Databean:
Uses `EVP_EncodedLength` to avoid a local implementation, and uses `EVP_DecodeBase64` to handle a corner case around null bytes at the end of input data for base64 encoding.

Counterpart to https://android-review.googlesource.com/c/device/google/cuttlefish/+/3481535

Bug: b/395180112
Test: bazel test :utils_test


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #887, opened 2025-02-14
Enable cvd powerbtn in HO API

ser-io:


frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -111,10 +113,10 @@ type HostOrchestratorService interface {
 	// Create cvd bugreport.
 	CreateBugReport(group string, opts CreateBugReportOpts, dst io.Writer) error
 
-	// Powerwash the device.
+	// [DEPRECATED] Use `ExecCommandWithNoOutput`
```
jemoreira, line 116: I actually prefer to keep these and make `ExecCommandWithNoOutput` private in the interface implementation.

That makes the host orchestrator service endpoints checked at compile time instead of specifying them as a string.

The other reason not to join these operations in a single endpoint is that the only thing common to all of them is that they don't return content (although they should return success/failure), but that's an implementation detail and not a strong reason to pool them together.

ser-io: I agree with your points, and we do as you mentioned in the HO code, every service endpoint is checked at compile time. 

The change in question happens in  the client implementation, where I started noticing that calls like `stop`, `powerwash` and `powerbtn` could be refactored on the client side having a rather smaller client API.

jemoreira: I was referring to the client too (should we find a better name for that client that "service"?)

Basically I don't want this code: `ho_client.Execute("powerbtn")`, because there is no compile-time check of the `"powerbtn"` string.

ser-io: Agree, removed the refactor commit from this PR keeping the focus on the new `powerbtn` functionality.

jemoreira (approved):

PR #890, opened 2025-02-26
Avoid using harcoded HO port values.

ser-io:


e2etests/orchestration/common/common.go
```diff
@@ -41,12 +39,16 @@ func Setup(port int) (*TestContext, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	ipAddr, err := dockerHelper.getIpAddr(id)
 	if err != nil {
 		return nil, err
 	}
 	result.DockerContainerID = id
-	result.ServiceURL = fmt.Sprintf("http://0.0.0.0:%d", port)
+	result.ServiceURL = fmt.Sprintf("http://%s:2080", ipAddr)
```
jemoreira, line 51: can 2080 not be hardcoded?

e2etests/orchestration/common/common.go
```diff
@@ -132,6 +133,29 @@ func (h *DockerHelper) RunContainer(img string, hostPort int) (string, error) {
...
+	var found *types.Container
+	for _, container := range listRes {
+		if container.ID == id {
+			found = &container
+			break
+		}
+	}
+	if found == nil {
+		return "", fmt.Errorf("container not found: %q", id)
+	}
```
jemoreira, line 151: Instead of looping over all containers here, could the client have a `ContainerById` method instead?

jemoreira (approved):

e2etests/orchestration/common/common.go
ser-io: Good call, used [ContainerInspect](https://github.com/docker/docker/blob/v28.0.1/client/container_inspect.go#L14).

e2etests/orchestration/common/common.go
ser-io: The docker image used in the e2e tests have the HO running at 2080 and we don't have logic in place to update the container and change the HO port. 

I used a const for the value though.

jemoreira (approved):

PR #913, opened 2025-03-12
Fix includes for environment.h

Databean:
`std::optional` is used but not included directly in environment.h.


jmacnak (approved):

Databean:
Thanks for reviewing!

PR #923, opened 2025-03-13
Import libcn-cbor

ser-io:
- b/402250403
- bazel test @libcn-cbor//:libcn-cbor_test

### DIFF
--- a/base/cvd/MODULE.bazel
+++ b/base/cvd/MODULE.bazel
@@ -89,3 +89,10 @@ git_repository(
     commit = "af6ac8265bbd07bcf977526458b60305c4304288",
     remote = "https://chromium.googlesource.com/libyuv/libyuv.git",
 )
+
+git_repository(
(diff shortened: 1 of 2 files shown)


base/cvd/MODULE.bazel
```diff
@@ -82,3 +82,12 @@ git_repository(
     commit = "952f776f6573aafbb62ea717d871cd1d6816c387", # v1.4.309
     remote = "https://github.com/KhronosGroup/Vulkan-Headers.git",
 )
+
+git_repository(
```
Databean, line 86: FYI on the internal side we'll need to do a go/thirdparty import for this library.

Databean (approved):

base/cvd/MODULE.bazel
ser-io: yes indeed, was looking into it and there's no libcn-cbor.

PR #942, opened 2025-03-16
RPMbuild Support

syslogic:
Have now started over, since the rebasing didn't work that well.
This also builds with: `./tools/buildutils/build_packages.sh`

Currently just unsure about the changed `bazel` build target.
And the TypeScript `server_connector.d.ts` might not be required.


daedroza:
Hello, thank you for putting this up - we've missed to update the instructions for installing for the rpm packages using dnf!

syslogic:
Since the build targets have changed, I'm not sure anymore, that everything required is being built. I mean, this would a) not only require a code review, but a resulting package review, in order to make sure everything is installed where it belongs and that there are no useless files contained.

And b), likely would also require a separate [`workflow`](https://github.com/google/android-cuttlefish/tree/main/.github/workflows), since such workflow runner is either Debian based or RHEL based.

syslogic:
That [`rpm-builder`](https://github.com/syslogic/android-cuttlefish/blob/rpm-build/.github/workflows/rpm-builder.yml) workflow is merely working. Haven't built any packages so far, because the build-time is rather limited. I'm trying to upload the built packages as runner artificats.

ser-io:
@syslogic thanks for your contribution. Converted the PR to Draft as it seems it's still work in progress and it's spamming the reviewers on every push. Feel free to mark it "Ready for review" once ready.

syslogic:
@ser-io I was extensively testing the workflow to provide the RHEL environment for building `*.rpm` from these `*.spec` files on GitHub, because the runner images are all Ubuntu. And it seems the `entrypoint` scripts are occassionally stuck at an obsolete version, which is a problem. I could even automate RPM package signature, when having a GPG `*.asc` signature and passphrase set up as secrets and upload to RPM repository. This PR now exceeds the maximum commit limit; will start a new PR.

PR #948, opened 2025-03-17
Migrate cvd_import_locations

cjreynol:
And fix a dependency error in `liblocation`, which was causing a link failure trying to build this target.

Bug: 401510752


cjreynol:
Commits folded into https://github.com/google/android-cuttlefish/pull/948 for the sake of having less PRs right now.

PR #949, opened 2025-03-17
Migrate run_cvd

cjreynol:
Also added a missing visibility declaration (previously #956) and updated the generated `build/version.h` to be shared as a library instead of a single header to be included privately.

Bug: 401510712


base/cvd/cuttlefish/host/commands/cvd/cli/commands/BUILD.bazel
```diff
@@ -60,7 +60,7 @@ clang_tidy_test(
 cc_library(
     name = "leaf_commands",
     srcs = [
-        "//:build_version_header",
+        #"//:build_version_header",
```
jemoreira: delete?

jemoreira (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/BUILD.bazel
cjreynol: Already deleted in a later commit, at the time of a merge conflict resolution.  Should have been in the earlier commit.

PR #955, opened 2025-03-18
Add swiftshader import and build

jmacnak:
Bug: b/404548466
Test: cd base/cvd && bazel build @swiftshader//:swiftshader


ser-io (approved):

cjreynol (approved):
From the commit to build:

`Test: cd base/cvd && bazel build @swiftshader//:vk_swiftshader`

PR #956, opened 2025-03-18
Add visibility declaration

cjreynol:
Otherwise there is a visibility error trying to include it.

```
ERROR: run_cvd/BUILD.bazel:26:11: in cc_library rule //cuttlefish/host/commands/run_cvd:librun_cvd: Visibility error:
target '//cuttlefish/host/frontend/webrtc:libcuttlefish_webrtc_command_channel' is not visible from
target '//cuttlefish/host/commands/run_cvd:librun_cvd'
Recommendation: modify the visibility declaration if you think the dependency is legitimate. For more info see https://bazel.build/concepts/visibility
```

Test: bazel build
//cuttlefish/host/frontend/webrtc:libcuttlefish_webrtc_commands_proto

PR #957, opened 2025-03-18
migrate webrtc signaling headers

jemoreira:


base/cvd/cuttlefish/host/frontend/webrtc_operator/constants/BUILD.bazel
```diff
@@ -0,0 +1,8 @@
+load("//:build_variables.bzl", "COPTS")
+
```
cjreynol, line 2: Needs the visibility declaration:

```
package(
    default_visibility = ["//:android_cuttlefish"],
)
```
https://github.com/google/android-cuttlefish/pull/956

cjreynol (approved):

base/cvd/cuttlefish/host/frontend/webrtc_operator/constants/BUILD.bazel
cjreynol: Also, we should probably `clang-tidy` this since it is source code?  We have for everything else, even if this is a simple file.

jemoreira: thanks! done

PR #963, opened 2025-03-19
Import wayland deps and build libcuttlefish_wayland_server

jmacnak:
Bug: b/402245771
Bug: b/402244573
Bug: b/402244575
Bug: b/402244120


base/cvd/MODULE.bazel
```diff
@@ -126,3 +126,47 @@ git_repository(
...
+)
+
+git_repository(
+    name = "expat",
+    build_file = "@//:BUILD.expat.bazel",
+    commit = "2691aff4304a6d7e053199c205620136481b9dd1", # 2.6.4
+    remote = "https://github.com/libexpat/libexpat/",
+    patch_strip = 1,
+    patches = [
+        "@//:PATCH.expat.patch",
```
Databean, line 151: Something that might be more convenient to maintain than patches for introducing entirely new files is cheating a bit on the dependencies between repositories.

```
# base/cvd/expat_config.h
# regular file, the one that would be introduced by the patch

# base/cvd/BUILD.expat.bazel
...
cc_library(
    name = "expat",
    srcs = [
        "@//:expat_config.h",
...

# base/cvd/BUILD.bazel
exports_files([".clang-tidy", "expat_config.h"])
```

I think this should work, but I'm having trouble validating since if I make this change locally, `bazel build @expat//:expat` succeeds for me even when `base/cvd/expat_config.h` has invalid C++ code.

Databean (approved):

base/cvd/MODULE.bazel
jmacnak: Done. I couldn't quite figure out how to work around needing the ".." with the wayland config and bazel's include setup.

base/cvd/external/expat/BUILD
```diff
@@ -0,0 +1,13 @@
+package(
```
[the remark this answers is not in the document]
jmacnak: Done.

base/cvd/BUILD.crosvm.bazel
```diff
@@ -0,0 +1,41 @@
...
+    outs = ["virtio-gpu-metadata-v1.c"],
+    cmd = "$(location @wayland//:wayland_scanner) -s private-code < $(location gpu_display/protocol/virtio-gpu-metadata-v1.xml) > $(location virtio-gpu-metadata-v1.c)",
+    tools = ["@wayland//:wayland_scanner"],
+)
+
+genrule(
+    name = "wayland_virtio_gpu_metadata_server_header",
+    srcs = ["gpu_display/protocol/virtio-gpu-metadata-v1.xml"],
+    outs = ["virtio-gpu-metadata-v1.h"],
+    cmd = "$(location @wayland//:wayland_scanner) -s server-header < $(location gpu_display/protocol/virtio-gpu-metadata-v1.xml) > $(location virtio-gpu-metadata-v1.h)",
```
ser-io, line 19: nit: as the rule has single file input and output we can use `$(SRCS)` and `$@` or `$(OUTS)` to avoid repeating the filenames: `$(location @wayland//:wayland_scanner) -s server-header < $(SRCS) > $(OUTS)"`

jmacnak: Done!

jemoreira (approved):

PR #968, opened 2025-03-20
Migrate licuttlefish_host_confui and remaining dependencies

jemoreira:
Dependencies list:
- confui_sign (previously part of libcuttlefish_security)
- screen_connector_common (previously part of screen_connector)
- common/libs/concurrency
- libteeui_localization


ser-io (approved):

base/cvd/cuttlefish/host/libs/confui/BUILD.bazel
```diff
@@ -0,0 +1,105 @@
...
+
+genrule(
+    name = "roboto_medium_font",
+    srcs = [
+        "Roboto-Medium.ttf",
+    ],
+    outs = [
+        "roboto_medium.cc",
+    ],
+    cmd = "./$(location generate_font_cc.sh) \"$<\" RobotoMedium >\"$@\"",
```
Databean, line 16: When running these against the internal build system, we'll have to use `$(location)` to get the paths to `xxd` and `sed` which are used inside the shell script. However IIUC `$(location)` expansion only works inside the `cmd` itself and not within the shell script.

Would it be possible to use a loop comprehension or macro rather than a shell script?

https://stackoverflow.com/a/48511351

Databean (approved):

base/cvd/cuttlefish/host/libs/confui/BUILD.bazel
jemoreira: After further discussion we agreed to keep it this way since it seems the internal build system also supports this way of calling the commands.

PR #970, opened 2025-03-21
Support bazel runfiles as host package migration symlink targets.

ser-io:
- Config example

```
symlinks: {
  target: "runfile:_main/cuttlefish/host/commands/logcat_receiver/logcat_receiver"
  link_name: "bin/logcat_receiver"
}
```

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
+++ b/base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
@@ -351,12 +351,31 @@ Result<void> SubstituteWithFlag(const std::string& target_dir, const std::vector
   return {};
 }
 
+static constexpr std::string_view kRunfilePrefix = "runfile:";
+
 Result<void> SubstituteWithMarker(const std::string& target_dir, const std::string& marker_file) {
   std::string content;
   CF_EXPECTF(android::base::ReadFileToString(marker_file, &content),
              "failed to read '{}'", marker_file);
   fetch::HostPkgMigrationConfig config;
   CF_EXPECT(google::protobuf::TextFormat::ParseFromString(content, &config), "failed parsing debian_substitution_marker file");
+  std::unique_ptr<Runfiles> runfiles = NULL;
+  auto resolve_runfile  = [&runfiles](const std::string& src) -> Result<std::string> {
+    if (src.rfind(kRunfilePrefix, 0) != 0) {
+      return src;
+    }
+    std::string self_path;
+    CF_EXPECT(android::base::Readlink("/proc/self/exe", &self_path));
+    CF_EXPECTF(self_path != fmt::format("{}/cvd", kInstallDir),
+              "runfiles cannot be resolved when executing installed cvd: '{}'", src);
+    if (runfiles == NULL) {
+       std::string runfiles_error;
+       runfiles.reset(
+          Runfiles::Create(self_path, BAZEL_CURRENT_REPOSITORY, &runfiles_error));
+       CF_EXPECTF(runfiles.get(), "Could not load runfiles: '{}'", runfiles_error);
+    }
+    return runfiles->Rlocation(src.substr(kRunfilePrefix.length()));
+  };
   for (int j = 0; j < config.symlinks_size(); j++) {
     const fetch::Symlink& symlink = config.symlinks(j);
     std::string full_link_name =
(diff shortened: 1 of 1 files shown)


base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
```diff
@@ -351,12 +351,31 @@ Result<void> SubstituteWithFlag(const std::string& target_dir, const std::vector
...
+static constexpr std::string_view kRunfilePrefix = "runfile:";
+
 Result<void> SubstituteWithMarker(const std::string& target_dir, const std::string& marker_file) {
   std::string content;
   CF_EXPECTF(android::base::ReadFileToString(marker_file, &content),
              "failed to read '{}'", marker_file);
   fetch::HostPkgMigrationConfig config;
   CF_EXPECT(google::protobuf::TextFormat::ParseFromString(content, &config), "failed parsing debian_substitution_marker file");
+  std::unique_ptr<Runfiles> runfiles = NULL;
+  auto resolve_runfile  = [&runfiles](const std::string& src) -> Result<std::string> {
```
jemoreira, line 363: Can this be a couple of functions instead of a lambda? One function to create the runfiles pointer and another to resolve a file given that pointer.

Databean:
> gathers all required binaries and symlinks them to the cvd directory

Note that if you want to reuse this with `cuttlefish-base.install`, it copies symlinks as-is without resolving them into the debian package, which is not what we want.

> that way cvd only needs to ask if the binaries are present in the same directory as itself and that would work in bazel

I'm assuming by "cvd directory" you mean the directory that the output executable of `bazel build :cvd` lands in, but we don't directly control that directory. If we add `data` dependencies, they end up in a `runfiles` subdirectory of the cvd directory. The closest I can come up with to what you're describing is this scheme in a new directory:

```
genrule(
    name = "exes_dir_rule",
    srcs = [
        "//cuttlefish/common/frontend/socket_vsock_proxy",
        "//cuttlefish/host/commands/cvd",
        "//cuttlefish/host/commands/start:cvd_internal_start",
    ],
    outs = [
        "exes_dir/cvd",
        "exes_dir/cvd_internal_start",
        "exes_dir/socket_vsock_proxy",
    ],
    cmd = "mkdir -p $(RULEDIR)/exes_dir && " +
        "cp $(location //cuttlefish/host/commands/cvd) $(RULEDIR)/exes_dir/cvd && " +
        "cp $(location //cuttlefish/common/frontend/socket_vsock_proxy) $(RULEDIR)/exes_dir/socket_vsock_proxy && " +
        "cp $(location //cuttlefish/host/commands/start:cvd_internal_start) $(RULEDIR)/exes_dir/cvd_internal_start",
)

sh_binary(
    name = "cvd",
    srcs = ["exes_dir/cvd"],
)
```

`bazel run :cvd` against the `sh_binary` target will still have an `argv[0]` pointing outside the bazel directory, but `/proc/self/exe` should correctly reference the copied version in `exes_dir`.

ser-io:
> I would prefer not to continue introducing dependencies to `bazel` in our cvd code. Those feel backward to me: the build system needs to know about the code, the code should be build system agnostic.
> 
> With that in mind my preferred way to solve the "run cvd in development" problem is to create a custom bazel rule that gathers all required binaries and symlinks them to the cvd directory, that way `cvd` only needs to ask if the binaries are present in the same directory as itself and that would work in bazel and when deployed as part of the debian package. cvd's cc_binary needs to depend on that rule to ensure it's built with it.

I agree. For development, as we can use a local `debian_substitution_marker` we can use the local binaries directly as targets, we just need to make sure we are using the latest version building the code `bazel build //cuttlefish/host/commands/...` before running fetch. We can do the following example:

```
symlinks: {
  target: "PATH_TO_CVD/bazel-bin/cuttlefish/host/commands/logcat_receiver/logcat_receiver"
  link_name: "bin/logcat_receiver"
}
```
1. bazel build //cuttlefish/host/commands/logcat_receiver

2. LOCAL_DEBIAN_SUBSTITUTION_MARKER_FILE=/tmp/debian_substitution_marker bazel run //cuttlefish/host/commands/cvd -- fetch --default_build=aosp-main/aosp_cf_x86_64_phone-trunk_staging-userdebug --target_directory=$HOME/dl

@Databean what do you think of removing the runfile logic in favor of local debian_substitution_marker.

Databean:
The problem I have with using a local version of `debian_substitution_marker` is that it's an unvalidated companion file that as written needs to have one version that links to `/usr/lib/cuttlefish-common` and one version that links into the `bazel-bin` output directory. The `bazel-bin` version will "happen to work" if you locally `blaze build` a dependent target, but can use a stale version at `bazel run :cvd` time if that build target left out of the `data` dependencies, which at compile time validates that the targets exist, but not that it is complete.

I tried to work out @jemoreira 's request in https://github.com/google/android-cuttlefish/pull/972 . It looks gross, but recreates the debian output files in a way that satisfies the properties of

> - `cvd` only needs to ask if the binaries are present in the same directory as itself
> - that would work in bazel and when deployed as part of the debian package

`cuttlefish-base.install` now depends on this target, so they won't diverge

> cvd's cc_binary needs to depend on that rule to ensure it's built with it.

(though the correct target to run for the last one becomes `//cuttlefish/debian_package:cvd`)

Databean:
Made a second attempt at this in https://github.com/google/android-cuttlefish/pull/976.

PR #981, opened 2025-03-26
Include //cuttlefish headers relative to the bazel workspace in run_cvd

Databean:
This lifts the requirements that libraries it includes have strip_include_prefix lines in their bazel build files.

The separate targets are no longer necessary now that run_cvd includes its own headers from paths relative to the bazel workspace.


jmacnak (approved):

Databean:
Thanks for reviewing!

PR #982, opened 2025-03-26
Enforce include-cleaner validation on kernel_log_monitor.

Databean:
Also makes `cuttlefish/` includes relative to the bazel workspace, lifting the requirement to use strip_include_prefix on dependencies.


jmacnak (approved):

ser-io (approved):

Databean:
Thanks for reviewing!

PR #983, opened 2025-03-26
Build and package lavapipe

jmacnak:
For use with `--gpu_mode=gfxstream_guest_angle_host_lavapipe`

Bug: b/406553848


jmacnak:
(The permission errors in kokoro was due to 2 different genrules trying to generate the same files. Fixed to have one genrule generate those files)

PR #988, opened 2025-03-27
Revert "Expose `gpu_mode` parameter in env config."

Databean:
This reverts commit 9a267341274b13f970ad1bcba97cc41dbf78d499.

The reason is a bad interaction with a gpu_mode setting in a legacy config.json file which sets flag defaults in cvd_internal_start at lower priority than an unset gpu_mode value in a canonical configuration.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/parser/instance/cf_graphics_configs.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/parser/instance/cf_graphics_configs.cpp
@@ -90,19 +90,11 @@ bool RecordScreen(const Instance& instance) {
   }
 }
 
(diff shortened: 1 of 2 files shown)


ser-io (commented):
Isn't a a way to fix forward this behavior? If a client is already using this field in their configurations, we are going to break them.

ser-io (approved):

Databean:
> Isn't a way to fix forward this behavior? If a client is already using this field in their configurations, we are going to break them.

The best strategy I can think of for that is to change the behavior of flag generation to not generate flags for fields that are not present . We could apply that specifically to the `gpu_mode` setting, but it seems safer to revert this first and figure out a new strategy going forwards.

PR #990, opened 2025-03-27
Separate run_cvd/launch into a separate bazel target.

Databean:
This allows each to have a smaller dependency list and parallelizes clang-tidy between the two of them, though `launch` takes significantly more time.


Databean:
Thanks for reviewing!

PR #995, opened 2025-03-28
Split `//cuttlefish/host/commands/cvd/cli/commands:leaf_commands` and enforce `misc-include-cleaner`

Databean:
https://github.com/google/android-cuttlefish/blob/2fb02a8034a2784495621dfcdabf9389a4779f00/base/cvd/cuttlefish/host/commands/cvd/cli/commands/BUILD.bazel#L60

This is one of the larger bazel targets in our repository.

There's also a limitation with clang_tidy_test that makes it single-threaded, which is slow for large targets. This target takes about two minutes on our desktops.

With smaller targets, there is a risk that header includes and dependencies desync from the true dependencies of the sources inside. Using misc-include-cleaner and layering_check together validate that both headers and dependencies are correct.

Bug: b/407081616


cjreynol (approved):

Databean:
Thanks for reviewing!

PR #997, opened 2025-03-30
RPM Support & Workflow

syslogic:
Now the `rpmbuild` runs `bazelisk` to build `cvd`, but only when running the container on my notebook. However, when running the container inside a GitHub runner, it does not seem to build anything, but it rather seems as if `docker` just goes away and `--entrypoint` is also not supported (therefore the two build-stages); maybe due to limited RAM? I can't get any output from `docker`, unless maybe mapping `stdout` to whatever logging facility there is availble by default.

    Attempting to copy to /tmp directory ...
    cp: cannot stat '/var/lib/docker/volumes/13ae35e7edade2d3ae5095f49d7abc787a3ed833bb93fc4913c1d85db68398e7/_data/android-cuttlefish-rpm.tar.gz': No such file or directory
    Error: Process completed with exit code 1.

These are the most relevant files:

- [`Dockerfile`](https://github.com/syslogic/android-cuttlefish/blob/rpm-build/docker/rpm-builder/Dockerfile)
- [`reusable-docker.yml`](https://github.com/syslogic/android-cuttlefish/blob/rpmbuild/.github/workflows/reusable-docker.yml)
- [`rpm-builder.yml`](https://github.com/syslogic/android-cuttlefish/blob/rpm-build/.github/workflows/rpm-builder.yml)
- [`rpm-builder-matrix.yml`](https://github.com/syslogic/android-cuttlefish/blob/rpmbuild/.github/workflows/rpm-builder-matrix.yml) is not in use anymore, but may be useful.

 ---

The image is also available on GitHub (this is the `builder` stage, which builds and terminates):

    ghcr.io/syslogic/rpm-builder:latest

When testing the container offline (usually with the `testing` stage):

 - The `docker run` run configuration requires build options `--target builder` or `--target testing` set.

 - On Linux, I'll also have to add build args `REPO_USER=syslogic` and `REPO_NAME=android-cuttlefish`.

I also have the R/W bind-mounts working there, which speeds up the rebuilding.<br>
In a GitHub runner, this mount is what should be cached.

 - `/home/github/android-cuttlefish/.cache:/root/.cache`
 - `/home/github/android-cuttlefish/.rpms/:/root/.rpms`

That `reusable-docker.yml` currently does not configure any volume / mount-point binds with `docker run`.

Docker Run Configuration

Possible runner limitations:

 - The GitHub runner seems to run out of RAM and exits rather quickly.
 - It may as well be the size of the cache, which is about `13.8` GB.

The problem with the runner is:

 - The `*.tar.gz` would instantly need to be uploaded, because **the file is gone, as soon as the runner step ends**.
 - `actions/upload-artifact@v4` does not find the file either, because it's another `job` or `step`; so this must happen while that shell is still connected. Porting `actions/upload-artifact` to shell script might solve the timing problem.

 ---

I've seen there are more files in `base/cvd/bazel-bin/cuttlefish/package` to be packaged.

In the current state, this is quite reproducible (despite not in a GitHub runner) and easy to extend.

 ---

The name of the ARM image, which I cannot rest would be: `arm64v8/rockylinux:9.3`


syslogic:
Meanwhile I have translated some more shell scripts, merely emulating the Debian behavior.
That's also why I'ce kept `rpm-builder-matrix.yml`, since there are 4 build-stages now again.
Also reorganized the directory and added `README-md`. It might not yet be perfect, as I don't know what to package or what to test, but it's now almost as testable as the Debian version.

syslogic:
Two architectures within one `Dockerfile` does not work, just alike missing local files in any target-stage does not work. Therefore I've created a new branch, where I've split apart the `Dockerfile`, in order to support x86 and ARM, as well as not having to comment out stages, to prevent the GitHub runner from failing. It is still unclear, if `actions/upload-artifact` can find the RPM packages from the bind-mount and it takes ages to build, especially without `actions/cache` to/from bind-mount (likely per architecture).

PR #1000, opened 2025-03-31
Add orchestration e2e tests for debian substitution marker logic.

ser-io:


e2etests/orchestration/common/common.go
```diff
@@ -203,6 +203,11 @@ func (h *DockerHelper) ExecADBShellCommand(id, adbBin, serial string, cmd []stri
 	return h.exec(id, append([]string{adbBin, "-s", serial, "shell"}, cmd...))
 }
 
+func (h *DockerHelper) VerifySymlink(id, filename string) error {
+	_, err := h.exec(id, []string{"readlink", filename})
```
jemoreira, line 207: why not [os.Readlink](https://pkg.go.dev/os#Readlink)?

e2etests/orchestration/create_local_debian_substitution_marker/main_test.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	if diff := cmp.Diff("Running", got.CVDs[0].Status); diff != "" {
+		t.Errorf("status mismatch (-want +got):\n%s", diff)
+	}
+	dh, err := common.NewDockerHelper()
+	if err != nil {
+		t.Fatal(err)
+	}
+	dir := filepath.Join("/var/lib/cuttlefish-common/user_artifacts", uploadDir)
+	link := filepath.Join(dir, "bin/adb")
```
jemoreira, line 102: IIUC this test
- modifies a build to add a custom marker file
- creates a cvd with the modified build
- checks that the cvd creation succeeds
- checks that the files mentioned in the marker file are symlinks

My first question is: Why is it being tested in orchestration at all? There is no orchestration feature being tested here, it's all `cvd`, so it should be tested in `cvd` tests. I understand that `cvd` is part of orchestration and these are e2e tests, but the actual feature being tested is not exposed to orchestration and meant to be transparent.

Secondly, this test is introducing the expectation that if a marker file contains the file `x` the resulting CVD's host files must have symlink `y`. This is introducing a higher level of coupling between the orchestration tests and `cvd` internals than I think is necessary.

e2etests/orchestration/common/common.go
ser-io: The `readlink` command is executed within the container, in order to use `os.Readlink` we need to make the file accessible to the host by binding a mount volume when running the container.

e2etests/orchestration/create_local_debian_substitution_marker/main_test.go
ser-io: Why is it being tested in orchestration at all? I started writing this test before https://github.com/google/android-cuttlefish/commit/49601c2e47ba08ac0c1e8eaf77da583a706cd74d when substitution only happened on `cvd fetch` logic,   it was meant to be an orchestration e2e tests aiming to have a test for substitution when artifacts were uploaded to the hosts which is where we can manipulate the the tarball. After that commit, it doesn't matter if the artifacts were downloaded or uploaded, discarded previous commits handling substitution at upload time, keeping the PR with 1 commit only. I agree that now it could be seen as `cvd` test only, however I don't think there's a need to duplicate the effort rewriting the logic as a `cvd` test only.

> Secondly, this test is introducing the expectation that if a marker file contains the file x the resulting CVD's host files must have symlink y. This is introducing a higher level of coupling between the orchestration tests and cvd internals than I think is necessary.

I think the substitution logic is a critical path that should be covered by an e2e tests the sooner the better. An idea to reduce the higher level of coupling could be inspecting the host to gather filepaths dynamically early on rather than using hard-coded filepaths but I don't think we might be changing these values that frequently to render the test problematic.

Databean: It doesn't seem like a significant modification to the [`cvd` tests](https://github.com/google/android-cuttlefish/blob/7fb47855a2c937e4531044437616bd82e11e3b2e/e2etests/cvd/launch_cvd_boot_test.sh#L57) to enable this workflow. Modifying the `cvd fetch` command at the end to have a `--host_substitutions` flag that comes from an variable will allow writing some bazel configurations for different host substitutions, similar to the way that the `cvd` tests [already pass `branch`, `target`](https://github.com/google/android-cuttlefish/blob/7fb47855a2c937e4531044437616bd82e11e3b2e/e2etests/cvd/BUILD.bazel#L6).

ser-io: From the host orchestrator perspective I'd like to have coverage for the upload path, which is heavily used in on-premises deployment, where the user upload its own artifacts, all the current cvd test are based on `cvd fetch`. I'd need to create a new `cvd test` using `cvd create -config_file=<config file>` having the canonical config pointing to local directories.

jemoreira: The HO only cares that the upload + run succeeds, it doesn't care whether the binaries are substituted or not.

From the cvd tests it should be possible to do something like `cvd fetch ...; <change some symlinks to actual files with no useful content>; cvd create ...`

ser-io: I think `cvd fetch` already does mixing if the package have the `debian_marker`. I'm aiming to test the flow where the user downloads the artifacts directly from go/ab  and then execute `cvd create --config_file=<config_file>` where the host package contains the debian marker. It's true the HO doesn't have to be in the middle for this. I'll close the PR in favor of a `cvd` direct logic.

PR #1001, opened 2025-03-31
cuttlefish: aaos: add Yocto build

gurchetansingh:
This adds containerized builds for Yocto.  Yocto is a popular paradigm for creating Automotive Grade Linux (AGL) distros.

BUG=b:407036509
TEST=podman run


jmacnak:
Can you share more info about how this might be used in the future and why it should be in the Cuttlefish host packages project? Being completely unfamiliar with yocto, the current PR does not seem to show any interaction between the Cuttlefish host package and the container...

gurchetansingh:
> Can you share more info about how this might be used in the future and why it should be in the Cuttlefish host packages project? Being completely unfamiliar with yocto, the current PR does not seem to show any interaction between the Cuttlefish host package and the container...

I was looking for a public place to put Yocto builds.  There has been prior art on the subject, for example:

https://android.googlesource.com/device/google/trout/+/refs/heads/main/agl_services_build/yocto-layer/meta-google

Those have bitrotted and the plan of record is delete device/google/trout entirely.  I suspect the need for Yocto builds will still persist.  The alternative is to do something like github/google/trout-builds, but since Trout is essentially a Cuttlefish derivative, adding another directory to an existing project (CF) is less costly.  

This does expand the scope of the Cuttlefish host package, from just Debian to Yocto hosts.

gurchetansingh:
Let me close this for now.  We do want this to be public, but there may be a non-public portion needed too.

PR #1002, opened 2025-03-31
Split `//cuttlefish/host/commands/run_cvd/launch:launch` and enforce `misc-include-cleaner`

Databean:
https://github.com/google/android-cuttlefish/blob/776324af3405b78ebbc8c6b634ffb109ac3bea2f/base/cvd/cuttlefish/host/commands/run_cvd/launch/BUILD.bazel#L9

This is one of the larger bazel targets in our repository.

There's also a limitation with clang_tidy_test that makes it single-threaded, which is slow for large targets. This target takes about four minutes on our desktops.

With smaller targets, there is a risk that header includes and dependencies desync from the true dependencies of the sources inside. Using misc-include-cleaner and layering_check together validate that both headers and dependencies are correct.

Bug: b/407136274


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #1003, opened 2025-04-01
Fetch error case fix and build+target cleanup

cjreynol:
I run my locally built `cvd` from the `bazel-out` directory, which was
failing on the previous `CF_EXPECT`.  This version skips running the
substitution logic if `cvd` is not run from `cuttlefish-common` where
the host tool binaries are all co-located.

Test: # failure case
Test: <repository>/base/cvd/bazel-bin/cuttlefish/host/commands/cvd/cvd fetch --target_directory=/tmp/cvd/fetch_test
--default_build=aosp-android-latest-release/aosp_cf_x86_64_only_phone-userdebug --host_substitutions=all
Test: # succeeds without --host_substitutions flag
Test: <repository>/base/cvd/bazel-bin/cuttlefish/host/commands/cvd/cvd fetch --target_directory=/tmp/cvd/fetch_test
--default_build=aosp-android-latest-release/aosp_cf_x86_64_only_phone-userdebug

Also, swap from `aosp-main` and `aosp_cf_x86_64_phone-trunk_staging-userdebug` to new public defaults.

LATER EDIT:  The `e2etests` check is failing because the `aosp-android-latest-release` build is too old.  Those changes carved out from the previous version of this PR are here: https://github.com/cjreynol/android-cuttlefish/tree/orchestrator_swap_from_aosp_main

From my commit message:

One of the e2etests fails if they are changed.
The build from `aosp-android-latest-release` is missing the flag from aosp/3451951.

Since the swap is not pressing, we can stay on the frozen `aosp-main`
build.  Once the next quarterly release comes out, or we use the CF host
tools from the debian package in tests, we can swap the rest.


Databean:
> I run my locally built cvd from the bazel-out directory

Please switch to building `//cuttlefish/package:cvd` and running `bazel-out/k8-fastbuild/bin/cuttlefish/package/cuttlefish-common/bin/cvd`. This one should be able to discover the related files.

base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
```diff
@@ -39,18 +39,25 @@ Result<void> SubstituteWithFlag(
...
+  std::string bin_dir_parent =
       android::base::Dirname(android::base::Dirname(self_path));
-  CF_EXPECTF(android::base::EndsWith(cuttlefish_common, "cuttlefish-common"),
-             "{}", cuttlefish_common);
+  if (!android::base::EndsWith(bin_dir_parent, "cuttlefish-common")) {
+    LOG(DEBUG) << "Binary substitution not available, run `cvd fetch` from "
+                  "`cuttlefish-common` package to enable.";
+    LOG(VERBOSE) << "Directory of executed `cvd fetch` binary: "
+                 << bin_dir_parent;
+    return {};
```
Databean, line 51: This makes substitution a "warn but continue" case when there are executables that are supposed to be substituted. Can this still be an explicit error when `host_substitutions` has members?

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
cjreynol: Agreed, done.

PR #1004, opened 2025-04-01
Extract //cuttlefish/common/libs/utils:proto

Databean:
This header file requires a `@protobuf` dependency, which was blocking most of the cuttlefish code behind building the proto compiler. Extracting this one target allows a significant amount of targets, including some of the long serial lint targets, to run concurrently with building the proto compiler.

Related to https://bazel.build/docs/bazel-and-cpp

Bug: b/407763557


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #1009, opened 2025-04-02
Use ubuntu 22.04 for consistency and update debian container images

Databean:
Bug: b/407791243

### DIFF
--- a/.github/workflows/host-image-ci.yml
+++ b/.github/workflows/host-image-ci.yml
@@ -22,9 +22,9 @@ on: [push]
 
 jobs:
   build-installer-iso-job:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     container:
-      image: debian@sha256:c2cedd7f80a4dd0f9f80d3699bd433ccf3de33ab63bfa2d4c4ba870c998222d6 # debian:bookworm-20240311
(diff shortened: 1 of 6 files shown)


ser-io (approved):

.github/workflows/host-image-ci.yml
```diff
@@ -46,12 +46,12 @@ jobs:
     uses: ./.github/workflows/reusable-build.yaml
 
   build-metapackage-deb-job:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
jemoreira, line 49: nit: We could avoid repeating this using a [variable](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables. Apparently those variables can be at [repo level ans shared between workflows](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#defining-configuration-variables-for-multiple-workflows).

jemoreira (approved):

.github/workflows/host-image-ci.yml
Databean: @jemoreira does this look good?

Variable values here: https://github.com/google/android-cuttlefish/settings/variables/actions

Databean: These kept giving me workflow parsing errors. Let's revisit this later.

PR #1011, opened 2025-04-02
Use libzip in host_bugreport

Databean:
[`host_bugreport` relies on](https://github.com/google/android-cuttlefish/blob/37f1b9789418bf9631140df43d6203c5fcd3c3b0/base/cvd/cuttlefish/host/commands/host_bugreport/BUILD.bazel#L18) an imported AOSP library [`libziparchive`](https://github.com/google/android-cuttlefish/blob/37f1b9789418bf9631140df43d6203c5fcd3c3b0/base/cvd/libziparchive/BUILD.bazel#L8), which is in the way of b/407854179 since it [uses c++20 features like `std::span`](https://github.com/google/android-cuttlefish/blob/37f1b9789418bf9631140df43d6203c5fcd3c3b0/base/cvd/libziparchive/zip_archive.cc#L280). Outside of AOSP, there is `libzip` which is a C library available in google3, the bazel registry, and debian.

Using an external library here allows us to avoid maintaining this import.

Bug: b/407864925


base/cvd/cuttlefish/host/commands/host_bugreport/main.cc
```diff
@@ -31,30 +33,34 @@
...
+      zip_source_file(archive, file_path.c_str(), 0, ZIP_LENGTH_TO_END);
+  if (source == nullptr) {
+    LOG(ERROR) << "Error in creating zip source file: "
+               << zip_strerror(archive);
+    return;
+  }
+  zip_int64_t index = zip_file_add(archive, zip_path.c_str(), source, 0);
+  if (index == -1) {
+    LOG(ERROR) << "Error in adding zip source file: " << zip_strerror(archive);
+    zip_source_free(source);
```
jemoreira, line 59: should this free be outside the `if`? we should consider a RAII wrapper around this interface too and hide this raw pointer handling in a library somewhere.

base/cvd/cuttlefish/host/commands/host_bugreport/main.cc
```diff
@@ -131,20 +137,30 @@ Result<void> CvdHostBugreportMain(int argc, char** argv) {
   CHECK(config) << "Unable to find the config";
 
   auto out_path = FLAGS_output.c_str();
-  std::unique_ptr<FILE, decltype(&fclose)> out(fopen(out_path, "wbe"), &fclose);
-  ZipWriter writer(out.get());
+  int err = 0;
+  std::unique_ptr<zip_t, decltype(&zip_discard)> archive(
+      zip_open(out_path, ZIP_CREATE | ZIP_TRUNCATE, &err), &zip_discard);
+  if (archive.get() == nullptr) {
```
jemoreira, line 143: nit: I think you can simply ask `archive == nullptr`

base/cvd/cuttlefish/host/commands/host_bugreport/main.cc
```diff
@@ -131,20 +137,30 @@ Result<void> CvdHostBugreportMain(int argc, char** argv) {
...
+        "zip_open failed: " + std::string(zip_error_strerror(&error));
+    zip_error_fini(&error);
+    return CF_ERR(msg);
+  }
 
-  auto save = [&writer, config](const std::string& path) {
-    SaveFile(writer, "cuttlefish_assembly/" + path, config->AssemblyPath(path));
+  auto save = [&archive, config](const std::string& path) {
+    SaveFile(archive.get(), "cuttlefish_assembly/" + path,
+             config->AssemblyPath(path));
```
jemoreira, line 154: nit: having a lambda to simplify two function calls seems overkill

base/cvd/cuttlefish/host/commands/host_bugreport/main.cc
Databean: Yes. `zip_file_add` [consumes the `zip_source_t*` on success](https://libzip.org/documentation/zip_file_add.html).

> _NOTE_: [`zip_source_free(3)`](https://libzip.org/documentation/zip_source_free.html) should not be called on a source after it was used successfully in a `zip_file_add` or `zip_file_replace` call.

ser-io (approved):

base/cvd/cuttlefish/host/commands/host_bugreport/main.cc
Databean: Changed to `!archive`.

base/cvd/cuttlefish/host/commands/host_bugreport/main.cc
Databean: Removed.

Databean:
Thanks for reviewing!

PR #1013, opened 2025-04-03
Make `flags_defaults.h` dependencies explicit

cjreynol:
Which required moving some declarations+definitions around and splitting out a couple of `host/libs/config` files into new targets, then dealing with the ripples of that. 
 `flags_defaults.h` was depending on declarations from `cuttlefish_config.h` and `flags.cc`, both in targets that depended on the `libassemble_cvd` target containing `flags_defaults.h`.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/BUILD.bazel
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/BUILD.bazel
@@ -91,6 +91,7 @@ cc_library(
         "//cuttlefish/host/libs/avb",
         "//cuttlefish/host/libs/command_util",
         "//cuttlefish/host/libs/config",
+        "//cuttlefish/host/libs/config:config_utils",
         "//cuttlefish/host/libs/config/adb",
         "//cuttlefish/host/libs/config/fastboot",
         "//cuttlefish/host/libs/image_aggregator",
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/BUILD.bazel
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/BUILD.bazel
@@ -96,6 +96,7 @@ cc_library(
         "//cuttlefish/host/libs/avb",
         "//cuttlefish/host/libs/command_util",
         "//cuttlefish/host/libs/config",
+        "//cuttlefish/host/libs/config:config_constants",
         "//cuttlefish/host/libs/config:config_utils",
         "//cuttlefish/host/libs/config/adb",
         "//cuttlefish/host/libs/config/fastboot",
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/BUILD.bazel
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/BUILD.bazel
@@ -12,6 +12,11 @@ cc_library(
     ],
     copts = COPTS + [ "-Werror=sign-compare" ],
     strip_include_prefix = "//cuttlefish",
+    deps = [
+        "//cuttlefish/common/libs/utils:environment",
+        "//cuttlefish/host/libs/config:config_constants",
+        "//cuttlefish/host/libs/config:config_utils",
+    ],
 )
 
 clang_tidy_test(
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
@@ -60,6 +60,7 @@
 #include "host/commands/assemble_cvd/misc_info.h"
 #include "host/commands/assemble_cvd/network_flags.h"
 #include "host/commands/assemble_cvd/touchpad.h"
+#include "host/libs/config/config_constants.h"
 #include "host/libs/config/config_flag.h"
 #include "host/libs/config/cuttlefish_config.h"
 #include "host/libs/config/display.h"
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
@@ -74,16 +74,6 @@
 #include "host/libs/vm_manager/qemu_manager.h"
 #include "host/libs/vm_manager/vm_manager.h"
 
-using cuttlefish::DefaultHostArtifactsPath;
-using cuttlefish::HostBinaryPath;
-using cuttlefish::TempDir;
-using cuttlefish::vm_manager::CrosvmManager;
-using google::FlagSettingMode::SET_FLAGS_DEFAULT;
-using google::FlagSettingMode::SET_FLAGS_VALUE;
-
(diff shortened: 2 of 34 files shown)


base/cvd/cuttlefish/host/libs/config/config_constants.h
```diff
@@ -60,4 +60,28 @@ inline constexpr char kApName[] = "crosvm_openwrt";
...
+inline constexpr char kGpuModeNone[] = "none";
+
+inline constexpr char kGpuVhostUserModeAuto[] = "auto";
+inline constexpr char kGpuVhostUserModeOn[] = "on";
+inline constexpr char kGpuVhostUserModeOff[] = "off";
+
+inline constexpr char kHwComposerAuto[] = "auto";
+inline constexpr char kHwComposerDrm[] = "drm_hwcomposer";
+inline constexpr char kHwComposerRanchu[] = "ranchu";
+inline constexpr char kHwComposerNone[] = "none";
```
Databean, line 86: Can these be `std::string_view` instances?

base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
```diff
@@ -74,16 +75,6 @@
 #include "host/libs/vm_manager/qemu_manager.h"
 #include "host/libs/vm_manager/vm_manager.h"
 
-using cuttlefish::DefaultHostArtifactsPath;
-using cuttlefish::HostBinaryPath;
-using cuttlefish::TempDir;
-using cuttlefish::vm_manager::CrosvmManager;
-using google::FlagSettingMode::SET_FLAGS_DEFAULT;
-using google::FlagSettingMode::SET_FLAGS_VALUE;
```
Databean, line 82: Is this a load-bearing change or was it a stylistic preference?

base/cvd/cuttlefish/host/commands/assemble_cvd/BUILD.bazel
```diff
@@ -12,6 +12,11 @@ cc_library(
     ],
     copts = COPTS + [ "-Werror=sign-compare" ],
     strip_include_prefix = "//cuttlefish",
+    deps = [
+        "//cuttlefish/common/libs/utils:environment",
+        "//cuttlefish/host/libs/config:config_constants",
+        "//cuttlefish/host/libs/config:config_utils",
+    ],
```
Databean, line 19: Can this change include fixes to some workarounds introduced because of incomplete dependencies?

https://github.com/google/android-cuttlefish/blob/6b1d18e8786e77610ec429f0b76fa2fbc906b465/base/cvd/cuttlefish/host/commands/cvd/cli/parser/instance/cf_vm_configs.cpp#L30

https://github.com/google/android-cuttlefish/blob/6b1d18e8786e77610ec429f0b76fa2fbc906b465/base/cvd/cuttlefish/host/commands/cvd/cli/parser/instance/cf_security_configs.cpp#L24

Databean (approved):

base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
cjreynol: Not load-bearing at all.

- the many uses of `DefaultHostArtifactsPath` are within the `namespace cuttlefish` block
- the single use of `HostBinaryPath` is also within that block
- there is a single use of `TempDir` outside of the `namespace cuttlefish` block, but that did not feel worth the `using`
- `CrosvmManager` and `SET_FLAGS_VALUE` are unused

Which leaves `SET_FLAGS_DEFAULT`, which was a personal choice based on:

- about a third of the declarations in the file were already using the fully qualified name
- almost all of the usages in other files (except 2 of the 19 outside of `flags.cc`) were using the fully qualified name, so a slight majority by pure happenstance
- the rest of the `using` statements would be gone
- and the most personal, I tend to prefer fully qualified names to distinguish local (to the current namespace) constants/functions from library ones

Interesting and slightly related, the `gflags` library exports the names under both the `google` and `gflags` namespaces: https://cs.android.com/android/platform/superproject/main/+/main:external/gflags/android/gflags/gflags_google.h;l=41;drc=b68ef53bfb513d5495d5fc421dd32f1411028ba3

I was looking because I was curious why we are not qualifying `SetCommandLineOptionWithMode`, since the `FlagSettingMode` errors without the `google::` or `gflags::` qualification.  It appears any of the `gflags::` functions can be used unqualified, but that does not hold for datatypes/enums.  Weird?

Since we are already cleaning up, let's pick one and be consistent across these usages.  Slight preference to qualified (though I would rather `gflags` than `google`) since we have the choice.  If we go with unqualified:

- the `using` statements should all go inside our `namespace cuttlefish` block
- we add statements for all of the `gflags` enums/types/functions

At this point I have thought about this enough we are going to make a sweeping change either way and unify it, so inertia is not much of a factor 😅

base/cvd/cuttlefish/host/commands/assemble_cvd/BUILD.bazel
cjreynol: Done.

base/cvd/cuttlefish/host/libs/config/config_constants.h
cjreynol: The `kVhostUser*` and `kGpuVhostUser*` values can, but the rest get converted/returned as `std::string` values and would need to be refactored.

I think we would have to change the return type of something like [`ConfigureGpuSettings`](https://github.com/google/android-cuttlefish/blob/6b1d18e8786e77610ec429f0b76fa2fbc906b465/base/cvd/cuttlefish/host/commands/assemble_cvd/graphics_flags.cc#L587) to be `constexpr` if we wanted to return the `inline constexpr std::string_view` versions of those `kGpuMode` values?  I do not think we would be able to do that with the `Result` type wrapper.

I guess we could convert to `std::string` before the `return`, but that seems like creating work for ourselves.

I am inclined to leave them for consistency, but happy to change the blocks that can be swapped without changes.

base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
Databean: For reference, Google has these positions on `using` statements:

- https://abseil.io/tips/119
- [go/totw/151](https://goto.google.com/totw/151)

> I was looking because I was curious why we are not qualifying `SetCommandLineOptionWithMode`, since the `FlagSettingMode` errors without the `google::` or `gflags::` qualification. It appears any of the `gflags::` functions can be used unqualified, but that does not hold for datatypes/enums. Weird?

I think this is [Argument-Dependent Lookup](https://en.cppreference.com/w/cpp/language/adl): when resolving functions, it includes the namespaces where the types of the function call argument instances are defined.

cjreynol: As much as I really want to go in and clean things up, for the sake of not delaying future tasks I am just going to call this "good enough" since it is at least consistently done.

PR #1014, opened 2025-04-03
Define toolchain options to use a host clang executable

Databean:
These toolchains are registered with lower priority than the downloaded prebuilt LLVM/clang, but can be selected using the `--extra-toolchains` flag, e.g.

```
bazel build //cuttlefish/package:cvd \
  --extra_toolchains=//toolchain:linux_local_clang_19
```

Note that if the clang executable for the requested version is missing, the toolchain will silently be excluded from toolchain resolution, so requesting `//toolchain:linux_local_clang_11` on a system without /usr/bin/clang-11 will fall through to the next supported toolchain following the normal resolution order.

See https://bazel.build/extending/toolchains#toolchain-resolution

Bug: b/407854179

### DIFF
--- a/base/cvd/MODULE.bazel
+++ b/base/cvd/MODULE.bazel
@@ -43,7 +43,6 @@ git_override(
     remote = "https://github.com/mikael-s-persson/bazel-compile-commands-extractor",
 )
 
-# Configure and register the toolchain.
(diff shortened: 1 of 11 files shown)


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #1016, opened 2025-04-03
Fix wifi and vhost_device_vsock invocation flows

Databean:
Wifi was broken because a missing preprocessor declaration caused the host wifi flow to not run.

vhost_device_vsock was broken because a missing preprocessor declaration caused an assertion intended for the guest to run on the host.

Bug: b/408245136
Test: /usr/bin/bazel run //cuttlefish/package:cvd -- create --host_substitutions=all --vhost_user_vsock=true
Test: /usr/bin/bazel run //cuttlefish/package:cvd -- create --host_substitutions=all && check wifi

### DIFF
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
@@ -1185,11 +1185,8 @@ Result<CuttlefishConfig> InitializeCuttlefishConfiguration(
   tmp_config_obj.set_enable_metrics(FLAGS_report_anonymous_usage_stats);
   // TODO(moelsherif): Handle this flag (set_metrics_binary) in the future
 
-#ifdef ENFORCE_MAC80211_HWSIM
+  // TODO: schuffelen - make this a device-specific android-info.txt setting
   tmp_config_obj.set_virtio_mac80211_hwsim(true);
-#else
-  tmp_config_obj.set_virtio_mac80211_hwsim(false);
-#endif
 
   if ((FLAGS_ap_rootfs_image.empty()) != (FLAGS_ap_kernel_image.empty())) {
     LOG(FATAL) << "Either both ap_rootfs_image and ap_kernel_image should be "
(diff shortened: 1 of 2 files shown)


jemoreira (approved):

cjreynol (approved):

Databean:
Thanks for reviewing!

base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
```diff
@@ -1184,11 +1184,7 @@ Result<CuttlefishConfig> InitializeCuttlefishConfiguration(
   tmp_config_obj.set_enable_metrics(FLAGS_report_anonymous_usage_stats);
   // TODO(moelsherif): Handle this flag (set_metrics_binary) in the future
 
-#ifdef ENFORCE_MAC80211_HWSIM
   tmp_config_obj.set_virtio_mac80211_hwsim(true);
-#else
-  tmp_config_obj.set_virtio_mac80211_hwsim(false);
-#endif
```
0405ysj, line 1191: Isn't this about losing ability to run Cuttlefish with legacy wifi option? I think it's still needed. Have you seen b/381980770?

0405ysj (changes requested):

base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
Databean: This won't work as written since with our debian package release there will only be a single version of the host executables, rather than a version that is compiled for each device target. Right now there's nothing setting this preprocessor constant in our build system. Until we develop a new strategy for this, "always enable mac80211 wifi" breaks less targets than "never enable mac80211 wifi".

0405ysj: Understood. Could you add TODO comment here later?

0405ysj (approved):

base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
Databean: Added now.

PR #1017, opened 2025-04-04
Libwebrtc and dependencies

jemoreira:
Imported dependencies:
- rnnoise
- libvpx (linked with host shared lib)
- libopus (linked with host shared lib)
- libpffft
- libaom (linked with host shared lib)
- libsrtp2


base/cvd/BUILD.libwebsockets.bazel
```diff
@@ -199,6 +199,6 @@ cc_library(
         "plugins",
     ],
     deps = [
-        "//external/libwebsockets:libwebsockets_config",
+        "@//build_external/libwebsockets:libwebsockets_config",
```
cjreynol, line 202: https://github.com/google/android-cuttlefish/pull/1005

@jmacnak Do the `@` symbols fail on the import?

cjreynol (approved):

ser-io (approved):

base/cvd/BUILD.libwebsockets.bazel
Databean: This file won't be used in the imported version. Internally we rely on downstream versions of these projects stored elsewhere in the source tree, while in our bazel build we have to supply some `BUILD.*.bazel` files to tell bazel how to build some of the third-party libraries we use.

base/cvd/BUILD.libwebrtc.bazel
```diff
@@ -0,0 +1,3310 @@
+package(
+    default_visibility = [
+        "@//:android_cuttlefish",
+    ],
+)
+
+cc_library(
+    name = "libwebrtc",
+    srcs = [
+        "api/adaptation/resource.cc",
```
Databean, line 10: Was this file automatically generated in some way? The combination of huge file lists and `select` blocks seems difficult to maintain.

base/cvd/build_external/libwebrtc/BUILD.bazel
```diff
@@ -0,0 +1,60 @@
...
+    ],
+)
+
+# These headers are needed because the libwebrtc code includes certain
+# header from these paths instead of the paths they're actually available at.
+cc_library(
+    name = "third_party",
+    srcs = [
+        "third_party/rnnoise/src/rnn_vad_weights.cc",
+    ],
```
Databean, line 24: `base/cvd/build_external/libwebrtc/third_party/crc32/src/src` includes a lot of files that aren't mentioned in this `BUILD.bazel` file, are they being used?

Databean (approved):

base/cvd/BUILD.libwebrtc.bazel
jemoreira: Yes, but not in a way that can be reproduced later. I basically copied the Android.bp file and modified it to become a python script that generated this file. The Android.bp file doesn't include headers, so it was easier to merge hundreds (thousands?) of targets into one and add headers to it than to try to find which target each header goes to.

That Android.bp file is generated from a script itself, I plan to use that script to generate build files for more recent versions of libwebrtc (to more closely follow the internal version which is basically the main branch), but I chose to start with the same version as in aosp because that's the one we know already works.

base/cvd/build_external/libwebrtc/BUILD.bazel
jemoreira: No, the @crc32c dependency should take care of that. I'll double check and delete as needed.

jemoreira:
This PR wasn't ready for review yet (build is failing). Sorry about the noise, I'll post a comment when it's ready.

base/cvd/BUILD.libwebrtc.bazel
Databean: If you're interested in running that script at build time, I found it was possible to write `Android.bp` files at external repository loading time [using starlark rules](https://github.com/google/android-cuttlefish/blob/793292a027951c7d02cc4bd44b20fd5dca44dac7/base/cvd/toolchain/file_detector.bzl) but I'm not sure if it's possible to invoke `python`. Maybe with [`repository_ctx.execute`](https://bazel.build/rules/lib/builtins/repository_ctx#execute).

base/cvd/build_external/libwebrtc/BUILD.bazel
jemoreira: Done

base/cvd/BUILD.libwebrtc.bazel
jemoreira: Parsing .gn files is not easy. The semi-automated process I know of actually builds with `gn` but generates a json file describing all the targets actually being built for the given configuration (arch, os, etc), then parsed those json files to generate the corresponding build file. Doing something like that at external repository loading time would be possible but inconvenient for webrtc because checking out the code requires around 24GB of disk space, that need to be downloaded first.

PR #1020, opened 2025-04-04
RPM Support & Workflow

syslogic:
Well, this is still quite the same, except now simulatanously running x86 and ARM builder images.<br/>
It depends, if the following `actions/upload-artifact` job can still find the files in the bind-mount.

Since it takes ages to build without uploading/downloading & mounting the build-cache, I'm still uncertain, if integeration and orchestration images can also be built on GitHub, as they need these local files restored before attempting to build the `Dockerfile`. The setup is in any case better than with the last attempt, which failed for two reasons.


syslogic:
Except some minor details, alike `hashFiles('/base/cvd/**.bazel')` not returning anything, Docker multi-platform manifest not being merged, RPM packages not being signed, or the ARM build for `cuttlefish-base` failing ...this now builds for `x86_64` and `aarch64`, caches the Bazel build and uploads the RPM package artifacts.

syslogic:
Meanwhile I can upload and download the Bazel build cache from/into the builder container, because function `hashFiles()` only works, when running `actions/checkout` before, which one unexpectedly has to do twice, due to the use of a reusable component. Also `${{ RUNNER.TEMP }}` is only available in the specific context of a `job`. Not building as `root` anymore, but as `runner:docker`, since I needed to immitate the GitHub runner's UID & GID. Still tryin to upload and restore the build-cache for `x86_64`, which is currently at `[7,544 / 11,282]`, past the mark where it previously bugged out (likely due to another cause).

syslogic:
Upon further inspection, it turned out why the build suddenly fails with `libzip` ...

```
WARNING: For repository 'zlib', the root module requires module version zlib@1.3.1.bcr.3,
but got zlib@1.3.1.bcr.4 in the resolved dependency graph.

Please update the version in your MODULE.bazel or set --check_direct_dependencies=off
```

syslogic:
This is now #1034 as a single commit.

PR #1021, opened 2025-04-04
Restore gpu_mode flag and propagate gpu_mode presence from cvd canonical configs

Databean:
This uses the "unset" keyword to let the layer below decide what default to apply, which can be influenced by legacy config.json values. This is load-bearing for some tests in CI.

Bug: b/408472915


ser-io (approved):

Databean:
Thanks for reviewing!

PR #1025, opened 2025-04-07
Import sensors_simulator command and its dependencies

dimorinny:
Bug: 402297215
Test: sensors simulator works with goldfish HAL
         bazel run //cuttlefish/package:cvd -- create --host_substitutions=all --vhost_user_vsock=true


base/cvd/cuttlefish/common/libs/sensors/sensors.h
ser-io, line 1: Existing code in AOSP was moved in the `staging` directory to keep logs history. Then we move out the relevant files from that directory rather than creating new files, e.g: https://github.com/google/android-cuttlefish/commit/82ce1df165eef90e7a37c59cf5c4a00f22504e71

In this case we need to mv https://github.com/google/android-cuttlefish/blob/main/staging/common/libs/sensors/sensors.h out of staging.

ser-io (changes requested):

base/cvd/cuttlefish/common/libs/sensors/sensors.h
dimorinny: Thank you, Sergio! I've updated the change - `sensors.h` and `sensors_simulator.h/cpp` are now moved instead of newly created from git's perspective.

ser-io (approved):

Databean:
CI was broken due to b/409672836

PR #1027, opened 2025-04-08
Allow separate substitution of `cvd_internal_start` and its subtools, fix `cvd start --help` exception failure

cjreynol:
Along with some error message improvements and other small cleanups related to the above.

Copied from the major commit messages:

- [Remove same directory requirement for subtools](https://github.com/google/android-cuttlefish/commit/12a7a8b038636e87af0045bec41ddcfa49b39ba7) 

Previously, `assemble_cvd` and `run_cvd` must be run from the same
directory as `cvd_internal_start`.  With our current substitution
strategy, this meant that substituting `cvd_internal_start` also forced
substitution of the others.

This version CAN (via environment variable or CWD) substitute only
`cvd_internal_start`.  `cvd create` and `cvd start` always set
`ANDROID_HOST_OUT`, for instance.

- [Fix cvd start --help in some cases](https://github.com/google/android-cuttlefish/commit/2a239d399248a446c62b94846a70f1b9a1e32d16) 

It expected `ANDROID_HOST_OUT` to be set, and looked it up in a way that
would throw an exception if it did not exist.  This version,
specifically for `start` help invocations, allows falling back to
different locations to find the host tools for the user's convenience.

That value needs to be propogated through to the subtool `Command`
invocations because of requirements on `etc/` files for outputting
flags.

Test: `cvd start --help` in a non-HOME directory with fetched host tools
and `ANDROID_HOST_OUT` unset.

### DIFF
--- a/base/cvd/cuttlefish/common/libs/utils/subprocess.cpp
+++ b/base/cvd/cuttlefish/common/libs/utils/subprocess.cpp
@@ -39,6 +39,8 @@
 #include <string>
 #include <thread>
 #include <type_traits>
+#include <unordered_map>
+#include <unordered_set>
 #include <utility>
 #include <vector>
 
@@ -46,6 +48,7 @@
 #include <android-base/strings.h>
 
 #include "common/libs/fs/shared_buf.h"
+#include "common/libs/utils/contains.h"
 #include "common/libs/utils/files.h"
 
 extern char** environ;
@@ -483,6 +486,18 @@ Subprocess Command::Start(SubprocessOptions options) const {
 }
 
 std::ostream& operator<<(std::ostream& out, const Command& command) {
+  std::unordered_set<std::string> to_show{"HOME",
+                                          "ANDROID_HOST_OUT",
+                                          "ANDROID_SOONG_HOST_OUT",
+                                          "ANDROID_PRODUCT_OUT",
+                                          "CUTTLEFISH_CONFIG_FILE",
+                                          "CUTTLEFISH_INSTANCE"};
+  for (const std::string& env_var : command.env_) {
+    std::vector<std::string> env_split = android::base::Split(env_var, "=");
+    if (!env_split.empty() && Contains(to_show, env_split.front())) {
+      out << env_var << " ";
+    }
+  }
   return out << android::base::Join(command.command_, " ");
 }
 
--- a/base/cvd/cuttlefish/host/commands/start/main.cc
+++ b/base/cvd/cuttlefish/host/commands/start/main.cc
@@ -15,6 +15,7 @@
 
 #include <fstream>
 #include <iostream>
+#include <optional>
 #include <sstream>
 #include <unordered_set>
 
@@ -26,6 +27,7 @@
 
 #include "common/libs/fs/shared_buf.h"
 #include "common/libs/fs/shared_fd.h"
+#include "common/libs/utils/environment.h"
 #include "common/libs/utils/files.h"
 #include "common/libs/utils/flag_parser.h"
 #include "common/libs/utils/subprocess.h"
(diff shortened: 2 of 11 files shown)


base/cvd/cuttlefish/common/libs/utils/subprocess.cpp
```diff
@@ -483,6 +484,18 @@ Subprocess Command::Start(SubprocessOptions options) const {
 }
 
 std::ostream& operator<<(std::ostream& out, const Command& command) {
+  std::vector<std::string> to_show{"HOME",
```
Databean, line 487: nit: `std::set`

base/cvd/cuttlefish/host/commands/start/flag_forwarder.cc
```diff
@@ -245,6 +246,13 @@ FlagForwarder::FlagForwarder(std::set<std::string> subprocesses,
   for (const auto& subprocess : subprocesses_) {
     Command cmd(subprocess);
     cmd.AddParameter("--helpxml");
+    //  set by `cvd` in some cases when it executes its internal binaries
+    //  assemble_cvd needs to find the host tools directory to read etc/ for
+    //  flag information
+    std::optional<std::string> host_out = StringFromEnv("ANDROID_HOST_OUT");
+    if (host_out) {
+      cmd.AddEnvironmentVariable("ANDROID_HOST_OUT", *host_out);
+    }
```
Databean, line 255: Is there another place that is unsetting environment variables? By default all environment variables are inherited from a parent process to a child process.

base/cvd/cuttlefish/host/commands/start/main.cc
```diff
@@ -70,12 +72,38 @@ namespace {
...
+    return *opt_path + "/bin";
+  }
+  opt_path = StringFromEnv("HOME");
+  if (opt_path && IsValidAndroidHostOutPath(*opt_path)) {
+    return *opt_path + "/bin";
+  }
+  std::string path = CurrentDirectory();
+  if (IsValidAndroidHostOutPath(path)) {
+    return path + "/bin";
+  }
```
Databean, line 89: This assumes the user is running `bin/launch_cvd` or `bin/cvd_internal_start` on the terminal?

base/cvd/cuttlefish/host/libs/config/config_utils.cpp
```diff
@@ -154,6 +155,17 @@ std::string DefaultGuestImagePath(const std::string& file_name) {
          file_name;
 }
 
+std::string DefaultEnvironmentPath(const char* environment_key,
+                                   const char* default_value,
+                                   const char* subpath) {
```
Databean, line 160: Can these be `const std::string&` arguments?

Databean (approved):

base/cvd/cuttlefish/common/libs/utils/subprocess.cpp
cjreynol: Agreed, though I went with `std::unordered_set` since we are not iterating through this collection.

base/cvd/cuttlefish/host/commands/start/main.cc
cjreynol: TLDR; Yes, but IMO that is an acceptable assumption.

Given how `IsValidAndroidHostOut`'s implementation is checking for `bin/cvd_internal_start` before returning a value from any of the new cases, if the assumption does not hold it should fall back to the previous logic (potentially with a new warning).

There are cases where they could execute `<non-bin>/cvd_internal_start`, their `$ANDROID_HOST_OUT`/`$HOME`/`$PWD` points to a directory with a `bin/cvd_internal_start` to satisfy `IsValidAndroidHostOut`, and they get a version of the executable they do not expect.  That could be confusing, but is also consistent with other parts of the codebase.  For example:
- [`cvd create`](https://github.com/google/android-cuttlefish/blob/e75ee7986498bc5f804f309dc9879b90511becfe/base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp#L97), 
- [`AndroidHostPath`](https://github.com/google/android-cuttlefish/blob/e75ee7986498bc5f804f309dc9879b90511becfe/base/cvd/cuttlefish/host/commands/cvd/utils/common.cpp#L50) called in `cvd start --help` and a half-dozen other sites
- [`DefaultHostArtifactsPath`](https://github.com/google/android-cuttlefish/blob/e75ee7986498bc5f804f309dc9879b90511becfe/base/cvd/cuttlefish/host/libs/config/config_utils.cpp#L108) functioning without any kind of validity check, called ~50 times


This should be a non-issue for anybody calling `cvd`.  `cvd start --help` already expected `ANDROID_HOST_OUT` to be set (which was the cause for the exception and it currently being broken) and sets it using `AndroidHostPath` during its execution anyways.

In the future I DO think we should move to a single implementation everywhere that also includes some kind of validation for the directory.  Probably two as an intermediate step, one using `StringFromEnv` for the non-`cvd` binaries and `AndroidHostPath` using `cvd_common::Envs`.  As I was digging into this I noticed that `run_cvd` is using its own versions ([eg [boot_state_machine`](https://github.com/cjreynol/android-cuttlefish/blob/9e632fb73f828a3f6df963162b8cb633f7512b9b/base/cvd/cuttlefish/host/commands/run_cvd/boot_state_machine.cc#L347)) of the previous `SubtoolPath` implementation[ 
 We will likely have unintended binary substitution there and can revisit this making this logic more consistent there.

cjreynol: Looks like `run_cvd` only uses `GetExecutablePath` for calling `adb` and `crosvm`.  When substituting those binaries will not be found and the fallback `HostBinaryPath` call will hit the typical sequence of environment variables, falling back to the current working directory.

base/cvd/cuttlefish/host/commands/start/flag_forwarder.cc
cjreynol: No, they are not being unset.  I think this change was an incorrect reading of [the doc page](https://linux.die.net/man/3/execvpe) without testing if it was actually needed.  Thanks for the catch!

base/cvd/cuttlefish/host/libs/config/config_utils.cpp
cjreynol: Done.

PR #1029, opened 2025-04-10
Bazel changes necessary to build on Arch Linux

Databean:


Databean:
Thanks for reviewing!

PR #1032, opened 2025-04-11
Disable android10-gsi boot test for now.

Databean:
The most recent android10-gsi build has been archived. Disable the test while we figure out if we can get another build.

Bug: b/409672836


Databean:
Thanks for reviewing!

PR #1033, opened 2025-04-12
Release 1.4.0

Databean:
Update changelog for PRs made since 1.3.0, including changelog updates in https://github.com/google/android-cuttlefish/pull/1026

### DIFF
--- a/base/debian/changelog
+++ b/base/debian/changelog
@@ -1,8 +1,28 @@
-cuttlefish-common (1.4.0) UNRELEASED; urgency=medium
+cuttlefish-common (1.4.0) stable; urgency=medium
 
-  * 
+  * Disable android10-gsi boot test for now. - https://github.com/google/android-cuttlefish/pull/1032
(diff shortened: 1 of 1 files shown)


ser-io (approved):

Databean:
Thanks for reviewing!

base/debian/changelog
```diff
@@ -4,6 +4,13 @@ cuttlefish-common (1.4.0) UNRELEASED; urgency=medium
 
  -- Chad Reynolds <chadreynolds@google.com>  Fri, 04 Apr 2025 13:05:10 -0700
 
+cuttlefish-common (1.3.1) stable; urgency=medium
```
ser-io, line 20: should this be only part of the v1.3.0 branch? I'm afraid that if a user checks out the main branch at this specific commit 985be6, it won't reproduce the right v1.3.1  package.

Databean: I've squashed the commits together so there is no longer a commit with 1.3.1 at the top and a mismatched repository state.

PR #1040, opened 2025-04-15
Add -fPIC to fix mesa_util compilation

cjreynol:
Mostly copy+paste from chat, for easier future reference:

I was testing out building with local Clang 19 and get a failure that does not appear when using the fixed version.  From the "build and package lavapipe" commit.

Building with the specific toolchain:
`bazel build //cuttlefish/package:cvd --extratoolchains=//toolchain:linux_local_clang_19`

The error:
`/usr/bin/ld: bazel-out/k8-fastbuild/bin/external/+_repo_rules2+mesa/libmesa_util.a(u_call_once.o): relocation R_X86_64_TPOFF32 against 'call_once_context' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: failed to set dynamic section sizes: bad value`

I do see some references to using the flag in external/mesa3d.

Reference from responses:
- would have expected all shared libraries to be compiled with that flag- https://github.com/google/android-cuttlefish/blob/eb2bef6af7b6e506e2bbf761e8daa604b3c2c2db/base/cvd/toolchain/cc_toolchain_config.bzl#L62

### DIFF
--- a/base/cvd/MODULE.bazel.lock
+++ b/base/cvd/MODULE.bazel.lock
@@ -248,8 +248,8 @@
     "https://bcr.bazel.build/modules/rules_java/7.3.2/MODULE.bazel": "50dece891cfdf1741ea230d001aa9c14398062f2b7c066470accace78e412bc2",
     "https://bcr.bazel.build/modules/rules_java/7.4.0/MODULE.bazel": "a592852f8a3dd539e82ee6542013bf2cadfc4c6946be8941e189d224500a8934",
     "https://bcr.bazel.build/modules/rules_java/7.6.1/MODULE.bazel": "2f14b7e8a1aa2f67ae92bc69d1ec0fa8d9f827c4e17ff5e5f02e91caa3b2d0fe",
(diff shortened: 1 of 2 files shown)


base/cvd/BUILD.mesa.bazel
```diff
@@ -913,7 +913,7 @@ cc_library(
     ] + glob([
         "src/util/**/*.h",
     ]),
-    copts = MESA_DEFAULT_COPTS,
+    copts = MESA_DEFAULT_COPTS + ["-fPIC",],
```
jmacnak, line 916: Can we just add it to `MESA_DEFAULT_COPTS` ?

cjreynol: Actually, @Databean found that the `-fPIC` flag was only being added for C++ compilations and not for C code.  I tested adding the extra action in the toolchain and this way it works universally.  My previous fix was only solving it for the `mesa` build and not for future C libraries.

jmacnak (approved):

Databean (approved):

PR #1044, opened 2025-04-17
Support building crosvm from source out of a different git repository using `rules_rust`

Databean:
It is possible to add crosvm to the list of executables included with host substitution, but it currently requires running with `--gpu_mode=guest_swiftshader` and `--enable_sandbox=false`.

Bug: b/402274999
Test: bazel run '@crosvm_bin//:crosvm__crosvm'


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #1054, opened 2025-04-18
Use system executables where it was trivial to do so

adelva1984:


cjreynol (approved):

base/debian/control
```diff
@@ -66,11 +67,14 @@ Depends: adduser,
...
          libz3-4,
          lsb-base,
+         lz4,
+         mtools,
          net-tools,
          openssl,
          opus-tools,
          python3,
          qemu-user-static [arm64],
+         selinux-utils,
```
Databean, line 77: FYI, none of these are included now in the Oxygen container environment which ignores our debian package dependencies.

Internal-specific debug information at https://paste.googleplex.com/4988993244233728

Databean: As written here, this blocks migrating `assemble_cvd` until the dependency is present in the system.

An alternative would be to first make sure the executables like `lz4` and `mmd` are available in the environments we use, and then updating `debian_substitutions_marker` to override the Android-provided `lz4` with a symlink to `/usr/bin/lz4`. That make sit possible to evaluate the change with Android presubmit rather than with the environment releases.

adelva1984: That seems like a good idea. Some of these dependencies are big and probably have to be imported, which is probably a good use of time right now. I'll update it with that suggestion.

adelva1984: Closing this PR because we decided to move the remaining few binaries in-house and provide them in the host package.

PR #1059, opened 2025-04-19
Migrate cf_vhost_user_input

cjreynol:
From the "Build cf_host_user_input" commit:

Translated the dependencies from the `Android.bp` into a `Cargo.toml` file.
That allows updating the build manifest like a typical Rust project
without needing to modify the `BUILD.bazel`.

Copied `inherited_fd.rs` from `librustutils` target.  It was the only
usage from that single dependency not available through the Rust
crate registry.  The full `rustutils` had a more complex build with FFI
binding generation that were unnecessary for this tool.  This file has
been stable since its introduction (aosp/3260174 in `rustutils`,
aosp/3253252 in AOSP).

Bug: 402272705


base/cvd/MODULE.bazel
```diff
@@ -99,14 +99,23 @@ python.toolchain(
...
+append_squashfs_overlay_crates.from_cargo(
     name = "append_squashfs_overlay_crates",
     manifests = [
         "//cuttlefish/host/commands/append_squashfs_overlay:Cargo.toml",
     ],
 )
-use_repo(crate, "append_squashfs_overlay_crates")
+use_repo(append_squashfs_overlay_crates, "append_squashfs_overlay_crates")
+
+vhost_user_input_crates = use_extension("@rules_rust//crate_universe:extensions.bzl", "crate")
```
jemoreira, line 111: why can't `append_squashfs_overlay_crates` (in its original name `crate`) be reused for this? A comment in the file with that reason would be nice.

base/cvd/cuttlefish/host/commands/vhost_user_input/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
```
jemoreira, line 1: FYI: the original Cargo.toml file looked like this:
```
[package]
name = "vhost_user_input"
version = "0.1.0"
edition = "2021"

[dependencies]
...
```
Is the `[[bin]]` section really needed? With rust if you have a `main.rs` file it creates an executable by default, a `lib.rs` generates a library.

jemoreira (approved):

base/cvd/cuttlefish/host/commands/vhost_user_input/Cargo.toml
cjreynol: Does that mean you want me to drop the `authors`?  I did not see a deleted `Cargo.toml`, so I copied the format of the `append_squashfs_overlay` file.

cjreynol: Yes, the `[[bin]]` was needed because `cargo` errored since the `main.rs` file in the same directory apparently does not match the expected convention.

The error:

```
Caused by:
  no targets specified in the manifest
  either src/lib.rs, src/main.rs, a [lib] section, or [[bin]] section must be present
```

base/cvd/MODULE.bazel
cjreynol: I think we could put everything into the same repo, especially in this setup, but I think it might end up more complicated in the future if we:

- tried to have the same crates with different feature sets or patches applied (specifically thinking of `crosvm` here)  I am also not sure how the
- need to use `from_cargo` extension for some dependencies and `spec` for others, I do not know how those interact
- what that means for the generated helpers.  The generated repo creates `all_crate_deps` and `aliases` that are called in the `rust_binary` targets `deps`, `aliases`, and `proc_macro_deps` field.  Is that impacted by mixing everything together? 

Rather than figuring those interactions out, my plan was creating a separate repo for each target to hopefully avoid any issues.  We could look into it more in the future, but I think this is sufficient for now.

As someone not biased by writing these lines, what do you think we should leave in a comment for future readers?  To me, if I see multiple examples of a single repo being created for each Rust target I would likely follow suit unless something was overly slow or incorrect.  Given that I have not looked deeply into the above interactions, I am not sure what technical details I could leave other than "Maybe not doing this is bad, this was easier".

base/cvd/cuttlefish/host/commands/vhost_user_input/Cargo.toml
Databean: Would moving the file into `src/main.rs` relative to this directory work instead of the `[[bin]]` block?

base/cvd/MODULE.bazel
Databean: It looks like these `use_repo` statements aren't fully independent to begin with. When adding `crosvm` in https://github.com/google/android-cuttlefish/pull/1044 , I started getting crossed wires between `append_squashfs_overlay_crates` and the new `crosvm_bin` target, with errors like
```
error: no matching package named `audio_streams` found
location searched: crates.io index
required by package `crosvm v0.1.0 (https://chromium.googlesource.com/crosvm/crosvm?rev=bc84c8bd6ce078e251594aa1af9e84932c5d0d81#bc84c8bd)`
    ... which satisfies git dependency `crosvm` of package `append_squashfs_overlay v0.1.0 (/tmp/.tmpKRJxv7)`
```

base/cvd/cuttlefish/host/commands/vhost_user_input/Cargo.toml
cjreynol: I did not realize we felt so strongly against `[[bin]]` blocks or I would have gone with that option to start :)

Will update it.

base/cvd/MODULE.bazel
cjreynol: Interesting.  I am digging through the docs now and see a reference to using separate repos (just above the `bindeps` example block):  https://bazelbuild.github.io/rules_rust/crate_universe_bzlmod.html#binary-dependencies

That is a different context, but it seems like the concept was at least considered in a case by the `rules_rust` developers.  

The [current code in the PR looks like it modifies](https://github.com/google/android-cuttlefish/pull/1044/files#diff-4778b97127cad9c2af996261ac65e4a96a1ac60fbcb0f09d3f6aa252513f9d73) the `append_squashfs_overlay_crates` repo declaration, but I do not understand why that repo needs any modifications for a new target.  Shouldn't the existing repo stay the same and all the changes go into the new `crosvm_bin` repo from that PR?

cjreynol: From our chat:  Those changes were to fix the issue.  Seems like the `use_extension` calls are returning the same object or different objects that share some state.

If we can get this merged to have the binary release unblocked (possibly/probably with the workaround from the `crosvm_bin` branch), then we can do a pass afterwards and combine everything into a single `crates` repo.

base/cvd/cuttlefish/host/commands/vhost_user_input/Cargo.toml
cjreynol: Done.

base/cvd/MODULE.bazel
cjreynol: Created b/413182531

PR #1064, opened 2025-04-21
Don't include append_squash_overlay in cvd package

jemoreira:
It breaks internal builds because append_squash_overlay is not yet being built there.


ser-io (approved):

Databean:
There is already support for excluding targets from internal builds without modifying external builds: http://shortn/_BGh3GEpXE5

PR #1069, opened 2025-04-22
Update e2etests so device logs are stored as test artifacts

cjreynol:
The `outputs.zip` now needs a flag to be created, even though this behavior used to be triggered automatically when files were written to `$TEST_UNDECLARED_OUTPUTS_DIR`.

Other minor fixes noted in commits.

Bug: 412696459


Databean (approved):
Was hoping to see these logs in sponge, but having them in placer is a significant improvement over not having them at all.

PR #1071, opened 2025-04-22
Build and package mkbootfs

adelva1984:
Locally fork fs_config from cutils as it is self-contained.

Bug: 402293224


base/cvd/mkbootfs/mkbootfs.cpp
```diff
@@ -0,0 +1,457 @@
+
```
Databean, line 1: I see the [original file](https://cs.android.com/android/platform/superproject/main/+/main:system/core/mkbootfs/mkbootfs.cpp;l=1;drc=e7f1d41237f2504a0aeb8bcf62c8954b48f282c5) doesn't have a license, should we add one here?

Databean (approved):

base/cvd/mkbootfs/mkbootfs.cpp
adelva1984: Done

PR #1073, opened 2025-04-22
Build and package mkbootimg and unpack_bootimg

adelva1984:
Bug: 402292787
Bug: 402299250


base/cvd/BUILD.mkbootimg.bazel
```diff
@@ -0,0 +1,14 @@
+package(
+    default_visibility = ["@//:android_cuttlefish"],
+)
+
+cc_library(
+    name = "bootimg_header",
+    hdrs = ["include/bootimg/bootimg.h"],
+    strip_include_prefix = "include/bootimg",
+)
```
Databean, line 9: It doesn't look like anything depends on this `cc_library` yet, is that planning to come later?

Databean (approved):

base/cvd/BUILD.mkbootimg.bazel
adelva1984: Yes, it'll be used by fastboot.

adelva1984: See PR #1075
Draft so I can let CI & merge complete on this dependency in case it needs to be reverted.

ser-io (approved):

PR #1076, opened 2025-04-23
Build and package lpadd and lpmake

adelva1984:


base/cvd/fec/BUILD.bazel
```diff
@@ -5,13 +5,6 @@ package(
 cc_binary(
     name = "fec",
     srcs = [
-        "ext4_utils/ext4_sb.cpp",
-        "ext4_utils/include/ext4_utils/ext4_extents.h",
-        "ext4_utils/include/ext4_utils/ext4.h",
-        "ext4_utils/include/ext4_utils/ext4_kernel_headers.h",
-        "ext4_utils/include/ext4_utils/ext4_sb.h",
-        "ext4_utils/include/ext4_utils/jbd2.h",
-        "ext4_utils/include/ext4_utils/xattr.h",
```
Databean, line 14: Should these files be deleted now if they're no longer being used?

Databean (approved):
Note that a deficiency of github as opposed to gerrit is that it doesn't track stacks of PRs like gerrit can track stacks of commits, so this becomes an independent PR with two commits and "files changed" view that represents the combination of both commits, rather than having any explicit relationship with https://github.com/google/android-cuttlefish/pull/1073.

base/cvd/cuttlefish/package/BUILD.bazel
```diff
@@ -34,8 +34,11 @@ package_files(
         "cuttlefish-common/bin/libvk_swiftshader.so": "@swiftshader//:vk_swiftshader",
         "cuttlefish-common/bin/log_tee": "//cuttlefish/host/commands/log_tee",
         "cuttlefish-common/bin/logcat_receiver": "//cuttlefish/host/commands/logcat_receiver",
+        "cuttlefish-common/bin/lpadd": "//liblp:lpadd",
+        "cuttlefish-common/bin/lpmake": "//liblp:lpmake",
         "cuttlefish-common/bin/metrics_launcher": "//cuttlefish/host/commands/metrics_launcher",
         "cuttlefish-common/bin/metrics": "//cuttlefish/host/commands/metrics",
+        "cuttlefish-common/bin/mkbootimg.py": "@mkbootimg//:mkbootimg.py",
```
jemoreira, line 41: Adding these targets with `@` to the packages list breaks the import to the internal repository. Consider not adding this here until a solution has been implemented internally. Example: b/413097319

adelva1984:
> Note that a deficiency of github as opposed to gerrit is that it doesn't track stacks of PRs like gerrit can track stacks of commits, so this becomes an independent PR with two commits and "files changed" view that represents the combination of both commits, rather than having any explicit relationship with #1073.

Yes, I can see. My plan is to rebase this once the dependent/separate PRs are merged. I think it's preferable for CI to have tested each change individually.

base/cvd/cuttlefish/package/BUILD.bazel
adelva1984: I see, but I have 5+ CLs WIP with this problem, I think it might be better to skip them on the internal repository and backfill, is that possible?

jemoreira: sure, that's possible.

Just wondering if it made sense to do everything in this PR except adding the binary to the :cvd package as that's not going to be used until the marker file in android is updated, which would be blocked on the internal build anyways.

base/cvd/fec/BUILD.bazel
adelva1984: Whoops, done

ser-io (approved):

base/cvd/cuttlefish/package/BUILD.bazel
adelva1984: Resolving the conversation and will work on the internal dependencies next week.

PR #1081, opened 2025-04-24
HO GCE credentials fix

ser-io:


e2etests/orchestration/create_with_gce_credentials_test/main_test.go
```diff
@@ -0,0 +1,64 @@
...
+		t.Fatal(err)
+	}
+	req := &hoapi.CreateCVDRequest{EnvConfig: envConfig}
+
+	_, err := srv.CreateCVD(req, &hoclient.AccessTokenBuildAPICreds{})
+
+	if err == nil {
+		t.Fatal("gce metadata setup is fake: no device should have been created")
+	}
+	const token = "DefinitelyNotARealTokenString123!"
```
jemoreira, line 60: IIUC this expects an error produced by the client to contain the GCE access token from the host orchestrator. That's a security vulnerability as the client is likely to not be in the GCE network and should otherwise not have access to the metadata server.

e2etests/orchestration/create_with_gce_credentials_test/main_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+func TestGCECredentials(t *testing.T) {
+	srv := hoclient.NewHostOrchestratorService(baseURL)
+	envConfig := make(map[string]interface{})
+	if err := json.Unmarshal([]byte(envConfigStr), &envConfig); err != nil {
+		t.Fatal(err)
+	}
+	req := &hoapi.CreateCVDRequest{EnvConfig: envConfig}
+
+	_, err := srv.CreateCVD(req, &hoclient.AccessTokenBuildAPICreds{})
```
jemoreira, line 55: I don't think it's a good idea to have a test making calls to the build api with fake credentials. At a minimum we should have asked for the api owner's permission to do this and negotiate the bandwidth, in the worst case we could end up with banned IP addresses for repeated calls with the same invalid token.

jemoreira: This test is passing `AccessTokenBuildAPICreds` to the HO client and expects the metadata server to be asked for the api credentials and the operation to fail with the actual token text in the error message. The problem there is that the token showing up in the error message is no guarantee that the api call was made with that token. It could be a validation error (that we add in the future), or something as simple as a bug that accessed the wrong variable when printing the message. It should directly test that the api was called.

A unit test is more suitable for this than an e2e one. In a unit test you could start the orchestrator server with a fake build api client and the fake metadata server, then make the call with the host orchestrator client and check that the fake build api received the correct call with the expected token.

e2etests/orchestration/create_with_gce_credentials_test/main_test.go
ser-io: Good point, the test is misleading. Rephrased the test as `cvd` does not print the access token. It prints the full json if it has an `error` field, used this approach instead.

e2etests/orchestration/create_with_gce_credentials_test/main_test.go
ser-io: Making unnecessary API requests against the Build API is a very good point. However, in this case no request is made against the Build API as `cvd` couldn't get a valid token. 

I'm looking forward to have a fake Build API in the e2e tests, this would enable tests covering final user credentials as well, not only gce service account credentials. The issue here is that we cannot control the Build API `cvd` uses yet. The HO doesn't call the Build API directly, those calls are always done through `cvd`. 

I understand it's ideal to have a fake Build API, but as we cannot use one at the moment, making sure  `cvd` attempts to load the gce credentials is the way to have some sort of coverage for this flow which is crucial in Oxygenation.

jemoreira (approved):

PR #1083, opened 2025-04-25
Migrate more files out of staging, particularly unittests left when migrating binary targets

cjreynol:
Delete files that were already migrated or unused.


staging/host/commands/config_server/main.cpp
```diff
@@ -1,55 +0,0 @@
-/*
```
Databean, line 1: We might need this for host compat, not sure.

staging/host/commands/metrics/unittest/clearcut_log_parser.sh
```diff
@@ -1,60 +0,0 @@
-#!/bin/bash
```
Databean, line 1: Might still need this for manual tests.

Databean (approved):

staging/host/commands/config_server/main.cpp
cjreynol: We have not been building this since [Delete config_server and device_config](https://android-review.git.corp.google.com/c/device/google/cuttlefish/+/3065664) a year ago.  I am not sure how the source file is being used for host compat.

staging/host/commands/metrics/unittest/clearcut_log_parser.sh
cjreynol: Editted the commit to leave the migrated script in `base/cvd/cuttlefish/host/commands/metrics/unittest/`

staging/host/commands/config_server/main.cpp
jmacnak: IIUC, our hwcomposer was the last user of the config server. At least as far back as android11-gsi, we are not dependant on the config server and instead read display properties from DRM: https://android.googlesource.com/device/generic/goldfish-opengl/+/refs/heads/android11-gsi/system/hwc2/DisplayFinder.cpp#270. I believe this should be safe to remove.

ser-io (approved):

PR #1091, opened 2025-04-30
cvd substitutions work when running from source

jemoreira:
When running the //cuttlefish/package:cvd target.

The marker file targets are ignored and cvd chooses what to replace each file with. The marker file shouldn't have specified the link targets in the first place as that requires knowledge of the host image it shouldn't possess.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
+++ b/base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
@@ -115,27 +115,19 @@ Result<std::string> GetCuttlefishCommonDir() {
   return cvd_exe.substr(0, cvd_exe.size() - std::string("/bin/cvd").size());
 }
 
-Result<void> Substitute(const std::string& target_dir, std::string link_name) {
-    static constexpr std::string_view kV1_2ModemSimulatorFiles =
-        "modem_simulator/files/";
-    static constexpr std::string_view kV1_3ModemSimulatorFiles =
-        "etc/modem_simulator/files/";
-    if (android::base::StartsWith(link_name, kV1_2ModemSimulatorFiles)) {
-      link_name = std::string(kV1_3ModemSimulatorFiles) +
-                  link_name.substr(kV1_2ModemSimulatorFiles.size());
-    }
+Result<void> Substitute(const std::string& target_dir,
+                        const std::string& link_name) {
+  static const std::string common_dir = CF_EXPECT(GetCuttlefishCommonDir());
+  std::string target = fmt::format("{}/{}", common_dir, link_name);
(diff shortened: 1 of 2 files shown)


base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
```diff
@@ -99,36 +102,53 @@ Result<void> SubstituteWithFlag(
...
-    const fetch::Symlink& symlink = config.symlinks(j);
-    std::string full_link_name =
-        fmt::format("{}/{}", target_dir, symlink.link_name());
+/**
+ * cvd needs to be run from a path ending in cuttlefish-common/bin/cvd. This
+ * function validates that and returns the path to the cuttlefish-common
+ * directory.
+ */
+Result<std::string> GetCuttlefishCommonDir() {
+  std::string cvd_exe = CF_EXPECT(GetExecutablePath(getpid()));
```
Databean, line 111: There's also [`android::base::GetExecutablePath()`](https://github.com/google/android-cuttlefish/blob/7da40940e58322998c8e9ba354dda3552d076827/base/cvd/libbase/file.cpp#L470) without the `getpid()`.

Databean, line 111: This is repeating the same IO operation to resolve the `/proc/self/exe` symlink on every iteration of the loop in `SubstituteWithMarker`, can it do it once?

base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
```diff
@@ -99,36 +102,53 @@ Result<void> SubstituteWithFlag(
...
+Result<void> SubstituteWithMarker(const std::string& target_dir,
+                                  const std::string& marker_file) {
+  std::string content;
+  CF_EXPECTF(android::base::ReadFileToString(marker_file, &content),
+             "failed to read '{}'", marker_file);
+  fetch::HostPkgMigrationConfig config;
+  CF_EXPECT(google::protobuf::TextFormat::ParseFromString(content, &config),
+            "failed parsing debian_substitution_marker file");
+  for (int j = 0; j < config.symlinks_size(); j++) {
+    CF_EXPECT(Substitute(target_dir, config.symlinks(j).link_name()));
```
Databean, line 151: If nothing is using the `symlinks.target` field, can it be removed the proto and replaced with a `reserved` directive? I don't know how this interact with textproto parsing so it might have to be delayed until the actual text files are updated.

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
jemoreira: Done

PR #1109, opened 2025-05-06
Add Rust test targets

cjreynol:
Commit messages:

[Add test target for append_squashfs_overlay](https://github.com/google/android-cuttlefish/commit/b13c6a944444c81635d5731047371ee0224ff455) 

Bug: 416096737

["Add" test target for copied inherited_fd.rs](https://github.com/google/android-cuttlefish/commit/3ff1edbc1d420e5ce6c66766954c31663a7dcb0d) 

The test is failing, leaving the target in the file to make it easier
to debug later.

Bug: 416096737


Databean:
> I tried adding a [dev dependency] stanza for the test-only depdency, but it was not being picked up. Something about the rules_rust generation, I assume.

Was that intended to be [`[dev-dependencies]`](https://doc.rust-lang.org/rust-by-example/testing/dev_dependencies.html)?

cjreynol:
I had tried the correct section name in the file, just a typo in the commit message.  Turns out there was a `normal_dev` parameter to use in the `all_crate_deps` helper to pick up that stanza.

Removed that section from the commit message and description now that they no longer apply.

jemoreira (approved):

PR #1111, opened 2025-05-07
Clean duplicated effort jobs.

ser-io:


.github/workflows/presubmit.yaml
```diff
@@ -82,34 +80,6 @@ jobs:
...
-    - name: Setup apt
-      run: apt update -y && apt upgrade -y
-    - name: Install dependencies
-      run: apt install -y git clang libcurl4-openssl-dev clang-tidy libtinfo5 xxd cmake liblzma-dev libvpx-dev libopus-dev libaom-dev autoconf libtool
-    - name: Install bazel
-      run: bash tools/buildutils/installbazel.sh
-    - name: Build cvd
-      run: cd base/cvd && bazel build --verbose_failures ...
-    - name: Test cvd
-      run: cd base/cvd && bazel test --sandbox_writable_path=$HOME --test_output=errors ...
```
Databean, line 107: Is `bazel test ...` being run in any other workflow right now?

ser-io: Good catch! Rephrased the job as run-cvd-unit-tests running only the cvd unit tests, no need to build cvd here.

Databean: IIUC in practice there won't be a huge time saving to remove the `bazel bulid '...'` step first since `bazel test '...'` will include most of that work.

Databean (approved):

.github/workflows/presubmit.yaml
Databean: I think the problem is that the `debuild` environment throws away the bazel cache, wasting the work of the first runner. [b/410052922](http://b/410052922) proposes using `pkg_deb` instead to address this issue, but that will be more work.

ser-io: I see, thanks.

PR #1133, opened 2025-05-09
e2e orchestration jobs improvements

ser-io:


.github/workflows/presubmit.yaml
```diff
@@ -101,6 +101,29 @@ jobs:
       with:
         name: debs
         path: debs.tar
+  e2e-tests-orchestration-build-image:
+    runs-on: ubuntu-24.04
+    if: ${{ always() && needs.build-debian-package.result == 'success' }}
```
jemoreira, line 106: maybe a dumb question: why is `always()` needed here? In fact, why is the whole conditional needed? IIUC `always()` says that the job should always run, but the other requirement says only if the dep succeeded, which means runs only when the dep succeeds, but that's the default behavior when having a `needs:` clause. Am I missing something?

.github/workflows/presubmit.yaml
```diff
@@ -101,6 +101,29 @@ jobs:
...
+    needs: [build-debian-package]
```
jemoreira, line 107: nit: If possible this should be moved one line up. IIUC the `if:` statement is using the dependency declared here, it's more readable to declare the dependency before referencing it.

jemoreira (approved):

PR #1135, opened 2025-05-09
Skip powerbtn_test due flakiness.

ser-io:


cjreynol:
Too much effort for one test, but if we start accumulating some amount of skipped tests due to flakiness we can make a new workflow job for them and just leave that particular check out of the "required" list for merging a PR.  Then we are still running them for potentially useful feedback, but they become non-blocking when flaking.

ser-io:
> Too much effort for one test, but if we start accumulating some amount of skipped tests due to flakiness we can make a new workflow job for them and just leave that particular check out of the "required" list for merging a PR. Then we are still running them for potentially useful feedback, but they become non-blocking when flaking.

Good point, however, for this case specifically, I think there's room for improvement, I'll discuss with the team if there's a more reliable way to assert the power button was pressed.

cjreynol (approved):

PR #1136, opened 2025-05-09
Build and package adb

adelva1984:
Heavily stripped source tree and build rules to build just the host adb binary/server and not the host adbd or target versions.

Dependency on libziparchive is removed and replaced with libzip code in incremental_utils.cpp, similar to fastboot.

The adb fastdeploy feature is removed, as it could not be disabled at compile time and depended on building jar files with Android dependencies that were bin2c'ed into the adb binary. Fortunately, fastdeploy is not enabled by default and is not required.

Bug: 402283039


base/cvd/cuttlefish/package/BUILD.bazel
```diff
@@ -5,6 +5,7 @@ package_files(
     executable = "cuttlefish-common/bin/cvd",
     package_file_to_src = {
         "cuttlefish-common/bin/acloud_translator": "//cuttlefish/host/commands/acloud_translator",
+	"cuttlefish-common/bin/adb": "//adb:adb",
```
jemoreira, line 8: nit: align

jemoreira (approved):

PR #1138, opened 2025-05-09
Move tests out of `//cuttlefish/host/commands/cvd/unittests`

Databean:
The [style guide](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) implies unit tests should be in the same directory.

Additionally, this makes it possible to run `bazel test :all` in the relevant directories while editing the source files.


cjreynol (approved):

ser-io (approved):

Databean:
Thanks for reviewing!

PR #1147, opened 2025-05-13
Use `chunk_offset_bytes` form value

ser-io:


0405ysj:
I understood the purpose of this PR(https://github.com/google/android-cuttlefish/pull/1147) as replacing https://github.com/google/android-cuttlefish/pull/1144. But I prefer to keep proceeding previous PR because,
- New HO upload API design haven't considered any compatibility with other legacy APIs including HO upload/extract API, as we discussed from the design doc before. Especially, our new HO upload API design is based on the immutable user artifact. If it's under same `UserArtifactManager`, mutable(legacy) and immutable artifacts are stored together unless introducing a new dedicated directory.
- Mixing the legacy code and the new implementation makes harder to understand the code. Though introducing `chunk_offset_bytes` is simple enough to reuse the legacy code, I think its benefit is not enough with considering below items.
  - It needs to have extra branches for whether to create new directory in the scope of `PUT /userartifacts/{dir}` or not.
  - It needs to have extra branches for which user(`cvd` or HO itself) interacts with filesystem, such as creating dir/file, untar and unzip.
  - It needs to have extra branches for whether to perform hash calculation or not.
  - AFAIR, we agreed with removing old HO user artifact APIs after completing to introduce new one.
  - I expect most of codes will be replaced after completing impl of the new one. Roughly estimates 20%-30% would be remained.

ser-io:
> I understood the purpose of this PR(#1147) as replacing #1144. But I prefer to keep proceeding previous PR because,
> 
> * New HO upload API design haven't considered any compatibility with other legacy APIs including HO upload/extract API, as we discussed from the design doc before. Especially, our new HO upload API design is based on the immutable user artifact. If it's under same `UserArtifactManager`, mutable(legacy) and immutable artifacts are stored together unless introducing a new dedicated directory.
> * Mixing the legacy code and the new implementation makes harder to understand the code. Though introducing `chunk_offset_bytes` is simple enough to reuse the legacy code, I think its benefit is not enough with considering below items.
>   
>   * It needs to have extra branches for whether to create new directory in the scope of `PUT /userartifacts/{dir}` or not.
>   * It needs to have extra branches for which user(`cvd` or HO itself) interacts with filesystem, such as creating dir/file, untar and unzip.
>   * It needs to have extra branches for whether to perform hash calculation or not.
>   * AFAIR, we agreed with removing old HO user artifact APIs after completing to introduce new one.
>   * I expect most of codes will be replaced after completing impl of the new one. Roughly estimates 20%-30% would be remained.

PR https://github.com/google/android-cuttlefish/pull/1144 adds 819 lines of code and doesn't reuse the existing code of uploading artifacts which is the main functionality here, this functionality does not change and it should be reused. If you think it needs improvements, you can add the improvements to the existing logic, so both flows benefits. I understand we can have branching regarding new logic like hash values manipulation, but this could be handle on the existing class and file.  The uploading artifacts logic is already in our codebase, it's being working for a while and it has e2e tests backing it up, it should be reused. It easier to review minimal changes of something that's already working than adding brand new 819 lines of code. If you reuse the code we should have less lines to add and less lines to review.

0405ysj:
The reasoning for writing https://github.com/google/android-cuttlefish/pull/1144 was based on the assumption below, and I thought we've already agreed with it when we discussed at the design doc and chat. Correct me if I'm wrong.
- Introducing endpoints with prefix `/v1/userartifacts`, not `/userartifacts` colliding with the existing one.
- Legacy endpoints with prefix `/userartifacts` will be completely removed after adopting `/v1/userartifacts`.
- User artifacts uploaded with newly introduced HO APIs should be immutable.
- `Owner` of `UserArtifactsManagerOpts` won't be used for the new impl.

I had analyzed benefits of reusing code before I worked on https://github.com/google/android-cuttlefish/pull/1144. Except constructors, I think most of legacy codes aren't available to reuse in the end.
- **UserArtifactsDirResolver**: Reusable for the both case.
- UserArtifactChunk: May have benefit to reuse with mixing fields.
- **UserArtifactsManager**: Have benefit to reuse.
- **UserArtifactsManagerOpts**: Have benefit to reuse.
- **UserArtifactsManagerImpl**: Have benefit to reuse.
- **NewUserArtifactsManagerImpl**: Have benefit to reuse.
- NewDir: No benefit to reuse, no caller would exist with the new design.
- ListDirs: No benefit to reuse, no caller would exist with the new design.
- **GetDirPath**: Have benefit to reuse.
- GetFilePath: No benefit to reuse, the caller only exists at testing functions. It can be replaced with filepath.Join.
- UpdateArtifact: May have benefit to reuse, but it requires branching.
- writeChunk: May have benefit to reuse, but it requires branching.
- ExtractArtifact: May have benefit to reuse, but it requires branching.
- Untar: No benefit to reuse, since owner won't be used.
- Unzip: No benefit to reuse, since owner won't be used.
- createNewUADir: No benefit to reuse, since owner won't be used.
- createUAFile: No benefit to reuse, since owner won't be used.

Also the reason why I asked for not reusing code via chat and got agreement was not to get too much implementation burden for rewriting whole code as we spent the last quarter for this.. Typically, I should write spaghetti codes for following PRs with reusing the legacy code. The legacy and new one aims for the same role, but its internal concept is so different. With the circumstance with removing legacy code someday, do we really have sufficient benefits to reuse legacy code though reusable pieces are small? Considering with introducing branches, I speculate the total number of lines can be more than https://github.com/google/android-cuttlefish/pull/1144. Of course, the code should be more complicated with this way.

> I understand we need branching regarding new logic like hash values manipulation, but this could be handle on the existing class and file avoiding the V1 suffix in the new names.

Since we're suggesting **immutable** artifacts with V1 APIs, at least any mechanism not to interfere V1 APIs by legacy APIs or vice versa is required. We don't want to allow any kind of mixed usage with interfering between legacy and new APIs, isn't it? For that specific concern, it's available to amend commit at https://github.com/google/android-cuttlefish/pull/1144 with alternative dir name such as `userartifacts` without underline(`_`).

ser-io:
> The reasoning for writing #1144 was based on the assumption below, and I thought we've already agreed with it when we discussed at the design doc and chat. Correct me if I'm wrong.
> 
> * Introducing endpoints with prefix `/v1/userartifacts`, not `/userartifacts` colliding with the existing one.
> * Legacy endpoints with prefix `/userartifacts` will be completely removed after adopting `/v1/userartifacts`.
> * User artifacts uploaded with newly introduced HO APIs should be immutable.
> * `Owner` of `UserArtifactsManagerOpts` won't be used for the new impl.
> 
> I had analyzed benefits of reusing code before I worked on #1144. Except constructors, I think most of legacy codes aren't available to reuse in the end.
> 
> * **UserArtifactsDirResolver**: Reusable for the both case.
> * UserArtifactChunk: May have benefit to reuse with mixing fields.
> * **UserArtifactsManager**: Have benefit to reuse.
> * **UserArtifactsManagerOpts**: Have benefit to reuse.
> * **UserArtifactsManagerImpl**: Have benefit to reuse.
> * **NewUserArtifactsManagerImpl**: Have benefit to reuse.
> * NewDir: No benefit to reuse, no caller would exist with the new design.
> * ListDirs: No benefit to reuse, no caller would exist with the new design.
> * **GetDirPath**: Have benefit to reuse.
> * GetFilePath: No benefit to reuse, the caller only exists at testing functions. It can be replaced with filepath.Join.
> * UpdateArtifact: May have benefit to reuse, but it requires branching.
> * writeChunk: May have benefit to reuse, but it requires branching.
> * ExtractArtifact: May have benefit to reuse, but it requires branching.
> * Untar: No benefit to reuse, since owner won't be used.
> * Unzip: No benefit to reuse, since owner won't be used.
> * createNewUADir: No benefit to reuse, since owner won't be used.
> * createUAFile: No benefit to reuse, since owner won't be used.
> 
> Also the reason why I asked for not reusing code via chat and got agreement was not to get too much implementation burden for rewriting whole code as we spent the last quarter for this.. Typically, I should write spaghetti codes for following PRs with reusing the legacy code. The legacy and new one aims for the same role, but its internal concept is so different. With the circumstance with removing legacy code someday, do we really have sufficient benefits to reuse legacy code though reusable pieces are small? Considering with introducing branches, I speculate the total number of lines can be more than #1144. Of course, the code should be more complicated with this way.
> 
> On the other hand, since we're suggesting **immutable** artifacts with new APIs, at least any mechanism not to interfere new APIs by legacy APIs or vice versa is required. We don't want to allow any kind of mixed usage with interfering between legacy and new APIs, isn't it?

Let's move this discussion to https://github.com/google/android-cuttlefish/pull/1144, this PR is a small improvement on the existing upload API.

jemoreira:
Is this backwards compatible? At first glance it seems it's replacing fixed sized chunks with byte_offset instead of extending the endpoint to support both, correct me if I'm wrong. I'd prefer not to break the existing flow with fixed sized chunks even if the new thing is better.

> Legacy endpoints with prefix /userartifacts will be completely removed after adopting /v1/userartifacts

This can take longer than you think, as we don't really know how this is being used in the wild. We may end up having to support some version of both endpoints for a while.

ser-io:
> Is this backwards compatible? At first glance it seems it's replacing fixed sized chunks with byte_offset instead of extending the endpoint to support both, correct me if I'm wrong. I'd prefer not to break the existing flow with fixed sized chunks even if the new thing is better.
> 
> > Legacy endpoints with prefix /userartifacts will be completely removed after adopting /v1/userartifacts
> 
> This can take longer than you think, as we don't really know how this is being used in the wild. We may end up having to support some version of both endpoints for a while.

Yes this change is backwards compatible. Clients using `chunk_number` , `chunk_total`,  `chunk_size_bytes` in the request's body would continue to upload artifacts using the ` /userartifacts` endpoint.

0405ysj:
> Let's move this discussion to https://github.com/google/android-cuttlefish/pull/1144, this PR is a small improvement on the existing upload API.

Then the purpose of this PR is just improving an existing API, regardless the conclusion at there, whether to keep working on https://github.com/google/android-cuttlefish/pull/1144 or make following PRs? Sounds good for moving the discussion space, since diverging discussion space makes harder to understand the flow.

ikicha:
What if cvdr is newer than HO? In this case, cvdr sends the chunk offset which old HO doesn't understand. Should it be a part of v1 API?

frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -539,36 +539,52 @@ func (h *createUpdateUserArtifactHandler) Handle(r *http.Request) (interface{},
...
+		if err != nil {
+			return nil, operator.NewBadRequestError(
+				fmt.Sprintf("Invalid chunk_offset_bytes form field value: %q", chunkOffsetBytesRaw), err)
+		}
+		chunk := UserArtifactChunk{
+			Name:             fheader.Filename,
+			File:             f,
+			ChunkOffsetBytes: chunkOffsetBytes,
+		}
+		return nil, h.m.UpdateArtifact(dir, chunk)
```
0405ysj, line 558: What's our benefit to support chunk offset at the legacy API? This change wouldn't be sufficient to land chunk offset feature. It needs to update all callers of `libhoclient` API and HO REST API. Unless, we should maintain both flow.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -142,7 +144,12 @@ func writeChunk(filename string, chunk UserArtifactChunk) error {
 		return err
 	}
 	defer f.Close()
-	if _, err := f.Seek(int64(chunk.ChunkNumber-1)*chunk.ChunkSizeBytes, 0); err != nil {
+	offset := chunk.ChunkOffsetBytes
+	if chunk.ChunkSizeBytes != 0 {
```
0405ysj, line 148: OOC, how do we treat if target upload file is empty?

frontend/src/host_orchestrator/orchestrator/userartifacts_test.go
0405ysj, line 1: Please add more tests for chunk offset feature. We're not assuming the serial case only. Host Orchestrator should work properly even when the order of delivered chunks are mixed randomly.

0405ysj (changes requested):

ser-io:
> What if cvdr is newer than HO? In this case, cvdr sends the chunk offset which old HO doesn't understand. Should it be a part of v1 API?

We aim for not breaking existing clients when new host images are rolled out keeping the API backwards compatible with previous client versions. 

If an operator rolls out a client with functionalities not supported in the server side yet, clients will fail. For example if you roll out a new version of cvdr using adb connect through web sockets and your  HO deployment doesn't support web sockets yet, `cvdr connect --connect_agent websockets` is going to fail.

frontend/src/libhoclient/httputils.go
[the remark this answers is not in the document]
ser-io: Answered the question here: https://github.com/google/android-cuttlefish/pull/1147#issuecomment-2886741139

frontend/src/host_orchestrator/orchestrator/controller.go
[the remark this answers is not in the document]
ser-io: The upload logic  is covered in most e2e tests. Unit tests are always welcome though.

frontend/src/host_orchestrator/orchestrator/controller.go
ser-io: chunk offset is a good idea the directory based upload api could stay for  a while while the hash upload based api gets to be stable, hence we can benefit from using chunk offset.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -38,11 +38,13 @@ type UserArtifactsDirResolver interface {
...
 
 type UserArtifactChunk struct {
-	Name           string
-	ChunkNumber    int
-	ChunkTotal     int
-	ChunkSizeBytes int64
-	File           io.Reader
+	Name             string
+	File             io.Reader
+	ChunkOffsetBytes int64
```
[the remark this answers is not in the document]
ser-io: If a client has a bug sending invalid chunks the final artifact won't be used due checksum mismatch, that's the ultimate test. It's true we can add checks to detect errors early on on the server side, you did this already so feel free to send the PRs adding these checks.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
0405ysj: Why don't we add a validation at the PR here? I cannot find any validation about `ChunkSizeBytes` at `controller.go`.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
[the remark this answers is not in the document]
0405ysj: I think it's better to introduce here because this PR is for introducing new form value.

frontend/src/host_orchestrator/orchestrator/controller.go
0405ysj, line 558: What about sending version information explicitly when calling `UpdateArtifact`? With that, we can avoid logic that checking fields of `UserArtifactChunk` internally.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
ser-io: Added `UseChunkOffset` field, there's no need to rely on the chunk size in this logic. 

Validation for `ChunkSizeBytes` could be added in a follow up PR as this PR is about chunk offset.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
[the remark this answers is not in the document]
0405ysj: Understood it's okay since this PR only introduces `chunk_offset_bytes` without file size information as we discussed before. I would that validation in following PRs.

0405ysj (approved):
Let's merge this PR as is. Currently lots of implementation is blocked, until merging this PR.

frontend/src/libhoclient/httputils.go
```diff
@@ -480,16 +480,11 @@ func writeMultipartRequest(writer *multipart.Writer, job uploadChunkJob) error {
...
 		return err
 	}
-	if err := addFormField(writer, "chunk_number", strconv.Itoa(job.ChunkNumber)); err != nil {
-		return err
-	}
-	if err := addFormField(writer, "chunk_total", strconv.Itoa(job.TotalChunks)); err != nil {
-		return err
-	}
-	if err := addFormField(writer, "chunk_size_bytes", strconv.FormatInt(job.ChunkSizeBytes, 10)); err != nil {
+	if err := addFormField(writer, "chunk_offset_bytes", strconv.FormatInt(offset, 10)); err != nil {
```
0405ysj, line 487: The modification here breaks the unit test at `libhoclient`. (`cd frontend/src/libhoclient && go test ./...`) I found it while cherry-picking commits in this PR to start work on https://github.com/google/android-cuttlefish/pull/1144...

- I cannot modify this PR by myself unless getting permission to do force push at https://github.com/ser-io/android-cuttlefish. Thus, modifying https://github.com/google/android-cuttlefish/pull/1144 is still blocked.
- I think we should have a new Github presubmit workflow for running unit tests.

0405ysj (changes requested):

frontend/src/libhoclient/httputils.go
ser-io: Done.

frontend/src/libhoclient/host_orchestrator_client_test.go
```diff
@@ -50,35 +50,35 @@ func TestUploadFileSucceeds(t *testing.T) {
...
 	xyzzyFile := createTempFile(t, tempDir, "xyzzy", []byte("abraca"))
 	mu := sync.Mutex{}
-	// expected uploads are keyed by format %filename %chunknumber of %chunktotal with the chunk content as value
+	// expected uploads are keyed by format %filename %chunk_offset,%chunk_size with the chunk content as value
 	uploads := map[string]struct{ Content []byte }{
 		// qux
-		"qux 1 of 3": {Content: []byte("lo")},
-		"qux 2 of 3": {Content: []byte("re")},
-		"qux 3 of 3": {Content: []byte("m")},
+		"qux 0,2": {Content: []byte("lo")},
```
0405ysj, line 56: Unlike previous key names, I think naming like "qux 0,2" doesn't look straightforward to understand. Though we have a comment above, could we have more meaningful key names?

e2etests/orchestration/upload_file_test/main_test.go
```diff
@@ -0,0 +1,125 @@
...
+	fileInfo, err := file.Stat()
+	if err != nil {
+		t.Fatal(err)
+	}
+	pipeReader, pipeWriter := io.Pipe()
+	writer := multipart.NewWriter(pipeWriter)
+	go func() {
+		defer pipeWriter.Close()
+		defer writer.Close()
+		if err := addFormField(writer, "chunk_number", strconv.Itoa(1)); err != nil {
```
0405ysj, line 59: Why don't we have E2E test based on `chunk_offset_bytes`? IMO, I think we need E2E tests both for `chunk_number` and `chunk_offset_bytes`?

e2etests/orchestration/upload_file_test/main_test.go
```diff
@@ -0,0 +1,125 @@
...
+			t.Log(err)
+		}
+		if err := addFormField(writer, "chunk_total", strconv.Itoa(1)); err != nil {
```
0405ysj, line 62: Wouldn't this be better to pass 2 or more chunks in E2E tests? For the real use case, we won't update `cvd-host_package.tar.gz` in a single chunk.

e2etests/orchestration/upload_file_test/main_test.go
```diff
@@ -0,0 +1,125 @@
...
+		if err != nil {
+			t.Log(err)
+		}
+		if _, err = io.Copy(fw, file); err != nil {
+			t.Log(err)
+		}
+	}()
+	url := fmt.Sprintf("%s/userartifacts/%s", baseURL, dir)
+	req, err := http.NewRequest(http.MethodPut, url, pipeReader)
+	req.Header.Set("Content-Type", writer.FormDataContentType())
```
0405ysj, line 78: What if goroutine above terminates before reaching at this line? At the end of goroutine above, `writer` will be closed.

e2etests/orchestration/upload_file_test/main_test.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+}
+
+func addFormField(writer *multipart.Writer, field, value string) error {
+	fw, err := writer.CreateFormField(field)
+	if err != nil {
+		return err
+	}
+	_, err = io.Copy(fw, strings.NewReader(value))
+	if err != nil {
```
0405ysj, line 108: nit: `if _, err = io.Copy(fw, strings.NewReader(value)); err != nil {`

e2etests/orchestration/upload_file_test/main_test.go
```diff
@@ -0,0 +1,125 @@
...
+	}()
+	url := fmt.Sprintf("%s/userartifacts/%s", baseURL, dir)
+	req, err := http.NewRequest(http.MethodPut, url, pipeReader)
+	req.Header.Set("Content-Type", writer.FormDataContentType())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, err = http.DefaultClient.Do(req)
+	if err != nil {
```
0405ysj, line 84: nit: `if _, err = http.DefaultClient.Do(req); err != nil {`

e2etests/orchestration/upload_file_test/main_test.go
```diff
@@ -0,0 +1,125 @@
...
+		if _, err = io.Copy(fw, file); err != nil {
+			t.Log(err)
+		}
+	}()
+	url := fmt.Sprintf("%s/userartifacts/%s", baseURL, dir)
+	req, err := http.NewRequest(http.MethodPut, url, pipeReader)
+	req.Header.Set("Content-Type", writer.FormDataContentType())
+	if err != nil {
+		t.Fatal(err)
+	}
```
0405ysj, line 81: nit: Better to check if `err != nil` when error is defined?

e2etests/orchestration/upload_file_test/main_test.go
ser-io: e2e tests **create_from_images_zip_test** and **create_local_image_test** are based on `chunk_offset_bytes`.

e2etests/orchestration/upload_file_test/main_test.go
ser-io: This test is just to validate form fields `chunk_total`, `chunk_size_bytes` and `chunk_number` are handled by the server and not to test how server handles the chunks internally, for that we have **create_from_images_zip_test** and **create_local_image_test** e2e tests.

e2etests/orchestration/upload_file_test/main_test.go
ser-io: Done

frontend/src/libhoclient/host_orchestrator_client_test.go
ser-io: Remove this test, there's no longer need to maintain a fake server as we have e2e tests now.

e2etests/orchestration/upload_file_test/main_test.go
ser-io: Good point. Moved the line on top of where the goroutine starts.

e2etests/orchestration/upload_file_test/main_test.go
ser-io: Done

frontend/src/libhoclient/host_orchestrator_client_test.go
0405ysj: It's still needed. Unit test is useful for the testing speed and efficiency, comparing to E2E test. Please recover this test.

ser-io: This is not a proper unit test, I wouldn't remove it if it were one. I created this tests because the absence of e2e tests at the time implementing a fake server to have some coverage. As I have implemented e2e tests now I don't need to maintain a fake implementation, I can use the real server.

0405ysj: >  when there's no bug in the main code and the test still fails.

Since you've added a new Github workflow for performing unit tests, I think this won't be happened easily like before.

> As I have implemented e2e tests now to cover this logic I don't need to maintain this fake server implementation any longer

I agree with test coverage is duplicated between this and E2E test. OTOH, this takes less than 1 second, comparing to the E2E test taking few minutes. So I've thought it's still useful for faster iteration for development.

If you still prefer to remove this test after recognizing this loss, I wouldn't be against on you. Out of curiosity, are you going to remove all tests under `host_orchestrator_client_test.go` afterwards, since it uses a fake server?

ser-io: > Out of curiosity, are you going to remove all tests under host_orchestrator_client_test.go afterwards, since it uses a fake server?

After a quick scan only `TestCreateCVD` can be removed, the others don't have an e2e counterpart yet.

ser-io:
@0405ysj Thanks for reviewing.

PR #1157, opened 2025-05-14
Expose default_vvmtruststore_file_name in env config

kisenshi:
ExtraPartition is created as a repeated file to be able to extend it in the future.

Bug: b/411021381


kisenshi:
This solution will be replaced by a generic approach. See b/418182565 for details

PR #1160, opened 2025-05-15
Build a minimal version of `secure_env`

Databean:
This is enough to boot, as the latest guest images have keymint and gatekeeper within the device.

Bug: b/401510857


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #1165, opened 2025-05-16
Adds `/v1/userartifacts/{hash}` endpoint.

ser-io:


frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -107,9 +107,11 @@ func (c *Controller) AddRoutes(router *mux.Router) {
 	router.Handle("/userartifacts",
 		httpHandler(&listUploadDirectoriesHandler{c.UserArtifactsManager})).Methods("GET")
 	router.Handle("/userartifacts/{name}",
-		httpHandler(&createUpdateUserArtifactHandler{c.UserArtifactsManager})).Methods("PUT")
+		httpHandler(&createUpdateUserArtifactHandler{m: c.UserArtifactsManager, v1: false})).Methods("PUT")
 	router.Handle("/userartifacts/{dir}/{name}/:extract",
 		httpHandler(&extractUserArtifactHandler{c.OperationManager, c.UserArtifactsManager})).Methods("POST")
+	router.Handle("/v1/userartifacts/{hash}",
+		httpHandler(&createUpdateUserArtifactHandler{m: c.UserArtifactsManager, v1: true})).Methods("PUT")
```
0405ysj, line 114: Let's avoid introducing new REST API, which doesn't have complete implementation yet. We're going to change its behavior from mutable user artifacts into immutable user artifacts, aren't we? If we introduce API endpoint in advance, we should be responsible on compatibility issue.

e2etests/orchestration/upload_artifact_v1_test/main_test.go
```diff
@@ -0,0 +1,45 @@
...
+	hoclient "github.com/google/android-cuttlefish/frontend/src/libhoclient"
+	"github.com/google/go-cmp/cmp"
+)
+
+const baseURL = "http://0.0.0.0:2080"
+
+func TestUploadArtifact(t *testing.T) {
+	srv := hoclient.NewHostOrchestratorService(baseURL)
+
+	if err := srv.UploadArtifact("../artifacts/cvd-host_package.tar.gz"); err != nil {
```
0405ysj, line 30: Why do we upload artifact into the outside of directory? Shouldn't it be prohibited?

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -78,6 +80,20 @@ func NewUserArtifactsManagerImpl(opts UserArtifactsManagerOpts) *UserArtifactsMa
 	}
 }
 
+func (m *UserArtifactsManagerImpl) UpdateArtifact(hash string, chunk UserArtifactChunk) error {
+	if err := createDir(m.RootDir); err != nil {
+		return err
+	}
+	filename := filepath.Join(m.RootDir, hash)
+	if err := createUAFile(filename, m.Owner); err != nil {
```
0405ysj, line 88: We shouldn't use `owner` for the new APIs.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -78,6 +80,20 @@ func NewUserArtifactsManagerImpl(opts UserArtifactsManagerOpts) *UserArtifactsMa
 	}
 }
 
+func (m *UserArtifactsManagerImpl) UpdateArtifact(hash string, chunk UserArtifactChunk) error {
```
0405ysj, line 83: It should update artifact only when dir name with `hash` is empty. Needs to check before writing chunks there.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -78,6 +80,20 @@ func NewUserArtifactsManagerImpl(opts UserArtifactsManagerOpts) *UserArtifactsMa
...
+		return err
+	}
+	filename := filepath.Join(m.RootDir, hash)
+	if err := createUAFile(filename, m.Owner); err != nil {
+		return err
+	}
+	if err := writeChunk(filename, chunk); err != nil {
+		return err
+	}
+	return nil
```
0405ysj, line 94: Duplicated?

frontend/src/host_orchestrator/orchestrator/userartifacts_test.go
0405ysj, line 1: Needs to introduce more tests, before introducing the new API endpoint.

frontend/src/libhoclient/host_orchestrator_client.go
0405ysj, line 1: Again, I think it's better to distinguish PR between HO REST API and HO client API.

0405ysj (changes requested):

frontend/src/host_orchestrator/orchestrator/controller.go
ser-io: I see your point, but this method is complete. Please let me know what's the problem you foresee.

e2etests/orchestration/upload_artifact_v1_test/main_test.go
ser-io: What do you mean? "../artifacts/cvd-host_package.tar.gz" is the path on the client side, not the place the file is uploaded to on the server side.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
ser-io: Every file has an owner. In the HO I prefer to keep the same ownership logic used for uploaded artifacts for consistency purposes. The ownership logic can change in the future, this is an internal logic, clients are not affected by this.

frontend/src/host_orchestrator/orchestrator/userartifacts_test.go
ser-io: Most the e2e test created devices by uploading artifacts. Feel free to send new PRs with more tests if needed.

frontend/src/libhoclient/host_orchestrator_client.go
ser-io: Having  HO and HO client changes in a single commit is a no-no because you cannot split commits. 

If you have commits different you can group them in the same PR if you consider they are related somehow. These changes are related hence I wanted to group them in the same PR.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
ser-io: filename values are computed differently compared with `UpdateArtifactWithDir`

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -48,14 +49,18 @@ type UserArtifactChunk struct {
 // Abstraction for managing user artifacts for launching CVDs.
 type UserArtifactsManager interface {
 	UserArtifactsDirResolver
+	// Update the artifact given the checksum.
+	UpdateArtifact(checksum string, chunk UserArtifactChunk) error
+	// Update the artifact given the checksum.
```
ikicha, line 54: please update the comment about StatArtifact

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -78,6 +83,56 @@ func NewUserArtifactsManagerImpl(opts UserArtifactsManagerOpts) *UserArtifactsMa
...
+	if err := createUAFile(filename, m.Owner); err != nil {
+		return err
+	}
+	if err := writeChunk(filename, chunk); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (m *UserArtifactsManagerImpl) StatArtifact(checksum string) (os.FileInfo, error) {
```
ikicha, line 100: I think we need a lock between UpdateArtifact and StatArtifact, if not, a client needs to take care of it. Even though we can assume every client is trusted, we'd better take care of a lock in the server.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -78,6 +83,56 @@ func NewUserArtifactsManagerImpl(opts UserArtifactsManagerOpts) *UserArtifactsMa
...
 
+func (m *UserArtifactsManagerImpl) UpdateArtifact(checksum string, chunk UserArtifactChunk) error {
+	if err := createDir(m.RootDir); err != nil {
+		return err
+	}
+	filename := filepath.Join(m.RootDir, checksum)
+	if err := createUAFile(filename, m.Owner); err != nil {
+		return err
+	}
+	if err := writeChunk(filename, chunk); err != nil {
```
ikicha, line 94: Can we avoid calling writeChunk once the file is already complete?

frontend/src/host_orchestrator/orchestrator/userartifacts.go
ser-io: Done.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
ser-io: This is a performance improvement that could be added, let's add it when it's needed.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
0405ysj: If we give write permission to the user executing `cvd`, `cvd` binary can modify that artifact while creating CF instances.(e.g. Resizing vbmeta images IIRC) Checksum can be changed, which leads different response for REST API call `GET /v1/userartifacts/{checksum}`. Since there's an E2E test, could you confirm if it's okay?

frontend/src/host_orchestrator/orchestrator/userartifacts.go
0405ysj: We need performance improvement with introducing new HO upload API, or somewhere else. Because, that's our motivation for doing this work. From the other comment thread, is your opinion as getting complexity on the client side, not HO itself?

> Aiming for simplicity, it's simpler to solve this by uploading first against one HO and signal the upload was finished and then start sending create requests

Copied from https://github.com/google/android-cuttlefish/pull/1165#discussion_r2098286920

frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -127,6 +130,11 @@ type HostOrchestratorService interface {
 	CreateSnapshot(groupName, instanceName string, req *hoapi.CreateSnapshotRequest) (*hoapi.CreateSnapshotResponse, error)
 }
 
+type UserArtifactsService interface {
+	// Upload artifact.
+	UploadArtifact(filename string) error
+}
+
```
0405ysj: Why do we keep all legacy methods at `HostOrchestratorService`, instead moving into `UserArtifactsService`? I know those code will be removed someday, but wouldn't it be better to refactor?

e2etests/orchestration/upload_artifact_v1_test/main_test.go
0405ysj: Ah, I misunderstood your code. Thanks for the correction.

frontend/src/host_orchestrator/api/v1/messages.go
```diff
@@ -152,3 +152,5 @@ type StartCVDRequest struct {
 	// Start from the relevant snaphost if not empty.
 	SnapshotID string `json:"snapshot_id,omitempty"`
 }
+
+type UserArtifactStatResponse struct{}
```
0405ysj, line 156: Unlike the design doc, won't we use `created_at` field in the response of `StatArtifact`? Noting that I'm not opposite on not using `created_at`. Just out of curiosity, I want to get confirmation if it's intended.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -94,6 +97,42 @@ func (m *UserArtifactsManagerImpl) UpdateArtifact(checksum string, chunk UserArt
 	return nil
 }
 
+func (m *UserArtifactsManagerImpl) StatArtifact(checksum string) (os.FileInfo, error) {
```
0405ysj, line 100: If we're not going to use `os.FileInfo` at the caller side, returning type as `error` would be okay? We can return `os.FileInfo` later.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -94,6 +97,42 @@ func (m *UserArtifactsManagerImpl) UpdateArtifact(checksum string, chunk UserArt
...
+	filename := filepath.Join(m.RootDir, checksum)
+	exist, err := fileExist(filename)
+	if err != nil {
+		return nil, err
+	}
+	if !exist {
+		return nil, operator.NewBadRequestError(fmt.Sprintf("user artifact with checksum %q does not exist", checksum), nil)
```
0405ysj, line 107: Status 404, instead of 400?

e2etests/orchestration/upload_artifact_v1_test/main_test.go
```diff
@@ -26,15 +27,31 @@ const baseURL = "http://0.0.0.0:2080"
...
+	for i := 0; i < 2; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			if err := srv.UploadArtifact("../artifacts/cvd-host_package.tar.gz"); err != nil {
+				t.Error(err)
+			}
+			if _, err := srv.StatArtifact(expected); err != nil {
+				t.Error(err)
+			}
```
0405ysj, line 50: IIUC, this test should fail if the implementation of `StatArtifact` is correct. If only a part of chunks are delivered, `StatArtifact` should return an error.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
0405ysj: I understood this makes sense because we removed status 409 from the reply of `PUT /v1/userartifacts/{checksum}`. I'll make this comment as unresolved if we need status 409 again.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
ser-io: `cvd` will not work with artifacts from the artifacts repository directly, artifacts needs to be duplicated first for `cvd` to use them.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
ser-io: The checksum based API solves the problem of avoiding to upload the same artifact twice allowing a client to reuse them when creating new devices. This is the performance improvement clients can achieve with the new API. To take the full benefit of the new API you need to wait for the artifact to be uploaded first.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
ser-io: Agree here, however using the `Stat` verb implies file info must be returned. This means I'd need work with a new method and signature, something like: `CheckArtifactExist(checksum string) (bool, error)`. In Golang the way to know if a file exist is using [os.Stat](https://pkg.go.dev/os#Stat) and discarding the file info, there's nothing as CheckIfFileExist, I'd like to keep the signature similar to os.Stat.

e2etests/orchestration/upload_artifact_v1_test/main_test.go
ser-io: If the previous `UploadArtifact` call doesn't fail, it means the file was upload correctly, it means all chunks were properly persisted on the server side and it means that the checksum of the file on the server side should match the checksum of the original file, hence the StatArtifact call should not fail and the overall test should pass.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
ser-io: Done.

PR #1167, opened 2025-05-16
Construct chunkState for efficiently managing concurrent updated chunks

0405ysj:
Design document: go/ho-upload-api-v1

At the newly proposed design of Host Orchestrator upload API, it supposes to handle concurrent chunks within one REST API `PUT /v1/userartifacts/{hash}`. With this API's given preconditions that the user artifact is immutable & API returns error status 409(conflict) when already it exists there, we typically calculate the hash sum before replying on REST API request when it actually writes a chunk.

However, calculating hash sum is too heavy task to do with this REST API, as the size of CF image(few gigabytes) & CVD host package(hundreds of megabytes) is too large comparing to the size of each chunk(tens of kilobytes) of the legacy APIs as reference. The estimated time complexity of each REST API is **O(artifact_size)** per REST API request, and total number of REST API requests is approximately **{size of artifact}/{size of chunk}**.

This PR suggests to minimize the number of calculating hash sum, by checking ranges of updated chunks. I imported `github.com/google/btree` because B-tree is efficient(**O(log N)**) for searching, inserting, and deleting elements. What I want to solve here is similar to the merge interval algorithm problem, but noting that the order of ranges are random since it's coming from the client concurrently.

With this PR, we can let `UserArtifactManager` to calculate hash sum only when updated chunk covers all bytes of the file. Without assuming ranges of chunks are heavily overlapped, the time complexity of `update(start int64, end int64)` is **O(log chunk_number)**. The time complexity of `isCompleted()` is just constant time(**O(1)**).

Also this struct is designed to be safe from concurrent execution.

### DIFF
--- a/frontend/src/host_orchestrator/go.mod
+++ b/frontend/src/host_orchestrator/go.mod
@@ -4,6 +4,7 @@ go 1.17
 
 require (
 	github.com/google/android-cuttlefish/frontend/src/liboperator v0.0.0-20240822182916-7bea0dafdbde
+	github.com/google/btree v1.1.3
 	github.com/google/go-cmp v0.5.9
 	github.com/google/uuid v1.3.0
 	github.com/gorilla/mux v1.8.0
--- a/frontend/src/host_orchestrator/orchestrator/chunkstate.go
+++ b/frontend/src/host_orchestrator/orchestrator/chunkstate.go
@@ -1,0 +1,138 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package orchestrator
+
+import (
+	"sync"
+	"fmt"
+
+	"github.com/google/btree"
+)
+
+// Structure for managing the state of updated chunk for efficiently knowing whether the user
+// artifact may need to calculate hash sum or not.
+type ChunkState struct {
+	fileSize int64
+	items    *btree.BTree
+	mutex    sync.RWMutex
+}
+
+type chunkStateItem struct {
+	// Starting byte offset of the continuous range having same state whether updated or not.
+	offset int64
+	// State description whether given byte offset of the user artifact is updated or not.
+	isUpdated bool
+}
+
+func (i chunkStateItem) Less(item btree.Item) bool {
+	return i.offset < item.(chunkStateItem).offset
+}
+
+func NewChunkState(fileSize int64) *ChunkState {
+	cs := ChunkState{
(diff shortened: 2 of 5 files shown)


frontend/src/host_orchestrator/orchestrator/chunkstate.go
```diff
@@ -0,0 +1,138 @@
...
+	offset int64
+	// State description whether given byte offset of the user artifact is updated or not.
+	isUpdated bool
+}
+
+func (i chunkStateItem) Less(item btree.Item) bool {
+	return i.offset < item.(chunkStateItem).offset
+}
+
+func newChunkState(fileSize int64) *chunkState {
```
ikicha, line 43: nits: Consider capitalizing the first case of the function which is expected to be used by a user and adding the comment for that. I know it's used in the same pkg, but for readability.

I think only newChunkState, update, isCompleted will be used by other component.

frontend/src/host_orchestrator/orchestrator/chunkstate.go
```diff
@@ -0,0 +1,138 @@
...
+	fileSize int64
+	items    *btree.BTree
+	mutex    sync.RWMutex
+}
+
+type chunkStateItem struct {
+	// Starting byte offset of the continuous range having same state whether updated or not.
+	offset int64
+	// State description whether given byte offset of the user artifact is updated or not.
+	isUpdated bool
```
ikicha, line 36: nits: isComplete might be better.

frontend/src/host_orchestrator/go.mod
```diff
@@ -4,6 +4,7 @@ go 1.17
 
 require (
 	github.com/google/android-cuttlefish/frontend/src/liboperator v0.0.0-20240822182916-7bea0dafdbde
+	github.com/google/btree v1.1.3
```
ikicha, line 7: you may need to update internal replacement config file for pkg name...

ikicha (approved):

frontend/src/host_orchestrator/orchestrator/chunkstate.go
0405ysj: Capitalized the first case for `NewChunkState`, `Update`, `IsCompleted` functions.

Also capitalized the struct `ChunkState`. Kept `chunkStateItem` as before, since I'm not intending to export that struct into outside.

frontend/src/host_orchestrator/orchestrator/chunkstate.go
0405ysj: We cannot say the full file is completed though the value of this boolean field is `true`. It's for marking whether the subrange of target artifact is updated or not.

frontend/src/host_orchestrator/go.mod
0405ysj: Thanks for notification, I would consider there :)

frontend/src/host_orchestrator/orchestrator/chunkstate.go
```diff
@@ -0,0 +1,138 @@
...
+
+	"github.com/google/btree"
+)
+
+// Structure for managing the state of updated chunk for efficiently knowing whether the user
+// artifact may need to calculate hash sum or not.
+type ChunkState struct {
+	fileSize int64
+	items    *btree.BTree
+	mutex    sync.RWMutex
```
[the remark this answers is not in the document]
0405ysj: Done.

PR #1183, opened 2025-05-20
Start building `//cuttlefish/host/commands/process_sandboxer`

Databean:
From `//cuttlefish/host/commands/process_sandboxer`, this includes:

- `credentialed_unix_server`
- `logs`
- `pidfd`
- `policies_header`
- `poll_callback`
- `proxy_common`
- `sandboxer_proxy`
- `signal_fd`

Includes a single policy, `policies:adb_connector`.

`sandboxed_api`'s repository has a `MODULE.bazel` but it is not listed on the bazel registry. This includes it by using a combination of a `bazel_dep` and a `git_override`, rather than a `git_repository`. It seems that a `git_repository` ignores the `MODULE.bazel` and doesn't isolate in the same way a `bazel_dep` does.

Headers from `sandboxed_api` and `abseil` insist on being `#include`d via quotes rather than angle brackets, so headers being built are updated to accommodate.


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #1197, opened 2025-05-23
Tooling to build the android-cuttlefish debs packages in containers.

ser-io:


tools/buildutils/cw/Containerfile
```diff
@@ -0,0 +1,17 @@
+FROM debian:stable AS base
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt update -y && apt upgrade -y
+RUN apt install -y devscripts config-package-dev debhelper-compat golang protobuf-compiler nasm
```
jemoreira, line 6: some of these packages aren't needed here since they're specified as build time dependencies of the packages being built and should be installed by the script that builds them.

tools/buildutils/cw/build_package.sh
```diff
@@ -0,0 +1,25 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#!/usr/bin/env bash
+
+set -o errexit -o nounset -o pipefail
+
+>&2 echo "Working dir: $(pwd)"
```
jemoreira, line 21: can this share code with tools/buildutils/build_packages.sh?

jemoreira (approved):

tools/buildutils/cw/Containerfile
0405ysj, line 1: Why don't we reuse existing [docker/Dockerfile](https://github.com/google/android-cuttlefish/blob/main/docker/Dockerfile) for building CF debian packages? It's workable for building `x86_64` and `arm64` CF debian packages.

0405ysj (changes requested):

tools/buildutils/cw/build_package.sh
jemoreira: Those two lines have 10 command line arguments in total, I'd say it's very likely that those commands are modified further in the future, so I prefer to keep them in a single place.

tools/buildutils/cw/Containerfile
0405ysj: I had a discussion with video chat, and we've agreed not to reuse existing [docker/Dockerfile](https://github.com/google/android-cuttlefish/blob/main/docker/Dockerfile) at this PR in advance.

Though I still speculate reusing existing Dockerfile can be the better approach with the perspective of maintenance cost for managing multiple Dockerfile or Containerfile, I think this can be done later on the other hand.

tools/buildutils/cw/build_package.sh
ser-io: Got it! Extracted building single package logic into its own script. PTAL.

tools/buildutils/build_package.sh
```diff
@@ -0,0 +1,29 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+function build_package() {
```
jemoreira: This doesn't need to be a function

tools/buildutils/cw/Containerfile
```diff
@@ -0,0 +1,17 @@
...
+FROM base AS with_bazel
+
+RUN --mount=source=.,target=/mnt,type=bind \
+  /mnt/tools/buildutils/installbazel.sh
+
+FROM with_bazel
+
+COPY ./tools/buildutils/build_package.sh /build_package.sh
+
+ENTRYPOINT ["/build_package.sh"]
```
jemoreira, line 13: IIUC this calls `build_package.sh` without parameters, but that script expects one. Either make the parameter optional (by giving it a default value in the script) or pass a parameter here.

tools/buildutils/cw/Containerfile
```diff
@@ -0,0 +1,17 @@
...
+
+RUN apt update -y && apt upgrade -y
+RUN apt install -y sudo devscripts
+
+FROM base AS with_bazel
+
+RUN --mount=source=.,target=/mnt,type=bind \
+  /mnt/tools/buildutils/installbazel.sh
+
+FROM with_bazel
```
jemoreira: what's the purpose of using multiple `FROM` instructions here?

tools/buildutils/cw/Containerfile
ser-io: The parameter should be passed into the `docker run` command, for example:

```
docker run -v=.:/mnt/build -w /mnt/build android-cuttlefish-build:latest [ARG]

docker run -v=.:/mnt/build -w /mnt/build android-cuttlefish-build:latest base

docker run -v=.:/mnt/build -w /mnt/build android-cuttlefish-build:latest frontend
```

jemoreira: In that case make the script fail when there is no parameter

tools/buildutils/build_package.sh
ser-io: True. Done!

tools/buildutils/cw/Containerfile
ser-io: Done!

tools/buildutils/cw/Containerfile
ser-io: Removed those statements. Added them based on a miss assumptions on how cache works when building images.

PR #1201, opened 2025-05-28
screen_connector_common.h refactoring

Databean:
- Move implementations to a source file, reducing header transitive includes.
- Reduce use of `auto`: abseil.io/tips/232
- Don't use `std::` prefixes on integer types: https://google.github.io/styleguide/cppguide.html#Integer_Types
- Delete some unused functions.
- Change `ScreenConnector` from a struct with static functions to a namespace with free functions, as there is no state.
- Replace a CHECK_GE with a CHECK_GT in one function.

Bug: b/420772194


jmacnak (approved):

Databean:
Thanks for reviewing!

PR #1203, opened 2025-05-29
Adds docker image presubmit check.

ser-io:


.github/workflows/presubmit.yaml
```diff
@@ -182,4 +182,22 @@ jobs:
     - name: Used disk space
       run: |
         df -h
+  docker-image-check:
+    runs-on: ubuntu-22.04
```
0405ysj, line 186: Could we have presubmit check for ARM64 docker image as well? Recently we're facing an issue that ARM64 docker image is not available to be built. We expect the root cause as missing files for ARM64 while building libffi? https://github.com/google/android-cuttlefish/pull/943

.github/workflows/presubmit.yaml
```diff
@@ -182,4 +182,22 @@ jobs:
...
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@a81bbbf8298c0fa03ea29cdc473d45769f953675 # aka v2
+    - name: Run check
+      shell: bash
+      run: |
+        cd docker
+        # TODO(b/420987840): Reuse deb packages created in `build-debian-package` job.
+        bash image-builder.sh -d
```
0405ysj, line 195: I think `./image-builder.sh -d` should be enough, since we're defining to use `bash` shell above?

.github/workflows/presubmit.yaml
0405ysj, line 1: What about E2E test coverage for testing each HO REST APIs on this docker image? At least we're facing at an issue on `systemd-journal-gatewayd`. Except E2E tests regarding that, I think we should recover test coverage not to occur any new issues breaking HO REST APIs on the docker image only?

.github/workflows/presubmit.yaml
ser-io: Done.

.github/workflows/presubmit.yaml
```diff
@@ -182,4 +182,31 @@ jobs:
...
+        docker images
+        sudo docker run --privileged -d -p 2080:2080 cuttlefish-orchestration:latest
+        # Wait for HO service to start.
+        sleep 10s
+        res=$( curl -v http://localhost:2080/cvds )
+        echo "response: ${res}"
+        test "${res}" = '{"cvds":[]}'
+        # Run create_from_images_zip_test e2e tests
+        cd e2etests
+        bazel test orchestration/create_from_images_zip_test:create_from_images_zip_test_test
```
0405ysj, line 210: About HO E2E tests running currently, would the logic around running podman instance be in the scope of bazel test module? My concern is about whether this Github action runs a podman instance or not, though I guess not.

0405ysj (approved):

.github/workflows/presubmit.yaml
ser-io: There's no podman instance here. This job creates a docker container exposing port 2080 to the host then the `create_from_images_zip_test` test reaches `http://localhost:2080` to execute its logic.

0405ysj: Thanks for your clarification :)

PR #1206, opened 2025-05-29
Use android/system/core to provide libcutils, liblp, libsparse

Databean:
Bug: b/421034556


adelva1984 (approved):

base/cvd/libbase/BUILD.bazel
```diff
@@ -34,6 +35,7 @@ cc_library(
         "mapped_file.cpp",
         "parsebool.cpp",
         "parsenetaddress.cpp",
+        "properties.cpp",
```
jemoreira, line 38: could we use libbase from android too? If you do that first you won't need to fix this copy.

jemoreira (approved):

Databean:
Going to break this into separate PRs for the different libraries as there is library-specific work to merge this into the google-internal version.

PR #1207, opened 2025-05-29
Support arm64 while building libffi

k311093:
Add arm64 related header.
Add arm64 target while selecting platform.


ser-io (approved):

base/cvd/BUILD.bazel
```diff
@@ -71,6 +71,17 @@ cc_library(
...
+cc_library(
+    name = "libffi_linux_arm64",
+    hdrs = [
+        "libffi/linux-arm64/ffi.h",
+        "libffi/linux-arm64/fficonfig.h",
+        "libffi/linux-arm64/ffitarget.h",
+    ],
+    includes = ["libffi/linux-arm64"],
+    visibility = ["//visibility:public"],
+)
```
jemoreira, line 83: this should be in `libffi/linux-arm64/BUILD`

base/cvd/BUILD.libffi.bazel
```diff
@@ -59,6 +73,7 @@ cc_library(
     name = "libffi",
     srcs = COMMON_SRCS + select({
         "@platforms//cpu:x86_64": SRCS_X86_64,
+        "@platforms//cpu:arm64": SRCS_ARM64,
         "//conditions:default": [],
     }),
     copts = [
```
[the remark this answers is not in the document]
k311093: I checked with the libffi at Bazel registry `bazel_dep(name = "libffi", version = "3.4.7.bcr.3")` and found that it also issue for arm64 build (looks like FFI_EXEC_TRAMPOLINE_TABLE treated as not defined while building arm64 even if the build rule is intented to set the value).

So using Bazel registry's module would be later when the issue is fixed.

base/cvd/BUILD.bazel
k311093: Moved cc library declaration to `libffi/linux-arm64/BUILD`
Also moved x86 related build rules to `libffi/linux-x86/BUILD`

PR #1208, opened 2025-05-29
Support arm64 while building libwebrtc

k311093:
Move x86 specific files into platform dependent list. Add neon related files for libyuv while building arm64 libyuv.


jemoreira (approved):
thanks!

PR #1217, opened 2025-06-02
Adds `gofmt` check in presubmit.

ser-io:


.github/workflows/presubmit.yaml
```diff
@@ -46,11 +46,24 @@ jobs:
       run: apt install -y git golang
     - name: Go version
       run: go version
+    - name: Set Github ENV variables
+      shell: bash
+      run: |
+        echo "GOPROJECTS=('host_orchestrator' 'libhoclient' 'liboperator' 'operator')" >> $GITHUB_ENV
```
jemoreira: can this be
```
env:
  GOPROJECTS: ('host_orchestrator' 'libhoclient' 'liboperator' 'operator')
```
?

ser-io: Got the following error using `env`.


```
Error
every step must define a `uses` or `run` key
```

Ref: https://github.com/google/android-cuttlefish/actions/runs/15399607234/workflow

Github documentation recommends using the `echo` command to set env variables: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-environment-variable

jemoreira: > Got the following error using env.

Every example I saw put `env` before the `jobs` list, not inside it.

ser-io: I see! Using `env` as part of the job now, not the step. Thanks!

jemoreira (approved):

PR #1220, opened 2025-06-03
Simplify some composite disk creation

Databean:
See commit messages.

Bug: b/422001084
Bug: b/422005116


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #1221, opened 2025-06-03
Build and package Gfxstream

jmacnak:
Bug: b/403290370
Test: `bazel build @crosvm_bin//:crosvm__crosvm`
Test: Uncomment the `# cuttlefish-common/bin/crosvm` line in cuttlefish/package/BUILD.bazel

```
bazel run //cuttlefish/package:cvd -- \
    create \
    --gpu_mode=gfxstream \
    --host_substitutions=all
```
Confirm in the host logs that the bazel built crosvm is being used.

Confirm in gdb that crosvm is using the bazel built libgfxstream_backend.so

### DIFF
--- a/base/cvd/BUILD.bazel
+++ b/base/cvd/BUILD.bazel
@@ -88,3 +88,19 @@ alias(
     actual = "@crosvm//:minijail_sources",
     visibility = ["//visibility:public"],
 )
+
+# Create an alias so that crosvm's rutabaga_gfx crate has visibility into the
+# gfxstream repository.
+alias(
+    name = "gfxstream_backend",
+    actual = "@gfxstream//host:gfxstream_backend",
+    visibility = ["//visibility:public"],
+)
+
+# Useful for `rust_library()`s which can not directly depend on a `cc_shared_library()`.
+# TODO: remove after ag/33719949 and update alias above to reference directly.
+cc_import(
+    name = "gfxstream_backend_import",
+    shared_library = ":gfxstream_backend",
+    visibility = ["//visibility:public"],
+)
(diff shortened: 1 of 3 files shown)


base/cvd/BUILD.bazel
```diff
@@ -88,3 +88,11 @@ alias(
     actual = "@crosvm//:minijail_sources",
     visibility = ["//visibility:public"],
 )
+
+# Create an alias so that crosvm's rutabaga_gfx crate has visibility into the
+# gfxstream repository.
```
Databean, line 93: Can this be literally an [`alias`](https://bazel.build/reference/be/general#alias) rule?

base/cvd/MODULE.bazel
```diff
@@ -35,6 +35,7 @@ bazel_dep(name = "platforms", version = "0.0.11")
 bazel_dep(name = "protobuf", version = "29.1")
 bazel_dep(name = "rules_cc", version = "0.1.1")
 bazel_dep(name = "rules_flex", version = "0.4")
+bazel_dep(name = "rules_license", version = "1.0.0")
```
Databean, line 38: How is this used? I don't see a `@rules_license` in this change.

base/cvd/BUILD.bazel
jmacnak: Updated. After ag/33719949 is resolved, we should have a cc_import to reference directly with the alias.

base/cvd/MODULE.bazel
jmacnak: This is needed because the Gfxstream "host" subdirectory's "package()" references the Gfxstream license. Without it, we see

```
external/+_repo_rules2+gfxstream/host/BUILD.bazel:280:10: every rule of type cc_binary implicitly depends upon the target '@@+_repo_rules2+gfxstream//:gfxstream_license', but this target could not be found because of: error loading package '@@+_repo_rules2+gfxstream//': Unable to find package for @@[unknown repo 'rules_license' requested from @@+_repo_rules2+gfxstream]//rules:license_kind.bzl: The repository '@@[unknown repo 'rules_license' requested from @@+_repo_rules2+gfxstream]' could not be resolved: No repository visible as '@rules_license' from repository '@@+_repo_rules2+gfxstream'.
```

PR #1222, opened 2025-06-03
Split HostOrchestratorService into smaller interfaces.

ser-io:


frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -72,61 +72,58 @@ type CreateBugReportOpts struct {
 	IncludeADBBugReport bool
 }
 
-// A client to the host orchestrator service running in a remote host.
-type HostOrchestratorService interface {
+type InstancesService interface {
```
jemoreira: nit: comments on each interface would be nice

frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -72,61 +72,58 @@ type CreateBugReportOpts struct {
...
 
+// A client to the host orchestrator service running in a remote host.
+type HostOrchestratorService interface {
+	InstancesService
+	UserArtifactsService
+	DeviceOperationsService
+	// Creates a webRTC connection to a device running in this host.
+	ConnectWebRTC(device string, observer wclient.Observer, logger io.Writer, opts ConnectWebRTCOpts) (*wclient.Connection, error)
+	// Connect to ADB WebSocket endpoint.
+	ConnectADBWebSocket(device string) (*websocket.Conn, error)
```
jemoreira, line 122: These two should be in `DeviceOperations`

frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -72,61 +72,58 @@ type CreateBugReportOpts struct {
...
-	// Creates a webRTC connection to a device running in this host.
-	ConnectWebRTC(device string, observer wclient.Observer, logger io.Writer, opts ConnectWebRTCOpts) (*wclient.Connection, error)
-
-	// Connect to ADB WebSocket endpoint.
-	ConnectADBWebSocket(device string) (*websocket.Conn, error)
-
-	// Wait for an operation, `result` will be populated with the relevant operation's result object.
-	WaitForOperation(name string, result any) error
-
+type DeviceOperationsService interface {
```
jemoreira: Use either "Instance" or "Device" but not both. Either rename this as `InstanceOperations` or rename the other one as `DeviceService`.

frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -72,61 +72,58 @@ type CreateBugReportOpts struct {
...
-
 	// Start the device.
 	Start(groupName, instanceName string, req *hoapi.StartCVDRequest) error
-
 	// Create device snapshot.
 	CreateSnapshot(groupName, instanceName string, req *hoapi.CreateSnapshotRequest) (*hoapi.CreateSnapshotResponse, error)
 }
 
+// A client to the host orchestrator service running in a remote host.
+type HostOrchestratorService interface {
```
jemoreira: nit: Since you're moving stuff around, consider replacing the "Service" prefix with "Client".

frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -72,61 +72,58 @@ type CreateBugReportOpts struct {
...
+	InstancesService
+	UserArtifactsService
+	DeviceOperationsService
+	// Creates a webRTC connection to a device running in this host.
+	ConnectWebRTC(device string, observer wclient.Observer, logger io.Writer, opts ConnectWebRTCOpts) (*wclient.Connection, error)
+	// Connect to ADB WebSocket endpoint.
+	ConnectADBWebSocket(device string) (*websocket.Conn, error)
+	// Wait for an operation, `result` will be populated with the relevant operation's result object.
+	WaitForOperation(name string, result any) error
```
jemoreira, line 128: `WaitForOperation` should probably be in its own interface. `OperationsService`?

frontend/src/libhoclient/host_orchestrator_client.go
ser-io: Good catch. Done.

frontend/src/libhoclient/host_orchestrator_client.go
ser-io: I created `ADBConnectionService` for these two. I think these two can be moved into its own abstraction.

frontend/src/libhoclient/host_orchestrator_client.go
ser-io: Renamed it to `InstanceOperationsService`.

frontend/src/libhoclient/host_orchestrator_client.go
ser-io: Good call! Handled this comment in a follow-up commit. PTAL

frontend/src/libhoclient/host_orchestrator_client.go
ser-io: Yes! Done.

frontend/src/libhoclient/host_orchestrator_client.go
jemoreira: webrtc is more than just ADB, that's not the right interface name for it. A more general name like InstanceConnections would be more appropriate

ser-io: Got it! Renames to `InstanceConnections` as suggested.

PR #1223, opened 2025-06-03
Remove the external DNS resolver capability.

Databean:
This was only present to support libc implementations that were missing DNS clients (bionic).

Bug: b/422242138


cjreynol (approved):

Databean:
Thanks for reviewing!

PR #1227, opened 2025-06-04
Remove support for booting protected VMs in cuttlefish.

Databean:
That ship has sailed.

Bug: b/422006866


adelva1984 (approved):

Databean:
Thanks for reviewing!

PR #1229, opened 2025-06-04
More logs in e2e tests output.

ser-io:


frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -128,13 +142,19 @@ type OperationsClient interface {
 	WaitForOperation(name string, result any) error
 }
 
+// Pulls service logs via http
+type JournaldClient interface {
```
jemoreira: nit: The fact that the ho logs are exposed with journald is an implementation detail. It could change if the ho is running in docker for example. This interface should not include "Journald" in the name.

Also, while not applicable to this PR, the URL where the logs are exposed shouldn't be `.../_journal/entries` but something more general that the implementation (nginx) then redirects or reverse-proxies to journald.

ser-io: Good point.

Rather than committing to a general anstraction to pull service logs, changed the PR to have rather an specific one targeting the [systemd-journal-gatewayd.service](https://www.freedesktop.org/software/systemd/man/latest/systemd-journal-gatewayd.service.html#Supported%20URLs) API, named it `SystemdJournalGatewaydClient`. The `HostOrchestratorClient` interface does not include this interface, but the `HostOrchestratorClientImpl` does implement the new interface method.

jemoreira: How does this address the fact that `HostOrchestratorClientImpl` is used for all host orchestrators and some may not have systemd, but a different implementation and url to get the logs?

ser-io: Moved the implementation out of `HostOrchestratorClientImpl` having a different struct `SystemdJournalGatewaydClientImpl` that implements it.

jemoreira: I prefer the first version of this PR, where getting HO logs was part of the HO client (with the caveat that the interface should not have journald in its name), which I approved.

This version not only doesn't address the fact that an implementation detail is being exposed more than it should, but also unnecessarily splits the HO client in two objects.

ser-io: Updated PR with offline discussion decision.

jemoreira (approved):

PR #1230, opened 2025-06-04
Use host credentials

ser-io:


frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -37,6 +37,7 @@ import (
 
 const HeaderBuildAPICreds = "X-Cutf-Host-Orchestrator-BuildAPI-Creds"
 const HeaderUserProject = "X-Cutf-Host-Orchestrator-BuildAPI-Creds-User-Project-ID"
+const HeaderUseHostBuildAPICreds = "X-Cutf-Host-Orchestrator-Use-Host-BuildAPI-Creds"
```
jemoreira: why is this header needed? why is it not the default when no other credential headers are present?

e2etests/orchestration/verify_use_host_credentials_test/main_test.go
```diff
@@ -0,0 +1,69 @@
...
+	if err == nil {
+		t.Fatal("error: no device should have been created")
+	}
+	log.Printf("got create error: %s\n", err)
+	// cvd pings `metadata.google.internal` aiming to retrieve host's service
+	// account credentials as such host is not available `srv.CreateCVD` fails
+	// with error mentioning that host: `metadata.google.internal` was not resolved.
+	const expected = "Could not resolve host: metadata.google.internal"
+	if !strings.Contains(fmt.Sprintf("%s", err), expected) {
+		t.Fatalf("%q part not found in error message", expected)
```
jemoreira: nit: print the actual error message too

frontend/src/host_orchestrator/orchestrator/controller.go
ser-io: The default is not using any credentials allowing to work with `aosp` targets seamlessly.

Not every deployment supports hosts having credentials. Currently, the only deployment is GCE, and even if deployed to GCE, the operators have to make extra steps to work with the service account credentials. Working with Host Credentials seems rather a very uncommon scenario, hence I don't think it would be a good default.

e2etests/orchestration/verify_use_host_credentials_test/main_test.go
ser-io: yes, error message is printed earlier in line 61.

frontend/src/host_orchestrator/orchestrator/controller.go
jemoreira: Whether uncommon or not, if the host has credentials authorized with the build api it should use them by default, the client doesn't need to know this.

The underlying problem here is that all GCE hosts have credentials, `cvd` always finds those credentials and doesn't check whether they have access to the build api or not and the build api denies access to public builds for credentials that don't have access.

The obvious solution would be to fix the build api, but that's beyond our control. I agree that the host orchestrator must be told to use the GCE credentials, but it should not be the client who decides that. There should be some configuration file in the GCE image or a metadata key that HO can read that tells it to use the GCE credentials or not.

ser-io: Added new HO configuration parameter `build_api_credentials_use_gce_metadata` as discussed offline and removed header and client responsibility. PTAL.

frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -726,3 +721,32 @@ func okHandler() http.Handler {
...
 	})
 }
+
+func getFetchCredentials(config BuildAPICredentialsConfig, r *http.Request) cvd.FetchCredentials {
+	if config.UseGCEMetadata {
+		log.Println("fetch credentials: using gce service account credentials")
+		return cvd.FetchCredentials{
+			UseGCEServiceAccountCredentials: true,
+		}
+	}
```
jemoreira, line 741: Would it make sense to use these credentials only if the user didn't pass their own? That would provide the most flexibility in my opinion: The user gets to choose whether to provide their own, if not the host orchestrator can use the GCE ones.

ser-io: I agree with the flexibility part, the current issue is that the HO is not smart enough to detect beforehand whether there's any valid credentials in the host, therefore I see it problematic to blindly default to use host credentials as this would result on misleading errors the final user is not responsible for, nor can do anything about. For now I prefer to leave that responsibility to the deployment's operator who has knowledge of whether the HO could use host credentials or not via config.

Also, in the current state of the HO, defaulting to `--credential_source=gce` will break the `aosp` artifacts flow, where no credentials are needed, for every deployment that's not GCE with a valid Service Account.

jemoreira: I'm not saying to use `--credential_source=gce` blindly by default.

This change uses `gce` when `config.UseGCEMetadata` is `true`, regardless of what client requested. I'm suggesting the host orchestrator should honor the client request to use their own credentials, and default to the configured behavior only if the client didn't provide their own credentials. So the different cases would look like this:
```
no user credentials + not configured to use GCE -> empty credentials
no user credentials + configured to use GCE -> credentials from GCE metadata
user provided credentials + (whatever was configured) -> credentials provided by user
```

ser-io: Got it now. Done! PTAL

jemoreira (approved):

PR #1231, opened 2025-06-04
Build and package cpio

adelva1984:
Use the libarchive version of cpio as the Gnu version is not in the BCR and uses a complicated series of git submodules; this also enables us to add bsdtar later which has some advantages such as a sparse extraction mode.

Bug: 402297826


base/cvd/cuttlefish/host/commands/assemble_cvd/boot_image_utils.cc
```diff
@@ -179,8 +179,7 @@ void UnpackRamdisk(const std::string& original_ramdisk_path,
   int cpio_status;
   do {
     LOG(ERROR) << "Running";
-    cpio_status = Command(HostBinaryPath("toybox"))
-                      .AddParameter("cpio")
+    cpio_status = Command(CpioBinary())
```
Databean, line 182: Will this block enabling substitution for `assemble_cvd` until `cpio` is available internally?

Can this get a symlink like ag/33707825 so that `cpio` will work out of `cvd-host_package.tar.gz` as well?

Databean (approved):

base/cvd/cuttlefish/host/commands/assemble_cvd/boot_image_utils.cc
adelva1984: Ack, fixed in ag/33922290

PR #1232, opened 2025-06-04
[HO] Fixes using latest green builds.

ser-io:


frontend/src/host_orchestrator/orchestrator/cvd/cvd.go
```diff
@@ -311,10 +312,32 @@ func (cli *CLI) PowerBtn(selector Selector) error {
...
-	BuildID     string
+	BuildID string
+	// Discarded if BuildID is set.
+	// When `Branch` field is set, the latest green build id will be used.
+	Branch      string
 	BuildTarget string
 }
 
+func (b *AndroidBuild) CvdFormat() string {
+	first := b.BuildID
```
jemoreira: nit: "first" renamed as "build" is more readable. And you don't need "second".

frontend/src/host_orchestrator/orchestrator/cvd/cvd.go
```diff
@@ -311,10 +312,32 @@ func (cli *CLI) PowerBtn(selector Selector) error {
...
 type AndroidBuild struct {
-	BuildID     string
+	BuildID string
+	// Discarded if BuildID is set.
+	// When `Branch` field is set, the latest green build id will be used.
+	Branch      string
 	BuildTarget string
 }
 
+func (b *AndroidBuild) CvdFormat() string {
```
jemoreira: It makes more sense for some `cvd` struct to have a method to format a build for its command than the build having a method to format itself how cvd needs it. Doesn't make much of a difference now because `AndroidBuild` is part of the `cvd` module, but it didn't have to be.

jemoreira (approved):

frontend/src/host_orchestrator/orchestrator/cvd/cvd.go
ser-io: Done.

frontend/src/host_orchestrator/orchestrator/cvd/cvd.go
ser-io: Good Point! ended up creating a new single function part of the module: `cliFormat(b AndroidBuild)`

PR #1234, opened 2025-06-05
Split `build_api.h` and extract class member `BuildApi::GetBuildZipName`

Databean:
`GetBuildZipName` does not use any state from the subclasses and does not have subclass-specific behavior.

Also removes `Result` from the `GetBuildZipName` return type since it never throws errors.

Bug: b/422579325


base/cvd/cuttlefish/host/libs/web/build_zip_name.h
```diff
@@ -0,0 +1,31 @@
...
+#include "cuttlefish/host/libs/web/android_build.h"
+
+namespace cuttlefish {
+
+/** Returns the name of one of the artifact target zip files.
+ *
+ * For example, for a target "aosp_cf_x86_phone-userdebug" at a build "5824130",
+ * the image zip file would be "aosp_cf_x86_phone-img-5824130.zip"
+ */
+std::string GetBuildZipName(const Build& build, const std::string& name);
```
cjreynol, line 29: NIT:  Feel free to ignore.

What about putting this in `build_api.h/cc` since it so tightly coupled to `Build` objects?

Databean: go/code-for-blaze#avoidstatics writes

> Warning: Avoid mixing `class` and free/`static` functions in the same `header` and `cc` files.

though this change may be taking it to an extreme level.

PR #1235, opened 2025-06-05
Introduce HO API GET /v1/userartifacts/{checksum}

0405ysj:
This PR covers the implementation of `GET /v1/userartifacts/{checksum}` at the level of HO REST API only.
Details are described at the design docuement. Its Golang client API will be covered at different PR.

Also, noting that this PR is based on commits listed at https://github.com/google/android-cuttlefish/pull/1144.

Design document: go/ho-upload-api-v1

### DIFF
--- a/frontend/src/host_orchestrator/orchestrator/controller.go
+++ b/frontend/src/host_orchestrator/orchestrator/controller.go
@@ -109,6 +109,8 @@ func (c *Controller) AddRoutes(router *mux.Router) {
 		httpHandler(&extractUserArtifactHandler{c.OperationManager, c.UserArtifactsManager})).Methods("POST")
 	router.Handle("/v1/userartifacts/{checksum}",
 		httpHandler(&createUpdateUserArtifactHandler{c.UserArtifactsManager, false})).Methods("PUT")
+	router.Handle("/v1/userartifacts/{checksum}",
+		httpHandler(&statUserArtifactHandler{c.UserArtifactsManager})).Methods("GET")
 	// Debug endpoints.
 	router.Handle("/_debug/varz", httpHandler(&getDebugVariablesHandler{c.DebugVariablesManager})).Methods("GET")
 	router.Handle("/_debug/statusz", okHandler()).Methods("GET")
@@ -607,6 +609,14 @@ func (h *createUpdateUserArtifactHandler) Handle(r *http.Request) (interface{}, 
 		}
 		return nil, h.m.UpdateArtifact(vars["checksum"], chunk)
 	}
+}
+
+type statUserArtifactHandler struct {
+	m UserArtifactsManager
+}
+
+func (h *statUserArtifactHandler) Handle(r *http.Request) (interface{}, error) {
+	return h.m.StatArtifact(mux.Vars(r)["checksum"])
 }
 
 type extractUserArtifactHandler struct {
--- a/frontend/src/host_orchestrator/orchestrator/controller.go
+++ b/frontend/src/host_orchestrator/orchestrator/controller.go
@@ -573,16 +573,22 @@ func (h *createUpdateUserArtifactHandler) Handle(r *http.Request) (interface{}, 
 			return nil, operator.NewBadRequestError("invalid directory name", err)
 		}
 		if err != nil {
-			log.Println("deprecated: use `chunk_offset_bytes`")
+			log.Println("use of deprecated `chunk_number`")
 			chunkNumberRaw := r.FormValue("chunk_number")
 			chunkNumber, err := strconv.Atoi(chunkNumberRaw)
 			if err != nil {
 				return nil, operator.NewBadRequestError(
-					fmt.Sprintf("Invalid chunk_number value: %q", chunkNumberRaw), err)
+					fmt.Sprintf("Invalid chunk_number form field value: %q", chunkNumberRaw), err)
+			}
+			if chunkNumber < 0 {
+				return nil, operator.NewBadRequestError(fmt.Sprintf("invalid value (chunk_number:%d)", chunkNumber), nil)
 			}
 			chunkSizeBytes, err := parseFormValueInt(r, "chunk_size_bytes")
 			if err != nil {
 				return nil, err
+			}
+			if chunkNumber < 0 {
+				return nil, operator.NewBadRequestError(fmt.Sprintf("invalid value (chunk_size_bytes:%d)", chunkSizeBytes), nil)
 			}
 			chunkOffsetBytes = int64(chunkNumber-1) * chunkSizeBytes
 		}
--- a/frontend/src/host_orchestrator/orchestrator/userartifacts.go
+++ b/frontend/src/host_orchestrator/orchestrator/userartifacts.go
@@ -51,6 +51,8 @@ type UserArtifactChunk struct {
 type UserArtifactsManager interface {
 	// Update artifact with the passed chunk
 	UpdateArtifact(checksum string, chunk UserArtifactChunk) error
+	// Stat artifact whether artifact with given checksum exists or not.
+	StatArtifact(checksum string) (*apiv1.StatArtifactResponse, error)
 
 	UserArtifactsDirResolver
 	// Creates a new directory for uploading user artifacts in the future.
@@ -185,6 +187,15 @@ func (m *UserArtifactsManagerImpl) UpdateArtifactWithDir(dir string, chunk UserA
 		return err
 	}
 	return nil
+}
+
+func (m *UserArtifactsManagerImpl) StatArtifact(checksum string) (*apiv1.StatArtifactResponse, error) {
+	if exists, err := dirExists(filepath.Join(m.RootDir, checksum)); err != nil {
+		return nil, fmt.Errorf("failed to check existence of directory: %w", err)
+	} else if !exists {
+		return nil, operator.NewNotFoundError(fmt.Sprintf("user artifact(checksum:%q) not found", checksum), nil)
+	}
+	return &apiv1.StatArtifactResponse{}, nil
 }
 
 func (m *UserArtifactsManagerImpl) getRWMutex(checksum string) *sync.RWMutex {
--- a/frontend/src/host_orchestrator/orchestrator/userartifacts.go
+++ b/frontend/src/host_orchestrator/orchestrator/userartifacts.go
@@ -36,7 +36,7 @@ import (
 // Resolves the user artifacts full directory.
 type UserArtifactsDirResolver interface {
 	// Given a directory name returns its full path.
-	GetDirPath(name string, isLegacy bool) string
+	GetDirPath(name string) string
 }
 
 type UserArtifactChunk struct {
@@ -135,12 +135,8 @@ func (m *UserArtifactsManagerImpl) ListDirs() (*apiv1.ListUploadDirectoriesRespo
 	return &apiv1.ListUploadDirectoriesResponse{Items: dirs}, nil
 }
 
-func (m *UserArtifactsManagerImpl) GetDirPath(name string, isLegacy bool) string {
-	if isLegacy {
-		return filepath.Join(m.LegacyRootDir, name)
-	} else {
-		return filepath.Join(m.RootDir, name)
-	}
+func (m *UserArtifactsManagerImpl) GetDirPath(name string) string {
+	return filepath.Join(m.LegacyRootDir, name)
 }
 
 func (m *UserArtifactsManagerImpl) GetFilePath(dir, filename string) string {
@@ -165,7 +161,7 @@ func (m *UserArtifactsManagerImpl) UpdateArtifact(checksum string, chunk UserArt
 		return fmt.Errorf("failed to update chunk: %w", err)
 	}
 	if mayMoveArtifact {
-		if err := m.moveArtifactIfNeeds(checksum, chunk); err != nil {
+		if err := m.moveArtifactIfNeeded(checksum, chunk); err != nil {
 			return fmt.Errorf("failed to move the user artifact from working directory: %w", err)
 		}
 	}
@@ -235,7 +231,8 @@ func writeChunk(filename string, chunk UserArtifactChunk) error {
 
 // Calculating the checksum of the user artifact is a heavy task. Instead, it records the state
 // of updated chunks to know whether it's ready to calculate checksum and move the user artifact or
-// not. This function returns the boolean value as true if it needs to calculate checksum afterwards.
+// not. This function returns the boolean value as true if the user artifact is fully updated so it
+// needs to calculate checksum later.
 func (m *UserArtifactsManagerImpl) writeChunkAndUpdateState(checksum string, chunk UserArtifactChunk) (bool, error) {
 	mu := m.getRWMutex(checksum)
 	// Reason for acquiring read lock is to allow updating multiple chunks concurrently, but to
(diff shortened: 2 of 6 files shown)


frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -522,48 +522,35 @@ func (h *createUpdateUserArtifactHandler) Handle(r *http.Request) (interface{},
...
-		chunk := UserArtifactChunk{
-			Name:             fheader.Filename,
-			File:             f,
-			UseChunkOffset:   true,
-			ChunkOffsetBytes: chunkOffsetBytes,
-		}
-		return nil, h.m.UpdateArtifactWithDir(dir, chunk)
 	} else {
-		// Deprecated: clients should use `chunk_offset_bytes` only
+		log.Println("deprecated: use `chunk_offset_bytes`")
```
jemoreira: nit: "use of deprecated `chunk_number`". The log message is not directed to the client that made the request.

frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -522,48 +522,35 @@ func (h *createUpdateUserArtifactHandler) Handle(r *http.Request) (interface{},
...
 		chunkNumberRaw := r.FormValue("chunk_number")
-		chunkTotalRaw := r.FormValue("chunk_total")
```
jemoreira: please validate that chunkNumber, chunkSize, etc are in the correct range (non-negative, >1, etc).

frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -522,48 +522,35 @@ func (h *createUpdateUserArtifactHandler) Handle(r *http.Request) (interface{},
 		return nil, err
 	}
 	chunkOffsetBytesRaw := r.FormValue("chunk_offset_bytes")
+	var chunkOffsetBytes int64
 	if chunkOffsetBytesRaw != "" {
```
jemoreira: nit: I know it's not introduced in this PR, but please encapsulate the `else` branch of this conditional into a `UserArtifactChunkFromDeprecatedRequest(*http.Request) (UserArtifactChunk, error)` and rewrite the `if` like this:
```
if chunkOffsetBytesRaw == "" {
    return UserArtifactChunkFromDeprecatedRequest(r)
}
chunkOffsetBytes, err := strconv.ParseInt(chunkOffsetBytesRaw, 10, 64)
if err != nil {
	return nil, operator.NewBadRequestError(
		fmt.Sprintf("Invalid chunk_offset_bytes form field value: %q", chunkOffsetBytesRaw), err)
}
...
```

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -142,12 +137,7 @@ func writeChunk(filename string, chunk UserArtifactChunk) error {
 		return err
 	}
 	defer f.Close()
-	offset := chunk.ChunkOffsetBytes
-	if !chunk.UseChunkOffset {
-		log.Println("deprecated: use `ChunkOffsetBytes`")
-		offset = int64(chunk.ChunkNumber-1) * chunk.ChunkSizeBytes
-	}
-	if _, err := f.Seek(offset, 0); err != nil {
+	if _, err := f.Seek(chunk.OffsetBytes, 0); err != nil {
```
jemoreira: use `os.SEEK_SET` instead of 0.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -24,60 +25,86 @@ import (
...
 	hoexec "github.com/google/android-cuttlefish/frontend/src/host_orchestrator/orchestrator/exec"
 	"github.com/google/android-cuttlefish/frontend/src/liboperator/operator"
+	"github.com/google/uuid"
 )
 
 // Resolves the user artifacts full directory.
 type UserArtifactsDirResolver interface {
 	// Given a directory name returns its full path.
-	GetDirPath(name string) string
+	GetDirPath(name string, isLegacy bool) string
```
jemoreira: Make it two functions instead:
```
GetDirPath(name string) string
GetLegacyDirPath(name string) string
```

frontend/src/host_orchestrator/orchestrator/userartifacts_test.go
```diff
@@ -369,3 +369,46 @@ func TestUpdateArtifactWithMultipleParallelChunkSucceeds(t *testing.T) {
 		t.Fatalf("artifact content mismatch (-want +got):\n%s", diff)
 	}
 }
+
+func TestStatArtifactFailsWithoutUpdateArtifact(t *testing.T) {
```
ser-io: nit: use name: `TestStatArtifactFailsArtifactNotFound`, it's gives a more general description of what this test is about.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -131,23 +187,126 @@ func (m *UserArtifactsManagerImpl) UpdateArtifactWithDir(dir string, chunk UserA
...
+	if err := createDir(m.commonWorkDir); err != nil {
+		return false, fmt.Errorf("failed to create working directory for all Host Orchestrators: %w", err)
+	}
+	if err := createDir(m.UuidWorkDir); err != nil {
+		return false, fmt.Errorf("failed to create working directory for this Host Orchestrator: %w", err)
+	}
+	workDir := filepath.Join(m.UuidWorkDir, checksum)
+	if err := createDir(workDir); err != nil {
+		return false, fmt.Errorf("failed to create working directory: %w", err)
+	}
```
jemoreira: These don't need to be protected by a mutex.

frontend/src/host_orchestrator/orchestrator/userartifacts_test.go
```diff
@@ -369,3 +369,46 @@ func TestUpdateArtifactWithMultipleParallelChunkSucceeds(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if _, err := uam.StatArtifact("foo"); err == nil {
+		t.Fatal("Expected an error")
+	}
+}
+
+func TestStatArtifactSucceedsWithUpdateArtifact(t *testing.T) {
```
ser-io: nit: name `TestStatArtifactSucceeds` is good enough. You don't need to describe how the artifact is there in the test name.

frontend/src/host_orchestrator/orchestrator/userartifacts_test.go
```diff
@@ -369,3 +369,46 @@ func TestUpdateArtifactWithMultipleParallelChunkSucceeds(t *testing.T) {
...
+	legacyRootDir := orchtesting.TempDir(t)
+	defer orchtesting.RemoveDir(t, legacyRootDir)
+	rootDir := orchtesting.TempDir(t)
+	defer orchtesting.RemoveDir(t, rootDir)
+	opts := UserArtifactsManagerOpts{LegacyRootDir: legacyRootDir, RootDir: rootDir}
+	uam, err := NewUserArtifactsManagerImpl(opts)
+	if err != nil {
+		t.Fatal(err)
+	}
+
```
ser-io: nit: remove empty line.

frontend/src/host_orchestrator/orchestrator/userartifacts_test.go
```diff
@@ -369,3 +369,46 @@ func TestUpdateArtifactWithMultipleParallelChunkSucceeds(t *testing.T) {
...
+		Name:          testFileName,
+		OffsetBytes:   0,
+		SizeBytes:     int64(len(testFileData)),
+		FileSizeBytes: int64(len(testFileData)),
+		File:          strings.NewReader(testFileData),
+	}
+	if err := uam.UpdateArtifact(checksum, chunk); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := uam.StatArtifact(checksum); err != nil {
```
ser-io, line 411: nit: add empty line about `StatArtifact` having a clear separation of the key method that's tested in this test.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -131,23 +187,126 @@ func (m *UserArtifactsManagerImpl) UpdateArtifactWithDir(dir string, chunk UserA
...
-		return err
+	if _, err := io.Copy(f, chunk.File); err != nil {
+		return fmt.Errorf("failed to copy chunk into working file: %w", err)
+	}
+	return nil
+}
+
+// Calculating the checksum of the user artifact is a heavy task. Instead, it records the state
+// of updated chunks to know whether it's ready to calculate checksum and move the user artifact or
+// not. This function returns the boolean value as true if it needs to calculate checksum afterwards.
```
jemoreira: nit: Rewrite this comment to describe the function's return value in terms of whether "the file has been fully written to disk" or "is completely available". Basically make it about the state of the file, not what will be done to it later.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -131,23 +187,126 @@ func (m *UserArtifactsManagerImpl) UpdateArtifactWithDir(dir string, chunk UserA
...
+		return false, nil
+	}
+	h := sha256.New()
+	if _, err := io.Copy(h, f); err != nil {
+		return false, err
+	}
+	return checksum == fmt.Sprintf("%x", h.Sum(nil)), nil
+}
+
+func (m *UserArtifactsManagerImpl) moveArtifactIfNeeds(checksum string, chunk UserArtifactChunk) error {
```
jemoreira: typo: `moveArtifactIfNeeded`

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -131,23 +187,126 @@ func (m *UserArtifactsManagerImpl) UpdateArtifactWithDir(dir string, chunk UserA
...
+	} else if !matches {
+		// Only if the checksum matches, it's ready to move the user artifact from working
+		// directory.
+		return nil
+	}
+	workDir := filepath.Join(m.UuidWorkDir, checksum)
+	if err := os.Rename(workDir, dir); err != nil {
+		return fmt.Errorf("failed to move the user artifact: %w", err)
 	}
+	m.chunkStates.LoadAndDelete(checksum)
```
jemoreira: You should remove `mu` from its map too. To make it in a thread safe manner you must remove it while holding the write lock AND you should check that `dir` doesn't yet exist every time any type of lock is acquired on it and remove it again if `dir` exists.

frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -501,12 +505,20 @@ func (h *listUploadDirectoriesHandler) Handle(r *http.Request) (interface{}, err
 }
 
 type createUpdateUserArtifactHandler struct {
-	m UserArtifactsManager
+	m        UserArtifactsManager
+	isLegacy bool
```
jemoreira: Where is this variable ever written to? I see it read in line 538 but never written, am I missing something?

0405ysj: That's used at line 106 and 110, where route endpoints with constructing `createUpdateUserArtifactHandler`.

frontend/src/host_orchestrator/orchestrator/userartifacts_test.go
0405ysj: Done.

frontend/src/host_orchestrator/orchestrator/userartifacts_test.go
0405ysj: Done.

frontend/src/host_orchestrator/orchestrator/userartifacts_test.go
0405ysj: Done.

frontend/src/host_orchestrator/orchestrator/userartifacts_test.go
0405ysj: Done.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
0405ysj: Current code uses `io.SeekStart` for it.

frontend/src/host_orchestrator/orchestrator/controller.go
0405ysj: Done.

frontend/src/host_orchestrator/orchestrator/controller.go
0405ysj: Done.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
0405ysj: Done.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
0405ysj: Done.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
0405ysj: Get rid of `isLegacy` field instead, since `isLegacy=false` is only used in testing `UserArtifactsManager`. I think we can introduce it later when it needs.

frontend/src/host_orchestrator/orchestrator/controller.go
0405ysj: It seems that's already outdated with https://github.com/google/android-cuttlefish/pull/1144. For current logic, it looks better without introducing new function for readability I think. Correct me if I'm wrong.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
0405ysj: It sounds for me as introducing double checked locking mechanism? Unless checking dir existence before trying to acquire any read & lock, mutex will be used forever.

BTW, I'd solve this comment in the later PR, since this comment is not related to this commit.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
0405ysj: With [another comment](https://github.com/google/android-cuttlefish/pull/1235#discussion_r2130487512) below, it sounds for me as introducing double checked locking mechanism. Creating directories doesn't require any locks though, HO wants to check dir existence in advance. Of course, checking dir existence should be done right after acquiring read & write lock at least. So, code below represents your request?

```
if exists, err := dirExists(filepath.Join(m.RootDir, checksum)); err != nil {
	return false, fmt.Errorf("failed to check existence of directory: %w", err)
} else if exists {
	return false, operator.NewConflictError(fmt.Sprintf("user artifact(checksum:%q) already exists", checksum), nil)
}
if err := createDir(m.RootDir); err != nil {
	return false, fmt.Errorf("failed to create root directory for storing user artifacts: %w", err)
}
if err := createDir(m.commonWorkDir); err != nil {
	return false, fmt.Errorf("failed to create working directory for all Host Orchestrators: %w", err)
}
if err := createDir(m.UuidWorkDir); err != nil {
	return false, fmt.Errorf("failed to create working directory for this Host Orchestrator: %w", err)
}
workDir := filepath.Join(m.UuidWorkDir, checksum)
if err := createDir(workDir); err != nil {
	return false, fmt.Errorf("failed to create working directory: %w", err)
}

mu := m.getRWMutex(checksum)
// Reason for acquiring read lock is to allow updating multiple chunks concurrently, but to
// restrict validating checksum or moving artifact into different directory when one or more
// chunks are being updated.
mu.RLock()
defer mu.RUnlock()

if exists, err := dirExists(filepath.Join(m.RootDir, checksum)); err != nil {
	return false, fmt.Errorf("failed to check existence of directory: %w", err)
} else if exists {
	return false, operator.NewConflictError(fmt.Sprintf("user artifact(checksum:%q) already exists", checksum), nil)
}
workFile := filepath.Join(workDir, chunk.Name)
if err := writeChunk(workFile, chunk); err != nil {
	return false, err
}
```

BTW, I'd solve this comment in the later PR, since this comment is not related to this commit.

PR #1237, opened 2025-06-05
Make UrlEscape a free function and not part of HttpClient

Databean:
It doesn't need the `CURL*` argument.

Bug: b/422588200


cjreynol (approved):

Databean:
Thanks for reviewing!

PR #1242, opened 2025-06-05
Remove lifetime management from `*BuildApi` classes, and split up `cvd/fetch` targets

Databean:
Initialization logic is extracted from `fetch_cvd.cc`.

Bug: b/422820430, b/422818275


Databean:
Thanks for reviewing!

PR #1243, opened 2025-06-05
Build and package mtools (mmd, mcopy)

adelva1984:
Bug: 402292564


base/cvd/cuttlefish/host/libs/config/esp.cpp
```diff
@@ -165,9 +165,8 @@ static bool MsdosMakeDirectories(const std::string& image_path,
 
 static bool CopyToMsdos(const std::string& image, const std::string& path,
                         const std::string& destination) {
-  const auto mcopy = HostBinaryPath("mcopy");
   const auto success =
-      Execute({mcopy, "-o", "-i", image, "-s", path, destination});
+      Execute({McopyBinary(), "-o", "-i", image, "-s", path, destination});
```
Databean, line 169: Can this also get an equivalent to ag/33922290?

Databean (approved):

base/cvd/cuttlefish/host/libs/config/esp.cpp
adelva1984: This time, it's already there, we don't need to add it. Because when I added mtools to AOSP, I created host symlinks in the Android.bp, so they're in the host package already. The other binaries didn't have "host symlinks", so we needed to add them.

PR #1247, opened 2025-06-06
Split interface for HttpClient

Databean:
https://github.com/google/android-cuttlefish/blob/22d461205c0eaf9806801b04c147096e32e09603/base/cvd/cuttlefish/host/libs/web/http_client/http_client.cc

The only novel behavior between implementations is in the `DownloadToCallback` implementation. It'll be easier to add a fake version for testing with a smaller interface.

Also, the behavior around files and json adds more dependencies that not all users may need.

Bug: b/422866225
Bug: b/422877302


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #1250, opened 2025-06-09
[libhoclient] upload artifacts

ser-io:


ser-io:
Reusing work done in https://github.com/google/android-cuttlefish/pull/1165

frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -513,6 +520,40 @@ func (c *HostOrchestratorClientImpl) UploadFile(uploadDir string, filename strin
...
+	}
+	return fmt.Sprintf("%x", h.Sum(nil)), nil
+}
+
+func (c *HostOrchestratorClientImpl) UploadArtifact(filename string) error {
+	checksum, err := SHA256Checksum(filename)
+	if err != nil {
+		return err
+	}
+	return c.upload("/v1/userartifacts/"+checksum, filename, DefaultUploadOptions())
```
0405ysj: `fmt.Sprintf("/v1/userartifacts/%s", checksum)`?

frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -513,6 +520,40 @@ func (c *HostOrchestratorClientImpl) UploadFile(uploadDir string, filename strin
 	return c.UploadFileWithOptions(uploadDir, filename, DefaultUploadOptions())
 }
 
+func (c *HostOrchestratorClientImpl) UploadFileWithOptions(uploadDir string, filename string, uploadOpts UploadOptions) error {
+	return c.upload("/userartifacts/"+uploadDir, filename, uploadOpts)
```
0405ysj, line 522: `fmt.Sprintf("/userartifacts/%s", uploadDir)`?

frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -103,6 +105,11 @@ type InstancesClient interface {
 
 // Manage artifacts created by the user.
 type UserArtifactsClient interface {
+	// Upload artifact into the artifacts repository.
+	// Artifacts are identified by their SHA256 checksum in the artifacts repository
+	UploadArtifact(filename string) error
+	// Stat artifacts in the artifacts repository.
+	StatArtifact(checksum string) (*hoapi.StatArtifactResponse, error)
```
0405ysj: I think `StatArtifact(checksum string)` is not enough, since there's no provided way for calculating checksum for given filename. IMO, either introducing `StatArtifact(filename string)` instead or introducing `CalculateChecksum(filename string)` is required. Among those, I prefer the latter option since it can reuse checksum calculation result when calling `StatArtifact` multiple times.

frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -513,6 +520,40 @@ func (c *HostOrchestratorClientImpl) UploadFile(uploadDir string, filename strin
 	return c.UploadFileWithOptions(uploadDir, filename, DefaultUploadOptions())
 }
 
+func (c *HostOrchestratorClientImpl) UploadFileWithOptions(uploadDir string, filename string, uploadOpts UploadOptions) error {
+	return c.upload("/userartifacts/"+uploadDir, filename, uploadOpts)
+}
+
+func SHA256Checksum(filename string) (string, error) {
```
0405ysj: As I said above, what about  adding this function into `UserArtifactsClient` interface?

frontend/src/libhoclient/httputils.go
```diff
@@ -363,7 +366,8 @@ func (u *FilesUploader) startWorkers(ctx context.Context, jobsChan <-chan upload
 }
 
 type uploadChunkJob struct {
-	Filename string
+	Filename      string
+	FileSizeBytes int64
 	// A number between 1 and `TotalChunks`. The n-th chunk represents a segment of data within the file with size
 	// `ChunkSizeBytes` starting the `(n-1) * ChunkSizeBytes`-th byte.
 	ChunkNumber    int
```
0405ysj, line 374: Why don't we use offset based field instead of `ChunkNumber` and `TotalChunks`?

0405ysj:
Though it's not the scope of this PR, are you willing to implement `libhoclient` code for HO extract API as well? I'm having a plan to implement that after this PR being merged, I want to know your plan before working on it. Anyway, thanks for your effort on libhoclient code!

frontend/src/libhoclient/host_orchestrator_client.go
ser-io: This is a very trivial concatenation, don't see any benefit for using `fmt.Sprintf` here.

frontend/src/libhoclient/host_orchestrator_client.go
ser-io: ditto.

frontend/src/libhoclient/httputils.go
ser-io: `offset` value is calculated in https://github.com/google/android-cuttlefish/blob/09d35e75f28d6e836310401b4fc5de78fbaae147/frontend/src/libhoclient/httputils.go#L484 based on `ChunkNumber` value. This `struct` could be changed to be created initially with the the offset value, but it should be on a different PR.

frontend/src/libhoclient/host_orchestrator_client.go
ser-io: `StatArtifact` method aligns with public API endpoint `GET /v1/userartifacts/{checksum}`, in this endpoint the only parameter is the SHA256 checksum value of the given artifact we want to stat.

>since there's no provided way for calculating checksum for given filename.

In the new upload API checksums are calculated using `SHA-256` hash function.

frontend/src/libhoclient/host_orchestrator_client.go
ser-io: This is fixed behavior, the server expects SHA256 checksum values.

ser-io:
> Though it's not the scope of this PR, are you willing to implement `libhoclient` code for HO extract API as well? I'm having a plan to implement that after this PR being merged, I want to know your plan before working on it. Anyway, thanks for your effort on libhoclient code!

I won't be implementing the modifications needed for the new extract API. You can go ahead and implement it.

frontend/src/libhoclient/host_orchestrator_client.go
0405ysj: Then, could you please unite the way into either `fmt.Sprintf("/userartifacts/%s", uploadDir)` or `"/userartifacts/"+uploadDir`? I can see `fmt.Sprintf` is used at `StatArtifact`.

frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -513,6 +520,40 @@ func (c *HostOrchestratorClientImpl) UploadFile(uploadDir string, filename strin
...
+	checksum, err := SHA256Checksum(filename)
+	if err != nil {
+		return err
+	}
+	return c.upload("/v1/userartifacts/"+checksum, filename, DefaultUploadOptions())
+}
+
+func (c *HostOrchestratorClientImpl) StatArtifact(checksum string) (*hoapi.StatArtifactResponse, error) {
+	res := &hoapi.StatArtifactResponse{}
+	path := fmt.Sprintf("/v1/userartifacts/%s", checksum)
```
0405ysj: Here uses `fmt.Sprintf("/v1/userartifacts/%s", checksum)`.

frontend/src/libhoclient/host_orchestrator_client.go
0405ysj: Please resolve this comment after choosing either way, using `fmt.Sprintf` or just concatenation.

frontend/src/libhoclient/httputils.go
0405ysj: SG. Agree with this doesn't need to be on this PR.

frontend/src/libhoclient/host_orchestrator_client.go
0405ysj: Agree with `StatArtifact` is aligned with `GET /v1/userartifacts/{checksum}`, but I think `StatArtifact(checksum string)` is not aligned with `UploadArtifact(filename string)`. I speculate user usually wants to call `StatArtifact` right after or before calling `UploadArtifact`, then users should study what `checksum` represents and how to get it. That's why I suggested some ways above.

If you still don't prefer to modify the interface, could you add a comment right above of `StatArtifact`, about what `checksum` represents at least? I know it's written at `UploadArtifact`, but I think explaining at `StatArtifact` is better.

frontend/src/libhoclient/host_orchestrator_client.go
0405ysj: Please feel free to resolve this comment, depending on the state of another comment [above](https://github.com/google/android-cuttlefish/pull/1250/files/147635d9c4a4fe9a13ec2ec69967c7e72ca69a17#r2136687544).

0405ysj:
Any updates?

frontend/src/libhoclient/host_orchestrator_client.go
ser-io: used concatenation

frontend/src/libhoclient/host_orchestrator_client.go
ser-io: used concatenation

frontend/src/libhoclient/host_orchestrator_client.go
ser-io: I see your point. I removed `StatArtifact(checksum string)`, let's to the public interface when it's needed. For now, I'm just adding `UploadArtifact(filename string)`.

e2etests/orchestration/upload_artifact_test/main_test.go
```diff
@@ -0,0 +1,72 @@
...
+			uploadErr = err
+		}
+		uploadDuration = time.Since(start)
+		log.Printf("upload duration: %d ms", uploadDuration.Milliseconds())
+	}()
+	// upload a second time before the first upload is completed
+	const secondRunnerWaitMs = 500
+	time.Sleep(secondRunnerWaitMs * time.Millisecond)
+	if err := client.UploadArtifact("../artifacts/images.zip"); err != nil {
+		log.Fatal(err)
```
0405ysj: nit: `t.Fatal(err)`

e2etests/orchestration/upload_artifact_test/main_test.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	wg.Wait()
+	if uploadErr != nil {
+		log.Fatal(uploadErr)
```
0405ysj: nit: `t.Fatal(err)`

e2etests/orchestration/upload_artifact_test/main_test.go
```diff
@@ -0,0 +1,72 @@
...
+	client := hoclient.NewHostOrchestratorClient(baseURL)
+	t.Cleanup(func() {
+		if err := common.CollectHOLogs(baseURL); err != nil {
+			log.Printf("failed to collect HO logs: %s", err)
+		}
+	})
+	var uploadDuration time.Duration
+	var uploadErr error
+	wg := sync.WaitGroup{}
+	wg.Add(1)
```
0405ysj: What about using channel instead of `sync.WaitGroup` if we still use single goroutine?

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -159,6 +159,9 @@ func (m *UserArtifactsManagerImpl) UpdateArtifact(checksum string, chunk UserArt
 	}
 	mayMoveArtifact, err := m.writeChunkAndUpdateState(checksum, chunk)
 	if err != nil {
+		if _, ok := err.(*operator.AppError); ok {
+			return err
+		}
```
0405ysj, line 164: Thanks! I missed that.

e2etests/orchestration/upload_artifact_test/main_test.go
```diff
@@ -0,0 +1,72 @@
...
+		start := time.Now()
+		if err := client.UploadArtifact("../artifacts/images.zip"); err != nil {
+			uploadErr = err
+		}
+		uploadDuration = time.Since(start)
+		log.Printf("upload duration: %d ms", uploadDuration.Milliseconds())
+	}()
+	// upload a second time before the first upload is completed
+	const secondRunnerWaitMs = 500
+	time.Sleep(secondRunnerWaitMs * time.Millisecond)
```
0405ysj: I think waiting for the fixed time(500ms) can lead this test as flaky. I speculate second runner can be completed earlier than the first runner.

Instead, what about running 2 runners simultaneously, and validate if only one of 2 runners returns an error? For the third runner, I think it can wait for both 2 previous runners.

0405ysj (commented):
I might miss your change written in `e2etests/orchestration/upload_artifact_test/main_test.go` before, sorry for that. Except E2E test, PR looks good to me.

e2etests/orchestration/upload_artifact_test/main_test.go
ser-io: Removed this part

e2etests/orchestration/upload_artifact_test/main_test.go
ser-io: Removed this part

e2etests/orchestration/upload_artifact_test/main_test.go
ser-io: Running 2 runners simultaneously didn't recreate the scenario I wanted deterministically, which was hitting 409 from the server. I moved the scenarios I wanted to test runner 2 and runner 3 to unit tests and removed the e2e test as is no longer relevant.

0405ysj: What about reviving E2E test with a single runner? I saw your HO client unit tests, and there's no testing scene for actual uploading with status 200. I think that's what we can do with E2E test, HDYT?

ser-io: 200 status on e2e tests would be covered when implementing the existing  `create_from_images_zip_test`  counterpart (https://github.com/google/android-cuttlefish/blob/main/e2etests/orchestration/create_from_images_zip_test/main_test.go) using the new API.

0405ysj: That E2E test is for testing legacy APIs, not  `PUT /v1/userartifacts/{checksum}`.

ser-io: Yes, it is. A **new test will be implemented**, similar to that one but using the new APIs. That's how we would cover 200 status.

0405ysj: Okay, then please implement that in the following PR or let me know if I should do.

0405ysj (approved):

e2etests/orchestration/upload_artifact_test/main_test.go
ser-io: The test would be implemented when we have implemented the ability to create instances from uploaded artifacts.

PR #1253, opened 2025-06-10
Introduce HO REST API `POST /v1/userartifacts/{checksum}/:extract`

0405ysj:
Details around this PR is at the design document(go/ho-api-extract).

This PR covers implementation of REST API of Host Orchestrator only.
Golang client API for this REST API is out of scope.

### DIFF
--- a/frontend/src/host_orchestrator/orchestrator/controller.go
+++ b/frontend/src/host_orchestrator/orchestrator/controller.go
@@ -104,13 +104,15 @@ func (c *Controller) AddRoutes(router *mux.Router) {
 	router.Handle("/userartifacts",
 		httpHandler(&listUploadDirectoriesHandler{c.UserArtifactsManager})).Methods("GET")
 	router.Handle("/userartifacts/{name}",
-		httpHandler(&createUpdateUserArtifactHandler{c.UserArtifactsManager, true})).Methods("PUT")
-	router.Handle("/userartifacts/{dir}/{name}/:extract",
-		httpHandler(&extractUserArtifactHandler{c.OperationManager, c.UserArtifactsManager})).Methods("POST")
-	router.Handle("/v1/userartifacts/{checksum}",
 		httpHandler(&createUpdateUserArtifactHandler{c.UserArtifactsManager, false})).Methods("PUT")
+	router.Handle("/userartifacts/{dir}/{name}/:extract",
+		httpHandler(&extractUserArtifactHandler{c.OperationManager, c.UserArtifactsManager, false})).Methods("POST")
+	router.Handle("/v1/userartifacts/{checksum}",
+		httpHandler(&createUpdateUserArtifactHandler{c.UserArtifactsManager, true})).Methods("PUT")
 	router.Handle("/v1/userartifacts/{checksum}",
 		httpHandler(&statUserArtifactHandler{c.UserArtifactsManager})).Methods("GET")
+	router.Handle("/v1/userartifacts/{checksum}/:extract",
+		httpHandler(&extractUserArtifactHandler{c.OperationManager, c.UserArtifactsManager, true})).Methods("POST")
 	// Debug endpoints.
 	router.Handle("/_debug/varz", httpHandler(&getDebugVariablesHandler{c.DebugVariablesManager})).Methods("GET")
 	router.Handle("/_debug/statusz", okHandler()).Methods("GET")
@@ -591,8 +593,8 @@ func (h *listUploadDirectoriesHandler) Handle(r *http.Request) (interface{}, err
 }
 
 type createUpdateUserArtifactHandler struct {
-	m        UserArtifactsManager
-	isLegacy bool
+	m    UserArtifactsManager
+	isV1 bool
 }
 
 func parseFormValueInt(r *http.Request, key string) (int64, error) {
@@ -622,7 +624,23 @@ func (h *createUpdateUserArtifactHandler) Handle(r *http.Request) (interface{}, 
 	}
 	defer f.Close()
 	chunkOffsetBytes, err := parseFormValueInt(r, "chunk_offset_bytes")
-	if h.isLegacy {
+	if h.isV1 {
+		if err != nil {
+			return nil, err
+		}
+		fileSizeBytes, err := parseFormValueInt(r, "file_size_bytes")
+		if err != nil {
+			return nil, err
+		}
+		chunk := UserArtifactChunk{
+			File:          f,
+			Name:          fheader.Filename,
+			OffsetBytes:   chunkOffsetBytes,
+			SizeBytes:     fheader.Size,
+			FileSizeBytes: fileSizeBytes,
+		}
+		return nil, h.m.UpdateArtifact(vars["checksum"], chunk)
+	} else {
 		dir := vars["name"]
 		if err := ValidateFileName(dir); err != nil {
 			return nil, operator.NewBadRequestError("invalid directory name", err)
@@ -653,22 +671,6 @@ func (h *createUpdateUserArtifactHandler) Handle(r *http.Request) (interface{}, 
 			OffsetBytes: chunkOffsetBytes,
 		}
 		return nil, h.m.UpdateArtifactWithDir(dir, chunk)
-	} else {
-		if err != nil {
-			return nil, err
-		}
-		fileSizeBytes, err := parseFormValueInt(r, "file_size_bytes")
-		if err != nil {
-			return nil, err
-		}
-		chunk := UserArtifactChunk{
-			File:          f,
-			Name:          fheader.Filename,
-			OffsetBytes:   chunkOffsetBytes,
-			SizeBytes:     fheader.Size,
-			FileSizeBytes: fileSizeBytes,
-		}
-		return nil, h.m.UpdateArtifact(vars["checksum"], chunk)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -626,23 +628,29 @@ func (h *statUserArtifactHandler) Handle(r *http.Request) (interface{}, error) {
 }
 
 type extractUserArtifactHandler struct {
-	om  OperationManager
-	uam UserArtifactsManager
+	om       OperationManager
+	uam      UserArtifactsManager
+	isLegacy bool
```
ser-io: Use field `isV1 bool` which is aligned with the url path value and you don't have to modify the `userartifacts/{dir}/{name}/:extract` handler code.

frontend/src/host_orchestrator/orchestrator/userartifacts_test.go
```diff
@@ -414,6 +416,165 @@ func TestStatArtifactSucceeds(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	data, err := ioutil.ReadFile("test_data/test.zip")
+	if err != nil {
+		t.Fatal(err)
+	}
+	expectedDir := orchtesting.TempDir(t)
+	defer orchtesting.RemoveDir(t, expectedDir)
+	if err := exec.Command("unzip", "test_data/test.zip", "-d", expectedDir).Run(); err != nil {
```
ser-io: Depending on the `unzip` tool makes this unit test non-hermetic, this will make the test fail when run in a different environment. 

You can avoid using the tool by verifying the expected extracted files directly as the input is know. It's also important, for security reasons, that the content of the compressed files is displayed in the test code itself, so we know what's inside of those compressed files.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -307,6 +335,37 @@ func (m *UserArtifactsManagerImpl) moveArtifactIfNeeded(checksum string, chunk U
 	return nil
 }
 
+func (m *UserArtifactsManagerImpl) findUpdatedFile(checksum string) (string, error) {
```
ser-io: Rename to `filepath(checksum string) (string, error)` having a clear name. I don't know what's an `UpdateFile` in `findUpdateFile` or in the context of the userattifacts.go  in general.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -198,6 +201,31 @@ func (m *UserArtifactsManagerImpl) StatArtifact(checksum string) (*apiv1.StatArt
 	return &apiv1.StatArtifactResponse{}, nil
 }
 
+func (m *UserArtifactsManagerImpl) ExtractArtifact(checksum string) error {
+	dirName := fmt.Sprintf("%s_extracted", checksum)
```
ser-io: Use the `%s_extracted` directory name only at the end when moving from the working directory in `if err := os.Rename(workDir, dir);` use a ioutil.TempDir(m.WorkDir, "") to create the directory to extract the artifacts into.

0405ysj: There's no code for verifying `checksum`. If you're asking about dir existence check, it's still needed for avoiding performing extraction once again.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
0405ysj: Renamed into `getFilePath`.

frontend/src/host_orchestrator/orchestrator/userartifacts.go
0405ysj: Done.

frontend/src/host_orchestrator/orchestrator/userartifacts_test.go
0405ysj: Removed dependency on `unzip` and `tar` tool. But I kept dependency on `diff` tool for testing.

frontend/src/host_orchestrator/orchestrator/controller.go
0405ysj: Modified the field to use isV1 instead. Also, modified `createUpdateUserArtifactHnadler` as well for uniting the style.

0405ysj:
Converting this PR as draft until it passes E2E test. I'm facing at this error, and currently checking culprits.

```
Executing tests from //orchestration/upload_file_test:upload_file_test_test
-----------------------------------------------------------------------------
--- FAIL: TestUploadFileUseChunkNumber (5.05s)
    main_test.go:94: open /var/lib/cuttlefish-common/user_artifacts/3180486197/cvd-host_package.tar.gz: permission denied

```

frontend/src/host_orchestrator/orchestrator/userartifacts_test.go
```diff
@@ -414,6 +416,155 @@ func TestStatArtifactSucceeds(t *testing.T) {
...
+		File:          bytes.NewReader(data),
+	}
+	if err := uam.UpdateArtifact(checksum, chunk); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := uam.ExtractArtifact(checksum); err != nil {
+		t.Fatal(err)
+	}
+	cmd := exec.Command("diff", "-r", "test_data/zip_extracted", filepath.Join(rootDir, fmt.Sprintf("%s_extracted", checksum)))
```
ser-io, line 448: There's no need of relying on an external tool like `diff` here. 

Make it simple by compressing only one file and then verify the content of that single file programatically using  "github.com/google/go-cmp/cmp" package.  Delete  the `zip_extracted` and `tar_gz_extracted` directories.

If you want to add more files, verify its existence and content programatically making the test hermetic.

ser-io (changes requested):

frontend/src/host_orchestrator/orchestrator/userartifacts_test.go
ser-io: Wait for the reviewer to approve resolve the conversation when the change is no trivial.

There's no need of relying on an external tool like diff here.

Make it simple by compressing only one file and then verify the content of that single file programatically using "github.com/google/go-cmp/cmp" package. Delete the zip_extracted and tar_gz_extracted directories.

If you want to add more files, verify its existence and content programatically making the test hermetic.

0405ysj: > Wait for the reviewer to approve resolve the conversation when the change is no trivial.

Sorry for that. I thought your request was about getting rid of `unzip`, not about `tar` or `diff`. That's why I resolved it.

> Make it simple by compressing only one file and then verify the content of that single file programatically using "github.com/google/go-cmp/cmp" package.

Can we say `unzip` or `untar` is safely maintained, with only single file inside? For instance, `untar` at `userartifacts.go` assumes nested dirs may exist, but not with `unzip`.

> Delete the zip_extracted and tar_gz_extracted directories.

If I delete those directories, there should be new impl written in Golang for zip/tar.gz compression. I wonder if implementing and maintaining those has sufficient benefit, comparing with introducing dirs like `zip_extracted`.

Also, I want to know if you prefer to remove compressed files or not.

> There's no need of relying on an external tool like diff here.

For similar reason, I wonder if getting rid of `diff` has sufficient benefits for us since there should be new impl for dir comparison. Besides there are already some other commands for running [tests](https://github.com/google/android-cuttlefish/blob/88238f712ba24ad06819b0cddbc158701bf1d88c/frontend/src/host_orchestrator/orchestrator/listcvdsaction_test.go#L75).

Could you share your thoughts? @ser-io

ser-io: You know the content of the compressed files before hand so you can have the following using "github.com/google/go-cmp/cmp" package.

```
expectedNames := []string{"alpha.txt", "bravo_dir/bravo.txt"}
expectedContents := map[string]string{
		"alpha.txt":     "This is alpha.",
		"bravo_dir/bravo.txt": "This is bravo",
}


got :=  ..  // read the extracted dir filenames and sort them if relevant

if diff := cmp.Diff(expectedNames, got); diff != "" {
      t.Errorf("names mismatch (-want +got):\n%s", diff)
}

got :=  ..  // read the files content
if diff := cmp.Diff(expectedContents, got); diff != "" {
      t.Errorf("content mismatch (-want +got):\n%s", diff)
}
```

There's no need to depend on an external tool like `diff` when we can code our verification in this specific scenario.

0405ysj: Got rid of `diff`, after the conversation few hours ago. Done, and resolving this comment since it seems as a trivial change for now.

frontend/src/host_orchestrator/orchestrator/userartifacts_test.go
```diff
@@ -414,6 +415,191 @@ func TestStatArtifactSucceeds(t *testing.T) {
 	}
 }
 
+func getContents(dir string) (map[string]string, error) {
```
ser-io, line 418: nit: move this helper function to the end of the file.

ser-io (approved):

frontend/src/host_orchestrator/orchestrator/userartifacts_test.go
```diff
@@ -414,6 +415,191 @@ func TestStatArtifactSucceeds(t *testing.T) {
...
+	expected := map[string]string{
+		"alpha.txt":   "This is alpha.\n",
+		"bravo.txt":   "This is bravo.\n",
+		"charlie.txt": "This is charlie.\n",
+		"delta.txt":   "This is delta.\n",
+	}
+	if got, err := getContents(filepath.Join(rootDir, fmt.Sprintf("%s_extracted", checksum))); err != nil {
+		t.Fatal(err)
+	} else if diff := cmp.Diff(expected, got); diff != "" {
+		t.Fatalf("chunk state mismatch (-want +got):\n%s", diff)
```
ser-io, line 480: update error message: `extracted zip mismatch`.

frontend/src/host_orchestrator/orchestrator/userartifacts_test.go
```diff
@@ -414,6 +415,191 @@ func TestStatArtifactSucceeds(t *testing.T) {
...
+	expected := map[string]string{
+		"foo/alpha.txt":   "This is alpha.\n",
+		"foo/bravo.txt":   "This is bravo.\n",
+		"bar/charlie.txt": "This is charlie.\n",
+		"delta.txt":       "This is delta.\n",
+	}
+	if got, err := getContents(filepath.Join(rootDir, fmt.Sprintf("%s_extracted", checksum))); err != nil {
+		t.Fatal(err)
+	} else if diff := cmp.Diff(expected, got); diff != "" {
+		t.Fatalf("chunk state mismatch (-want +got):\n%s", diff)
```
ser-io, line 522: update error message

frontend/src/host_orchestrator/orchestrator/userartifacts_test.go
0405ysj: Is it about moving private functions at the end? There are some other helper functions in this file.
- getSha256Sum(data []byte) string
- constructSeparatedChunks(name, data string) []UserArtifactChunk
- getChunkStateItemList(cs *chunkState) []chunkStateItem

IMO, all helper functions should be considered together. If you let me know your style, I'd apply it for the later PR.

ser-io: When reading a test file I expect to to see all the Test* functions next to each other avoiding anything between them.

0405ysj: Acknowledged. I'd work on it after https://github.com/google/android-cuttlefish/pull/1277 is being merged.

PR #1254, opened 2025-06-10
Add a `FakeHttpClient` implementation for tests, and its own test

Databean:
Bug: b/423756056


Databean:
Thanks for reviewing!

PR #1257, opened 2025-06-10
Swap from remaining `aosp-main` builds

cjreynol:
Use `aosp-android-latest-release` instead, which is now more updated.

Bug: 408034504

### DIFF
--- a/e2etests/orchestration/bugreport_test/main_test.go
+++ b/e2etests/orchestration/bugreport_test/main_test.go
@@ -116,7 +116,7 @@ func assertAdbBugReportIsIncluded(filename string) error {
 		return err
 	}
 	defer r.Close()
-	re, err := regexp.Compile(`^bugreport-aosp_cf_x86_64_phone-MAIN\..*\.zip$`)
+	re, err := regexp.Compile(`^bugreport-aosp_cf_x86_64_only_phone.*\.zip$`)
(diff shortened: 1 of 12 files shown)


e2etests/orchestration/bugreport_test/main_test.go
```diff
@@ -116,7 +116,7 @@ func assertAdbBugReportIsIncluded(filename string) error {
 		return err
 	}
 	defer r.Close()
-	re, err := regexp.Compile(`^bugreport-aosp_cf_x86_64_phone-MAIN\..*\.zip$`)
+	re, err := regexp.Compile(`^bugreport-aosp_cf_x86_64_only_phone-MAIN\..*\.zip$`)
```
ser-io: Using the new target `bugreport-aosp_cf_x86_64_only_phone-BP2A.250605.031.A2.12bafa40-2025-06-13-14-28-11.zip  ` is an example of the bugreport name I'm getting. 

Let's remove the `-MAIN` part from the regex then.

ser-io (approved):

e2etests/orchestration/bugreport_test/main_test.go
cjreynol: Done, thanks for looking into that!

PR #1263, opened 2025-06-11
Package cf_vhost_user_input and fix its tests

jemoreira:


base/cvd/cuttlefish/host/commands/vhost_user_input/src/inherited_fd.rs
```diff
@@ -154,120 +154,89 @@ mod test {
...
     #[test]
-    fn happy_case() -> Result<()> {
-        let fixture = Fixture::setup(2)?;
-        let f0 = fixture.fds[0];
-        let f1 = fixture.fds[1];
-
-        // SAFETY: assume files opened by Fixture are inherited ones
-        unsafe {
-            init_once()?;
+    fn all_in_one() -> Result<()> {
```
Databean, line 157: Can this use [`std::sync::Once`](https://doc.rust-lang.org/std/sync/struct.Once.html) to set up the fixture once for all tests? ([stackoverflow](https://stackoverflow.com/a/58006287))

jemoreira: there is at least one test that needs to run before it has been called

PR #1266, opened 2025-06-12
Split fetch flags

Databean:
This file attracts new changes, it should be split up to provide multiple places for changes to land.

Bug: b/424280895


Databean:
Thanks for reviewing!

PR #1267, opened 2025-06-12
Extract enums from `CuttlefishConfig::InstanceSpecific`

Databean:
These don't need to be declared inside the class.

Bug: b/424305767


Databean:
Thanks for reviewing!

PR #1271, opened 2025-06-13
Create a type to represent the metadata image.

Databean:
The type provides access to the path of the metadata image, given that the file satisfies the constraints we impose on it. This means that accessing the file path guarantees that the file is created with the right constraints.

More specifically:
- Remove metadata-specific members from `CuttlefishConfig::InstanceSpecific`.
- Remove the `--blank_metadata_image_mb` flag. This has no users in Android CI.
- Enforce the `metadata.img` name is only referenced in one place.

Bug: b/424863187


base/cvd/cuttlefish/host/commands/assemble_cvd/disk/metadata_image.cc
```diff
@@ -16,27 +16,54 @@
...
+
+  return MetadataImage(std::move(path));
+}
+
+Result<MetadataImage> MetadataImage::Reuse(
+    const CuttlefishConfig::InstanceSpecific& instance) {
+  std::string path = instance.PerInstancePath(Name());
+
+  CF_EXPECT(FileExists(path));
+  CF_EXPECT_EQ(FileSize(path), kMetadataImageMb << 20);
```
cjreynol: Very minor NIT: 

My personal preference would be another constant calculated from `kMetadataImageMb` so no thought is needed to recognize this as a bytes value.  Feel free to close with no changes.

cjreynol: It seems part of the "issue" here is our uses of various prefixed sizes for different operations.  Looking at `CreateBlankImage` and (its `NewfsMsdos` call) we seems to be converting to bytes for different operations at various levels (eg `FileSize` call here or `SharedFD::Truncate` in `CreateBlankImage`), then using the MB value for error and log messages.

jemoreira: is the image size always an even multiple of 2^20 bytes?

base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd.cc
```diff
@@ -369,8 +369,14 @@ Result<const CuttlefishConfig*> InitFilesystemAndCreateConfig(
...
     for (const auto& instance : config.Instances()) {
-      auto os_builder = OsCompositeDiskBuilder(config, instance);
-      creating_os_disk |= CF_EXPECT(os_builder.WillRebuildCompositeDisk());
+      Result<MetadataImage> metadata = MetadataImage::Reuse(instance);
+      if (metadata.ok()) {
+        DiskBuilder os_builder =
+            OsCompositeDiskBuilder(config, instance, *metadata);
+        creating_os_disk |= CF_EXPECT(os_builder.WillRebuildCompositeDisk());
+      } else {
+        creating_os_disk = true;
```
cjreynol, line 378: I believe we can add a `continue` after this statement to escape the loop, since every other operation on the variable within the loop is an OR.

base/cvd/cuttlefish/host/commands/assemble_cvd/disk/metadata_image.cc
Databean: In the past, it was mandated to be an even multiple because of the `--blank_metadata_image_mb` accepted a value in MiB. It has to be exactly 64 MiB as of this CL. We can raise it later if we need to.

base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd.cc
Databean: Assuming you mean a `break`.

base/cvd/cuttlefish/host/commands/assemble_cvd/disk/metadata_image.cc
Databean: Created `kMetadataImageBytes`.

Databean:
Thanks for reviewing!

PR #1273, opened 2025-06-14
Split out more parts of `fetch_cvd.cc`

Databean:
Bug: b/424021019


base/cvd/cuttlefish/host/commands/cvd/fetch/host_tools_target.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include <optional>
+#include <string>
+
+#include "cuttlefish/host/commands/cvd/fetch/fetch_cvd_parser.h"
+#include "cuttlefish/host/libs/web/android_build_string.h"
+
+namespace cuttlefish {
+
+static constexpr char kHostToolsSubdirectory[] = "host_tools";
```
cjreynol: I believe this still needs to be marked `inline` to ensure there is only one copy of the variable for all source file includes of the header.  The implicit `inline` of `static constexpr` variables only applies to data members of a class.

From https://en.cppreference.com/w/cpp/language/static.html "Constant static members" section:

"""
A constexpr static data member is implicitly inline and does not need to be redeclared at namespace scope. This redeclaration without an initializer (formerly required) is still permitted, but is deprecated.
"""

and from https://en.cppreference.com/w/cpp/language/inline.html in the first section:

"""
The inline specifier, when used in a [decl-specifier-seq](https://en.cppreference.com/w/cpp/language/declarations.html#Specifiers) of a variable with static storage duration (static class member or namespace-scope variable), declares the variable to be an inline variable.

A static data member declared constexpr on its first declaration is implicitly an inline variable.
"""

cjreynol (approved):

base/cvd/cuttlefish/host/commands/cvd/fetch/host_tools_target.h
Databean: Thanks, updated.

PR #1276, opened 2025-06-16
Add reasonable defaults to `base/cvd/.bazelrc`

Databean:
With this change, `bazel build '//...'` no longer runs clang-tidy, and test results don't take up many lines of output.

The clang-tidy targets are somewhat inconvenient that they are evaluated during the `build` step, but only report results at `test` time.

Bug: b/425437146


adelva1984 (approved):

Databean:
Thanks for reviewing!

PR #1278, opened 2025-06-17
Add C++ wrapper classes for libzip

Databean:
There's a lot of raw pointer management in existing usage:

- https://github.com/google/android-cuttlefish/blob/4cf78f84bf8b17af7e74e48b6b4f97f99b034c24/base/cvd/cuttlefish/host/libs/web/remote_zip.cc
- https://github.com/google/android-cuttlefish/blob/4cf78f84bf8b17af7e74e48b6b4f97f99b034c24/base/cvd/cuttlefish/host/libs/web/remote_zip_test.cc
- https://github.com/google/android-cuttlefish/blob/4cf78f84bf8b17af7e74e48b6b4f97f99b034c24/base/cvd/cuttlefish/host/commands/host_bugreport/main.cc

Bug: b/425534816


base/cvd/cuttlefish/host/libs/zip/remote_zip.h
```diff
@@ -17,19 +17,15 @@
...
 #include "cuttlefish/host/libs/web/http_client/http_client.h"
+#include "cuttlefish/host/libs/zip/zip_cc.h"
 
 namespace cuttlefish {
 
-Result<std::unique_ptr<zip_t, void (*)(zip_t*)>> ZipFromUrl(
-    HttpClient&, const std::string&, uint64_t size,
-    std::vector<std::string> headers);
-
+Result<Zip> ZipFromUrl(HttpClient&, const std::string&, uint64_t size,
```
jemoreira: What's the string parameter for? I'm sure it can be deduced from reading the implementation but that shouldn't be necessary. A name and maybe some comments would make this easier to use.

base/cvd/cuttlefish/host/libs/zip/zip_cc.h
```diff
@@ -0,0 +1,107 @@
...
+      std::unique_ptr<SeekableZipSourceCallback>);
+  static Result<ZipSource> FromData(const void* data, size_t size);
+  static Result<ZipSource> FromFile(const std::string& path);
+  static Result<ZipSource> FromZip(class Zip);
+
+  ZipSource(ZipSource&&);
+  ~ZipSource();
+  ZipSource& operator=(ZipSource&&);
+
+  Result<class ZipSourceReader> Reader();
```
jemoreira, line 92: IIUC the only purpose of `ZipSource` is to get a `ZipSourceReader` out of it. Can the two classes be merged into one?

base/cvd/cuttlefish/host/libs/zip/zip_cc.h
```diff
@@ -0,0 +1,107 @@
...
+
+class ZipSourceReader {
+ public:
+  friend class ZipSource;
+
+  ZipSourceReader(ZipSourceReader&&);
+  ~ZipSourceReader();
+  ZipSourceReader& operator=(ZipSourceReader&&);
+
+  Result<uint64_t> Read(void* data, uint64_t length);
```
jemoreira, line 113: Does a return of `{0}` mean `EOF`? If that's the case mention it in a comment.

base/cvd/cuttlefish/host/libs/zip/zip_cc.h
```diff
@@ -0,0 +1,107 @@
...
+ public:
+  friend class ZipSource;
+
+  static Result<Zip> CreateFromSource(ZipSource);
+  static Result<Zip> OpenFromSource(ZipSource);
+
+  Zip(Zip&&);
+  ~Zip();
+
+  Result<void> AddFile(const std::string& name, ZipSource);
```
jemoreira, line 138: `AddFile` and `Finalize` sound like actions to do when creating a zip file, but a `Zip` can only be created from a `ZipSource`, which can only be read from, so how can `Zip` "write" anything?. Am I missing something?

adelva1984:
Can you update adb and fastboot as well? I forked both to use libzip as requested, and they are both C++.

base/cvd/cuttlefish/host/libs/zip/zip_cc.h
Databean: A `ZipSource` can be written to. `cvd_host_bugreport` uses this strategy to create a `ZipSource` for a new file and then write to it. That's the `CreateFromSource` call.

base/cvd/cuttlefish/host/libs/zip/zip_cc.h
Databean: There's a state change in the underlying `zip_source_t` that it can be "opened for read" or "opened for write", which are mutually exclusive operations so they can fail. The intention here is to tie the state change to the lifetime of the `ZipReader` instance, but I can see it being excessive at a time when the write operations are not exposed.

base/cvd/cuttlefish/host/libs/zip/zip_cc.h
Databean: Added comments throughout the file.

base/cvd/cuttlefish/host/libs/zip/remote_zip.h
Databean: Added the parameter name and a comment.

Databean:
> I forked both to use libzip as requested

I'm starting to regret bringing that up.

IIUC `adb` uses `ZipArchive` to [find the offsets where zip file members start](https://cs.android.com/android/platform/superproject/main/+/main:packages/modules/adb/client/incremental_utils.cpp;l=347;drc=959eb4a096fc51bf2814e3374cb12b41d32a045d). I'm not sure I trust `zip_source_tell` to find these locations while a `zip_t` is moving around the seek offsets in the underlying `zip_source_t`, and I don't see anything else in libzip that exposes this information.

https://www.github.com/nih-at/libzip/issues/82 explicitly says this is unsupported in libzip.

jemoreira (approved):

adelva1984:
> > I forked both to use libzip as requested
...
> https://www.github.com/nih-at/libzip/issues/82 explicitly says this is unsupported in libzip.

I wasn't sure about this either, and honestly, I didn't test it, but it looks like adb is permitted to fall back to --no-incremental if the device doesn't support incfs (we do), so I think it would be better to just remove or disable the incremental code altogether and delete the dependency.

PR #1282, opened 2025-06-18
Create a type to represent misc.img creation

Databean:
There is currently an implicit ordering that [InitializeMiscImage](https://github.com/google/android-cuttlefish/blob/f6d691bf5cb8e337462929351ce45386af83424c/base/cvd/cuttlefish/host/libs/config/data_image.cpp#L251) must be run before [`BuildCompositeDiskIfNecessary`](https://github.com/google/android-cuttlefish/blob/05001207984a8a91e4a860fb7ac4b6dfcffe71a8/base/cvd/cuttlefish/host/commands/assemble_cvd/disk_flags.cc#L755) which is not obvious to a reader.

Hiding the file path behind a type will make it possible to enforce that the file is created before other code can access the type.


Databean:
Thanks for reviewing!

PR #1285, opened 2025-06-18
Add runner's architecture in bazel cache name

k311093:


ikicha (approved):

k311093:
Checked caching at https://github.com/google/android-cuttlefish/pull/1289, it shows that cache for debian build does not speeds up cvd unit test, update cache for each task.

It might also need to find a way to docker-image-check task which runs bazel while building docker image.

k311093:
Looks like it's due to different directory name, just changed directory name

PR #1300, opened 2025-06-19
libzip C++ wrapper: More types, minus ZipBuilder

Databean:
This avoids errors around "instance doesn't support this operation" by presenting more distinct types with different subsets of operations.

The ZipBuilder class wasn't adding any value over holding a WritableZip directly..

Bug: b/425534816


base/cvd/cuttlefish/host/libs/zip/zip_cc.h
```diff
@@ -56,101 +62,154 @@ struct ZipStat {
...
+
+  /* Returns a RAII instance that puts this instance in an "open for reading"
+   * state. Can fail. Should not outlive this instance. */
+  Result<class SeekingZipSourceReader> Reader();
+
+ private:
+  SeekableZipSource(std::unique_ptr<Impl>);
+};
+
+class WritableZipSource : public SeekableZipSource {
```
jemoreira, line 117: This inheritance implies that all `WritableZipSource`s are also readable, but that's not always true. For example, when writing zip-compressed data to a network stream.

jemoreira (approved):

base/cvd/cuttlefish/host/libs/zip/zip_cc.h
Databean: I don't believe libzip has the concept of a write-only `zip_source_t` instance.

https://libzip.org/documentation/zip_source_function_create.html

> Depending on the uses, there are three useful sets of commands to be supported by a `zip_source_callback`:
>
> - read source
> - seekable read source
> - read/write source

Databean:
Thanks for reviewing!

PR #1308, opened 2025-06-20
Fix memory leak in zip_cc.cc

Databean:
Bug: b/426664911


Databean:
Thanks for reviewing!

PR #1309, opened 2025-06-21
Split `//cuttlefish/host/commands/cvd/cli/selector`

Databean:
See https://bazel.build/docs/bazel-and-cpp

Bug: b/426677072


rmuthiah (approved):

Databean:
Thanks for reviewing!

PR #1316, opened 2025-06-23
Bring nasm into bazel, since libvpx requires it.

3405691582:
libvpx will complain about requiring nasm or yasm in its configure script. Instead of requiring it to be installed on the host, bring the dependency into the bazel build and use that.


3405691582:
@ser-io mind taking a look?

jemoreira (approved):

base/cvd/BUILD.libvpx.bazel
```diff
@@ -17,12 +17,14 @@ genrule(
...
         "INITIAL_WD=`pwd`",
+        "AS_PATH=$$(dirname $$(realpath $(location @nasm)))",
         "cd `dirname $(location configure)`",
-        "./configure",
+        "env PATH=$$AS_PATH:$$PATH ./configure",
         "make",
         "cp libvpx.a $$INITIAL_WD/$(RULEDIR)",
     ]),
     visibility = ["//visibility:private"],
+    tools = ["@nasm"],
```
jemoreira, line 27: This is the only place AFAIK that `nasm` is used. Could you remove the dependency from the control file (base/debian/control) and the github workflows (.github/workflows/*.yaml)?

3405691582: Done this in a separate pr (#1325) for separability.

PR #1318, opened 2025-06-24
Use clang instead of gcc to build libvpx

jemoreira:


jemoreira:
I tried using the clang toolchains from the genrule, but that fails because they need to define a `TemplateVariableInfo` but don't.

base/cvd/BUILD.libvpx.bazel
```diff
@@ -18,7 +18,7 @@ genrule(
     cmd = " && ".join([
         "INITIAL_WD=`pwd`",
         "cd `dirname $(location configure)`",
-        "./configure",
+        "CC=clang CXX=clang++ ./configure",
```
Databean: Is this using the distro clang?

Databean (approved):

base/cvd/build_external/libvpx/generic/vpx_scale_rtcd.h
```diff
@@ -0,0 +1,80 @@
...
+ *  Copyright (c) 2025 The WebM project authors. All Rights Reserved.
+ *
+ *  Use of this source code is governed by a BSD-style license
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree. An additional intellectual property rights grant can be found
+ *  in the file PATENTS.  All contributing project authors may
+ *  be found in the AUTHORS file in the root of the source tree.
+ */
+
+// This file is generated. Do not edit.
```
Databean: Is there a way to make bazel generate these `// This file is generated.` at build time, rather than committing the files to source?

base/cvd/build_external/libvpx/PATCH.libvpx_config.patch
```diff
@@ -0,0 +1,5675 @@
+diff --git a/tools_common.h b/tools_common.h
```
Databean: Having such a large patch file seems pretty brittle. Is there a command or script that can re-generate this patch file?

base/cvd/BUILD.libvpx.bazel
```diff
@@ -1,34 +1,915 @@
...
+        "//conditions:default": ["@platforms//:incompatible"],
+    }),
+)
+
+cc_library(
     name = "libvpx",
+    srcs = select({
+        "@platforms//cpu:aarch64": libvpx_arm64_c_srcs,
+        "@platforms//cpu:x86_64": libvpx_x86_64_c_srcs + libvpx_x86_64_c_hdrs,
+        "//conditions:default": libvpx_generic_c_srcs,
```
Databean, line 621: These lists have a lot of overlap, would it be possible to define a list of the common files so that they're not listed multiple times?

base/cvd/build_external/libvpx/generic/vpx_scale_rtcd.h
jemoreira: Done

base/cvd/build_external/libvpx/PATCH.libvpx_config.patch
jemoreira: Left a comment on MODULE.bazel with the command used to generate it.

base/cvd/BUILD.libvpx.bazel
jemoreira: Done

PR #1323, opened 2025-06-25
Implement HO golang client API `ExtractArtifact(filename string)`

0405ysj:


frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -590,6 +593,24 @@ func (c *HostOrchestratorClientImpl) ExtractFile(uploadDir string, filename stri
 	return result, nil
 }
 
+func (c *HostOrchestratorClientImpl) ExtractArtifact(filename string) (*hoapi.Operation, error)  {
+	checksum, err := sha256Checksum(filename)
+	if err != nil {
+		return err
+	}
+	path := "/v1/userartifacts/" + checksum + "_extracted"
+	res := &hoapi.StatArtifactResponse{}
```
0405ysj: Currently I'm starting implementation about HO golang client API for extracting user artifact, but I think it's better to have more support on `StatArtifactResponse` so that retriving an answer whether artifact is extracted or not.

```
type StatArtifactResponse struct{
  Extracted boolean `json:"extracted"`
}
```

@jemoreira @ser-io  To get agreement before implementation, is it good to proceed implementation without additional review regarding this issue on our previous design doc? Thanks!

ser-io: You can receive `Extracted: false` and then receive 409 in the following POST /:extract request, as the server state could change in between those calls.

You  rather need to handle 409 response on POST /:extract  and don't return error to the function caller when the server returns 409. By doing so, there's no need to have the new field `Extracted boolean `json:"extracted"`` just rely on the server response on POST /:extract request.

ser-io (changes requested):
https://github.com/google/android-cuttlefish/pull/1323/files#r2167565156

frontend/src/libhoclient/host_orchestrator_client.go
0405ysj: Acknowledged. Done with handling 409 error in the scope of client function `ExtractArtifact`.

frontend/src/libhoclient/host_orchestrator_client_test.go
```diff
@@ -177,6 +177,75 @@ func TestUploadArtifactReceive409WhenUploadingChunks(t *testing.T) {
...
+		case "POST /v1/userartifacts/d2c055002a6cdf8dd9edf90c7a666cb5f7f2d25da8519ec206f56777d74e0c7d/:extract":
+			writeErr(w, http.StatusConflict)
+		default:
+			t.Fatal("unexpected endpoint: " + ep)
+		}
+	}))
+	defer ts.Close()
+	client := NewHostOrchestratorClient(ts.URL)
+
+	if op, err := client.ExtractArtifact(testFile); err != nil {
```
ikicha: err == nil?

ikicha: or message below is wrong.

frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -590,6 +593,26 @@ func (c *HostOrchestratorClientImpl) ExtractFile(uploadDir string, filename stri
...
+	if err != nil {
+		return nil, err
+	}
+	res := &hoapi.StatArtifactResponse{}
+	if err := c.HTTPHelper.NewGetRequest("/v1/userartifacts/" + checksum).JSONResDo(res); err != nil {
+		return nil, err
+	}
+	op := &hoapi.Operation{}
+	if err := c.HTTPHelper.NewPostRequest("/v1/userartifacts/"+checksum+"/:extract", nil).JSONResDo(op); err != nil {
+		if apiCallError, ok := err.(*ApiCallError); ok && apiCallError.HTTPStatusCode == 409 {
```
ikicha: maybe we'd better use `http.StatusConflict` instead of 409?

frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -108,6 +108,9 @@ type UserArtifactsClient interface {
 	// Upload artifact into the artifacts repository.
 	// Artifacts are identified by their SHA256 checksum in the artifacts repository
 	UploadArtifact(filename string) error
+	// Extract artifact into the artifacts repository.
+	// Artifacts are identified by their SHA256 checksum in the artifacts repository
```
ser-io, line 112: Document returning `nil, nil`. Returning a nil operation pointer when there's no error means that the artifact was already extracted.

ser-io (approved):

frontend/src/libhoclient/host_orchestrator_client_test.go
ser-io: nit: avoid `if else` clause here and assert both are nil in independent follow up if clauses. `t.Fatal("expected error")` would early return, hence there's no need for the else clause which makes the code harder to follow.

```
op, err := client.ExtractArtifact(testFile)

// assert err is nil

// assert op is nil
```

ser-io (approved):

frontend/src/libhoclient/host_orchestrator_client.go
0405ysj: Done.

frontend/src/libhoclient/host_orchestrator_client.go
0405ysj: Done.

frontend/src/libhoclient/host_orchestrator_client_test.go
0405ysj: The message was wrong. Done.

0405ysj: Done.

ikicha (approved):

PR #1324, opened 2025-06-25
Create bazel cache from scratch at regular intervals

k311093:
This prevents incresing size for Bazel disk cache.

### DIFF
--- a/.github/workflows/update-bazel-cache.yaml
+++ b/.github/workflows/update-bazel-cache.yaml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
+  schedule:
+    - cron: '0 */3 * * *'
 
 jobs:
   update-cvd-test-bazel-cache:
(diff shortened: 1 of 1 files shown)


.github/workflows/update-bazel-cache.yaml
```diff
@@ -62,6 +72,12 @@ jobs:
           ${{ runner.os }}-${{ runner.arch }}-bazel-disk-cache-${{ env.CACHE_VERSION }}-debian-package-${{ github.ref_name }}-
```
ikicha: this only difference is about restore-keys, isn't it? can we merge them by using comparison operator? restore-keys: ${{ github.event_name == 'schedule' && runner.os + '-' ~~~(maybe we can use format here) || '' }} 
https://docs.github.com/en/actions/reference/evaluate-expressions-in-workflows-and-actions#operators

k311093: Changed to use if expression for restore-keys, used format string for all keys.

PR #1326, opened 2025-06-25
Use free-tier runners.

ser-io:


ser-io:
host has to be cleaned up first and then use containers, steps https://github.com/google/android-cuttlefish/blob/main/tools/buildutils/cw/README.md can be used here.

PR #1327, opened 2025-06-25
Bump LLVM version to 19.1.

3405691582:
Otherwise, the clang in the toolchain requires libtinfo to be installed on the build host, and this dependency no longer exists in later versions.

The runtime dependency in the package and workflow builds can probably be removed too, but will do that in a later separate pr (since this change should be applied first).

### DIFF
--- a/.github/workflows/presubmit.yaml
+++ b/.github/workflows/presubmit.yaml
@@ -136,7 +136,7 @@ jobs:
   build-debian-package:
     runs-on: ubuntu-22.04-4core
     container:
-      image: debian@sha256:823849b88ae7e9b6ceb605fbdf51566499c234a9cfca8da1e4f22234fd65a09c # debian:bullseye-20250317 (amd64)
+      image: debian@sha256:defe49aa5a875725310dff6b398aa7b49b595121adf9ddabc6f69274ffcceca3 # debian:bookworm-20250610 (amd64)
     steps:
     - name: Check for dockerenv file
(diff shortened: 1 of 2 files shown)


Databean (approved):

.github/workflows/presubmit.yaml
```diff
@@ -136,7 +136,7 @@ jobs:
   build-debian-package:
     runs-on: ubuntu-22.04-4core
     container:
-      image: debian@sha256:823849b88ae7e9b6ceb605fbdf51566499c234a9cfca8da1e4f22234fd65a09c # debian:bullseye-20250317 (amd64)
+      image: debian@sha256:defe49aa5a875725310dff6b398aa7b49b595121adf9ddabc6f69274ffcceca3 # debian:bookworm-20250610 (amd64)
```
Databean, line 139: We intentionally support older debian / ubuntu releases, is this going to force users to upgrade?

3405691582: Possibly. Let me see if we can use a select or similar to make this backport more gracefully.

3405691582: Drat, we can't use select in MODULE.bazel. The ideal way might be to use the [bring-your-own LLVM](https://github.com/bazel-contrib/toolchains_llvm?tab=readme-ov-file#bring-your-own-llvm) functionality but that's a big lift for now. I'll close this for now and revisit at a later date.

PR #1328, opened 2025-06-25
Use build macros in //cuttlefish/host/libs/web/http_client

Databean:
Bug: b/407624875


Databean:
Thanks for reviewing!

PR #1329, opened 2025-06-25
Use build macros in `//cuttlefish/host/libs/web`

Databean:
Also fixes clang-tidy errors in `oauth2_consent.cpp`

Bug: b/407624875


base/cvd/cuttlefish/host/libs/web/BUILD.bazel
```diff
@@ -1,36 +1,38 @@
-load("//:build_variables.bzl", "COPTS")
-load("//tools/lint:linters.bzl", "clang_tidy_test")
+# Copyright (C) 2025 The Android Open Source Project
```
cjreynol: We have not been adding the header to `BUILD.bazel` files.

base/cvd/cuttlefish/host/libs/web/oauth2_consent.cpp
```diff
@@ -46,11 +55,11 @@ Result<std::string> AuthorizationCodeFromUrl(const std::string_view url) {
   std::string_view code = url;
 
   static constexpr std::string_view kCodeEq = "code=";
-  std::size_t code_eq_pos = code.find(kCodeEq);
+  size_t code_eq_pos = code.find(kCodeEq);
```
cjreynol, line 58: No opinion here, merely trying to get on the same page.  In the past I have used [`std::size_t`](https://en.cppreference.com/w/cpp/types/size_t.html) from the `<cstdlib>` header instead.  Does that trigger a warning?  I vaguely recall a past conversation about types being exported by certain headers as a library implementation detail, even if that was not a part of the standard, but I do not remember the specifics.

base/cvd/cuttlefish/host/libs/web/BUILD.bazel
Databean: Maybe we should?

https://opensource.google/documentation/reference/patching#license_headers_and_copyright_notices

> All published source files should have some form of license header with at least one copyright notice.

Depends if the bazel files are "source files".

base/cvd/cuttlefish/host/libs/web/oauth2_consent.cpp
Databean: https://google.github.io/styleguide/cppguide.html#Integer_Types

> The standard library header <stdint.h> defines types like int16_t, uint32_t, int64_t, etc. You should always use those in preference to short, unsigned long long and the like, when you need a guarantee on the size of an integer. Prefer to omit the std:: prefix for these types, as the extra 5 characters do not merit the added clutter. Of the built-in integer types, only int should be used. When appropriate, you are welcome to use standard type aliases like size_t and ptrdiff_t.

base/cvd/cuttlefish/host/libs/web/BUILD.bazel
Databean: Other Google projects aren't consistent on this.

Tensorflow doesn't put license headers in build files: https://github.com/tensorflow/tensorflow/blob/67b2dae91e4cb0690764d6f21288122f12df4482/tensorflow/core/data/BUILD#L1

Abseil does:
https://github.com/abseil/abseil-cpp/blob/2a2d6aad626f75a32c8b7bb775fbff73d558b8b1/absl/algorithm/BUILD.bazel#L1

Databean: I've removed it now for consistency.

Databean:
Thanks for reviewing!

PR #1340, opened 2025-06-27
Use nasm with bazel deps while building libvpx

k311093:
It tried to use installed nasm while make because PATH is not specified. Add PATH to use nasm correctly.


k311093:
The error was not catched in the presubmit (only detected in clean build https://github.com/google/android-cuttlefish/actions/runs/15914949288). because Bazel cache didn't aware of installation of packages.

Let me add logic for it.

0405ysj (approved):
Talked with PR author, and got answer that currently `cvd` is not available to build in the local machine without this PR.

PR #1345, opened 2025-06-27
Import ag/34252503

Databean:
Bug: b/424853343


Databean:
Thanks for reviewing!

PR #1346, opened 2025-06-27
Validate build files in presubmit

Databean:
- Should use `cf_cc_*` targets instead of `cc_*` targets inside `//cuttlefish`
- Build files should be formatted correctly
- There should be no unused load statements

A failed presubmit will look like this:

https://github.com/Databean/android-cuttlefish/actions/runs/15935645391/job/44954735155

Bug: b/428245967


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #1359, opened 2025-07-01
Remove the `init_boot` flag.

Databean:
I don't believe there are any users of this flag.

Bug: b/429022737


cjreynol (approved):
Should we also remove this from `cuttlefish_config.h`, `assemble_cvd_flags.h`, etc?

Databean:
Removing from `cuttlefish_config.h` will be more work, but you're right that it should be removed from `assemble_cvd_flags.h` here.

Databean:
Thanks for reviewing!

PR #1360, opened 2025-07-01
Create a type to represent the factory-reset-protected partition

Databean:
https://github.com/google/android-cuttlefish/blob/c79c39ceb2035ff7e9c1e2b732070aaef7312a88/base/cvd/cuttlefish/host/commands/assemble_cvd/disk/factory_reset_protected.cc#L27

https://github.com/google/android-cuttlefish/blob/c79c39ceb2035ff7e9c1e2b732070aaef7312a88/base/cvd/cuttlefish/host/commands/assemble_cvd/disk/initialize_instance_composite_disk.cc#L89

There's a dependency here to enforce the ordering, but the file path is accessible at any time from the config. Creating a type to represent the creation of the image enforces an ordering that accessing the path implies the file is already created.


cjreynol (approved):

Databean:
Thanks for reviewing!

PR #1362, opened 2025-07-02
Request remote zip file sizes.

Databean:
Also includes:

- Support HTTP HEAD requests in the HTTP client and always return HTTP response headers
- Utilities to find a response by its case-insensitive key

Bug: b/423755602


cjreynol (approved):

Databean:
Thanks for reviewing!

PR #1364, opened 2025-07-03
HO can launch CF instances with user artifact v1 HO APIs

0405ysj:
This PR is to support HO launching CF instances with canonical config & new user artifact APIs.

Noting that **writing E2E test is out of scope** from this PR. I'm going to write in following PR with getting confirmation about this change.

Corresponding change for `cvdr` is suggested at https://github.com/google/cloud-android-orchestration/pull/435.

Internally, this PR suggests passing canonical config about user artifacts uploaded & extracted by new HO APIs with given structure. And below are the reasons why I suggested this structure and creating directories with storing symlinks inside.
```
"@userartifacts/{{.Checksum_1}},{{.Checksum_2}},{{.Checksum_3}},{{.Checksum_n}}"
```
- The design of new user artifact upload API assumes not uploading multiple files in a single directory. However, some execution flows of `cvdr create` requires to upload multiple images. For instance, `cvdr create --local_image` requires to upload all files listed at `device/google/cuttlefish/required_images`. It was able to upload those into single directory with legacy APIs, but not available with new APIs.
- Canonical config doesn't support to configure multiple directories. The structure of canonical config is defined at [load_config.proto](https://github.com/google/android-cuttlefish/blob/main/base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_config.proto), but there's no repeated string field to configure. Even making a patch to support this functionality isn't simple, since lots of variables defined under `assemble_cvd` are written with assuming single directory per instance, such as [here](https://github.com/google/android-cuttlefish/blob/main/base/cvd/cuttlefish/host/commands/assemble_cvd/flags/system_image_dir.h).
- Copying user artifact is still one of the valid options. But creating symlinks instead is much faster to launch CF instances.
- HO should receive the list of checksums inside canonical config. One of the alternative structures would be passing json array under `default_build` of canonical config, but this is different from the definition of canonical config defined at [load_config.proto](https://github.com/google/android-cuttlefish/blob/main/base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_config.proto).
- HO cannot reuse `@user_artifacts` which is used for representing user artifacts uploaded/extracted with legacy APIs. Parent directory is different, and it requires to expose `{checksum}_extracted` dir name structure to the outside for that case.

### DIFF
--- a/frontend/src/host_orchestrator/orchestrator/createcvdaction.go
+++ b/frontend/src/host_orchestrator/orchestrator/createcvdaction.go
@@ -15,17 +15,21 @@
 package orchestrator
 
 import (
-	"bytes"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"log"
 	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
 	"sync/atomic"
 
 	apiv1 "github.com/google/android-cuttlefish/frontend/src/host_orchestrator/api/v1"
 	"github.com/google/android-cuttlefish/frontend/src/host_orchestrator/orchestrator/cvd"
 	"github.com/google/android-cuttlefish/frontend/src/host_orchestrator/orchestrator/exec"
 	"github.com/google/android-cuttlefish/frontend/src/liboperator/operator"
+	"github.com/hashicorp/go-multierror"
 )
 
 type CreateCVDActionOpts struct {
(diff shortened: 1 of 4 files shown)


jemoreira:
> "@userartifacts/{{.Checksum_1}},{{.Checksum_2}},{{.Checksum_3}},{{.Checksum_n}}"

Why a list inside a string?

IIUC that string is going in a field(s) of the canonical config passed to the create endpoint. Those canonical configs don't accept lists inside a string. What am I missing?

frontend/src/host_orchestrator/orchestrator/createcvdaction.go
```diff
@@ -104,11 +108,24 @@ func (a *CreateCVDAction) launchWithCanonicalConfig(op apiv1.Operation) (*apiv1.
...
-	configFile, err := createTempFile("cvdload*.json", data, 0640)
+	config := string(data)
+	log.Printf("environment config:\n%s", config)
+	regex, err := regexp.Compile("\"" + apiv1.EnvConfigUserArtifactsVarV1 + "/[0-9a-zA-Z_,]+\"")
+	if err != nil {
+		return nil, fmt.Errorf("failed to compile regex: %w", err)
+	}
+	var errs *multierror.Error
+	config = regex.ReplaceAllStringFunc(config, func(base string) string {
+		checksums := strings.Split(strings.TrimPrefix(strings.Trim(base, "\""), apiv1.EnvConfigUserArtifactsVarV1), ",")
```
jemoreira, line 119: I think this leaves the first checksum with a starting `'/'` while the other ones don't have it. It's easier to just trim it here before splitting than to deal with the differences later on.

0405ysj:
Close this PR, as we would introduce a new HO API instead. https://github.com/google/android-cuttlefish/pull/1375

PR #1370, opened 2025-07-07
Delete the `--use_16k` flag.

Databean:
As far as I can tell, no device build configurations are active where this flag works correctly.

Bug: b/427829326
Test: Run with substitution


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #1373, opened 2025-07-07
Remove the `--hibernation_image` flag

Databean:
This flag was added in b/339688542 but in practice the hibernation image is always the same file, `hibernation_swap.img` in the system image directory. I could not find any customization of this flag in CI.

`hibernation_image` is also removed from the config, and the default path relative to `system_image_dir` is now the only path.

`SystemImageDirFlag` is passed into new code, but this should be relevant for multiple partition images.

Bug: b/430131136
Test: launch with substitution


Databean:
Thanks for reviewing!

PR #1375, opened 2025-07-08
Introduce PrepareImageDirectory under UserArtifactsManager of HO

0405ysj:
As the concern about suggestion at https://github.com/google/android-cuttlefish/pull/1364 was arisen with passing a list at the string field(`default_build`) of canonical config, this PR is for introducing new HO API to access multiple user artifacts from single directory by creating some symlinks. Also this PR introduces its golang client API, to be used for `cvdr`.

Below is the summary of API structure.

HO REST API endpoint:
```
POST /v1/userartifacts/:prepare_img_dir
```

Request struct of HO REST API:
```
type PrepareImageDirectoryRequest struct {
	Checksums []string `json:"checksums"`
}
```

Response struct of HO REST API:
```
type PrepareImageDirectoryResponse struct {
	Dir string `json:"dir"`
}
```

HO golang client API:
```
// Output string is the directory name under user artifact root dir.
PrepareImageDirectory(filenames []string) (string, error)
```

### DIFF
--- a/frontend/src/host_orchestrator/api/v1/messages.go
+++ b/frontend/src/host_orchestrator/api/v1/messages.go
@@ -136,6 +136,10 @@ type ListUploadDirectoriesResponse struct {
 
 type StatArtifactResponse struct{}
 
+type PrepareImageDirectoryRequest struct {
+	Checksums []string `json:"checksums"`
+}
+
 type PrepareImageDirectoryResponse struct {
 	// [Output Only]
 	Dir string `json:"dir"`
--- a/frontend/src/host_orchestrator/api/v1/messages.go
+++ b/frontend/src/host_orchestrator/api/v1/messages.go
@@ -136,6 +136,11 @@ type ListUploadDirectoriesResponse struct {
 
 type StatArtifactResponse struct{}
 
+type PrepareImageDirectoryResponse struct {
+	// [Output Only]
+	Dir string `json:"dir"`
+}
+
 type CreateSnapshotRequest struct {
 	// [Optional]
 	// Value must match regex: "^([a-z0-9\\-]+)$".
--- a/frontend/src/host_orchestrator/orchestrator/userartifacts.go
+++ b/frontend/src/host_orchestrator/orchestrator/userartifacts.go
@@ -57,6 +57,9 @@ type UserArtifactsManager interface {
 	StatArtifact(checksum string) (*apiv1.StatArtifactResponse, error)
 	// Extract artifact with the given checksum
 	ExtractArtifact(checksum string) error
+	// Prepare image directory containing symlinks of user artifacts under multiple directories
+	// named with different checksum each.
+	PrepareImageDirectory(checksums []string) (*apiv1.PrepareImageDirectoryResponse, error)
 
 	UserArtifactsDirResolver
 	// Creates a new directory for uploading user artifacts in the future.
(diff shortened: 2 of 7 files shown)


frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -235,6 +238,34 @@ func (m *UserArtifactsManagerImpl) ExtractArtifact(checksum string) error {
...
+	}
+	imgDirName, err := imageDirName(dirs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get checksum from directory names: %w", err)
+	}
+	dstDir := filepath.Join(m.RootDir, imgDirName)
+	if exists, err := dirExists(dstDir); err != nil {
+		return nil, fmt.Errorf("failed to check existence of directory: %w", err)
+	} else if exists {
+		return nil, operator.NewConflictError("image directory with same set of user artifacts already exists", nil)
```
ser-io, line 254: why do you care whether there's an image directory with the same set of artifacts already?

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -57,6 +57,9 @@ type UserArtifactsManager interface {
 	StatArtifact(checksum string) (*apiv1.StatArtifactResponse, error)
 	// Extract artifact with the given checksum
 	ExtractArtifact(checksum string) error
+	// Prepare image directory containing symlinks of user artifacts under multiple directories
```
ser-io, line 60: Remove the new method from the `UserArtifactsManager` interface. Responsibility of creating images directories to created cvd devices later don't need to be added to the user artifacts manager abstraction. This new logic could be perfectly isolated as it requires a list of checksum and an instance of `UserArtifactsDirResolver` as input and a directory name as output.

To implement the new endpoint isolate its logic in a new `Action` struct. You can see action implementations in files like: `fetchartifactsaction.go`, `createsnapshotaction.go`, `startcvdaction.go` and so on, follow the same pattern used in those implementations with the following key components:


```
// CreateImagesDirAction struct

type CreateImagesDirAction struct {
	req                   *apiv1.CreateImagesRequest
        uaDirResolver  UserArtifactsDirResolver
        rootDir string
        ....
}

// Run method.
func (a *CreateImagesDirAction) Run() (apiv1.Operation, error) {
}
```

frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -235,6 +238,34 @@ func (m *UserArtifactsManagerImpl) ExtractArtifact(checksum string) error {
 	return nil
 }
 
+func (m *UserArtifactsManagerImpl) PrepareImageDirectory(checksums []string) (*apiv1.PrepareImageDirectoryResponse, error) {
```
ser-io, line 241: Return an apiv1.Operation. HO methods dealing with file management should be non-blocking. The operation will be completed with the relevant `apiv1.PrepareImageDirectoryResponse`.

frontend/src/host_orchestrator/api/v1/messages.go
```diff
@@ -136,6 +136,11 @@ type ListUploadDirectoriesResponse struct {
 
 type StatArtifactResponse struct{}
 
+type PrepareImageDirectoryResponse struct {
+	// [Output Only]
+	Dir string `json:"dir"`
```
ser-io, line 145: `Name`

jemoreira (changes requested):
Please make the image directory a resource in the RESTful api:

```
POST /cvd_imgs_dirs - create
GET /cvd_imgs_dirs - list
GET /cvd_imgs_dirs/{id} - details
DELETE /cvd_imgs_dirs/{id} - delete

optional:
PUT /cvd_imgs_dirs/{id} - modify, to add files to an existing one, for example
```

The list and delete endpoints are needed to clean up in case `cvdr` crashes between preparing the directory and launching the device.

I apologize for not thinking of this yesterday, hopefully most of this PR can be reused and only some modifications are needed for the api endpoints.

0405ysj:
Close this PR. API set is discussed again, and handling all of 5 APIs at this PR is not efficient to manage.

PR #1378, opened 2025-07-08
Remove `strip_include_prefix` from `base/cvd/cuttlefish` and use workspace-relative include paths

Databean:
Similar to b/427554449, the `compile_commands.json` file created for the repository has to be regenerated to correctly pick up on alternate include paths. Using workspace-relative paths makes it easier to reuse an existing hedron-created `compile_commands.json`.

Additionally, this was being done piecemeal on individual files as they were being edited.


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #1379, opened 2025-07-09
Introduce types for persistent_vbmeta.img and ap_vbmeta.img

Databean:
There is soft dependency ordering that `GeneratePersistentVbmeta` should be called before `InitializeInstanceCompositeDisk`. This change makes it a hard dependency by encoding the file names in new dedicated types that are accepted as inputs by `InitializeInstanceCompositeDisk`.

https://github.com/google/android-cuttlefish/blob/19c4bd3b6ce47a59b6caf778d19cd0fc3b5b362a/base/cvd/cuttlefish/host/commands/assemble_cvd/disk/initialize_instance_composite_disk.cc#L50

https://github.com/google/android-cuttlefish/blob/19c4bd3b6ce47a59b6caf778d19cd0fc3b5b362a/base/cvd/cuttlefish/host/commands/assemble_cvd/disk/initialize_instance_composite_disk.cc#L72

https://github.com/google/android-cuttlefish/blob/19c4bd3b6ce47a59b6caf778d19cd0fc3b5b362a/base/cvd/cuttlefish/host/commands/assemble_cvd/disk/generate_persistent_vbmeta.cpp#L54

This removes `vbmeta_path` from `CuttlefishConfig`, which was also easy to mistake for `vbmeta_image`. `vbmeta_path` was used for the "persistent disk", while `vbmeta_image` was used for the "operating system disk".

Bug: b/430677564


Databean:
Thanks for reviewing!

PR #1380, opened 2025-07-10
Warning fixes

adelva1984:


base/cvd/BUILD.libsrtp2.bazel
```diff
@@ -56,15 +56,6 @@ cc_library(
         "include/srtp_priv.h",
         "include/ut_sim.h",
     ],
-    copts = [
-        "-fPIC",
-        "-Wall ",
-        "-pedantic",
-        "-Wstrict-prototypes",
```
Databean, line 63: If I understand correctly these are the only `copts` that are related to warnings, while `-fPIC` is related to linking, `-O4` and `-funroll-loops` are performance. Not sure what `-c` is doing there.

base/cvd/build_external/gfxstream/PATCH.gfxstream.3683847.patch
```diff
@@ -0,0 +1,836 @@
+commit 3683847ab3d60808a606d29843ad26840c8b32d6
```
Databean, line 1: I see this commit is upstream here:

https://github.com/google/gfxstream/commit/3683847ab3d60808a606d29843ad26840c8b32d6

Is it possible to uprev the gfxstream commit hash in `base/cvd/MODULE.bazel` rather than pull in this individual commit?

Databean (approved):

base/cvd/BUILD.libsrtp2.bazel
adelva1984: That's true, but I think -fPIC does nothing (this can be a shared library or part of one, which must be PIC on Linux), -O4 generates a warning from clang as -O3 is the maximum supported optimization level, and I don't know if SRTP has any business selecting special optimization with -funroll-loops, given what it does, so I just removed all of these as they were probably just taken from the upstream build system and all feel a bit cargo culty. If you'd prefer I split this out, or update the message, I can do that.

base/cvd/build_external/gfxstream/PATCH.gfxstream.3683847.patch
adelva1984: It is, but I pinged Jason about this option and we agreed to do it this way, so that we don't have to preempt the validation of a new gfxstream, and we don't have to do more work on either side just for the sake of warning fixes. When upstream comes in, the patches can just be dropped.

base/cvd/BUILD.libsrtp2.bazel
Databean: That's fair, I imagine a lot of specific compiling optimization flags are "this made it faster on my machine once 10 years ago" at best.

PR #1383, opened 2025-07-10
Create types for InstanceCompositeDisk and ApCompositeDisk

Databean:
Having these types in hand will guarantee that the files they represent on disk have been initialized.

Bug: b/431041073


cjreynol (approved):

Databean:
Thanks for reviewing!

PR #1384, opened 2025-07-11
Create types for `BootloaderEnvPartition` and `ApBootloaderEnvPartition`

Databean:
Having these types in hand implies that the files these types represent have beeen created on the disk.

Bug: b/431041079


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #1385, opened 2025-07-11
Fix some mistakes in remote_zip.cc

Databean:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Range

- The header is named "Accept-Ranges".
- Ranges given are inclusive, not exclusive.
- Successful reads should update the seek location.

Bug: b/423755602


cjreynol (approved):

Databean:
Thanks for reviewing!

PR #1392, opened 2025-07-12
Improve cvd fetch error messages

cjreynol:
and try to clean up the CLI output by not filling it with long API error responses.

Also cleanups to `AndroidBuildApi` to catch more error cases, lessen duplication, and in case we need to update API URLs in the future.  See commit messages for specifics.

Bug: 427744621

### DIFF
--- a/base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
+++ b/base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
@@ -40,6 +40,7 @@
 #include "cuttlefish/common/libs/utils/contains.h"
 #include "cuttlefish/common/libs/utils/environment.h"
 #include "cuttlefish/common/libs/utils/files.h"
+#include "cuttlefish/common/libs/utils/json.h"
 #include "cuttlefish/common/libs/utils/result.h"
 #include "cuttlefish/host/libs/web/android_build.h"
 #include "cuttlefish/host/libs/web/android_build_string.h"
@@ -233,10 +234,7 @@ Result<std::optional<std::string>> AndroidBuildApi::LatestBuildId(
                 << target << "\" on \"" << branch << "\", but received "
                 << json["builds"].size() << ". Full response:\n"
                 << json);
-  CF_EXPECT(json["builds"][0].isMember("buildId"),
-            "\"buildId\" member missing from response.  Full response:\n"
-                << json);
-  return json["builds"][0]["buildId"].asString();
+  return CF_EXPECT(GetValue<std::string>(json["builds"][0], { "buildId" }));
 }
 
 Result<std::string> AndroidBuildApi::BuildStatus(const DeviceBuild& build) {
@@ -263,7 +261,7 @@ Result<std::string> AndroidBuildApi::BuildStatus(const DeviceBuild& build) {
             "Response had \"error\" but had http success status. Received \""
                 << json << "\"");
 
-  return json["buildAttemptStatus"].asString();
+  return CF_EXPECT(GetValue<std::string>(json, { "buildAttemptStatus" }));
 }
 
 Result<std::string> AndroidBuildApi::ProductName(const DeviceBuild& build) {
@@ -290,8 +288,7 @@ Result<std::string> AndroidBuildApi::ProductName(const DeviceBuild& build) {
             "Response had \"error\" but had http success status. Received \""
                 << json << "\"");
 
-  CF_EXPECT(json.isMember("target"), "Build was missing target field.");
-  return json["target"]["product"].asString();
+  return CF_EXPECT(GetValue<std::string>(json, { "target", "product" }));
 }
 
 Result<std::unordered_set<std::string>> AndroidBuildApi::Artifacts(
@@ -331,7 +328,8 @@ Result<std::unordered_set<std::string>> AndroidBuildApi::Artifacts(
       page_token = "";
     }
     for (const auto& artifact_json : json["artifacts"]) {
-      artifacts.emplace(artifact_json["name"].asString());
+      artifacts.emplace(
+          CF_EXPECT(GetValue<std::string>(artifact_json, {"name"})));
     }
   } while (!page_token.empty());
   return artifacts;
@@ -387,9 +385,7 @@ Result<std::string> AndroidBuildApi::GetArtifactDownloadUrl(
   CF_EXPECT(!json.isMember("error"),
             "Response had \"error\" but had http success status. "
                 << "Received \"" << json << "\"");
-  CF_EXPECT(json.isMember("signedUrl"),
-            "URL endpoint did not have json path: " << json);
-  return json["signedUrl"].asString();
+  return CF_EXPECT(GetValue<std::string>(json, { "signedUrl" }));
 }
 
 Result<void> AndroidBuildApi::ArtifactToFile(const DeviceBuild& build,
--- a/base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
+++ b/base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
@@ -101,15 +101,12 @@ std::ostream& operator<<(std::ostream& out, const Build& build) {
 
 AndroidBuildApi::AndroidBuildApi(HttpClient& http_client,
                                  CredentialSource* credential_source,
-                                 std::string api_key,
+                                 AndroidBuildUrl* android_build_url,
                                  const std::chrono::seconds retry_period,
-                                 std::string api_base_url,
-                                 std::string project_id,
                                  CasDownloader* cas_downloader)
     : http_client(http_client),
       credential_source(credential_source),
-      android_build_url_(std::move(api_base_url), std::move(api_key),
-                         std::move(project_id)),
+      android_build_url_(android_build_url),
       retry_period_(retry_period),
       cas_downloader_(cas_downloader) {}
 
@@ -188,7 +185,7 @@ Result<std::vector<std::string>> AndroidBuildApi::Headers() {
 Result<std::optional<std::string>> AndroidBuildApi::LatestBuildId(
     const std::string& branch, const std::string& target) {
   const std::string url =
-      android_build_url_.GetLatestBuildIdUrl(branch, target);
+      android_build_url_->GetLatestBuildIdUrl(branch, target);
   auto response =
       CF_EXPECT(HttpGetToJson(http_client, url, CF_EXPECT(Headers())));
   const auto& json = response.data;
@@ -214,7 +211,7 @@ Result<std::optional<std::string>> AndroidBuildApi::LatestBuildId(
 
 Result<std::string> AndroidBuildApi::BuildStatus(const DeviceBuild& build) {
   const std::string url =
-      android_build_url_.GetBuildStatusUrl(build.id, build.target);
+      android_build_url_->GetBuildStatusUrl(build.id, build.target);
   auto response =
       CF_EXPECT(HttpGetToJson(http_client, url, CF_EXPECT(Headers())));
   const auto& json = response.data;
@@ -231,7 +228,7 @@ Result<std::string> AndroidBuildApi::BuildStatus(const DeviceBuild& build) {
 
 Result<std::string> AndroidBuildApi::ProductName(const DeviceBuild& build) {
   const std::string url =
-      android_build_url_.GetProductNameUrl(build.id, build.target);
+      android_build_url_->GetProductNameUrl(build.id, build.target);
   auto response =
       CF_EXPECT(HttpGetToJson(http_client, url, CF_EXPECT(Headers())));
   const auto& json = response.data;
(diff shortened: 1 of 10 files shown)


base/cvd/cuttlefish/host/libs/web/android_build_url.h
```diff
@@ -0,0 +1,49 @@
...
+#include <string_view>
+#include <vector>
+
+namespace cuttlefish {
+
+class AndroidBuildUrl {
+ public:
+  AndroidBuildUrl() = delete;
+  AndroidBuildUrl(AndroidBuildUrl&&) = delete;
+  virtual ~AndroidBuildUrl() = default;
```
Databean: nits:

- Is there anything against making this type copyable / movable? Copying or moving `std::string`s should be safe.
- The no-argument constructor is deleted automatically when there is a constructor with arguments, not counting copy or move constructors.
- The destructor only needs to be virtual if there are going to be subclasses, so that line can be left out if the non-virtual default destructor is good enough.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
```diff
@@ -202,138 +204,93 @@ Result<std::vector<std::string>> AndroidBuildApi::Headers() {
...
-            "Expected to receive 1 build for \""
-                << target << "\" on \"" << branch << "\", but received "
-                << json["builds"].size() << ". Full response:\n"
-                << json);
-  CF_EXPECT(json["builds"][0].isMember("buildId"),
-            "\"buildId\" member missing from response.  Full response:\n"
-                << json);
-  return json["builds"][0]["buildId"].asString();
+
+  CF_EXPECTF(json["builds"].size() == 1,
```
Databean: Maybe check that `json["builds"]` is an array as well?

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
```diff
@@ -60,25 +61,27 @@ bool StatusIsTerminal(const std::string& status) {
...
-  }
-  return result;
-}
-
 struct CloseDir {
   void operator()(DIR* dir) { closedir(dir); }
 };
 
+Result<Json::Value> GetResponseJson(const HttpResponse<Json::Value>& response,
+                                    const bool allow_redirect = false) {
```
Databean, line 69: Why are redirects disallowed?

Come to think of it, maybe we should be using [`CURLOPT_FOLLOWLOCATION`](https://curl.se/libcurl/c/CURLOPT_FOLLOWLOCATION.html) so that redirects are handled inside libcurl, and a redirect leaking out would definitely be an error.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
```diff
@@ -60,25 +61,27 @@ bool StatusIsTerminal(const std::string& status) {
...
+  //  debug information in error responses floods stderr with too much text
+  //  logged at a level that still ends up in the log file
+  LOG(DEBUG) << "API response data:\n" << response.data;
+  const bool allowed_response_code =
```
Databean: nit: maybe `has_allowed_response_code` or `response_code_allowed` since this variable isn't the response code but is a property of the response code?

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
```diff
@@ -122,10 +125,9 @@ AndroidBuildApi::AndroidBuildApi(HttpClient& http_client,
                                  CasDownloader* cas_downloader)
     : http_client(http_client),
       credential_source(credential_source),
-      api_key_(std::move(api_key)),
+      android_build_url_(std::move(api_base_url), std::move(api_key),
+                         std::move(project_id)),
```
Databean: Can `AndroidBuildUrl` be part of the API, so the caller is explicitly passing in this semantic object rather than 3 strings?

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
```diff
@@ -202,138 +204,93 @@ Result<std::vector<std::string>> AndroidBuildApi::Headers() {
...
+    //    - user provided a build ID (not an error)
+    //    - user provided a branch with a typo
+    //    - user provided a branch without the necessary authentication
+    //    (for example, internal branches)
+    // This message is a best attempt at helping the user in third case
+    no_auth_error_message =
+        "\n\nThis fetch was run unauthenticated, which could be the "
+        "problem.\nTry `cvd help login`";
+  }
+  const Json::Value& json = CF_EXPECT(
```
Databean: Declaring this a `const Json::Value&` may be legal by const reference lifetime extension ( https://abseil.io/tips/101 and https://abseil.io/tips/107 ) but rather than trying to reason through it, can this be a non-reference declaration?

Especially considering `CF_EXPECT` uses compiler extensions internally, doesn't seem worth the risk.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
```diff
@@ -202,138 +204,93 @@ Result<std::vector<std::string>> AndroidBuildApi::Headers() {
...
-            "Error fetching the product name of \""
-                << build << "\". The server response was \"" << json
-                << "\", and code was " << response.http_code);
-  CF_EXPECT(!json.isMember("error"),
-            "Response had \"error\" but had http success status. Received \""
-                << json << "\"");
-
-  CF_EXPECT(json.isMember("target"), "Build was missing target field.");
-  return json["target"]["product"].asString();
+  const Json::Value& json = CF_EXPECT(GetResponseJson(response),
```
Databean: Declaring this a `const Json::Value&` may be legal by const reference lifetime extension ( https://abseil.io/tips/101 and https://abseil.io/tips/107 ) but rather than trying to reason through it, can this be a non-reference declaration?

Especially considering `CF_EXPECT` uses compiler extensions internally, doesn't seem worth the risk.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
```diff
@@ -202,138 +204,93 @@ Result<std::vector<std::string>> AndroidBuildApi::Headers() {
...
-    const auto& json = response.data;
-    CF_EXPECT(response.HttpSuccess(),
-              "Error fetching the artifacts of \""
-                  << build << "\". The server response was \"" << json
-                  << "\", and code was " << response.http_code);
-    CF_EXPECT(!json.isMember("error"),
-              "Response had \"error\" but had http success status. Received \""
-                  << json << "\"");
+
+    const Json::Value& json = CF_EXPECT(GetResponseJson(response),
```
Databean: Declaring this a `const Json::Value&` may be legal by const reference lifetime extension ( https://abseil.io/tips/101 and https://abseil.io/tips/107 ) but rather than trying to reason through it, can this be a non-reference declaration?

Especially considering `CF_EXPECT` uses compiler extensions internally, doesn't seem worth the risk.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
```diff
@@ -362,34 +319,16 @@ Result<std::unordered_set<std::string>> AndroidBuildApi::Artifacts(
...
-                                           << "\". The server response was \""
-                                           << json << "\", and code was "
-                                           << response.http_code);
-  CF_EXPECT(!json.isMember("error"),
-            "Response had \"error\" but had http success status. "
-                << "Received \"" << json << "\"");
-  CF_EXPECT(json.isMember("signedUrl"),
-            "URL endpoint did not have json path: " << json);
-  return json["signedUrl"].asString();
+  const Json::Value& json =
```
Databean: Declaring this a `const Json::Value&` may be legal by const reference lifetime extension ( https://abseil.io/tips/101 and https://abseil.io/tips/107 ) but rather than trying to reason through it, can this be a non-reference declaration?

Especially considering `CF_EXPECT` uses compiler extensions internally, doesn't seem worth the risk.

base/cvd/cuttlefish/host/libs/web/android_build_url.cpp
```diff
@@ -0,0 +1,175 @@
...
+namespace cuttlefish {
+
+namespace {
+
+struct UrlBuilder {
+  std::string base_url;
+  std::vector<std::string> query_string;
+
+  void AddQueryParameter(std::string_view key, std::string_view value) {
+    query_string.push_back(fmt::format("{}={}", key, UrlEscape(value)));
```
Databean: Should the keys be escaped as well?

base/cvd/cuttlefish/host/libs/web/android_build_url.cpp
```diff
@@ -0,0 +1,175 @@
...
+    }
+    return result.str();
+  }
+};
+
+std::string BuildNameRegexp(
+    const std::vector<std::string>& artifact_filenames) {
+  // surrounding with \Q and \E treats the text literally to avoid
+  // characters being treated as regex
+  auto it = artifact_filenames.begin();
```
Databean, line 105: This assumes there is at least one member in `artifact_filenames`, which is not validated in the public API in `AndroidBuildUrl::GetArtifactUrl` either.

base/cvd/cuttlefish/host/libs/web/android_build_url.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+UrlBuilder GetArtifactDownloadBaseUrl(std::string_view api_base,
+                                      std::string_view id,
+                                      std::string_view target,
+                                      std::string_view artifact) {
+  return UrlBuilder{.base_url = fmt::format(
+                        "{}/builds/{}/{}/attempts/latest/artifacts/{}/url",
+                        api_base, UrlEscape(id), UrlEscape(target),
+                        UrlEscape(artifact))};
+}
```
Databean: If these were `static` members of `UrlBuilder`, then `base_url` and `query_string` could be private members of `UrlBuilder`.

base/cvd/cuttlefish/host/libs/web/android_build_url.h
cjreynol: Done.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
cjreynol: Done.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
cjreynol: Done.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
cjreynol: Done.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
cjreynol: Done.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
cjreynol: Agreed.  Done.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
cjreynol: Done.

base/cvd/cuttlefish/host/libs/web/android_build_url.cpp
cjreynol: No.  The keys should all have fixed string values that are part of AB's interface, not user-provided values.  

I do think that distinction is less clear with this version where the query string information is more distinct from the URL, even the fixed values we always use.  Clearly it is obscure enough in this form to raise the question.  I'll leave a short comment for the future.

cjreynol: Done.

base/cvd/cuttlefish/host/libs/web/android_build_url.cpp
cjreynol: `BuildNameRegexp(artifact_filenames)` is only called when `!artifact_filenames.empty()`.  If there were no artifacts then that query parameter would not be appended to the URL.  

Granted, we only call `Artifacts` privately with one or two values in `DownloadFile` or `DownloadFileWithBackup`.  That interface could probably be fixed to the actual allowed number of parameters, something like `Artifacts(const Build& build, std::string name, std::optional<std::string> othername = std::nullopt)` rather than a vector. 

I think it is sufficient in its current form right now, but the next pass at this we can always add more validation.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
cjreynol: I made a bug to look into the `cURL` option.  It looks non-trivial enough to be its own commit/PR to aid tracking down the behavior change later if there are errors.  For now I am going to keep the behavior the same as before.

b/431852235

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
cjreynol: Done, in a new commit after adding the one `AndroidBuildUrl`.

base/cvd/cuttlefish/host/libs/web/android_build_url.cpp
cjreynol: Done.

PR #1393, opened 2025-07-14
Introduce ImageDirectoriesManager in Host Orchestrator

0405ysj:
From https://github.com/google/android-cuttlefish/pull/1375#pullrequestreview-2999123088, there was a discussion for introducing API set around managing image directories. As the result of discussion, 5 APIs is getting introduced into HO.

- [P1] `POST /cvd_imgs_dirs`: Create an empty image directory. Reply `apiv1.Operation` in advance. As the operation result, ID is returned.
- [P1] `PUT /cvd_img_dirs/{id}`: Create or modify symlink under image directory. Reply `apiv1.Operation` in advance. If artifact doesn't exist, return 404. If symlink already exists, modify it.
- [P2] `GET /cvd_imgs_dirs`: List IDs of image directories.
- [P2] `DELETE /cvd_img_dirs/{id}`: Delete the image directory. Reply `apiv1.Operation` in advance.
- [P2] `GET /cvd_img_dirs/{id}`: Describe details of the image directories.

This PR introduces HO API `POST /cvd_imgs_dirs` and its corresponding golang client API. About the rest of APIs listed above, those will be implemented in different PRs.

### DIFF
--- a/frontend/src/libhoclient/host_orchestrator_client.go
+++ b/frontend/src/libhoclient/host_orchestrator_client.go
@@ -120,6 +120,12 @@ type UserArtifactsClient interface {
 	ExtractFile(uploadDir string, filename string) (*hoapi.Operation, error)
 }
 
+// Manage image directories created by the user.
+type ImageDirectoriesClient interface {
+	// Create an empty image directory.
+	CreateImageDirectory() (*hoapi.Operation, error)
+}
+
 // Operations that could be performend on a given instance.
 type InstanceOperationsClient interface {
 	// Create cvd bugreport.
@@ -160,6 +166,7 @@ type SnapshotsClient interface {
 type HostOrchestratorClient interface {
 	InstancesClient
 	UserArtifactsClient
+	ImageDirectoriesClient
 	InstanceOperationsClient
 	InstanceConnectionsClient
 	OperationsClient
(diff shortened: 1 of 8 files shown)


frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -609,6 +616,14 @@ func (c *HostOrchestratorClientImpl) ExtractArtifact(filename string) (*hoapi.Op
 	return op, nil
 }
 
+func (c *HostOrchestratorClientImpl) CreateImageDirectory() (*hoapi.Operation, error) {
```
jemoreira, line 619: This is going to break the orchestration build because the fakeHostOrchestratorClient there doesn't have this method. Please create a PR to add it there too.

jemoreira (approved):

frontend/src/libhoclient/host_orchestrator_client.go
0405ysj: Is it `fakeHostService`? I searched all structs using `HostOrchestratorClient` interface, and this was the only one I found. https://github.com/google/cloud-android-orchestration/blob/91fae399a37f09478b0fc1f6c351e9da0ad6d52d/pkg/cli/testing.go#L48

I'll create a PR there. To merge this PR, resolving this comment.

0405ysj: https://github.com/google/cloud-android-orchestration/pull/439 is corresponding PR for it.

PR #1394, opened 2025-07-14
cuttlefish: delete dead gfx code

gurchetansingh:
With the entire industry moving to Vulkan, it makes sense to delete OpenGL-based renderers whenever possible.  virgl was deleted from AOSP in 2024, so this Cuttlefish code path is essentially dead code anyways.

We hope to delete gfxstream GLES in the near future, but users still rely on it.


gurchetansingh:
CC: @jmacnak

PR #1395, opened 2025-07-14
Add a buffering layer for SeekableZipSource.

Databean:
libzip in some cases does many small reads. When reads map directly to HTTP range requests, it causes a large number of round-trips to pull all the data. Introducing a buffer that rounds up small reads and can read in sequential data preempts some of the reads.

Bug: b/431837315


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #1399, opened 2025-07-15
Support extracting files with libzip

Databean:
Will come in useful with reading zip files directly from the build API.


Databean:
Thanks for reviewing!

FYI renamed the `Clone` function to `Copy`.

PR #1405, opened 2025-07-16
Extract qcow2 code from a free function into a new type.

Databean:
This type will fit into a type hierarchy of disk image representations including raw images, android-sparse images, and composite disks that can describe representations of vitual disks that consist of multiple host files.

The "private implementation" idiom is used to hide implementation details of the qcow2 header type, but this may be excessive.

Bug: b/432327319


Databean:
Thanks for reviewing!

PR #1406, opened 2025-07-16
Remove the misc_info_txt flag

Databean:
https://github.com/google/android-cuttlefish/blob/81d945157656e3fc56f386c5291c6d28b667bef0/base/cvd/cuttlefish/host/libs/config/cuttlefish_config.h#L657

https://github.com/google/android-cuttlefish/blob/81d945157656e3fc56f386c5291c6d28b667bef0/base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd_flags.cpp#L521

The flag value is only stored in `CuttlefishConfig` but nothing reads it from `CuttlefishConfig`. It was originally introduced in aosp/2359203 but then not used in the follow-up CL aosp/2318835.

Bug: b/432331402


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #1407, opened 2025-07-16
Deprecate load_config.proto fields

cjreynol:
And suppress the build warnings from those deprecations.

Bug: 432329063


base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_config.proto
```diff
@@ -74,7 +76,7 @@ message Build {
 }
 
 message Connectivity {
-  optional Vsock vsock = 1;
+  optional Vsock vsock = 1 [deprecated = true];
```
Databean: `vsock` is in the "proposed for updated config" list in go/cf-config-flags-list .

base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_config.proto
```diff
@@ -84,7 +86,7 @@ message Vsock {
 message Graphics {
   repeated Display displays = 1;
   optional bool record_screen = 2;
-  optional string gpu_mode = 3;
+  optional string gpu_mode = 3 [deprecated = true];
```
Databean, line 89: go/cf-config-flags-list mentions this should still be present, but limited to "auto" or "off" rather than the full set of options supported by our internal tools.

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_config.proto
```diff
@@ -111,11 +113,11 @@ message Super {
 message Security {
   optional string serial_number = 1;
   optional bool use_random_serial = 2;
-  optional bool guest_enforce_security = 3;
+  optional bool guest_enforce_security = 3 [deprecated = true];
 }
 
 message Streaming {
-  optional string device_id = 1;
+  optional string device_id = 1 [deprecated = true];
```
jemoreira: This one is meant to be use by the orchestrators. Don't deprecate.

jemoreira (changes requested):

base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_config.proto
cjreynol: Yes.  I thought we would change this to a new enum type.  From what I read in the proto best practices about type changes I believe that change might be an issue.  My plan was to deprecate this and introduce a new, limited field once we get to the point of removal.

I can leave a comment with that information, unless you disagree with the above plan.

EDIT:  Found a reference, `string` to `enum` is listed as an unsafe change.

base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_config.proto
cjreynol: Removed the `deprecated` option.

base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_config.proto
cjreynol: Thanks for the catch, that was a mistake!  Removed the `deprecated` option.

cjreynol:
Looks like this triggers `clang-tidy`'s `clang-diagnostic-deprecated-declarations`.  If that is going to trigger for every usage as we try to migrate users it may not be worth trying to signal our intentions until we are closer to removing fields?

Databean:
According to go/deprecation-policy it's valid to mark something as deprecated if there is a replacement available at the time of deprecation.

PR #1419, opened 2025-07-18
Add rootcanal.

jacky8hyf:
* Used a workaround for https://github.com/bazelbuild/bazel/issues/26578 by replacing local_repository() calls in rootcanal's MODULE.bazel.
* Replaced pdlc calls to be from the pdl-compiler crate.

Bug: b/402296262


base/cvd/build_external/rootcanal/0001-Add-local.bzl.patch
```diff
@@ -0,0 +1,90 @@
...
+Subject: [PATCH 1/3] Add local.bzl.
+
+Originally from
+https://cs.android.com/android/kernel/superproject/+/common-android-mainline:build/kernel/kleaf/impl/local_repository.bzl
+hence licence header is kept.
+
+This replaces @bazel_tools//cuttlefish/bazel:local.bzl to address
+https://github.com/bazelbuild/bazel/issues/26578
+---
+ local.bzl | 65 +++++++++++++++++++++++++++++++++++++++++++++++++++++++
```
Databean: Rather than using a patch to overlay this newly introduced `.bzl` file into the rootcanal repository, is it possible for the main android-cuttlefish repository to have this `.bzl` file as a regular file, and reference it from within the patched rootcanal repository?

That will make the file easier to maintain, without having to edit it through diff syntax.

Databean (approved):

base/cvd/build_external/rootcanal/0001-Add-local.bzl.patch
jacky8hyf: I tried that approach but it is unfortunately I don't find a clean way. The reason is that we can only determine the root of the **module** via this `_get_module_dir` function, which calculates the path by `repository_ctx.path(Label(local.bzl)`. If I were to move local.bzl to another repository, I can't think of a good way to calculate `_get_module_dir()`.

`repository_ctx.path(".")` doesn't work; that gives the path to the repository (e.g. fmtlib in this case), not the path to the module.

Do you have any suggestions?

jacky8hyf: Actually nevermind, I think I found a good workaround. I'll upload a patch shortly.

Databean: I patched in this PR and tried some variations on `:local.bzl`: it seems to me like at `repository_ctx` evaluation time it doesn't check that any target with this name actually exists. `repository_ctx.path` seems to work even for made up targets that don't exist, returning the path of where the target _would be_, assuming it exists.

https://bazel.build/rules/lib/builtins/repository_ctx#path writes

> Note that remote repositories and module extensions are executed during the analysis phase and thus cannot depends on a target result (the label should point to a non-generated file).

Maybe this is related, as generated files would be named differently?

In any case, if it doesn't need to refer specifically to `:local.bzl` but can instead refer to some arbitrary target that doesn't exist, then the `local.bzl` file doesn't need to actually be present in the repository that uses these rules.

jacky8hyf: Done, PTAL at the latest approach. I do this by turning "path" into an attr.label() not an attr.string(); then `third_party/fmtlib` effectively becomes `Label(//third_party/fmtlib:fmtlib)`, which, when fed into `repository_ctx.path()`, would yield the correct result.

Databean (approved):

PR #1420, opened 2025-07-19
Create a ChromeOsStateImage type

Databean:
This type will represent creating the file on disk, so accepting an argument of the type establishes an ordering between creating the file and using the file.

Bug: b/432833719


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #1425, opened 2025-07-22
Move *CompositeDiskBuilder out of disk_flags.cc

Databean:
`ApCompositeDiskBuilder` and `OsCompositeDiskBuilder` historically were in `disk_flags.cc` to access `FLAGS_*` values only available in `disk_flags.cc`. With the introduction of the `assemble_cvd_flags.h` header, these functions could be placed anywhere.

`ap_composite_disk.cc` and `os_composite_disk.cc` had a public interface that was only used by the respective `CompositeDiskBuilder`.

Bug: b/433348982


cjreynol (approved):

Databean:
Thanks for reviewing!

PR #1430, opened 2025-07-22
Split disk_flags.cc

Databean:
The functions in `disk_flags.cc` are not very related, and could be in separate files.

https://github.com/google/android-cuttlefish/blob/942da230f383bb8b5e5f9b65edc1fe412ecb0f8b/base/cvd/cuttlefish/host/commands/assemble_cvd/disk_flags.cc#L4

Bug: b/433548151


Databean:
Thanks for reviewing!

PR #1431, opened 2025-07-22
Read avb metadata straight into memory rather than going through a temporary directory and file

Databean:
https://github.com/google/android-cuttlefish/blob/287ffbb338b31b12517544022c8116377ce4ef36/base/cvd/cuttlefish/host/commands/assemble_cvd/boot_image_utils.cc#L504

https://github.com/google/android-cuttlefish/blob/287ffbb338b31b12517544022c8116377ce4ef36/base/cvd/cuttlefish/host/commands/assemble_cvd/boot_image_utils.cc#L194

https://github.com/google/android-cuttlefish/blob/287ffbb338b31b12517544022c8116377ce4ef36/base/cvd/cuttlefish/host/libs/avb/avb.cpp#L101

https://github.com/google/android-cuttlefish/blob/287ffbb338b31b12517544022c8116377ce4ef36/base/cvd/cuttlefish/host/libs/avb/avb.cpp#L91

The temporary directory only exists to hold one file. The file is only created in order to immediately read it from disk. The directory and file are immediately deleted afterwards.

Bug: b/433586538

### DIFF
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/boot_image_utils.cc
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/boot_image_utils.cc
@@ -26,7 +26,6 @@
 #include <string>
 
 #include <android-base/logging.h>
-#include <android-base/scopeguard.h>
 #include <android-base/strings.h>
 
 #include "cuttlefish/common/libs/fs/shared_fd.h"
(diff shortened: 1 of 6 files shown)


cjreynol (approved):

Databean:
Thanks for reviewing!

PR #1432, opened 2025-07-23
Return bootparams in-memory from UnpackBootImage

Databean:
It's only written to a file to be immediately read.

https://github.com/google/android-cuttlefish/blob/287ffbb338b31b12517544022c8116377ce4ef36/base/cvd/cuttlefish/host/commands/assemble_cvd/boot_image_utils.cc#L292

https://github.com/google/android-cuttlefish/blob/287ffbb338b31b12517544022c8116377ce4ef36/base/cvd/cuttlefish/host/commands/assemble_cvd/boot_image_utils.cc#L206

Doing this in-memory saves on IO.

Bug: b/433594073

### DIFF
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/boot_image_utils.cc
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/boot_image_utils.cc
@@ -32,6 +32,7 @@
 #include "cuttlefish/common/libs/utils/files.h"
 #include "cuttlefish/common/libs/utils/result.h"
 #include "cuttlefish/common/libs/utils/subprocess.h"
+#include "cuttlefish/common/libs/utils/subprocess_managed_stdio.h"
 #include "cuttlefish/host/libs/avb/avb.h"
 #include "cuttlefish/host/libs/config/config_utils.h"
 #include "cuttlefish/host/libs/config/known_paths.h"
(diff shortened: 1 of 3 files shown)


base/cvd/cuttlefish/host/commands/assemble_cvd/boot_image_utils.cc
```diff
@@ -201,24 +201,25 @@ bool GetAvbMetadataFromBootImage(const std::string& boot_image_path,
   return true;
 }
 
-Result<void> UnpackBootImage(const std::string& boot_image_path,
-                             const std::string& unpack_dir) {
-  SharedFD output_file = SharedFD::Creat(unpack_dir + "/boot_params", 0666);
```
cjreynol, line 193: Since these files are not temporary, perhaps this `boot_params` file is created as a log for verifying or debugging?  Some form of the output is logged after being run through `ExtractValue` in the updated `RepackBootImage` here, but not in the `gem5_image_unpacker.cpp::Gem5ImageUnpacker` function.  Could we at least add a `LOG(DEBUG)` there, or one in the helper function here?

Actually, looking at the gem5 usage it is followed by a `UnpackVendorBootImageIfNotUnpacked` call that has similar logic creating a `vendor_boot_params` file used as a signal to avoid duplicate unpacking.  I wonder if this file was used as a signal at some point in the past?  I do not see any relevant hits searching for `boot_params` in our code or elsewhere in Android that seems relevant to us, but maybe @dimorinny would know.

base/cvd/cuttlefish/host/commands/assemble_cvd/boot_image_utils.cc
```diff
@@ -201,24 +201,25 @@ bool GetAvbMetadataFromBootImage(const std::string& boot_image_path,
...
+  Command unpack_cmd = Command(UnpackBootimgBinary())
+                           .AddParameter("--boot_img")
+                           .AddParameter(boot_image_path)
+                           .AddParameter("--out")
+                           .AddParameter(unpack_dir);
+
+  std::string standard_out;
+  std::string standard_err;
+
+  int exit_code = RunWithManagedStdio(std::move(unpack_cmd), nullptr,
```
cjreynol: NIT:

Can we wrap up this `RunWithManagedStdio` block into a helper, since this pattern is somewhat common?  My first thought is:

```c++
struct Captured {
  int exit_code;
  std::string standard_out;
  std::string standard_error;
};

Result<Captured> RunAndCapture(Command&& command, int expected_exit_code = 0);
```

where the definition handles lines 212-220 here generically.

Databean: Will work on this in separate PRs first.

https://www.github.com/google/android-cuttlefish/pull/1437 kicks this off.

Databean: Done.

base/cvd/cuttlefish/host/commands/assemble_cvd/boot_image_utils.cc
Databean: Added a `LOG(DEBUG)` message for this.

Databean:
Thanks for reviewing!

cjreynol (approved):

Databean:
Thanks for reviewing!

PR #1433, opened 2025-07-23
Create a type to represent the `--boot_image` flag.

Databean:
There's an ordering dependency that
[`ResolveInstanceFiles`](https://github.com/google/android-cuttlefish/blob/287ffbb338b31b12517544022c8116377ce4ef36/base/cvd/cuttlefish/host/commands/assemble_cvd/disk_flags.cc#L129) needs to be called before other code that uses `FLAGS_boot_image`.

Bug: b/433570553


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #1436, opened 2025-07-23
Add ms-tpm-20-ref

jacky8hyf:
- Dropped -Wno-error=deprecated-declarations because we don't need it when building agianst BoringSSL
- Applied a few patches to make it build properly against boringssl.

Bug: b/402294836

### DIFF
--- a/base/cvd/BUILD.ms-tpm-20-ref.bazel
+++ b/base/cvd/BUILD.ms-tpm-20-ref.bazel
@@ -1,0 +1,126 @@
+load("@rules_cc//cc:cc_binary.bzl", "cc_binary")
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+
+# From PLATFORM_INC in TPMCmd/Makefile.am
+PLATFORM_INC = [
+    "TPMCmd/Platform/include",
+    "TPMCmd/Platform/include/prototypes",
+]
+
+# From SIMULATOR_INC in TPMCmd/Makefile.am
+SIMULATOR_INC = [
+    "TPMCmd/Simulator/include",
+    "TPMCmd/Simulator/include/prototypes",
+]
+
+# From TPM_INC in TPMCmd/Makefile.am
+TPM_INC = [
+    "TPMCmd/tpm/include",
+    "TPMCmd/tpm/include/platform_interface",
+    "TPMCmd/tpm/include/platform_interface/prototypes",
+    "TPMCmd/tpm/include/private",
+    "TPMCmd/tpm/include/private/prototypes",
+    "TPMCmd/tpm/include/public",
+    "TPMCmd/tpm/cryptolibs",
+    "TPMCmd/tpm/cryptolibs/common/include",
+    "TPMCmd/tpm/cryptolibs/Ossl/include",
+    "TPMCmd/tpm/cryptolibs/TpmBigNum/include",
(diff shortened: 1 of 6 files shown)


base/cvd/BUILD.ms-tpm-20-ref.bazel
```diff
@@ -0,0 +1,126 @@
...
+    "TPMCmd/tpm/include/platform_interface/prototypes",
+    "TPMCmd/tpm/include/private",
+    "TPMCmd/tpm/include/private/prototypes",
+    "TPMCmd/tpm/include/public",
+    "TPMCmd/tpm/cryptolibs",
+    "TPMCmd/tpm/cryptolibs/common/include",
+    "TPMCmd/tpm/cryptolibs/Ossl/include",
+    "TPMCmd/tpm/cryptolibs/TpmBigNum/include",
+    "TPMCmd/TpmConfiguration",
+]
```
Databean: nit: while `DEFINES` and `COMPILER_FLAGS` are reused, these constants don't contribute as much. https://bazel.build/build/share-variables writes

> Code duplication in BUILD files is usually fine. This can make the file more readable: each declaration can be read and understood without any context. This is important, not only for humans, but also for external tools. For example, a tool might be able to read and update BUILD files to add missing dependencies. Code refactoring and code reuse might prevent this kind of automated modification.
> 
> If it is useful to share values (for example, if values must be kept in sync), you can introduce a variable.

In particular, `buildozer` can more easily update arrays in rule values than arrays in constants / variables.

Can these be inlined into the rule using them?

Databean (approved):

base/cvd/BUILD.ms-tpm-20-ref.bazel
jacky8hyf: sure, let me do that.

base/cvd/BUILD.ms-tpm-20-ref.bazel
```diff
@@ -0,0 +1,135 @@
...
+        ":platform_headers",
+        ":tpm_headers",
+        "@boringssl//:crypto",
+        "@boringssl//:ssl",
+    ],
+)
+
+cc_binary(
+    name = "simulator",
+    srcs = glob(["TPMCmd/Simulator/**/*.c"]),
```
Databean, line 107: Sorry only occurs to me now: `secure_env` uses parts of `TPMCmd/Simulator` as a library:

https://github.com/google/android-cuttlefish/blob/393fc6ea415c2e0f28982f03dc5903b3e5856650/base/cvd/cuttlefish/host/commands/secure_env/in_process_tpm.cpp#L35C11-L35C23

However here `TPMCmd/Simulator/**/*.c` is only built as part of a `cc_binary`, so it doesn't seem like it will be possible to link in.

jacky8hyf: Got it. Let me build a separate cc_library() with the necessary sources.

jacky8hyf: Maybe it would be easier to define this extra cc_library when we write the corresponding rules of secure_env? I could add both TPMCmdp.c and TcpServer.c to the library now, but I am not sure if that would be too much.

jacky8hyf: Anyways, see #1442 .

PR #1445, opened 2025-07-24
Convenient wrappers for `RunWithManagedStdio`

Databean:
https://www.github.com/google/android-cuttlefish/pull/1432/files#r2226344612 points out that many callers have the same logic around calling RunWithManagedStdio. We should have a nicer interface for the common case.

### DIFF
--- a/base/cvd/cuttlefish/common/libs/utils/subprocess_managed_stdio.cc
+++ b/base/cvd/cuttlefish/common/libs/utils/subprocess_managed_stdio.cc
@@ -51,7 +51,7 @@ class ThreadJoiner {
 
 }  // namespace
 
-int RunWithManagedStdio(Command&& cmd_tmp, const std::string* stdin_str,
+int RunWithManagedStdio(Command cmd_tmp, const std::string* stdin_str,
                         std::string* stdout_str, std::string* stderr_str,
                         SubprocessOptions options) {
   /*
@@ -138,6 +138,20 @@ int RunWithManagedStdio(Command&& cmd_tmp, const std::string* stdin_str,
     return -1;
   }
   return code;
+}
+
+Result<std::string> RunAndCaptureStdout(Command command) {
+  std::string standard_out;
+  std::string standard_err;
+  std::string command_str = command.GetShortName();
+  int exit_code = RunWithManagedStdio(std::move(command), nullptr,
+                                      &standard_out, &standard_err);
+  LOG(DEBUG) << "Ran " << command_str << " with stdout:\n" << standard_out;
+  LOG(DEBUG) << "Ran " << command_str << " with stderr:\n" << standard_err;
+  CF_EXPECTF(exit_code == 0,
+             "Failed to execute '{}' <args>: exit code = {}, stdout = '{}', stderr = '{}'",
+             command_str, exit_code, standard_out, standard_err);
+  return standard_out;
 }
 
 }  // namespace cuttlefish
--- a/base/cvd/cuttlefish/common/libs/utils/subprocess_managed_stdio.h
+++ b/base/cvd/cuttlefish/common/libs/utils/subprocess_managed_stdio.h
@@ -17,9 +17,13 @@
 
 #include <string>
 
+#include "cuttlefish/common/libs/utils/result.h"
 #include "cuttlefish/common/libs/utils/subprocess.h"
 
 namespace cuttlefish {
+
+// Success is defined by the exit code being `0`.
+Result<std::string> RunAndCaptureStdout(Command);
 
 /*
  * Consumes a Command and runs it, optionally managing the stdio channels.
@@ -33,8 +37,8 @@ namespace cuttlefish {
  * If some setup fails, `command` fails to start, or `command` exits due to a
  * signal, the return value will be negative.
  */
-int RunWithManagedStdio(Command&& cmd_tmp, const std::string* stdin,
-                        std::string* stdout, std::string* stderr,
+int RunWithManagedStdio(Command, const std::string* stdin, std::string* stdout,
+                        std::string* stderr,
                         SubprocessOptions options = SubprocessOptions());
 
 }  // namespace cuttlefish
--- a/base/cvd/cuttlefish/common/libs/utils/BUILD.bazel
+++ b/base/cvd/cuttlefish/common/libs/utils/BUILD.bazel
@@ -349,6 +349,7 @@ cf_cc_library(
     hdrs = ["subprocess_managed_stdio.h"],
     deps = [
         "//cuttlefish/common/libs/fs",
+        "//cuttlefish/common/libs/utils:result",
         "//cuttlefish/common/libs/utils:subprocess",
         "//libbase",
     ],
--- a/base/cvd/cuttlefish/common/libs/utils/archive.cpp
+++ b/base/cvd/cuttlefish/common/libs/utils/archive.cpp
@@ -69,14 +69,9 @@ Result<std::vector<std::string>> ExtractFiles(
   for (const auto& extract : to_extract) {
     bsdtar_cmd.AddParameter(extract);
   }
-  bsdtar_cmd.RedirectStdIO(Subprocess::StdIOChannel::kStdOut,
-                           Subprocess::StdIOChannel::kStdErr);
-  std::string bsdtar_output;
-  int bsdtar_ret = RunWithManagedStdio(std::move(bsdtar_cmd), nullptr, nullptr,
-                                       &bsdtar_output);
+  std::string bsdtar_output =
+      CF_EXPECT(RunAndCaptureStdout(std::move(bsdtar_cmd)));
   LOG(DEBUG) << bsdtar_output;
-  CF_EXPECTF(bsdtar_ret == 0, "bsdtar extraction failed on '{}', '''{}'''",
-             archive, bsdtar_output);
 
   std::vector<std::string> outputs = android::base::Split(bsdtar_output, "\n");
   for (std::string& output : outputs) {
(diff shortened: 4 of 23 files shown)


base/cvd/cuttlefish/common/libs/utils/subprocess_managed_stdio.cc
```diff
@@ -140,4 +140,20 @@ int RunWithManagedStdio(Command&& cmd_tmp, const std::string* stdin_str,
   return code;
 }
 
+Result<std::string> CaptureSuccessfulStdout(Command command) {
+  std::string standard_out;
+  std::string standard_err;
+  int exit_code = RunWithManagedStdio(std::move(command), nullptr,
+                                      &standard_out, &standard_err);
+  CF_EXPECTF(exit_code == 0,
+             "Command failure: exit code = {}, stdout = '{}', stderr = '{}'",
```
cjreynol: I know we should have the full invocation in the log from the `Command.Start` call, but that detail could possibly change in the future. Can we include a name here to make it easier to connect the error and invocation when reviewing logs?

base/cvd/cuttlefish/common/libs/utils/subprocess_managed_stdio.h
```diff
@@ -17,10 +17,14 @@
 
 #include <string>
 
+#include "cuttlefish/common/libs/utils/result.h"
 #include "cuttlefish/common/libs/utils/subprocess.h"
 
 namespace cuttlefish {
 
+Result<void> RunsSuccessfully(Command);
```
jemoreira: nit: `RunsSuccessfully(Command)` sounds like checking that the command runs instead of actually running it. How about just `Run(Command)` or `Execute(Command)`?

base/cvd/cuttlefish/common/libs/utils/subprocess_managed_stdio.h
```diff
@@ -17,10 +17,14 @@
...
+Result<std::string> CaptureSuccessfulStdout(Command);
```
jemoreira: nit: Same idea here. Consider renaming to something more "execution" related, like `RunAndCaptureStdout`.

base/cvd/cuttlefish/common/libs/utils/subprocess_managed_stdio.cc
```diff
@@ -140,4 +140,20 @@ int RunWithManagedStdio(Command&& cmd_tmp, const std::string* stdin_str,
...
+             exit_code, standard_out, standard_err);
+  return standard_out;
+}
+
+Result<void> RunsSuccessfully(Command command) {
+  CF_EXPECT(CaptureSuccessfulStdout(std::move(command)));
```
jemoreira: Calling `CaptureSuccessulStdout` here means the command's stdout won't go to the parent's stdout like stderr presumably does.

base/cvd/cuttlefish/common/libs/utils/subprocess_managed_stdio.cc
```diff
@@ -140,4 +140,20 @@ int RunWithManagedStdio(Command&& cmd_tmp, const std::string* stdin_str,
   return code;
 }
 
+Result<std::string> CaptureSuccessfulStdout(Command command) {
+  std::string standard_out;
+  std::string standard_err;
+  int exit_code = RunWithManagedStdio(std::move(command), nullptr,
+                                      &standard_out, &standard_err);
```
jemoreira, line 148: I'd prefer not to capture stderr here since the name only mentions stdout. The returned error doesn't need the stderr output because when not captured it goes to the parent's stderr which is probably where the returned error goes to too. Also, the subprocess' stderror may be quite large making the error output harder to read.

jemoreira (approved):

base/cvd/cuttlefish/common/libs/utils/subprocess_managed_stdio.h
cjreynol: I think the name is trying to imply the exit code is checked.  I'm not sure `Run` or `Execute` provide that extra bit of context.

base/cvd/cuttlefish/host/commands/cvd/instances/status_fetcher.cpp
```diff
@@ -139,31 +139,26 @@ Result<Json::Value> FetchInstanceStatus(LocalInstance& instance,
...
   Command command = CF_EXPECT(ConstructCommand(construct_cmd_param));
 
-  std::string serialized_json;
-
-  int res = RunWithManagedStdio(std::move(command), nullptr, &serialized_json,
-                                nullptr /*stderr*/);
-
   // old branches will print nothing
-  if (serialized_json.empty() && res == 0) {
-    serialized_json = "[{\"warning\" : \"cvd-status-unsupported device\"}]";
```
cjreynol: We lose this output in the new logic.

A successful invocation on an old branch (according to the above comment) will return an empty string from `stdout`, but in the old logic this JSON was provided, there is a successful `ParseJson` call, and that parsed value would be the new value of `instance_status_json`.  Then the users gets a (mild) hint to they were on an old branch.

In the new logic a successful invocation with an empty output makes it into the following conditional block, maybe fails `ParseJson` (did not look if the empty string is accepted), and definitely fails the `isArray` check, returning an error `Result`.

I think this case is a change for better, but we also lose the case where the invocation fails, captures invalid JSON or a non-singular JSON array, then would return an error `Result`.  The capture-all-failure conditional provides more information.

Maybe we just leave it as is?  This was a quick version I came up with, and I do not really like it:

```
  // ...
  if (!serialized_json.ok()) {
    LOG(ERROR) << "Status tool failed:\n"
               << serialized_json.error().FormatForEnv();
    instance.set_state(cvd::INSTANCE_STATE_UNREACHABLE);
    instance_status_json["warning"] = "cvd status failed";
  } else {
    std::string to_parse = *serialized_json;
    if (to_parse.empty()) {
      to_parse = "[{\"warning\" : \"cvd-status-unsupported device\"}]";
    }
    Json::Value json_array = CF_EXPECT(ParseJson(to_parse), "Status tool returned invalid JSON");
    CF_EXPECTF(json_array.isArray(), "Status tool returned unexpected output (not an array): {}", to_parse);
    CF_EXPECT_EQ(json_array.size(), 1ul);
    instance_status_json = json_array[0];
  }
  OverrideInstanceJson(instance, instance_status_json);
  return instance_status_json;
}
```

base/cvd/cuttlefish/common/libs/utils/subprocess_managed_stdio.h
jemoreira: A method like `Result<whatever> Run()` says that it will "run" (from the name) and may "succeed" or not (from the return type). It has the same information as "RunSuccessfully".

My comment is not so much about removing the "Successfully" part (although I think it's redundant), but the use of "Runs" instead of "Run".

base/cvd/cuttlefish/common/libs/utils/subprocess_managed_stdio.cc
Databean: > The returned error doesn't need the stderr output because when not captured it goes to the parent's stderr which is probably where the returned error goes to too.

Our current error reporting system is not that straightforward. `LOG` statements in `assemble_cvd` are configured to [write to a temporary file and stderr](https://github.com/google/android-cuttlefish/blob/ef53fac08041ed71e78e0544c16461700781119e/base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd.cc#L306). In `run_cvd` without `--daemon` mode, `LOG` statements will similarly [write output to stderr and the `launcher.log` file](https://github.com/google/android-cuttlefish/blob/ef53fac08041ed71e78e0544c16461700781119e/base/cvd/cuttlefish/host/commands/run_cvd/main.cc#L249). In `run_cvd` with `--daemon` mode specifically, there is [stdio redirection](https://github.com/google/android-cuttlefish/blob/ef53fac08041ed71e78e0544c16461700781119e/base/cvd/cuttlefish/host/commands/run_cvd/boot_state_machine.cc#L208) so that child process stderr output is sent to the `launcher.log` file.

Without stderr redirection or capturing output, the subprocess output doesn't make it to log files. Additionally, not capturing it can put some distance between the `Result` output and the `stderr` of the failing subprocess in the log messages. I'd rather capture it for completeness in the returned object.

jemoreira: Makes sense, could you capture it and print it to the parent's stderr then?

Databean: Capturing it and `LOG`-ing it is doable, and shouldn't be a behavioral regression. In practice the logs will likely appear right before the failing `Result` output as most `Result`s fail all the way back to `main` in our implementation.

This still feels like a semantic regression. Reporting the errors via `LOG`-ing replaces a return value with a side effect. A caller in general may choose to handle a failing `Result` without itself failing, in which case reporting the error output could be helpful but could also be misleading.

base/cvd/cuttlefish/common/libs/utils/subprocess_managed_stdio.h
Databean: The intention here is to distinguish this function from the [existing comparable ones](https://github.com/google/android-cuttlefish/blob/7ba5e7f22ce674d4c572d0e754267233db31569e/base/cvd/cuttlefish/common/libs/utils/subprocess.h#L342):
```c++
int Execute(const std::vector<std::string>& commands);
int Execute(const std::vector<std::string>& commands,
            const std::vector<std::string>& envs);
```
I don't have a strong opinion about what this function is called, but the main distinction between the new function and the existing ones is whether non-zero exit codes are semantically considered a "failure".

Maybe `RunExpectSuccess`?

base/cvd/cuttlefish/common/libs/utils/subprocess_managed_stdio.cc
Databean: It'll look a bit strange at callsites, but it seems less confusing to drop this function entirely.

Callers could instead invoke `CaptureSuccessfulStdout` / `RunAndCaptureStdout` which has more obvious semantics from the name, and drop the captured output.

base/cvd/cuttlefish/common/libs/utils/subprocess_managed_stdio.h
Databean: Done.

base/cvd/cuttlefish/common/libs/utils/subprocess_managed_stdio.cc
Databean: Removed this function.

base/cvd/cuttlefish/common/libs/utils/subprocess_managed_stdio.cc
Databean: Done.

base/cvd/cuttlefish/host/commands/cvd/instances/status_fetcher.cpp
Databean: Reverted back to the original version of this file to postpone the issue.

base/cvd/cuttlefish/common/libs/utils/subprocess_managed_stdio.h
Databean: Removed this function.

cjreynol (approved):

base/cvd/cuttlefish/common/libs/utils/subprocess_managed_stdio.cc
Databean: Added logging.

Databean:
Thanks for reviewing!

PR #1450, opened 2025-07-27
Make `cvd host_bugreport` more robust

jemoreira:
- The internal subcommand no longer fails when it manages to gather at least some information. It still logs any problems it encounters.
- The final bugreport archive includes the fetch.log file when it's available
- Enables appending to an existing file with the libzip wrapper

### DIFF
--- a/base/cvd/cuttlefish/host/commands/host_bugreport/main.cc
+++ b/base/cvd/cuttlefish/host/commands/host_bugreport/main.cc
@@ -14,8 +14,6 @@
  * limitations under the License.
  */
 
-#include <stdio.h>
-
 #include <string>
 
 #include <android-base/file.h>
@@ -47,7 +45,7 @@ void LogError(Result<T> res) {
   }
 }
 
-void AddNetsimdLogs(WritableZip& archive) {
+Result<void> AddNetsimdLogs(WritableZip& archive) {
   // The temp directory name depends on whether the `USER` environment variable
   // is defined.
   // https://source.corp.google.com/h/googleplex-android/platform/superproject/main/+/main:tools/netsim/rust/common/src/system/mod.rs;l=37-57;drc=360ddb57df49472a40275b125bb56af2a65395c7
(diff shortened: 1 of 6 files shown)


base/cvd/cuttlefish/host/libs/zip/zip_cc.cc
```diff
@@ -504,15 +504,20 @@ Result<SeekableZipSource> ReadableZip::GetFile(uint64_t index) {
 
 ReadableZip::ReadableZip(std::unique_ptr<Impl> impl) : impl_(std::move(impl)) {}
 
-Result<WritableZip> WritableZip::FromSource(WritableZipSource source) {
+Result<WritableZip> WritableZip::FromSource(WritableZipSource source,
+                                            bool truncate) {
```
Databean: Can the public interface of this either use an `enum` or be two functions, like `OpenSource` and `TruncateSource`?

See https://abseil.io/tips/94

base/cvd/cuttlefish/host/commands/host_bugreport/main.cc
```diff
@@ -180,34 +178,38 @@ Result<void> CvdHostBugreportMain(int argc, char** argv) {
...
+  }
+
+  auto config = CuttlefishConfig::Get();
+  CHECK(config) << "Unable to find the config";
+
+  WritableZip archive = CF_EXPECT(ZipOpenReadWrite(FLAGS_output));
+
+  // This function will gather as much as it can and log what it couldn't. It
+  // won't fail because the bugreport log file needs to be added to the archive
+  // and the archive needs to be finalized.
```
Databean: "it won't fail because" doesn't sound exactly right here: is "failures are ignored because" more accurate?

base/cvd/cuttlefish/host/commands/host_bugreport/main.cc
```diff
@@ -180,34 +178,38 @@ Result<void> CvdHostBugreportMain(int argc, char** argv) {
...
+  TakeHostBugreport(config, archive);
```
Databean, line 209: Failures in `adb bugreport` weren't fatal before this change either, but error messages were logged with `LogError` or `LOG(ERROR)`.

Can this report the error from a failed `Result`?

Databean (approved):

base/cvd/cuttlefish/host/commands/host_bugreport/main.cc
jemoreira: Done

base/cvd/cuttlefish/host/commands/host_bugreport/main.cc
jemoreira: It logs failures in the function itself and returns void. It has to continue trying other things after something fails, that's why it doesn't return error.

base/cvd/cuttlefish/host/libs/zip/zip_cc.cc
jemoreira: Done

PR #1451, opened 2025-07-28
Import ag/34790658

Databean:
Bug: b/281744537


Databean:
Thanks for reviewing!

PR #1453, opened 2025-07-29
Create a DiskImage subclass for android-sparse images

Databean:
Bug: b/434797867


Databean:
Thanks for reviewing!

PR #1455, opened 2025-07-29
root-canal -> rootcanal

adelva1984:
The upstream project prefers the non-hyphenated name.


adelva1984:
ag/34835563 for the corresponding internal change

Databean (approved):

PR #1456, opened 2025-07-29
Use `Result` types in `image_aggregator.cc`

Databean:
This allows it to report errors without aborting, and reported errors include stack trace information rather than only the line number of the failing assertion.

image_aggregator at one time had a non-Cuttlefish dependency, which was a reason to not use the Result type throughout. The dependent code has since stopped using image_aggregator, and the image_aggregator code has moved to a new repository (GitHub).

Bug: b/434998699


base/cvd/cuttlefish/host/libs/image_aggregator/image_aggregator.cc
```diff
@@ -333,11 +327,8 @@ class CompositeDiskBuilder {
    * This method is not deterministic: some data is generated such as the disk
    * uuids.
    */
-  GptBeginning Beginning() const {
-    if (partitions_.size() > GPT_NUM_PARTITIONS) {
-      LOG(FATAL) << "Too many partitions: " << partitions_.size();
-      return {};
-    }
+  Result<GptBeginning> Beginning() const {
+    CF_EXPECT_LT(partitions_.size(), GPT_NUM_PARTITIONS, "Too many partitions");
```
cjreynol: `s/CF_EXPECT_LT/CF_EXPECT_LE/`, since the previous error mode was only `>`, not `>=`.

cjreynol (approved):

base/cvd/cuttlefish/host/libs/image_aggregator/image_aggregator.cc
Databean: Good catch, fixed.

Databean:
Thanks for reviewing!

PR #1458, opened 2025-07-30
Create a `VmManagerFlag` type

Databean:
This expresses the dependency in choosing a default VMM based on the guest architecture.

It's currently instantiated in two places to avoid changing type signatures on the `flags.cc` members.

Bug: b/435031349


Databean:
Thanks for reviewing!

PR #1460, opened 2025-07-30
sparse_image_utils acquires flock on where real image is located at

0405ysj:
Description: b/435041392

### DIFF
--- a/base/cvd/cuttlefish/host/libs/image_aggregator/sparse_image.cc
+++ b/base/cvd/cuttlefish/host/libs/image_aggregator/sparse_image.cc
@@ -38,7 +38,10 @@ namespace {
 
 constexpr std::string_view kAndroidSparseImageMagic = "\x3A\xFF\x26\xED";
 
-Result<SharedFD> AcquireLock(const std::string& tmp_lock_image_path) {
+Result<SharedFD> AcquireLockForImage(const std::string& image_path) {
+  std::string image_realpath;
+  CF_EXPECT(android::base::Realpath(image_path, &image_realpath));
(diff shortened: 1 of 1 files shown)


base/cvd/cuttlefish/host/libs/image_aggregator/sparse_image_utils.cc
```diff
@@ -61,10 +64,10 @@ Result<bool> IsSparseImage(const std::string& image_path) {
...
-
-  SharedFD fd = CF_EXPECT(AcquireLock(tmp_lock_image_path));
-
+  if (!CF_EXPECT(IsSparseImage(image_path))) {
+    return {};
+  }
+  SharedFD fd = CF_EXPECT(AcquireLockForImage(image_path));
   if (!CF_EXPECT(IsSparseImage(image_path))) {
     return {};
   }
```
Databean: nit: this check is redundant with the new one at the beginning of the function.

Databean (approved):

base/cvd/cuttlefish/host/libs/image_aggregator/sparse_image_utils.cc
Databean: On second thought, I see the first check is relevant as a fast short-circuit, but the second check avoids races between two processes that simultaneously check the condition and then enter the critical section.

My mistake.

0405ysj: Ah, yes. That was my intention. Thanks for the review!

PR #1461, opened 2025-07-30
Expose image type detection from image_aggregator

Databean:
Instances of the raw image type can only be created from the image detection logic, since a raw image is defind by not having the magic prefix of any special image format.

Bug: b/435218921


base/cvd/cuttlefish/host/libs/image_aggregator/detect_image.h
```diff
@@ -0,0 +1,28 @@
...
+
+#include <memory>
+#include <string>
+
+#include "cuttlefish/common/libs/utils/result.h"
+#include "cuttlefish/host/libs/image_aggregator/disk_image.h"
+
+namespace cuttlefish {
+
+Result<std::unique_ptr<DiskImage>> DetectImage(const std::string& path);
```
jemoreira: nit: I'd expect a function named `DetectImage` to return the image type or something like that, but this return type looks like a full image instead of just a type. Consider `ImageFromFile(...)` or `DiskImage::FromFile(...)`.

jemoreira (approved):

base/cvd/cuttlefish/host/libs/image_aggregator/detect_image.h
Databean: Renamed to `ImageFromFile`.

Unfortunately this can't be in the `DiskImage` class as it would create a circular dependency between the `:disk_image` bazel target and the subclass bazel targets.

Databean:
Thanks for reviewing!

PR #1462, opened 2025-07-30
Lift `GetGuestConfigAndSetDefaults` components out

Databean:
The dependency ordering is easier to express by lifting out the inner function calls, and it slightly splits up the large argument lists.

Bug: b/435249755


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #1467, opened 2025-07-31
Implement GET /cvd_imgs_dirs and DELETE /cvd_imgs_dirs/{id}

0405ysj:
Context: b/431651822

### DIFF
--- a/frontend/src/host_orchestrator/api/v1/messages.go
+++ b/frontend/src/host_orchestrator/api/v1/messages.go
@@ -136,9 +136,18 @@ type ListUploadDirectoriesResponse struct {
 
 type StatArtifactResponse struct{}
 
+type ImageDirectory struct {
+	ID string `json:"id"`
+}
+
 type CreateImageDirectoryResponse struct {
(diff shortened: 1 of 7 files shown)


frontend/src/host_orchestrator/api/v1/messages.go
```diff
@@ -145,6 +145,11 @@ type CreateImageDirectoryResponse struct {
 	ID string `json:"id"`
 }
 
+type ListImageDirectoriesResponse struct {
+	// [Output Only] Identifiers of image directories.
+	IDs []string `json:"ids"`
```
jemoreira: Please use an array of a struct with an `ID` field in it.

We may in the future want to return more information about the image directories when listing, for example any user provided labels/names. Adding fields to an existing struct is better than turning a string into a struct with multiple values.

0405ysj: Thanks. Done!

jemoreira (approved):

PR #1473, opened 2025-08-01
Preserve proxy environment variables

Radon10043:
Hi,

The `debuild` command can fail in environments requiring a network proxy. This is because `debuild` runs in a sanitized environment by default, which strips essential variables like `http_proxy` and `https_proxy`.

This commit addresses the issue by detecting if these proxy variables are set in the shell and then explicitly forwarding them to the `debuild` process using the `-e` flag. This ensures network-dependent steps in the build can succeed behind a proxy.


Radon10043:
Hi @jemoreira ,

The workflow for this PR is awaiting maintainer approval. Could you please approve it to run the checks when you get a chance?

Thank you!

PR #1478, opened 2025-08-01
Move libcutils to the imported version from AOSP.

jacky8hyf:
Also run buildifier on touched files.


jacky8hyf:
This does need an internal change to work. @Databean should I merge the internal change first?

Databean:
If you have a working internal change in hand I think it works out better to submit here first, and then handle the internal change after the public one is merged here.

base/cvd/BUILD.android_system_core.bazel
```diff
@@ -77,3 +80,32 @@ cc_binary(
     srcs = ["libsparse/simg2img.cpp"],
     deps = [":libsparse"],
 )
+
+cc_library(
+    name = "libcutils",
```
Databean, line 85: Actually one thing first: before you submit this CL here, please make sure we can get `BUILD` visibility on the internal change. Something like cl/765004401 to be sure the internal change will land successfully.

jacky8hyf: Done, I've uploaded a change to change the visibility there (cl/791783546). Once internal changes are in I'll submit this.

PR #1479, opened 2025-08-01
Move `mbr.h` to `image_aggregator`

Databean:
Bug: b/435760503


base/cvd/cuttlefish/host/libs/image_aggregator/mbr.h
```diff
@@ -15,24 +15,36 @@
...
   MbrPartitionEntry partitions[4];
-  std::uint8_t boot_signature[2];
+  uint8_t boot_signature[2];
 };
 
 static_assert(sizeof(MasterBootRecord) == kSectorSize);
+
+/**
+ * Creates a "Protective" MBR Partition Table header. The GUID
+ * Partition Table Specification recommends putting this on the first sector
```
3405691582, line 44: I think this should go back on mbr.cc, and/or reword this comment.

Databean: https://google.github.io/styleguide/cppguide.html#Function_Comments writes

> Declaration comments describe use of the function (when it is non-obvious); comments at the definition of a function describe operation.

It seems relevant to the use of the function to describe what the output is going to be?

3405691582: As far as I understood it before, "this" was referring to the specifics of the MBR header, and guessing there was something special about it. Staring at it a little more, it is just a plain old MBR header, so as worded it is probably okay in retrospect.

PR #1480, opened 2025-08-02
Expose a zip file accessor from `BuildApi`

Databean:
The non-caching implementation uses a lazy-loading mechanism to only download needed parts of the zip file. The caching implementation eagerly downloads the entire file. Optimizing the cached version is TODO.

Bug: b/435770397


cjreynol (approved):

Databean:
Thanks for reviewing!

PR #1485, opened 2025-08-04
Empty the acloud_cvd_temp dir on cvd reset.

3405691582:
The acloud_cvd_temp directory contains instance lock files. If we are resetting with cvd reset, we should remove these, since we are resetting state completely anyway.

This will become more important later when we are making changes to the way instance locks are acquired, and starting from a clean slate with cvd reset is more convenient than manually removing lock files.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/reset.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/reset.cpp
@@ -158,6 +158,11 @@ class CvdResetCommandHandler : public CvdCommandHandler {
     }
     CF_EXPECT(KillAllCuttlefishInstances(
         /* clear_instance_dirs*/ options.clean_runtime_dir));
+
+    if (DirectoryExists("/tmp/acloud_cvd_temp/")) {
+      CF_EXPECT(RecursivelyRemoveDirectory("/tmp/acloud_cvd_temp/"));
+    }
+
     return {};
   }
   cvd_common::Args CmdList() const override { return {kResetSubcmd}; }


base/cvd/cuttlefish/host/commands/cvd/cli/commands/reset.cpp
```diff
@@ -158,6 +158,11 @@ class CvdResetCommandHandler : public CvdCommandHandler {
     }
     CF_EXPECT(KillAllCuttlefishInstances(
         /* clear_instance_dirs*/ options.clean_runtime_dir));
+
+    if (DirectoryExists("/tmp/acloud_cvd_temp/")) {
+      CF_EXPECT(RecursivelyRemoveDirectory("/tmp/acloud_cvd_temp/"));
```
Databean, line 163: Does it seem worth it to delete the contents of this directory without deleting the directory itself?

I could imagine a user changing the directory permissions of this to facilitate some multi-user interaction, though it would have to be reset on every reboot anyway since `/tmp` is not going to be persistent. Maybe not that significant in that context.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/reset.cpp
```diff
@@ -158,6 +158,11 @@ class CvdResetCommandHandler : public CvdCommandHandler {
     }
     CF_EXPECT(KillAllCuttlefishInstances(
         /* clear_instance_dirs*/ options.clean_runtime_dir));
+
+    if (DirectoryExists("/tmp/acloud_cvd_temp/")) {
```
Databean, line 162: As there are already [3 non-test occurrences](https://github.com/search?q=repo%3Agoogle%2Fandroid-cuttlefish%20acloud_cvd_temp&type=code) of this string, maybe worth making a shared constant in a new file somewhere?

3405691582: That'll be the next pr after this one. ;)

base/cvd/cuttlefish/host/commands/cvd/cli/commands/reset.cpp
3405691582: It may not be worth the effort. If we want to support the multi-user case better, then we probably need to improve the directory structure here and maybe put some of these files elsewhere, but that's a larger change. 

But I'm happy to make the adjustment if you think it's worthwhile.

Databean: Changing the structure has other downsides, it breaks compatibility with [python acloud](https://cs.android.com/android/platform/superproject/main/+/main:tools/acloud/list/instance.py;l=221;drc=86219821d73808761a317e1b48a0c0cebf3c1fd5). Might as well delete it then.

PR #1490, opened 2025-08-04
Fixup json annotation for `cvd display list`

jmacnak:


Databean (approved):

frontend/src/host_orchestrator/api/v1/messages.go
```diff
@@ -191,11 +191,11 @@ type DisplayMode struct {
 type Display struct {
 	DPI           []int       `json:"dpi"`
 	Mode          DisplayMode `json:"mode"`
-	RefreshRateHZ int         `json:"refresh-rate"`
+	RefreshRateHZ int         `json:"refresh_rate_hz"`
```
Databean, line 194: This doesn't look consistent with oxygen? See http://shortn/_HtRMXuO8rf

Databean: My mistake, I mixed this up with `load_config.proto`.

jmacnak: Notes for posterity:

We have `refresh_rate_hz` in the `cvd create --display=<>` command line flags: https://github.com/google/android-cuttlefish/blob/main/base/cvd/cuttlefish/host/libs/config/display.cpp#L82 

but then `refresh_rate_hertz` for the `cvd load` https://github.com/google/android-cuttlefish/blob/7354bfd1ad910ab140859b6c94593feb9d423d7d/base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_config.proto#L94

This specifically is used to provide the host orchestrator interface to commands like `cvd display list` which is after you already have a device.

PR #1493, opened 2025-08-04
Remove the start_webrtc flag behavior

Databean:
Instead make the behavior dependent on the VMM.  webrtc is required to run with crosvm, and nonfunctional with gem5. I'm not sure if there is a dependency on it with qemu, so this leaves it running for qemu.

The flag and config values are left to support some interaction with python acloud.

https://cs.android.com/android/platform/superproject/main/+/main:tools/acloud/internal/lib/cvd_runtime_config.py;l=149;drc=70921dc34d0c06b760159a62f43258ab2ec7fd06

https://cs.android.com/android/platform/superproject/main/+/main:tools/acloud/create/local_image_local_instance.py;l=631;drc=e1d40e5e18750a85a623b15411bd0f72aa6082e2

The current behavior had some complex default logic that ultimately amounted to "enable it by default for crosvm and qemu+swiftshader", unless the user passes a flag value. The new behavior simplifies is it to "always run for crosvm and qemu".

Bug: b/436356145
Test: launch with crosvm
Test: launch with qemu

### DIFF
--- a/base/cvd/cuttlefish/host/commands/run_cvd/launch/streamer.cpp
+++ b/base/cvd/cuttlefish/host/commands/run_cvd/launch/streamer.cpp
@@ -268,7 +268,17 @@ class WebRtcServer : public virtual CommandSource,
 
   // SetupFeature
   bool Enabled() const override {
-    return sockets_.Enabled() && instance_.enable_webrtc();
+    if (!sockets_.Enabled()) {
+      return false;
+    }
+    switch (config_.vm_manager()) {
+      case VmmMode::kCrosvm:
+      case VmmMode::kQemu:
+        return true;
+      case VmmMode::kGem5:
+      case VmmMode::kUnknown:
+        return false;
+    }
   }
 
  private:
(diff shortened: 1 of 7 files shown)


base/cvd/cuttlefish/host/commands/run_cvd/launch/streamer.cpp
```diff
@@ -268,7 +268,7 @@ class WebRtcServer : public virtual CommandSource,
 
   // SetupFeature
   bool Enabled() const override {
-    return sockets_.Enabled() && instance_.enable_webrtc();
+    return sockets_.Enabled() && config_.vm_manager() != VmmMode::kGem5;
```
jemoreira: nit: This should be `&& (vmm == crosvm || vmm == qemu)` in case we add a new arch in the future.

jemoreira: Using a `switch` would probably be better as that would case a build failure if a new arch is added to the enum but not handled here.

Databean: Done.

Databean:
Thanks for reviewing!

PR #1494, opened 2025-08-05
Simplify the `Avb` class interface

Databean:
- Always provide an algorithm and key to the Avb class
- Merge `GetDefaultAvb()` into the default constructor of `Avb`

Bug: b/436367852


cjreynol (approved):

Databean:
Thanks for reviewing!

PR #1496, opened 2025-08-05
Expose GPT types from `image_aggregator.cc`

Databean:
Bug: b/436625775


Databean:
Thanks for reviewing!

PR #1497, opened 2025-08-05
Allow for locks to be acquired incrementally.

3405691582:
Since we currently statically allocate tap interfaces for the maximum number of instances we anticipate running, the current code tries to acquire a number of instance locks for the number of tap interfaces currently on the system.

If we want to only acquire tap interfaces when instances are created, this scheme has to be modified. Here, we provide some functionality to examine what locks have been acquired and to acquire additional locks.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.cpp
@@ -250,4 +250,16 @@ Result<void> InstanceLockFileManager::RemoveLockFile(int instance_num) {
   return {};
 }
 
+Result<std::set<InstanceLockFile>> InstanceLockFileManager::AcquireUnusedLocks(
+    unsigned int number) {
+  std::set<InstanceLockFile> more;
+  for (int i = 1; more.size() < number; i++) {
+    auto lock = CF_EXPECT(TryAcquireLock(i));
+    if (lock && CF_EXPECT(lock->Status()) == InUseState::kNotInUse) {
+      more.emplace(std::move(*lock));
+    }
+  }
+  return more;
+}
+
 }  // namespace cuttlefish
--- a/base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.h
+++ b/base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.h
@@ -51,6 +51,7 @@ class InstanceLockFileManager {
 
   Result<InstanceLockFile> AcquireLock(int instance_num);
   Result<std::set<InstanceLockFile>> AcquireLocks(const std::set<int>& nums);
+  Result<std::set<InstanceLockFile>> AcquireUnusedLocks(unsigned int number);
 
   Result<std::optional<InstanceLockFile>> TryAcquireLock(int instance_num);
   Result<std::set<InstanceLockFile>> TryAcquireLocks(const std::set<int>& nums);


base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.cpp
```diff
@@ -250,4 +252,37 @@ Result<void> InstanceLockFileManager::RemoveLockFile(int instance_num) {
   return {};
 }
 
+Result<std::set<int>> InstanceLockFileManager::GetLocks() {
+  std::stringstream path;
+  path << InstanceLocksPath();
```
Databean: Does this need the `std::stringstream`? `InstanceLocksPath()` is already a `std::string`, so this can be

```c++
std::string path = InstanceLocksPath();
```
or
```c++
if (!DirectoryExists(InstanceLocksPath())) { ... }
```

base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.cpp
```diff
@@ -250,4 +252,37 @@ Result<void> InstanceLockFileManager::RemoveLockFile(int instance_num) {
...
+
+  std::set<int> r;
+  std::vector<std::string> contents = CF_EXPECT(DirectoryContents(path.str()));
+  for (const std::string& filepath : contents) {
+    // "local-instance-%d.lock"
+    if (!absl::StartsWith(filepath, "local-instance-") ||
+        !absl::EndsWith(filepath, ".lock")) {
+      continue;
+    }
+    auto tokens = android::base::Tokenize(filepath, "-.");
```
Databean: If this uses [`absl::ConsumePrefix`](https://github.com/abseil/abseil-cpp/blob/81dff3b24feae5d6ac39026081dd08e13d23dbba/absl/strings/strip.h#L48) and `absl::ConsumeSuffix` it can combine the steps of validation and parsing. That also wouldn't accept strange names like `local-instance-10-other-nonsense.lock`.

base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.cpp
```diff
@@ -250,4 +252,37 @@ Result<void> InstanceLockFileManager::RemoveLockFile(int instance_num) {
...
+    r.insert(std::stoi(tokens[2]));
```
Databean: nit: [`absl::SimpleAtoi`](https://github.com/abseil/abseil-cpp/blob/81dff3b24feae5d6ac39026081dd08e13d23dbba/absl/strings/numbers.h#L65) to validate that this string is actually numeric?

base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.cpp
```diff
@@ -250,4 +252,37 @@ Result<void> InstanceLockFileManager::RemoveLockFile(int instance_num) {
...
+Result<std::set<InstanceLockFile>> InstanceLockFileManager::LockAnother(
+    unsigned int number) {
+  std::set<int> all = CF_EXPECT(GetLocks());
+
+  std::set<int> more;
+  for (int i = 1; more.size() <= number; i++) {
+    if (all.find(i) != all.end()) continue;
+    more.insert(i);
+  }
+  return CF_EXPECT(TryAcquireLocks(more));
```
Databean: Note that `TryAcquireLocks` may return [less than the number of requested locks](https://github.com/google/android-cuttlefish/blob/94dcd9af969991f1be742284e8ace80e48d8f2d3/base/cvd/cuttlefish/host/commands/cvd/instances/lock/lock_file.cpp#L154). It could happen in this case if there is a race with another process invoking `LockAnother`, if both processes invoke `open(..., O_CREAT)` and the other process wins the race to `flock` the file.

base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.h
```diff
@@ -65,6 +65,9 @@ class InstanceLockFileManager {
   // calling this. It is a quick fix for b/316824572
   Result<void> RemoveLockFile(int instance_num);
 
+  Result<std::set<int>> GetLocks();
+  Result<std::set<InstanceLockFile>> LockAnother(unsigned int number);
```
jemoreira: nit: "Another" sounds like only one more. `LockMore` ?

3405691582: I just overloaded `AcquireLocks`, I couldn't think of a better option.

base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.cpp
3405691582: Good point. Used AcquireLocks instead.

base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.cpp
3405691582: Done.

base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.cpp
3405691582: Done.

base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.cpp
3405691582: Done.

base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.h
Databean: `AllocateLocks` since it is creating new lock files?

base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.cpp
Databean: The downside of using `AcquireLocks` is that it opens the risk of deadlocks. I think this ordering between two concurrent invocations could cause a deadlock:

- A: `AcquireLocks(3)`
- A: `GetLocks()` reports N locks
- A: `AcquireLocks({N+1, N+2, N+3})`
- A: `AcquireLock(N+1)`
- B: `AcquireLocks(3)`
- B: `GetLocks()` reports N+1 locks
- B: `AcquireLocks({N+2, N+3, N+4})`
- B: `AcquireLock(N+2)`
- A: `AcquireLock(N+2)`

Is it important that locks files are both allocated and acquired in the same operation?

Databean: On second thought: `A` is blocked but `B` can still run to completion.

base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.h
3405691582: The lock files are opened `O_CREAT`, so creation can happen with the existing Acquire methods, though.

base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.cpp
3405691582: While deadlocking may be unlikely, there is a point that concurrent calls to AcquireLocks may mean that a `cvd create` issued by multiple users could race on an instance lock, which would lead to the suboptimal situation where one `cvd create` would block on the other instance's lifetime. 

I think it's important to have `AcquireLocks(int)` uphold the contract that it returns the number of instance locks that were requested, so I've tweaked this again. Here, we just keep trying higher lock numbers if a `TryAcquireLock` call failed.

3405691582:
@jemoreira any other comments?

base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.cpp
```diff
@@ -250,4 +253,42 @@ Result<void> InstanceLockFileManager::RemoveLockFile(int instance_num) {
...
+  return r;
+}
+
+Result<std::set<InstanceLockFile>> InstanceLockFileManager::AcquireLocks(
+    unsigned int number) {
+  std::set<int> all = CF_EXPECT(GetLocks());
+
+  std::set<InstanceLockFile> more;
+  for (int i = 1; more.size() < number; i++) {
+    if (all.find(i) != all.end()) continue;
```
jemoreira: Why is this skipping the existing files? The contract so far is that the `id` is "locked" if the file exists and has an `"I"` in it. An empty file or a file with an `"N"` are not holding the lock.
We can't just change that behavior because it's shared with `acloud`.

3405691582: It seemed like acquiring new lock files that would be guaranteed to be unused would be simpler and a slight optimization. In fact, if we should be testing potentially existing lock files to see if they are unused, then we don't need `GetLocks` at all. 

I've also adjusted the name to better match the intent: i.e., that we're acquiring unused locks.

jemoreira: That may work only if everyone working with those locks follows that same idea, but like I said, these locks are shared with `acloud` and it doesn't do that.

3405691582: Fair enough. To be clear, I've made the change to remove `GetLocks` and only attempted to acquire locks that were not in use.

jemoreira (approved):

PR #1500, opened 2025-08-06
Mark 1.18 as stable

jmacnak:
Bug: b/435727090


Databean:
Did the Oxygen release from last week get sent to production? I only see it running on beta.

jmacnak:
Abandon'ing this. Due to timing, Oxygen just went ahead with the 1.19 release.

PR #1505, opened 2025-08-07
cvdalloc binary.

3405691582:
This process--cvdalloc--will eventually be invoked from run_cvd, intended to be invoked before the vmm starts. Because modifying the network requires elevated permissions, we must be setuid root. Network resources are allocated using the existing allocd implementation we exposed earlier.

We must ensure that interfaces are set up and available before the vmm starts. To synchronize around this, run_cvd will call socketpair(2) twice for signaling important events (essentially, using these like semaphores), the first event being that cvdalloc has finished allocating resources and thus the vmm can properly start, and the other event being that the vmm has exited and cvdalloc can tear down the resources that it had originally created.

The run_cvd process manages the lifecycle of the instance and is responsible for starting and stopping ancillary binaries--such as cvdalloc--that make the guest perform as intended. As the instance is being stopped, there are mechanisms that allow us to cleanly stop cvdalloc so run_cvd could correctly signal that resources could be torn down. However, since shutdown order is not necessarily guaranteed, it can be difficult to ensure that cvdalloc's shutdown signal is received after the vmm exits.

Instead, cvdalloc forks and then blocks on the teardown socketpair. Once run_cvd--which holds the other end of the socketpair--exits, that end is implictly shut down, which unblocks cvdalloc to complete the teardown. This ensures teardown executes once run_cvd exits. This is not a guarantee that teardown executes once the vmm exits; it may be the case that run_cvd exits abnormally, triggering cvdalloc to attempt to shutdown, but the vmm may still be running. For now, we ignore this case, but later we can use retries to try and mitigate this problem.

This version is very basic which will be iterated on in future commits; it is not wired up to anything yet.


base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -0,0 +1,143 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include <string.h>
+#include <unistd.h>
+
+#include <android-base/logging.h>
+
+#include "allocd/alloc_utils.h"
+#include "cuttlefish/common/libs/fs/shared_fd.h"
```
Databean: nit: in the [categorization of header includes](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes), abseil and android-base are both "7. Other libraries' `.h` files" while these are "9. Your project's `.h` files."

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -0,0 +1,143 @@
...
+  if (pid > 0) {
+    int r = setuid(0);
+    if (r == -1) {
+      LOG(ERROR) << "Couldn't setuid root: " << strerror(errno);
+      return 1;
+    }
+
+    LOG(INFO) << "cvdalloc: allocating network resources";
+
+    cuttlefish::CreateBridge("cvd-br");
```
Databean: nit: consider defining a substitute `main` function inside the `cuttlefish` namespace so that it doesn't need to use `cuttlefish::` all over, similar to `assemble_cvd`:

https://github.com/google/android-cuttlefish/blob/613f81afb382cf7fb8b0d8793b9a4f0ab0986981/base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd.cc#L680

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -0,0 +1,143 @@
...
+    if (!sock->IsOpen()) {
+      LOG(ERROR) << "cvdalloc: allocate socket is closed: " << sock->StrError();
+      return 1;
+    }
+    sock->Shutdown(SHUT_RDWR);
+
+    return 0;
+  }
+
+  if (pid == 0) {
```
Databean: Consider writing this as a [guard](https://en.wikipedia.org/wiki/Guard_(computer_science)) so it doesn't need as much nesting.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
3405691582: Fixed.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
3405691582: Done.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
3405691582: I've dropped the check. (Reordering parent and child will be awkward; it reads nicer to have allocate at the top of the file and teardown at the bottom).

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+  if (absl::GetFlag(FLAGS_id) == 0 ||
+      absl::GetFlag(FLAGS_teardown_event_socket) == 0 ||
+      absl::GetFlag(FLAGS_allocate_event_socket) == 0) {
+    usage();
+    return 1;
+  }
+
+  /*
+   * Explicit setuid calls seem to be required.
```
jemoreira, line 86: The setuid bit only changes the effective user id, not the real user id. You can check those values with `geteuid()` and `getuid()` respectively.

The effective user id is the one that should be used for permissions. Why do you need to set the real id as well?

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -0,0 +1,147 @@
...
+   * Save our current uid, so we can restore it to drop privileges later.
+   */
+  uid_t orig = getuid();
+
+  int id = absl::GetFlag(FLAGS_id);
+
+  pid_t pid = fork();
+  if (pid < 0) return -1;
+
+  if (pid > 0) {
```
jemoreira: nit: This will be easier to read if you put the following code in two different functions, one to execute in the parent and the other in the child.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -0,0 +1,147 @@
...
+    CreateEthernetIface(absl::StrFormat("cvd-etap-%02d", id), "cvd-br", true,
+                        true, false);
+
+    r = setuid(orig);
+    if (r == -1) {
+      LOG(WARNING) << "cvdalloc: couldn't drop privileges";
+      return 1;
+    }
+
+    auto sock = SharedFD::Dup(absl::GetFlag(FLAGS_allocate_event_socket));
```
jemoreira: The file descriptor from `FLAGS_allocate_event_socket` needs to be closed explicitly after `SharedFD::Dup`

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -0,0 +1,147 @@
...
+    if (!sock->IsOpen()) {
+      LOG(ERROR) << "cvdalloc: allocate socket is closed: " << sock->StrError();
+      return 1;
+    }
+    sock->Shutdown(SHUT_RDWR);
```
jemoreira, line 113: What's this for? IIUC the next line just returns from this function which causes the process to exit, so the file descriptor will be closed completely anyways.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+    return 0;
+  }
+
+  LOG(INFO) << "cvdalloc (teardown): waiting to teardown";
+
+  auto sock = SharedFD::Dup(absl::GetFlag(FLAGS_teardown_event_socket));
```
jemoreira: Need to close `FLAGS_teardown_event_socket` after confirming `sock` is open

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -0,0 +1,147 @@
...
+  if (!sock->IsOpen()) {
+    LOG(ERROR) << "cvdalloc (teardown): teardown socket is closed: "
+               << sock->StrError();
+    return 1;
+  }
+  int i;
+  if (sock->Read(&i, sizeof(i)) < 0) {
+    sock->Close();
```
jemoreira: It's rarely needed to explicitly close a `SharedFD`, they close automatically when the destructor is called on the last `SharedFD` object pointing to the underlying fd.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -0,0 +1,147 @@
...
+    LOG(ERROR) << "Couldn't setuid root: " << strerror(errno);
+    return 1;
+  }
+
+  LOG(INFO) << "cvdalloc (teardown): tearing down resources";
+
+  DestroyMobileIface(absl::StrFormat("cvd-mtap-%02d", id), id, kMobileIp);
+  DestroyMobileIface(absl::StrFormat("cvd-wtap-%02d", id), id, kWirelessIp);
+  DestroyEthernetIface(absl::StrFormat("cvd-etap-%02d", id), true, true, false);
+  DestroyBridge("cvd-br");
```
jemoreira: This list is repeated here and in the parent's code. This should be encapsulated somehow to avoid that duplication.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -0,0 +1,147 @@
...
+  auto sock = SharedFD::Dup(absl::GetFlag(FLAGS_teardown_event_socket));
+  if (!sock->IsOpen()) {
+    LOG(ERROR) << "cvdalloc (teardown): teardown socket is closed: "
+               << sock->StrError();
+    return 1;
+  }
+  int i;
+  if (sock->Read(&i, sizeof(i)) < 0) {
+    sock->Close();
+    return 1;
```
jemoreira: Shouldn't it still be torn down anyways?

I think that no matter what you do, there will always be a way for the parent to create the interfaces and the child to not be notified about it, which would cause a resource leak. Unless you create the interfaces in the parent **before** calling `fork()`, that way the child only executes if the interfaces were successfully created.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
Databean: How about separate functions for parent and child?

```c++
if (pid > 0) {
  return AllocateResources(pid);
} else {
  return WaitForTermination();
}
```

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -0,0 +1,147 @@
...
+#include "absl/flags/flag.h"
+#include "absl/flags/parse.h"
+#include "absl/strings/str_format.h"
+
+#include "allocd/alloc_utils.h"
+#include "cuttlefish/common/libs/fs/shared_fd.h"
+
+ABSL_FLAG(int, id, 0, "Id");
+ABSL_FLAG(int, teardown_event_socket, 0, "Socket");
+ABSL_FLAG(int, allocate_event_socket, 0, "Socket");
```
Databean: Do these need to be separate sockets? Presumably if `run_cvd` closes `allocate_event_socket` it also isn't interested in using the resources.

Databean (approved):

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -0,0 +1,147 @@
...
+    int r = setuid(0);
+    if (r == -1) {
+      LOG(ERROR) << "Couldn't setuid root: " << strerror(errno);
+      return 1;
+    }
+
+    LOG(INFO) << "cvdalloc: allocating network resources";
+
+    CreateBridge("cvd-br");
+    CreateMobileIface(absl::StrFormat("cvd-mtap-%02d", id), id, kMobileIp);
```
jemoreira: This is probably work for another PR as it potentially involves changing allocd's interface.

These functions should return a `Result` as creating a network interface could theoretically fail. Handling of those failures should involve tearing down the interfaces already created. I think the best way to do this, while addressing my other comment about code duplication, is to have these functions return `Result<SomeInterfaceObject>` where `SomeInterfaceObject` uses RAII to deallocate the underlying resources when its destructor runs.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
jemoreira: If your intention is to let the caller know the other end of the file descriptor closed this won't work because the child also has a copy the fd and `run_cvd` won't be notified until all copies of this end of the connection are closed.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
3405691582: I agree that it should not be necessary, but it appears to be required since it won't work otherwise: the ip command's underlying TUNSETIFF ioctl will fail with EPERM. I managed to track this down to a reverted kernel commit that affected the way that permissions are examined that might be the culprit, but didn't actually verify it since I had better things to do and didn't want to jump down that rabbit hole.

This point will be moot shortly after this anyway, since there's a pr soon in the pipeline to use Linux capabilities instead of setuid.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
3405691582: There was a simplicity in using two sockets, but I think it shouldn't be too bad to just use one. Done.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
3405691582: Done.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
3405691582: Done; I had forgotten some things about how dup worked.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
3405691582: It worked as intended given `shutdown` changes the behavior of all fds associated with the socket, but it's moot now given the behavioral change.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
3405691582: Removed.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
3405691582: Good catch; added `Teardown` calls in each guard clause.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
3405691582: I think the `Allocate`/`Teardown` functions are sufficient for now; the list is too small and not simple enough for deduplicating the list. That can change if this list grows or changes in the future.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
3405691582: Done.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
Databean: They do return `bool`s already, maybe worth at least logging failures here or some kind of fail-fast logic even without full cleanup.

3405691582: For now, I've peppered in `CF_EXPECT`s across `Allocate` and call `Teardown` if any of them fail. 

(There is no `CF_EXPECT` in `Teardown` yet, since we might as well try and tear down the other interfaces if one of them fails; retrying and other mitigation for leaking resources will come in a later commit.)

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -0,0 +1,145 @@
...
+    Teardown(id);
+    return CF_ERRNO("Allocation failed");
+  }
+
+  auto sock = SharedFD::Dup(absl::GetFlag(FLAGS_socket));
+  if (!sock->IsOpen()) {
+    Teardown(id);
+    return CF_ERRNO("cvdalloc: socket is closed: " << sock->StrError());
+  }
+  r = close(absl::GetFlag(FLAGS_socket));
```
jemoreira: In practice this close can only fail with `EINTR` since the dup above succeeded. That could be handled by wrapping the close call like this: `TEMP_FAILURE_RETRY(close(...))`

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -0,0 +1,145 @@
...
+    Teardown(id);
+    return CF_ERRNO("Read: " << strerror(errno));
+  }
+
+  Teardown(id);
+
+  r = setuid(orig);
+  if (r == -1) {
+    return CF_ERRNO("cvdalloc: couldn't drop privileges: " << strerror(errno));
+  }
```
jemoreira: nit: you can use a [scoped guard](https://github.com/google/android-cuttlefish/blob/162bc1ff05e89a4d071747108ac2f3d3b7bf797d/base/cvd/libbase/include/android-base/scopeguard.h#L27) immediately after the `setuid(0)` call to ensure this is done before return.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -0,0 +1,145 @@
...
+  int id = absl::GetFlag(FLAGS_id);
+
+  int r = setuid(0);
+  if (r == -1) {
+    return CF_ERRNO("Couldn't setuid root: " << strerror(errno));
+  }
+
+  Result<void> result = Allocate(id);
+  if (!result.ok()) {
+    Teardown(id);
```
jemoreira: nit: All the calls to `Teardown(id)` could be handled with a [scope guard](https://github.com/google/android-cuttlefish/blob/162bc1ff05e89a4d071747108ac2f3d3b7bf797d/base/cvd/libbase/include/android-base/scopeguard.h#L27) created just before the call to `Allocate(id)`. It will ensure that `Teardown(id)` will be called before returning, whether it's an early return or a successful one.

jemoreira (approved):

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
3405691582: Done.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
3405691582: Done.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
3405691582: Done.

PR #1515, opened 2025-08-09
Add a DisjointRangeSet type representing which parts of a space are covered

Databean:
Bug: b/437447876

### DIFF
--- a/base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
+++ b/base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
@@ -1,0 +1,155 @@
+//
+// Copyright (C) 2025 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#include "cuttlefish/host/libs/zip/disjoint_range_set.h"
+
+#include <stdint.h>
+
+#include <algorithm>
+#include <memory>
+#include <set>
+#include <utility>
+#include <vector>
+
+#include <android-base/logging.h>
+
+namespace cuttlefish {
+namespace {
+
+class Range {
+ public:
+  Range(uint64_t start, uint64_t end) : start_(start), end_(end) {}
+
+  bool Contains(const Range& other) const {
+    return start_ <= other.start_ && end_ >= other.end_;
+  }
+
+  bool CanMerge(const Range& other) const {
+    return start_ <= other.end_ && other.start_ <= end_;
+  }
+
+  // Preconditions: `CanMerge(other)`
+  Range Merge(const Range& other) const {
+    return Range(std::min(start_, other.start_), std::max(end_, other.end_));
+  }
+
+  bool operator<(const Range& other) const { return start_ < other.start_; }
+  bool operator==(const Range& other) const {
+    return start_ == other.start_ && end_ == other.end_;
+  }
+
+  std::pair<uint64_t, uint64_t> AsPair() const {
+    return std::make_pair(start_, end_);
+  }
+
+ private:
+  uint64_t start_;
+  uint64_t end_;
+};
+
+using Iterator = std::set<Range>::iterator;
+
+}  // namespace
+
+struct DisjointRangeSet::Impl {
+  std::set<Range> ranges_;
+};
+
+DisjointRangeSet::DisjointRangeSet() : impl_(new Impl) {}
+
+DisjointRangeSet::DisjointRangeSet(const DisjointRangeSet& other)
+    : impl_(new Impl) {
+  impl_->ranges_ = other.impl_->ranges_;
+}
+
+DisjointRangeSet::DisjointRangeSet(DisjointRangeSet&& other) : impl_(new Impl) {
+  std::swap(impl_, other.impl_);
+}
+
+DisjointRangeSet& DisjointRangeSet::operator=(const DisjointRangeSet& other) {
+  impl_->ranges_ = other.impl_->ranges_;
+  return *this;
+}
+
+DisjointRangeSet& DisjointRangeSet::operator=(DisjointRangeSet&& other) {
+  impl_ = std::move(other.impl_);
+  other.impl_.reset(new Impl);
+  return *this;
+}
+
+DisjointRangeSet::~DisjointRangeSet() = default;
+
+bool DisjointRangeSet::ContainsRange(uint64_t start, uint64_t end) const {
+  CHECK_LE(start, end) << "Invalid range: expected start <= end";
+
+  const std::set<Range>& ranges = impl_->ranges_;
+
+  if (ranges.empty()) {
+    return false;
+  }
+  Range range(start, end);
+
+  Iterator range_it = ranges.upper_bound(range);
+  if (range_it != ranges.begin()) {
+    range_it--;
+  }
+  return range_it->Contains(range);
+}
+
+void DisjointRangeSet::InsertRange(uint64_t start, uint64_t end) {
+  CHECK_LE(start, end) << "Invalid range: expected start <= end";
+
+  if (ContainsRange(start, end)) {
+    return;
+  }
+
+  std::set<Range>& ranges = impl_->ranges_;
+  Range new_range(start, end);
+
+  // Merge in ranges that start after the newly inserted range starts, if the
+  // newly inserted range would overlap with them. `it` always points to the
+  // range with the next highest starting location.
+  Iterator it = ranges.upper_bound(new_range);
+  while (it != ranges.end() && new_range.CanMerge(*it)) {
+    new_range = new_range.Merge(*it);
+    it = ranges.erase(it);
+  }
+  // Merge a range that start before the newly inserted range, if the newly
+  // inserted range overlaps with the previous one. `it` initially points at the
+  // range with the next highest starting location, and `--it` moves it to point
+  // to the element directly before that. This works differently than the
+  // forwards merge because while `it` can safely be the `end()` sentinel past
+  // the end of the collection, there is no equivalent before-the-start sentinel
+  // value. There can be at most one range before that overlaps with this one.
+  if (it != ranges.begin() && new_range.CanMerge(*(--it))) {
+    new_range = new_range.Merge(*it);
+    it = ranges.erase(it);
+  }
+  ranges.emplace(new_range);
+}
+
+std::vector<std::pair<uint64_t, uint64_t>> DisjointRangeSet::AllRanges() const {
+  std::vector<std::pair<uint64_t, uint64_t>> pairs;
+  pairs.reserve(impl_->ranges_.size());
+  for (const Range& range : impl_->ranges_) {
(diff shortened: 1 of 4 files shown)


base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
```diff
@@ -0,0 +1,162 @@
...
+
+DisjointRangeSet& DisjointRangeSet::operator=(DisjointRangeSet&& other) {
+  impl_.reset(new Impl);
+  std::swap(impl_, other.impl_);
+  return *this;
+}
+
+DisjointRangeSet::~DisjointRangeSet() = default;
+
+bool DisjointRangeSet::ContainsRange(uint64_t start, uint64_t end) const {
```
jemoreira, line 93: Please handle the case where `end` <= `start`. I'm guessing a `CHECK` would suffice.

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set_test.cc
```diff
@@ -0,0 +1,122 @@
...
+TEST(DisjointRangeSet, SingleMember) {
+  DisjointRangeSet set;
+
+  set.InsertRange(5, 10);
+
+  EXPECT_FALSE(set.ContainsRange(4, 5));
+  EXPECT_FALSE(set.ContainsRange(4, 6));
+  EXPECT_TRUE(set.ContainsRange(5, 6));
+  EXPECT_TRUE(set.ContainsRange(7, 9));
+  EXPECT_FALSE(set.ContainsRange(9, 11));
```
jemoreira, line 46: Add `EXPECT_TRUE(set.ContainsRange(5, 10))` and `EXPECT_TRUE(set.ContainsRange(9, 10))` to cover the full range and smaller range with same end.

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set_test.cc
```diff
@@ -0,0 +1,122 @@
...
+  EXPECT_TRUE(set.ContainsRange(16, 18));
+  EXPECT_TRUE(set.ContainsRange(7, 17));
+
+  EXPECT_FALSE(set.ContainsRange(18, 22));
+
+  std::vector<std::pair<uint64_t, uint64_t>> expected = {{5, 20}};
+  EXPECT_EQ(set.AllRanges(), expected);
+}
+
+TEST(DisjointRangeSet, MergingSameStart) {
```
jemoreira, line 120: add `MergingSameEnd` too

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
```diff
@@ -0,0 +1,162 @@
...
+
+  bool Contains(uint64_t pnt) const { return start_ <= pnt && pnt < end_; }
+  bool Contains(const Range& other) const {
+    return start_ <= other.start_ && end_ >= other.end_;
+  }
+
+  std::optional<Range> MergeIfPossible(const Range& other) const {
+    bool intersects = Contains(other.start_) || Contains(other.end_) ||
+                      other.Contains(start_) || other.Contains(end_);
+    bool adjacent = start_ == other.end_ || end_ == other.start_;
```
jemoreira: nit:
`!intersects && !adjacent` is equivalent to `this.start_ > other.end_ || other.start_ > this.end_`. i.e. For two ranges not to touch each other one must start **after** the other ends.

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
```diff
@@ -0,0 +1,162 @@
...
+  std::swap(impl_, other.impl_);
+}
+
+DisjointRangeSet& DisjointRangeSet::operator=(const DisjointRangeSet& other) {
+  impl_->ranges_ = other.impl_->ranges_;
+  return *this;
+}
+
+DisjointRangeSet& DisjointRangeSet::operator=(DisjointRangeSet&& other) {
+  impl_.reset(new Impl);
```
jemoreira: why is this needed here but not in the constructor on line 79?

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
```diff
@@ -0,0 +1,162 @@
...
+void DisjointRangeSet::InsertRange(uint64_t start, uint64_t end) {
+  if (ContainsRange(start, end)) {
+    return;
+  }
+
+  std::set<Range>& ranges = impl_->ranges_;
+  Range new_range(start, end);
+
+  Iterator it = ranges.emplace(new_range).first;
+  if (std::optional<Range> merge = it->MergeIfPossible(new_range); merge) {
```
jemoreira: It's not self evident why this is needed, especially because `MergeIfPossible` will always merge here.

This could be avoided by falling back to comparing `end_` in the range comparison operator when `start_` is the same.

Alternatively, consider using the returned boolean value for the `if` conditions and commenting that a range with the same `start_` would prevent insertion of this range.

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
```diff
@@ -0,0 +1,162 @@
...
+    if (std::optional<Range> merge = it->MergeIfPossible(new_range); merge) {
+      it--;
+      it = ranges.erase(it);
+      ranges.erase(it);
+      it = ranges.emplace(new_range = *merge).first;
+      it++;
+    } else {
+      break;
+    }
+  }
```
jemoreira: This would be simpler to implement and easier to read if you remove and merge all relevant ranges before inserting the new (merged) range:
```
auto it = ranges.upper_bound(new_range);
while (it != ranges.end() && new_range.Touches(*it)) {
    new_range = new_range.MergeIfPossible(*it);
    it = ranges.erase(it)
}
if (it != ranges.begin() && new_range.Touches(*(--it))) {
    new_range = new_range.MergeIfPossible(*it);
    it = ranges.erase(it);
}
ranges.emplace(new_range);
```

jemoreira (approved):

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
Databean: The move constructor `DisjointRangeSet(DisjointRangeSet&&)` uses a member initializer list `: impl_(new Impl)` which is unavailable on this move assignment operator for an already-constructed instance, so it does the equivalent `impl_.reset(new Impl)`.

Since this is assigning to both anyway, changed it to use regular assignment rather than being fancy with `std::swap`.

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
Databean: Done.

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set_test.cc
Databean: Done.

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set_test.cc
Databean: Done.

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
Databean: Thanks, that's a lot cleaner.

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
Databean: Removed this in favor of using [implementation you proposed](https://github.com/google/android-cuttlefish/pull/1515/files/724d6a009955b5e7eee5d67ae1519d90b2cc82e3#r2268137711).

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
Databean: It looks a lot nicer for sure but it did take a while to convince myself this is correct. Changed to use this implementation.

Databean (commented):
Thanks for reviewing!

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
```diff
@@ -0,0 +1,145 @@
...
+  uint64_t start_;
+  uint64_t end_;
+};
+
+using Iterator = std::set<Range>::iterator;
+
+}  // namespace
+
+struct DisjointRangeSet::Impl {
+  std::set<Range> ranges_;
```
jemoreira, line 66: nit: This could be `std::set<const Range>` now to make sure those ranges are not modified while in the set. Modifying the `start_` member is particularly dangerous because it risks breaking the set.

jemoreira (approved):

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
```diff
@@ -0,0 +1,145 @@
...
+
+  Iterator range_it = ranges.upper_bound(range);
+  if (range_it != ranges.begin()) {
+    range_it--;
+  }
+  return range_it->Contains(range);
+}
+
+void DisjointRangeSet::InsertRange(uint64_t start, uint64_t end) {
+  CHECK_LE(start, end) << "Invalid range: expected start <= end";
```
jemoreira, line 111: nit: Could skip this since `ContainsRange` does this check

Databean: But then `DisjointRangeSet` becomes responsible for maintaining the function preconditions of `Range`, though admittedly it is already a private type...

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
Databean: For some reason this breaks the `impl_->ranges_ = other.impl_->ranges_;` lines in the copy constructor and copy assignment operator. I suspect it is related to this:

https://en.cppreference.com/w/cpp/container/set/operator=.html

> ... the memory owned by `*this` may be reused when possible. In any case, the elements originally belonging to `*this` may be either destroyed or replaced by element-wise copy-assignment.

Though the implementation I've written using these types doesn't actually need the copy constructor / copy assignment operator, `gtest` uses it in the assertions that compare equality.

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
Databean: Left a comment about the precondition.

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
Databean: I tracked down where this comes from in the standard. In c++14 23.2.1 Table 99 -- Allocator-aware container requirements, it writes:

> ...X denotes an allocator-aware container class with a value_type of T... a and b denote non-const lvalues of type X, t denotes an lvalue or a const rvalue of type X...
>
> - Expression: `a = t`
> - Return type: `X&`
> - Requires: T is CopyInsertable into X and CopyAssignable.
> - post: a == t
> - Complexity: linear

Databean: This leaves 3 options:

1. Make it not `const`
2. Remove the copy constructors, and work around it in the unit test file
3. Manually implement a slower copy constructor

jemoreira: not const is fine

PR #1519, opened 2025-08-11
Create a type for the `--mcu_config_path` flag

Databean:
As with other flags using `SetCommandLineOptionWithMode`, this flag was used as global mutable state.

Bug: b/437989846


base/cvd/cuttlefish/host/commands/assemble_cvd/flags/mcu_config_path.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include "cuttlefish/common/libs/utils/result.h"
+
+namespace cuttlefish {
+
+class McuConfigPathFlag {
+ public:
+  static McuConfigPathFlag FromGlobalGflags();
+
+  Result<Json::Value> JsonForIndex(size_t) const;
```
jmacnak, line 33: minor nit: i think you want `#include <cstddef>` for `size_t`

jmacnak (approved):

base/cvd/cuttlefish/host/commands/assemble_cvd/flags/mcu_config_path.h
Databean: Good catch, thanks.

Since the [style guide](https://google.github.io/styleguide/cppguide.html#Integer_Types) mentions using the types without the `std::` prefix, used `<stddef.h>` rather than `<csstddef>`.

Databean:
Thanks for reviewing!

PR #1521, opened 2025-08-12
Create a flag to represent `--super_image`

Databean:
There was an implicit ordering dependency involving global state that `ResolveInstanceFiles` needs to be called before
`DiskImageFlagsVectorization`.

Bug: b/438296342


Databean:
Thanks for reviewing!

PR #1523, opened 2025-08-13
Add a LazilyLoadedFile type that loads fragments from external source as needed

Databean:
Bug: b/438315638


base/cvd/cuttlefish/host/libs/zip/lazily_loaded_file.cc
```diff
@@ -0,0 +1,159 @@
...
+  CF_EXPECT_EQ(contents_file_->LSeek(seek_pos_, SEEK_SET), seek_pos_,
+               contents_file_->StrError());
+  if (already_downloaded_.ContainsRange(seek_pos_, seek_pos_ + size)) {
+    ssize_t data_read = contents_file_->Read(data, size);
+    CF_EXPECT_GE(data_read, 0, contents_file_->StrError());
+    seek_pos_ += data_read;
+    return data_read;
+  }
+  CF_EXPECT(callback_->Seek(seek_pos_));
+  size_t data_read = CF_EXPECT(callback_->Read(data, size));
```
jemoreira, line 159: This can potentially read again a portion that has already been read if we already have part of the requested range.

To avoid multiple reads of the same range we could add functionality to `DisjointRangeSet` to return the first start and end of range it has at or after a given position. With that in hand we can request more data if we don't have the beginning of the range to read and then return as much data as we have, but possibly less than the caller asked for.

For example, let's say we have [3, 5) and [7, 10) and we're asked to read 10 bytes at position 0. Then we read with the callback [0, 3) and return to the user [0,5). They want 10 bytes, so they read again 5 bytes at position 5, we download [5, 7) and finally return the bytes from [5, 10).

Databean: Added a new interface `std::optional<uint64_t> EndOfContainingRange(uint64_t start) const;` and used it here.

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
```diff
@@ -32,7 +33,7 @@ class Range {
   Range(uint64_t start, uint64_t end) : start_(start), end_(end) {}
 
   bool Contains(const Range& other) const {
-    return start_ <= other.start_ && end_ >= other.end_;
+    return start_ <= other.start_ && end_ >= other.end_ && other.start_ < end_;
```
jemoreira: Is this addition really needed? The first two conditions guarantee that `end_ >= other.end_` and it's known that `other.end_ > other.start_` for it to be a valid interval, so `end_ > other.start_` should be true if the first two are also true.

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
```diff
@@ -139,6 +142,20 @@ void DisjointRangeSet::InsertRange(uint64_t start, uint64_t end) {
   ranges.emplace(new_range);
 }
 
+std::optional<uint64_t> DisjointRangeSet::EndOfContainingRange(
+    uint64_t start) const {
+  const std::set<Range>& ranges = impl_->ranges_;
+
+  Range test(start, start);
```
jemoreira: Should this be `(start, start + 1)`?

base/cvd/cuttlefish/host/libs/zip/lazily_loaded_file.cc
jemoreira: This still may download the same data more than once. In the example I originally provided it will download the entire [0, 10) range even though it already has parts of it.

This minimizes the number of calls to the read callback as opposed to how much data is gotten from it. If that's the objective, please add a comment that mentions it.

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
Databean: The problem I was running into is that the ranges are supposed to be right-open intervals, and the tests were reporting `[5, 10)]` contains `[10, 10)`.

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
Databean: Yeah maybe that's why I needed that extra conditional...

base/cvd/cuttlefish/host/libs/zip/lazily_loaded_file.cc
Databean: The objective is to minimize round trips against the input source rather than minimizing bandwidth.

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
Databean: Removed this since the issue was the `(start, start)` range below.

base/cvd/cuttlefish/host/libs/zip/disjoint_range_set.cc
Databean: Fixed

base/cvd/cuttlefish/host/libs/zip/lazily_loaded_file.cc
Databean: Added a comment.

Databean:
Thanks for reviewing!

PR #1524, opened 2025-08-13
Split out a lot of `//cuttlefish/host/commands/modem_simulator:modem_simulator_lib`

Databean:
The remaining 5 files in `modem_simulator_lib` have deeper circular dependencies with each other.

Some circular dependencies were resolved by splitting apart header and implementation files, and also by creating a virtual superclass of `ModemSimulator` that can be included without pulling in the dependencies of the implementation.

Bug: b/438343307


ser-io (approved):

Databean:
Thanks for reviewing!

PR #1526, opened 2025-08-13
Add a use_cvdalloc flag to the config.

3405691582:
To ensure that we can manage changes without disruption, we should have a feature flag to isolate new behavior from old. We could just repurpose the allocd flag, but it might be more readable to add a flag here and remove the allocd flag later in a different commit.


jemoreira:
> To ensure that we can manage changes without disruption, we should have a feature flag to isolate new behavior from old. We could just repurpose the allocd flag, but it might be more readable to add a flag here and remove the allocd flag later in a different commit.

I don't think we need flags at all, but if you really want one I'd prefer it to be a build time flag. Once we add a user visible runtime flag it becomes hard to remove over time.

Databean:
An `assemble_cvd` flag seems fine to me but we definitely don't want it in the `load_config.proto` configuration.

3405691582:
Yes, this won't be necessary to embed in any protos. Eventually this should become default, preferred behavior.

PR #1532, opened 2025-08-14
Create a type to represent the `--vendor_boot_image` flag

Databean:
This removes one part of the ordering dependency between `ResolveInstanceFiles` and `DiskImageFlagsVectorization`.

Bug: b/438587104


Databean:
Thanks for reviewing!

PR #1534, opened 2025-08-14
Use incremental locking when use_cvdalloc.

3405691582:
The existing locking scheme examines the number of preexisting tap interfaces on the system as created by cuttlefish-host-resources, then acquires the number of locks as there are tap interfaces. This scheme won't work when we want to incrementally create tap interfaces. Instead, acquire as many instance locks as requested when we use_cvdalloc.

The case where instance numbers can be explicitly requested is slightly interesting; here we use TryAcquireLocks. Arguably, this could be used in the non-use_cvdalloc case, but in that case, the instance numbers are bounded by the number of tap interfaces, so will have different results. Branch the behavior all the same, though.


base/cvd/cuttlefish/host/commands/cvd/cli/selector/creation_analyzer.cpp
```diff
@@ -157,8 +157,14 @@ CreationAnalyzer::AnalyzeInstanceIdsInternal(
     }
     return instance_info;
   }
+
+  const auto incremental = selector_options_parser_.UseCvdalloc();
+  std::set<int> requested(requested_instance_ids.begin(),
+                          requested_instance_ids.end());
   auto acquired_all_file_locks =
-      CF_EXPECT(instance_lock_file_manager_.LockAllAvailable());
+      incremental
```
jemoreira: Can you make it always acquire locks incrementally? regardless of whether cvdalloc should be used or not.

jemoreira (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/selector/creation_analyzer.cpp
3405691582: Sure; done.

PR #1535, opened 2025-08-14
Pass DefaultNetworkInterfaces the config.

3405691582:
The IfaceConfig as returned by DefaultNetworkInterfaces informs the vmm about the network interfaces it can use, including the interface names, which embed the instance number. Instead of passing the interface number directly, pass the CuttlefishConfig instead.

We will soon use the flag we generated earlier to tell the vmm to expect interfaces with different names when we are using cvdalloc, so that cvdalloc doesn't try to modify existing tap interfaces generated by the cuttlefish-host-resources service. We could that this by detecting if the service is running, but doing that effectively will be a pain.


base/cvd/cuttlefish/host/commands/assemble_cvd/alloc.cc
```diff
@@ -31,7 +31,10 @@ static std::string StrForInstance(const std::string& prefix, int num) {
   return stream.str();
 }
 
-IfaceConfig DefaultNetworkInterfaces(int num) {
+IfaceConfig DefaultNetworkInterfaces(
+    const CuttlefishConfig::InstanceSpecific& instance) {
+  int num = std::stoi(instance.id());
```
Databean: `std::stoi` throws exceptions on failing to parse an integer, can this use `absl::SimpleAtoi` ?

jemoreira (approved):

base/cvd/cuttlefish/host/commands/assemble_cvd/alloc.cc
3405691582: Done.

PR #1539, opened 2025-08-15
Mark release 1.20.0 as unstable

rmuthiah:


base/debian/changelog
```diff
@@ -1,8 +1,33 @@
-cuttlefish-common (1.20.0) UNRELEASED; urgency=medium
-
-  * 
-
- -- Alistair Delva <adelva@google.com>  Fri, 08 Aug 2025 15:15:33 -0700
+cuttlefish-common (1.20.0) unstable; urgency=medium
+
+  * Start development on v1.20 by @adelva1984 in https://github.com/google/android-cuttlefish/pull/1513
```
jemoreira, line 3: It shouldn't be the same list on both files. Some changes only apply to cuttlefish-common and some to cuttlefish-fronted

jemoreira (approved):

PR #1543, opened 2025-08-17
Add host orchestrator endpoint for cvd reset

jemoreira:
Bug: b/362543051


e2etests/orchestration/reset_test/BUILD.bazel
```diff
@@ -0,0 +1,30 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@rules_go//go:def.bzl", "go_test")
+
+go_test(
+    name = "reset_test_test",
+    srcs = ["main_test.go"],
+    env = {
```
ser-io: In order to use a single android build target among tests in this suite we have target "//orchestration/artifacts:images_zip" target. By using this target we have to upload the artifacts first using the upload API and then create the device. 

Unless you need this specific build id target or fetching is important in your test, let's reuse the same target among the tests.


data = [
  "//orchestration/artifacts:cvd_host_package",
  "//orchestration/artifacts:images_zip",
],

frontend/src/libhoclient/host_orchestrator_client.go
ser-io, line 1: Let's split backend and client code into different commits. If we are introducing a bug on the backend code and want to revert that change there's no need to revert the client change with it.

e2etests/orchestration/reset_test/BUILD.bazel
jemoreira: That failed when I tried to run the tests locally because it tried to extract the same zip twice.

Why do we prefer uploading image over fetch?

frontend/src/libhoclient/host_orchestrator_client.go
jemoreira: Done

e2etests/orchestration/reset_test/BUILD.bazel
jemoreira: Done

PR #1544, opened 2025-08-18
Use ubuntu-22.04-arm as Github workflow runner

0405ysj:
Context: https://docs.github.com/en/enterprise-cloud@latest/actions/concepts/runners/github-hosted-runners#runner-images

Now ARM64 runners are in standard Github-hosted runners.

### DIFF
--- a/.github/workflows/artifacts.yaml
+++ b/.github/workflows/artifacts.yaml
@@ -29,7 +29,7 @@ jobs:
         name: docker-image-x86_64
         path: ${{ env.x86_64_image_path }}
   export-docker-image-arm64:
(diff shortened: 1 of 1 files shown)


syslogic:
Somehow the `4core` part is missing, which might still be required. Ever tried, if it exists - or is even needed?

0405ysj:
I've tried, and it passed for building docker image for arm64.(Artifacts / export-docker-image-arm64) FYI, the number of core was same as before. The difference is `ubuntu-22.04-arm` is a kind of standard Github-hosted runner. https://docs.github.com/en/billing/concepts/product-billing/github-actions#free-use-of-github-actions

PR #1547, opened 2025-08-18
Validate cvdalloc binary file permissions.

3405691582:
If we are trying to run cvdalloc with a freshly-built binary, then it will not work since it will not be minted already setuid root for obvious reasons. However, we should at least detect this somewhat early so that you don't find out later when the vmm dies because cvdalloc couldn't set up interfaces because it lacked permissions.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
+++ b/base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
@@ -15,9 +15,15 @@
 
 #include "cuttlefish/host/commands/run_cvd/launch/cvdalloc.h"
 
+#include <errno.h>
+#include <fcntl.h>
+#include <string.h>
 #include <sys/socket.h>
+#include <sys/stat.h>
+
 #include <chrono>
 #include <string>
+#include <string_view>
 #include <unordered_set>
 #include <utility>
 #include <vector>
@@ -50,9 +56,11 @@ class Cvdalloc : public vm_manager::VmmDependencyCommand {
 
   // CommandSource
   Result<std::vector<MonitorCommand>> Commands() override {
+    std::string path = CvdallocBinary();
+    CF_EXPECT(BinaryIsValid(path));
     auto nice_stop = [this]() { return Stop(); };
 
-    Command cmd(CvdallocBinary(), KillSubprocessFallback(nice_stop));
+    Command cmd(path, KillSubprocessFallback(nice_stop));
     cmd.AddParameter("--id=", instance_.id());
     cmd.AddParameter("--socket=", their_socket_);
     std::vector<MonitorCommand> commands;
@@ -82,6 +90,21 @@ class Cvdalloc : public vm_manager::VmmDependencyCommand {
         CF_EXPECT(SharedFD::SocketPair(AF_LOCAL, SOCK_STREAM, 0));
     socket_ = std::move(p.first);
     their_socket_ = std::move(p.second);
+    return {};
+  }
+
+  Result<void> BinaryIsValid(std::string_view path) {
+    struct stat st;
+    int r = stat(path.data(), &st);
+    CF_EXPECT(r == 0, "Could not stat the cvdalloc binary at "
+                          << path << ": " << strerror(errno));
+
+    CF_EXPECTF((st.st_mode & S_ISUID) != 0 && st.st_uid == 0,
+        "cvdalloc binary does not have permissions to allocate resources.\n"
+        "As root, please\n\n    chown root {}\n    chmod u+s {}\n\n"
+        "and start the instance again.",
+        path, path);
+
     return {};
   }
 


base/cvd/cuttlefish/host/commands/cvd/cli/selector/creation_analyzer.cpp
```diff
@@ -291,6 +300,23 @@ Result<std::string> CreationAnalyzer::AnalyzeHome() const {
...
+  struct stat st;
+  int r = stat(path.c_str(), &st);
+  CF_EXPECT(r == 0, "Could not stat the cvdalloc binary at "
+                        << path << ": " << strerror(errno));
+
+  auto no_suid_msg = absl::StrFormat(
+      "cvdalloc binary does not have permissions to allocate resources.\n"
+      "As root, please\n\n    chown root %s\n    chmod u+s %s",
+      path, path);
+  CF_EXPECT((st.st_mode & S_ISUID) != 0 && st.st_uid == 0, no_suid_msg);
```
Databean: fyi, there is also [`CF_EXPECTF`](https://github.com/google/android-cuttlefish/blob/6196f77e317aa8f7696bb17a738a9e9f31352fe2/base/cvd/cuttlefish/common/libs/utils/result.h#L417) which uses [`fmtlib`](https://fmt.dev/11.1/) to handle a format string and final format arguments.

base/cvd/cuttlefish/host/commands/cvd/cli/selector/creation_analyzer.cpp
```diff
@@ -291,6 +300,23 @@ Result<std::string> CreationAnalyzer::AnalyzeHome() const {
   return auto_generated_home;
 }
 
+Result<void> CreationAnalyzer::AnalyzeCvdallocPermissions(
```
Databean: nit: maybe `Validate` rather than `Analyze` since it only reports success / failure?

base/cvd/cuttlefish/host/commands/cvd/cli/selector/creation_analyzer.cpp
```diff
@@ -86,6 +88,8 @@ class CreationAnalyzer {
    */
   Result<std::string> AnalyzeHome() const;
 
+  Result<void> AnalyzeCvdallocPermissions(const std::string& path) const;
```
jemoreira: I don't think this is the best place to check this:

The fact that we use the `setuid` bit is just an implementation detail, subject to change (would it change if capabilities are used instead?). I think this check should be closer to where the extra permissions are needed. For example, `cvdalloc` should check if `geteui()` is `0`.

jemoreira: > so that you don't find out later when the vmm dies because cvdalloc couldn't set up interfaces because it lacked permissions

The VMM shouldn't even be started if cvdalloc fails to allocate resources. `run_cvd` should just fail at that point.

3405691582: If we have cvdalloc check its own permissions, then the user would only be able to see this by examining the launcher log. As (I believe) an unrelated issue, if cvdalloc doesn't have permissions and fails to allocate resources, the vmm dies, but the create attempt never finishes/crashes out and hangs, so there isn't immediate feedback that the user should act and investigate other than the create attempt seems to take a long time. 

(To also answer your question: when we use capabilities, then yes, this implementation will change.)

jemoreira: > if cvdalloc doesn't have permissions and fails to allocate resources, the vmm dies...

Why would the vmm run at all if `cvdalloc` failed?

I think that question is at the core of whether this is needed or not. If `run_cvd` detects that the interfaces couldn't be created (i.e cvdalloc failed) it should report it and stop right there. The message should be clear to the user there as it would be the last thing in launcher.log.

I understand it's a little less convenient for the user to have to go into launcher.log to find the root cause, but that's the case for all other launch failures. I don't think it would be feasible to check for all possible failure causes in cvd's `CreationAnalyzer`, and if we can't do for all why do it for just one? Instead, we should try to make `cvd create`'s output more useful by removing the extra noise and maybe surfacing the relevant messages from launcher.log or at least pointing the user to the file's location (assemble_cvd or run_cvd print this, but there is no guarantee that they will if they fail, cvd should print this instead). I had planned to do that in b/324653441 and b/324652982 but never got around to it.

3405691582: The problem isn't so much that it's inconvenient, it's that `cvd create` hangs, so you don't immediately know that the underlying process has failed.

If cvdalloc fails, this would be reported in the `WaitForAvailability` stage as a [prerequisite of the vmm](https://github.com/google/android-cuttlefish/blob/main/base/cvd/cuttlefish/host/libs/vm_manager/crosvm_manager.cpp#L515). But we [don't seem to actually do anything about it](https://github.com/google/android-cuttlefish/blob/main/base/cvd/cuttlefish/common/libs/utils/subprocess.cpp#L419), so the vmm starts and fails. The parent run_cvd process doesn't end, so there are still child processes running and logging to the launcher.log and aren't terminated when the vmm fails.

Why doesn't run_cvd and `cvd create` just die? I suspect this is because if critical processes fail, [that isn't treated as a bad Result](https://github.com/google/android-cuttlefish/blob/main/base/cvd/cuttlefish/host/libs/process_monitor/process_monitor.cc#L144); there's also some sort of issue with [this](https://github.com/google/android-cuttlefish/blob/main/base/cvd/cuttlefish/host/libs/process_monitor/process_monitor.cc#L441) that I haven't looked into either. 

I could go down this rabbit hole. Should I?

base/cvd/cuttlefish/host/commands/cvd/cli/selector/creation_analyzer.cpp
3405691582: Done.

base/cvd/cuttlefish/host/commands/cvd/cli/selector/creation_analyzer.cpp
3405691582: Done.

base/cvd/cuttlefish/host/commands/cvd/cli/selector/creation_analyzer.cpp
jemoreira: > But we [don't seem to actually do anything about it](https://github.com/google/android-cuttlefish/blob/main/base/cvd/cuttlefish/common/libs/utils/subprocess.cpp#L419), so the vmm starts and fails.

Sounds like that's the real problem that needs to be fixed here.

> Why doesn't run_cvd and cvd create just die? 

`cvd create` is just waiting for `run_cvd`

> I suspect this is because if critical processes fail, [that isn't treated as a bad Result](https://github.com/google/android-cuttlefish/blob/main/base/cvd/cuttlefish/host/libs/process_monitor/process_monitor.cc#L144)

That sounds right. It used to be that all processes were restarted, but then it was changed to only non-critical processes are restarted. `cvdalloc` needs to be treated as a critical process.

> I could go down this rabbit hole. Should I?

At least as far down as to make the vmm not be launched when cvdalloc fails. That's a more permanent fix than checking at cvd create time.

3405691582: This has been moved to run_cvd. Thankfully, doing it here means that if the check fails, the user-facing cvd create process balks immediately and the error message is last in the launcher log. 

Unfortunately, this means that the instance has been partially created, and that repeating the command won't work, so you have to `cvd start` once you fix the permissions. I'll see if this flow can be improved later.

jemoreira: > Unfortunately, this means that the instance has been partially created, and that repeating the command won't work, so you have to cvd start once you fix the permissions.

This is by design: When the instance is deleted all the logs are lost with it.

PR #1548, opened 2025-08-18
Name cvdalloc-managed interfaces differently.

3405691582:
Tell the vmm to expect cvdalloc-managed tap interfaces to be named differently from preallocated tap interfaces. This way, we avoid conflicts on a host where cuttlefish-host-resources may be running.

While we are doing this, we remove the padding requirements for the interface number in the naming scheme for cvdalloc-managed interfaces. This makes the interface naming more idiomatic and avoids the potential problem in the future if we have more than 99 instances on a host.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/alloc.cc
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/alloc.cc
@@ -20,40 +20,46 @@
 #include <sstream>
 
 #include "absl/strings/numbers.h"
+#include "absl/strings/str_format.h"
 
 #include "cuttlefish/common/libs/fs/shared_fd.h"
+#include "cuttlefish/host/commands/cvdalloc/interface.h"
 #include "cuttlefish/host/libs/allocd/request.h"
 #include "cuttlefish/host/libs/allocd/utils.h"
 
-
 namespace cuttlefish {
 
-static std::string StrForInstance(const std::string& prefix, int num) {
-  std::ostringstream stream;
-  stream << prefix << std::setfill('0') << std::setw(2) << num;
-  return stream.str();
+static Result<std::string> InterfaceName(
+    const CuttlefishConfig::InstanceSpecific& instance,
+    const std::string& name) {
+  int num;
+  CF_EXPECT(absl::SimpleAtoi(instance.id(), &num));
+
+  if (instance.use_cvdalloc()) {
+    return CvdallocInterfaceName(name, num);
+  }
+
+  return absl::StrFormat("%s-%s-%02d", kDefaultInterfacePrefix, name, num);
 }
 
 Result<IfaceConfig> DefaultNetworkInterfaces(
     const CuttlefishConfig::InstanceSpecific& instance) {
   IfaceConfig config{};
 
-  int num;
-  CF_EXPECT(absl::SimpleAtoi(instance.id(), &num));
-
-  config.mobile_tap.name = StrForInstance("cvd-mtap-", num);
+  config.mobile_tap.name = CF_EXPECT(InterfaceName(instance, "mtap"));
   config.mobile_tap.resource_id = 0;
   config.mobile_tap.session_id = 0;
 
-  config.bridged_wireless_tap.name = StrForInstance("cvd-wtap-", num);
+  config.bridged_wireless_tap.name = CF_EXPECT(InterfaceName(instance, "wtap"));
   config.bridged_wireless_tap.resource_id = 0;
   config.bridged_wireless_tap.session_id = 0;
 
-  config.non_bridged_wireless_tap.name = StrForInstance("cvd-wifiap-", num);
+  config.non_bridged_wireless_tap.name =
+      CF_EXPECT(InterfaceName(instance, "wifiap"));
   config.non_bridged_wireless_tap.resource_id = 0;
   config.non_bridged_wireless_tap.session_id = 0;
 
-  config.ethernet_tap.name = StrForInstance("cvd-etap-", num);
+  config.ethernet_tap.name = CF_EXPECT(InterfaceName(instance, "etap"));
   config.ethernet_tap.resource_id = 0;
   config.ethernet_tap.session_id = 0;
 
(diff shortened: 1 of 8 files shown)


base/cvd/cuttlefish/host/commands/cvdalloc/interface.h
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <string>
+
+namespace cuttlefish {
+
+const std::string kCvdallocInterfacePrefix = "cvd-pi";
```
Databean: nit: this should be a `constexpr std::string_view` because that is trivially destructible, unlike `std::string`. ([style guide](https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables))

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.h
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <string>
+
+namespace cuttlefish {
+
+const std::string kCvdallocInterfacePrefix = "cvd-pi";
```
Databean: nit: this should be a `constexpr std::string_view` because that is trivially destructible, unlike `std::string`. ([style guide](https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables))

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -37,26 +39,28 @@ void Usage() {
   LOG(ERROR) << "Should only be invoked from run_cvd.";
 }
 
-Result<void> Allocate(int id) {
+Result<void> Allocate(int id, const std::string &bridge_name) {
   LOG(INFO) << "cvdalloc: allocating network resources";
 
-  CreateBridge("cvd-br");
+  CreateBridge(bridge_name);
```
Databean, line 45: Not related to this change, but can `CreateBridge` also be validated with `CF_EXPECT`?

base/cvd/cuttlefish/host/commands/assemble_cvd/alloc.cc
```diff
@@ -20,40 +20,45 @@
...
-  return stream.str();
+static Result<std::string> InterfaceName(
+    const CuttlefishConfig::InstanceSpecific& instance,
+    const std::string& name) {
+  int num;
+  CF_EXPECT(absl::SimpleAtoi(instance.id(), &num));
+
+  if (!instance.use_cvdalloc()) {
+    return absl::StrFormat("%s-%s-%02d", kDefaultInterfacePrefix, name, num);
+  }
```
Databean: nit: following [go/tott-685](http://go/tott-685), can the conditional cases be reversed so there is less negation? i.e.

```c++
if (instance.use_cvdalloc()) {
  ...
}
```

base/cvd/cuttlefish/host/commands/assemble_cvd/alloc.cc
```diff
@@ -20,40 +20,45 @@
...
-  int num;
-  CF_EXPECT(absl::SimpleAtoi(instance.id(), &num));
-
-  config.mobile_tap.name = StrForInstance("cvd-mtap-", num);
+  config.mobile_tap.name = CF_EXPECT(InterfaceName(instance, "mtap"));
   config.mobile_tap.resource_id = 0;
   config.mobile_tap.session_id = 0;
 
-  config.bridged_wireless_tap.name = StrForInstance("cvd-wtap-", num);
+  config.bridged_wireless_tap.name = CF_EXPECT(InterfaceName(instance, "wtap"));
```
Databean, line 53: @jemoreira to confirm: I vaguely remember the original intent was to keep `cvdalloc` potentially compatible with older versions of `run_cvd`, but now I don't think that matters anymore? Considering our host executables will be updated in lockstep from this repository, and should eventually be mixed only with older guest images rather than with older host executables.

base/cvd/cuttlefish/host/commands/assemble_cvd/alloc.h
```diff
@@ -25,6 +25,8 @@
 
 namespace cuttlefish {
 
+const std::string kDefaultInterfacePrefix = "cvd";
```
Databean: nit: this should be a `constexpr std::string_view` because that is trivially destructible, unlike `std::string`. ([style guide](https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables))

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
3405691582: Not right now. Since bridge creation is not idempotent, the second time the bridge creation is attempted for a second instance it will fail unnecessarily. We should definitely get to improving the underlying logic at some point, but for now, ignoring the error should be harmless.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.h
3405691582: Done.

base/cvd/cuttlefish/host/commands/assemble_cvd/alloc.cc
3405691582: Done.

base/cvd/cuttlefish/host/commands/assemble_cvd/alloc.h
3405691582: Done.

base/cvd/cuttlefish/host/commands/cvdalloc/interface.h
3405691582: Done.

base/cvd/cuttlefish/host/commands/assemble_cvd/alloc.cc
jemoreira: SGTM

jemoreira (approved):

PR #1549, opened 2025-08-18
Expose lower-level zip source types from BuildApi and remote_zip.h

Databean:
Creating the low-level types and wrapping them before returning it gets in the way of expressing the correct ordering of IO wrappers.

The new function `OpenZip(BuildApi&, Build, std::string)` now expresses the intended layering of Buffer(Cache(Remote)).

Bug: b/439624498


Databean:
Thanks for reviewing!

PR #1551, opened 2025-08-18
Don't fail if the artifact is already extracted

jemoreira:
If the artifact was already extracted successfully once the current request should succeed immediately instead of failing.

Bug: b/439670070

### DIFF
--- a/frontend/src/host_orchestrator/orchestrator/userartifacts.go
+++ b/frontend/src/host_orchestrator/orchestrator/userartifacts.go
@@ -150,7 +150,7 @@ func (m *UserArtifactsManagerImpl) ExtractArtifact(checksum string) error {
 	if exists, err := dirExists(dir); err != nil {
 		return fmt.Errorf("failed to check existence of directory: %w", err)
 	} else if exists {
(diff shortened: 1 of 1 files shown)


frontend/src/host_orchestrator/orchestrator/userartifacts.go
```diff
@@ -150,7 +150,7 @@ func (m *UserArtifactsManagerImpl) ExtractArtifact(checksum string) error {
 	if exists, err := dirExists(dir); err != nil {
 		return fmt.Errorf("failed to check existence of directory: %w", err)
 	} else if exists {
-		return operator.NewConflictError(fmt.Sprintf("user artifact(checksum:%q) already extracted", checksum), nil)
```
ser-io, line 153: Actually, I think it's ok to return 409 in this case. However we should rather modify the client to NOT return error if 409 was received.

jemoreira: That was my initial approach, but doing that was not trivial because the client needs to return an operation, but it doesn't get one from the server in these cases. The correct approach here is to change the client interface to not return operations at all, I'll handle that in a different PR.

PR #1552, opened 2025-08-18
netsimd: Add dependencies of libnetsim_daemon

jacky8hyf:
Steps to build netsimd.

Bug: b/402294843

### DIFF
--- a/base/cvd/BUILD.android_tools_netsim.bazel
+++ b/base/cvd/BUILD.android_tools_netsim.bazel
@@ -1,3 +1,4 @@
+load("@@//build_external/netsim:pdl_rust.bzl", "pdl_rust")
 load("@bazel_skylib//rules:run_binary.bzl", "run_binary")
 load("@netsim_crates//:defs.bzl", "all_crate_deps")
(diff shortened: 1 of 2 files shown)


base/cvd/BUILD.android_tools_netsim.bazel
```diff
@@ -179,3 +253,32 @@ cc_library(
...
+        ":libnetsim_hostapd_rs",
+        ":libnetsim_http_proxy",
+        ":libnetsim_libslirp_rs",
+        ":libnetsim_packets",
+        ":libnetsim_proto",
+        # pica is manually excluded from Cargo.toml due to missing required
+        # version on crates.io, so we need to manually add it here.
+        "@netsim_crates//:pica",
+        # This is missing from Cargo.toml, so manually add it.
+        "@netsim_crates//:openssl",
```
jacky8hyf: This is probably wrong (because grpc++ uses boringssl instead). We need an alternative.

jacky8hyf: I dropped the last commit from this PR for now.

Databean (approved):

PR #1553, opened 2025-08-19
Remove operations from the host orchestrator client api

jemoreira:
Also makes ExtractArtifact not fail when it as already been extracted.

Bug: b/439670070
Bug: b/439672811


0405ysj (approved):

0405ysj:
Thanks for fixing bugs. Hiding operation from API surface sounds good to me.

PR #1555, opened 2025-08-19
Fix Debian version of container image

0405ysj:
As Debian 13 is newly introduced, this container image began to use it but haven't fitted with the new version. This commit is for fixing Debian version as 12(bookworm) which we used before, and we can upgrade into Debian 13 later.

### DIFF
--- a/tools/buildutils/cw/Containerfile
+++ b/tools/buildutils/cw/Containerfile
@@ -1,4 +1,4 @@
-FROM debian:stable AS base
+FROM debian:12 AS base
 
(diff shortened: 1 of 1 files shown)


0405ysj:
@ser-io Was there any specific reason to use `debian:stable` rather than fixed version? I've found an error recently, with trixie(debian 13) being introduced into public.

ser-io:
> @ser-io Was there any specific reason to use `debian:stable` rather than fixed version? I've found an error recently, with trixie(debian 13) being introduced into public.

No specific reason, it should have been fixed to 12, good catch.

PR #1558, opened 2025-08-19
Download META/ab_partitions.txt from the target files zip.

Databean:
The full target files zip file is not downloaded unless it is requested by other flags.

Bug: b/439877215

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
+++ b/base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
@@ -48,10 +48,13 @@
 #include "cuttlefish/host/libs/web/android_build.h"
 #include "cuttlefish/host/libs/web/android_build_api.h"
 #include "cuttlefish/host/libs/web/android_build_string.h"
+#include "cuttlefish/host/libs/web/build_api_zip.h"
 #include "cuttlefish/host/libs/web/build_zip_name.h"
 #include "cuttlefish/host/libs/web/chrome_os_build_string.h"
 #include "cuttlefish/host/libs/web/http_client/curl_global_init.h"
 #include "cuttlefish/host/libs/web/luci_build_api.h"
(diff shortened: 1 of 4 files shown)


base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
```diff
@@ -248,10 +251,11 @@ Result<void> FetchDefaultTarget(BuildApi& build_api, const Builds& builds,
     trace.CompletePhase("Desparse image files");
   }
 
+  std::string target_files_name =
+      GetBuildZipName(*builds.default_build, "target_files");
+  ReadableZip target_files_zip = ReadableZip::Invalid();
```
cjreynol, line 256: Rather than adding `ReadableZip::Invalid` to allow the `target_files_zip` to be the cross-condition variable, could we extract the `ab_partitions.txt` logic to be a helper we call in both blocks.  Then we could have the resulting filepath be the shared variable.  I think intentionally adding an invalid state is adding mental overhead to future updates, and in this case I think avoiding it does not make things worse.

Roughly thinking:

```c++
std::string ExtractAbPartitions(const std::string& target_directory, ReadableZip target_files) {
  //...
}
```

then the conditional becomes:

```c++
std::string ab_partitions_filepath;
if (...) {
  //...
  ReadableZip target_files = ...;
  ab_partitions_filepath = ExtractAbPartitions(target_directories.default_target_files, target_files);
} else {
  ReadableZip target_files = ...;
  ab_partitions_filepath = ExtractAbPartitions(...); 
}
CF_EXPECT(config.AddFilesToConfig(...));
```

Databean: Thanks for the call out.

I forgot that I had already made this PR earlier and rewrote this code in https://github.com/google/android-cuttlefish/pull/1738 . It doesn't include `ReadableZip::Invalid` and instead uses a `? :` operator to initialize the `ReadableZip` value.

PR #1559, opened 2025-08-20
Use standard runner to build CF debian packages for x86 on Github workflow

0405ysj:
Currently `run-cvd-unit-tests`, `build-debian-package`, and their corresponding workflows for updating cache respectively are using `ubuntu-22.04-4core` which isn't a standard runner and costs.

The reason for using `ubuntu-22.04-4core` was to retrieve enough disk space, and I've found the action `jlumbroso/free-disk-space@v1.3.1` is used from another job(`e2e-tests-orchestration-runner-special`). However, we cannot utilize that action when the whole steps in the job are running on the container, though we need to build debian packages on debian, not on ubuntu. So I brought `tools/buildutils/cw/Containerfile` to build debian packages from debian environment.

### DIFF
--- a/.github/workflows/presubmit.yaml
+++ b/.github/workflows/presubmit.yaml
@@ -192,29 +192,22 @@ jobs:
       with:
         name: cvd-unit-tests-logs
         path: base/cvd/bazel-out/k8-fastbuild/testlogs
-  build-debian-package:
-    runs-on: ubuntu-22.04-4core
-    container:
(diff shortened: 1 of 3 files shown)


k311093 (approved):

0405ysj:
Calculated estimated cost for running large runners in this repository for recent 30 days. It costed $641/mo. I expect it'll be saved with this and some other PRs.

0405ysj:
I'm not sure why `build-debian-package` is still a pending check. I've renamed `build-debian-package` into `build-debian-package-x86` at all. So merging with bypassing that check.

PR #1560, opened 2025-08-20
Use standard runner to build CF debian packages for ARM on Github workflow

0405ysj:
Note: This PR contains commits suggested at https://github.com/google/android-cuttlefish/pull/1559. I would rebase this PR before submission, after merging https://github.com/google/android-cuttlefish/pull/1559.

Primary reason for building CF debian packages for ARM is to upload debian packages into Artifact Registry. To achieve that, I want to introduce a presubmit or postsubmit check for building them.

- Estimated time to run job: 10 minutes after caching is applied.
- Estimated time change to complete workflow: No change, as this job won't be in the critical path to complete.
- Cost for running job: Free-tier as it's using `ubuntu-22.04-arm`, which is the standard runner.

### DIFF
--- a/.github/actions/run-cw-sharded-e2e-test/action.yaml
+++ b/.github/actions/run-cw-sharded-e2e-test/action.yaml
@@ -8,7 +8,7 @@ runs:
   using: "composite"
   steps:
     - name: Free disk space
-      uses: jlumbroso/free-disk-space@v1.3.1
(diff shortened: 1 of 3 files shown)


0405ysj:
Q: How do you think executing this job during either presubmit or postsubmit? I suggested for the presubmit in advance, but I'm still hesitating it. Presubmit may be better to catch ARM-specific error early, but it's not critical for the primary reason to upload deb packages into Artifact Registry.

PR #1563, opened 2025-08-20
Try to use standard runner for performing cvd unit tests

0405ysj:


0405ysj:
Note
- Cannot run on debian container defined for a whole job, because of executing `jlumbroso/free-disk-space@v1.3.1`.
- Cannot run on ubuntu-22.04, because of missing support around `P_PIDFD`.
- Cannot run on ubuntu-24.04, because of missing support around libtinfo5. If I add manually download .deb file for it, guess as workable.

k311093 (approved):

PR #1565, opened 2025-08-20
Create abstraction for cvdalloc socket signaling.

3405691582:
Soon, run_cvd will need to also unblock and block on cvdalloc's socket. To do this reliably, we should have some stronger checks and some formalizations around the signaling that we're doing with the socket. Since we're treating these like semaphores, introduce a Post and Wait to do to this.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvdalloc/sem.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvdalloc/sem.cpp
@@ -1,0 +1,38 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "cuttlefish/host/commands/cvdalloc/sem.h"
+
+#include "cuttlefish/host/libs/command_util/util.h"
+
+namespace cuttlefish {
+
+Result<void> Post(const SharedFD socket) {
+  char i = 0;
+  CF_EXPECT(socket->Write(&i, sizeof(char)) != -1);
+  return {};
+}
+
+Result<void> Wait(const SharedFD socket, std::chrono::seconds timeout) {
+  CF_EXPECT(socket->IsOpen(), "IsOpen: " << socket->StrError());
+  CF_EXPECT(WaitForRead(socket, timeout.count()), "WaitForRead");
+  char i = -1;
+  int r = socket->Read(&i, sizeof(i));
+  CF_EXPECT_GT(r, 0, socket->StrError());
+  CF_EXPECT(i == 0, "Unexpected read");
+  return {};
+}
+
+}  // namespace
--- a/base/cvd/cuttlefish/host/commands/cvdalloc/sem.h
+++ b/base/cvd/cuttlefish/host/commands/cvdalloc/sem.h
@@ -1,0 +1,43 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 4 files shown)


base/cvd/cuttlefish/host/commands/cvdalloc/sem.h
```diff
@@ -0,0 +1,39 @@
...
+#include "cuttlefish/common/libs/utils/result.h"
+
+namespace cuttlefish {
+
+/*
+ * Write into the socket.
+ *
+ * Any other processes Wait'ing to read on the socket will be unblocked.
+ */
+Result<void> Post(const SharedFD &socket);
```
Databean: nit: `SharedFD` behaves like a `std::shared_ptr` over a file descriptor, so it should be passed by value the same way `std::shared_ptr` usually is.

base/cvd/cuttlefish/host/commands/cvdalloc/sem.cpp
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+#include "cuttlefish/host/commands/cvdalloc/sem.h"
+
+#include "cuttlefish/host/libs/command_util/util.h"
+
+namespace cuttlefish {
+
+Result<void> Post(const SharedFD &socket) {
+  int i = 0;
```
Databean: To avoid data being left in the socket buffer, can this be a `char` rather than an `int`? Otherwise it's theoretically possible for `socket->Read(...)` below to return only some of the bytes in one call, and more bytes in another call.

base/cvd/cuttlefish/host/commands/cvdalloc/sem.h
```diff
@@ -0,0 +1,39 @@
...
+
+/*
+ * Wait on the socket.
+ *
+ * The process wil block until the socket can be read from, the socket is
+ * shut down, or the timeout is reached. The Result is ok if the socket
+ * can be successfully read from or is shut down.
+ */
+Result<void> Wait(const SharedFD &socket, int timeout);
```
Databean: Can `timeout` be in `std::chrono::seconds` so the units are obvious to the caller?

base/cvd/cuttlefish/host/commands/cvdalloc/sem.cpp
```diff
@@ -0,0 +1,38 @@
...
+  CF_EXPECT(socket->Write(&i, sizeof(int)) != -1);
+  return {};
+}
+
+Result<void> Wait(const SharedFD &socket, int timeout) {
+  CF_EXPECT(socket->IsOpen(), "IsOpen");
+  CF_EXPECT(WaitForRead(socket, timeout), "WaitForRead");
+  int i = -1;
+  int r = socket->Read(&i, sizeof(i));
+  CF_EXPECT(r > 0, "Read");
```
Databean: The function comment says "The Result is ok if the socket can be successfully read from or is shut down." but a shutdown will result in `r == 0` for the end-of-file condition.

Maybe `CF_EXPECT_GE(r, 0, socket->StrError())`?

base/cvd/cuttlefish/host/commands/cvdalloc/sem.cpp
```diff
@@ -0,0 +1,38 @@
...
+namespace cuttlefish {
+
+Result<void> Post(const SharedFD &socket) {
+  int i = 0;
+  CF_EXPECT(socket->Write(&i, sizeof(int)) != -1);
+  return {};
+}
+
+Result<void> Wait(const SharedFD &socket, int timeout) {
+  CF_EXPECT(socket->IsOpen(), "IsOpen");
```
Databean: Maybe worth reporting `socket->StrError()` as well, which is usually recorded in the case that a `SharedFD::Open` or similar encounters an error.

Databean (approved):

base/cvd/cuttlefish/host/commands/cvdalloc/sem.cpp
3405691582: Ah, I actually meant to amend the function comment (and thought I did). If the socket is shutdown, that should be an error condition.

jemoreira (approved):

base/cvd/cuttlefish/host/commands/cvdalloc/sem.cpp
3405691582: Done.

base/cvd/cuttlefish/host/commands/cvdalloc/sem.h
3405691582: Done.

base/cvd/cuttlefish/host/commands/cvdalloc/sem.cpp
3405691582: Used CF_EXPECT_GT.

base/cvd/cuttlefish/host/commands/cvdalloc/sem.cpp
3405691582: Done.

base/cvd/cuttlefish/host/commands/cvdalloc/sem.h
3405691582: Done.

PR #1566, opened 2025-08-20
base: debian: control: change libtinfo5 to libncurses-dev

grandpaul:
Debian currently did a new release (Debian 13 Trixie). And libtinfo5 is upgraded to libtinfo6. I think we can depends on libncurses-dev as this development package will install libtinfoX by its dependencies. So we can make this package buildable on both Debian 12 Bookworm and Debian 13 Trixie.

Can we use clang from the system instead? It seems to me that it is the clang binary that depends on libtinfo5.


0405ysj:
FYI: You may want to change `libtinfo5` into `libncurses-dev` at `.github/workflows/*.yaml`.

https://github.com/search?q=repo%3Agoogle%2Fandroid-cuttlefish%20libtinfo5&type=code

grandpaul:
This is not working. I'll investigate more.

PR #1567, opened 2025-08-21
Post before shutdown on clean termination.

3405691582:
The semantics of shutdown on our semaphores have changed to better signify abnormal termination cases. This will eventually allow run_cvd to determine if the notification it receives is because the allocation process has completed successfully or whether cvdalloc has died and should propagate the error to prevent the vmm from starting.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
@@ -18,7 +18,7 @@
 
 #include <android-base/logging.h>
 #include <android-base/macros.h>
-#include <android-base/scopeguard.h>
+#include "absl/cleanup/cleanup.h"
 #include "absl/flags/flag.h"
 #include "absl/flags/parse.h"
 #include "absl/strings/str_format.h"
@@ -87,6 +87,10 @@ Result<int> CvdallocMain(int argc, char *argv[]) {
     return CF_ERRNO("close: " << strerror(errno));
   }
 
+  absl::Cleanup shutdown = [sock]() {
+    sock->Shutdown(SHUT_RDWR);
+  };
+
   /*
    * Explicit setuid calls seem to be required.
    *
@@ -102,12 +106,12 @@ Result<int> CvdallocMain(int argc, char *argv[]) {
    */
   uid_t orig = getuid();
 
-  auto drop_privileges = android::base::ScopeGuard([orig]() {
+  absl::Cleanup drop_privileges = [orig]() {
     int r = setuid(orig);
     if (r == -1) {
       LOG(ERROR) << "cvdalloc: couldn't drop privileges: " << strerror(errno);
     }
-  });
+  };
 
   r = setuid(0);
   if (r == -1) {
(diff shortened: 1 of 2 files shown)


jemoreira:
> to determine if the notification it receives is because the allocation process has completed successfully or whether cvdalloc has died

Can't the exit code be used for that?

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -131,6 +139,9 @@ Result<int> CvdallocMain(int argc, char *argv[]) {
   CF_EXPECT(Wait(sock, kSemNoTimeout));
 
   /* Teardown invoked by scopeguard above. */
+  /* This isn't a dead store. */
+  // NOLINTNEXTLINE(clang-analyzer-deadcode.DeadStores)
```
jemoreira: IIUC you want to `Post` after teardown, before shutdown and only if everything went ok.

Consider encapsulating all the code between the teardown guard and the return below into a function and calling it from here like this:
```
CF_EXPECT(ThatNewFunction());
(void)Post(sock);
```
That avoids the linting exception and makes your intention clearer IMHO.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -131,6 +139,9 @@ Result<int> CvdallocMain(int argc, char *argv[]) {
   CF_EXPECT(Wait(sock, kSemNoTimeout));
 
   /* Teardown invoked by scopeguard above. */
+  /* This isn't a dead store. */
```
Databean: The anonymous function used with `ScopeGuard` above is taking `clean_teardown` by value rather than by reference. To make sure modifications are propagated, it needs to be `&clean_teardown` in the capture list. Does that fix the lint error as well?

Databean (approved):

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
3405691582: Ah, of course.

3405691582:
> Can't the exit code be used for that?

AIUI, run_cvd's `VmmDependencyCommand`s don't get visibility into the command's running state, since the processes are managed separately, so we need to use the socketpair we created.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
3405691582: I took a slightly different tack. I think this reads much better, while still having the intended behavior.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -116,21 +120,19 @@ Result<int> CvdallocMain(int argc, char *argv[]) {
...
   CF_EXPECT(Post(sock));
 
   LOG(INFO) << "cvdalloc: waiting to teardown";
 
   CF_EXPECT(Wait(sock, kSemNoTimeout));
+  CF_EXPECT(Teardown(id, bridge_name));
+  CF_EXPECT(Post(sock));
 
-  /* Teardown invoked by scopeguard above. */
+  teardown.Disable();
```
jemoreira: Please move this to right before the `Teardown` function is called. It may still run if `Teardown` itself or `Post` return error.

Ideally `ScopeGuard` should have an `Execute` or similar method to force it to run early, but that's not the case.

jemoreira: Actually, could you use abseil's [Cleanup](https://github.com/abseil/abseil-cpp/blob/master/absl/cleanup/cleanup.h) instead. I know I suggested ScopeGuard in the first place, but I didn't know about Cleanup then and Cleanup has `Invoke` which you can call in line 132 instead of the function directly.

3405691582: No problem; standardizing on absl is probably a good idea anyway. Done.

PR #1572, opened 2025-08-25
Pass sve=[auto=true] option for crosvm under arm64

k311093:
While running image that uses sve instruction, boot failed with error about handling sve instruction even the CPU supports sve instruction.

It needs to enable sve instruction at crosvm side.

b/438945282

change for testing: ag/35226572

### DIFF
--- a/base/cvd/cuttlefish/host/libs/vm_manager/crosvm_builder.cpp
+++ b/base/cvd/cuttlefish/host/libs/vm_manager/crosvm_builder.cpp
@@ -95,7 +95,11 @@ Result<void> CrosvmBuilder::AddCpus(const Json::Value& vcpu_config_json) {
 }
 
 void CrosvmBuilder::AddCpus(size_t cpus) {
-  command_.AddParameter("--cpus=", cpus);
+  if (HostArch() == Arch::Arm64) {
+    command_.AddParameter("--cpus=", cpus, ",sve=[auto=true]");
+  } else {
+    command_.AddParameter("--cpus=", cpus);
+  }
 }
 
 void CrosvmBuilder::AddHvcSink() {


0405ysj (approved):

ikicha:
Is there a case for this to run older crosvm binary which doesn't have sve param?

k311093:
> Is there a case for this to run older crosvm binary which doesn't have sve param?

It uses crosvm from `$ANDROID_HOST_OUT/bin` or `$HOME/bin`

For the case of `$HOME/bin`, not sure but might be okay for running because it is the case that running using cvd package which crosvm is being used that built with the cvd command line.

But it could be the the case that running with Android Source Tree `$ANDROID_HOST_OUT/bin` which does not have https://chromium-review.googlesource.com/c/crosvm/crosvm/+/6180906 (this could be happen if not synced for a while).

It might better to check if crosvm has option for sve, but crosvm help message does not shows for auto flag

```
sve=[enable=bool] - SVE Config.
                    (aarch64 only)                   Examples:                   
                                                     sve=[enable=true] - Enables
                    SVE on device. Will fail is SVE unsupported.         default
                    value = false.
```

and might not able to use `crosvm version` becuase the version has not changed from 0.1.0.

Databean:
> Is there a case for this to run older crosvm binary which doesn't have sve param?

Right now `cvd` can be used for android checkouts other than `main`, like `android14-gsi`. This PR is making code changes in `run_cvd` though, not `cvd`. In the current code it's possible to run an older guest image with a newer `run_cvd` by using `cvd fetch --host_substitutions=all` which will aggressively replace host executables from the android build using the migrated versions from the debian package. The debian package version of `crosvm` is incomplete so that one won't be substituted even with `--host_substitutions=all`, while the latest version of `run_cvd` will be used. We only have automatic substitution enabled for the latest internal code, controlled by the less ambitious `debian_substitution_marker` file, which is a list of substitutions we've validated to be correct on that branch.

Ultimately the point is that it is technically reachable via the right invocation to have a combination of a newer `run_cvd` with an older `crosvm`, though it's not on an obvious path.

It is also possible to use a configuration like `cvd fetch --host_package_build=aosp-android-latest-release --host_substitutions=all` which would bring in the latest `crosvm` with an older guest image. A configuration like this is validated by the tests at http://cs/file:cloud-android-devs/host-compat .

PR #1574, opened 2025-08-25
Continue refactoring files under tools/baseimage

0405ysj:
This PR continues refactoring after https://github.com/google/android-cuttlefish/pull/1561.

As this PR is quite large either and contains some commits are already reviewed at https://github.com/google/android-cuttlefish/pull/1561, please review per each commit. Thanks!

### DIFF
--- a/tools/baseimage/cmd/create_gce_x86_64_image/main.go
+++ b/tools/baseimage/cmd/create_gce_x86_64_image/main.go
@@ -125,9 +125,9 @@ func createImageMain(project, zone string) error {
 	if err := gce.WaitForInstance(project, zone, insName); err != nil {
 		return fmt.Errorf("waiting for instance error: %v", err)
 	}
-	log.Printf("deleting instance %q...", insName)
-	if err := h.DeleteInstance(insName); err != nil {
-		return fmt.Errorf("error deleting instance: %v", err)
+	log.Printf("stopping instance %q...", insName)
+	if err := h.StopInstance(insName); err != nil {
+		return fmt.Errorf("error stopping instance: %v", err)
 	}
 	if err := h.CreateImage(insName, attachedDiskName, outImageName); err != nil {
 		return fmt.Errorf("failed to create image: %w", err)
--- a/tools/baseimage/pkg/gce/helper.go
+++ b/tools/baseimage/pkg/gce/helper.go
@@ -122,6 +122,14 @@ func (h *GceHelper) CreateInstance(name string) (*compute.Instance, error) {
 	return payload, nil
 }
 
+func (h *GceHelper) StopInstance(name string) error {
+	op, err := h.Service.Instances.Stop(h.Project, h.Zone, name).Do()
+	if err != nil {
+		return err
+	}
+	return h.waitForOperation(op)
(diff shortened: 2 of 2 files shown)


tools/baseimage/cmd/create_gce_x86_64_image/main.go
ser-io, line 1: Let's keep these log message in the binary. These log messages are a feature of the binary, not the library. We shouldn't force other binaries using this library to have these messages written in their output without some sort of  acknowledgement/setup.

I foresee you might need similar logging message in your new binary. It's ok to have them duplicated, we don't need to introduce any logging parameters to the gce helper for now.

tools/baseimage/cmd/create_gce_x86_64_image/main.go
```diff
@@ -46,30 +46,32 @@ func createImageMain(project, zone string) error {
 	}
 	insName := outImageName
 	attachedDiskName := fmt.Sprintf("%s-attached-disk", insName)
+	disk, err := h.CreateDisk(debianSourceImageProject, debianSourceImage, attachedDiskName)
+	if err != nil {
+		return fmt.Errorf("failed to create disk: %w", err)
+	}
 	defer func() {
```
ser-io: In this function `createImageMain` I'd like to keep the cleanup logic centralized in a single defer block of code rather than isolated defer calls all over the function, as I read this logic as whole and not individual GCP calls. I understand `DeleteDisk` might be called without the disk existing, but this is not an issue.

In your new binary, feel free to have the defer calls split all over your function, but let's not have pattern on how this binaries needs to be written and give the author flexibility.

tools/baseimage/cmd/create_gce_x86_64_image/main.go
```diff
@@ -44,7 +45,7 @@ func createImageMain(project, zone string) error {
 	if err != nil {
 		return fmt.Errorf("failed to create GCE helper: %w", err)
 	}
-	insName := outImageName
+	insName := fmt.Sprintf("cf-%s", uuid.NewString())
```
ser-io: Keep the constant name for now in this binary.

I haven't decided how to solve the naming problem yet. Names like fmt.Sprintf("cf-%s", uuid.NewString()) are non-memorable and non-descriptive and both are properties I want the names generated by this script to have.

tools/baseimage/cmd/create_gce_x86_64_image/main.go
0405ysj: Acknowledged your requests. I'd understand it should be okay to have duplicated code across the existing one and the new one. I've gotten rid of last 3 commits overall.

tools/baseimage/cmd/create_gce_x86_64_image/main.go
0405ysj: ditto

tools/baseimage/cmd/create_gce_x86_64_image/main.go
0405ysj: ditto

ser-io (approved):

PR #1576, opened 2025-08-25
Wire cvdalloc up to run_cvd.

3405691582:
The instance id is passed to cvdalloc here so that cvdalloc can just simply trust that it is the appropriate instance ID and let that be managed elsewhere.

Recall that we need run_cvd to invoke socketpair(2) for acting on key events with the cvdalloc process: when allocation has completed, when the instance is stopping and resources should be torn down, and when teardown is complete.

We must wait on allocation completion so that we know we can start up the vmm without issue. To do that we need to override the WaitForAvailability method provided through VmmDependencyCommand; all classes implementing VmmDependencyCommand will have their WaitForAvailability methods called synchronously before the vmm is started, which achieves the intended result.

When run_cvd is shutting down cleanly, we could potentially just have run_cvd close its end of the socket, let cvdalloc invoke its teardown sequence, and have run_cvd abandon the cvdalloc process (letting init(1) or equivalent adopt it), but this loses the ability to see cvdalloc's log messages. Instead, we wait explicitly on teardown completion for a clean shutdown.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
+++ b/base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
@@ -1,0 +1,118 @@
+//
+// Copyright (C) 2025 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "cuttlefish/host/commands/run_cvd/launch/cvdalloc.h"
+
+#include <sys/socket.h>
+#include <chrono>
+#include <string>
+#include <unordered_set>
+#include <utility>
+#include <vector>
+
+#include <android-base/logging.h>
+#include <fruit/component.h>
+#include <fruit/fruit_forward_decls.h>
+#include <fruit/macro.h>
+
+#include "cuttlefish/common/libs/fs/shared_fd.h"
+#include "cuttlefish/common/libs/utils/result.h"
+#include "cuttlefish/common/libs/utils/subprocess.h"
+#include "cuttlefish/host/commands/cvdalloc/sem.h"
+#include "cuttlefish/host/libs/config/cuttlefish_config.h"
+#include "cuttlefish/host/libs/config/known_paths.h"
+#include "cuttlefish/host/libs/feature/command_source.h"
+#include "cuttlefish/host/libs/feature/feature.h"
+#include "cuttlefish/host/libs/vm_manager/vm_manager.h"
+
+namespace cuttlefish {
+namespace {
+
+constexpr std::chrono::seconds kCvdAllocateTimeout = std::chrono::seconds(30);
+constexpr std::chrono::seconds kCvdTeardownTimeout = std::chrono::seconds(2);
+
+class Cvdalloc : public vm_manager::VmmDependencyCommand {
+ public:
+  INJECT(Cvdalloc(const CuttlefishConfig::InstanceSpecific &instance))
+      : instance_(instance) {}
+
+  // CommandSource
+  Result<std::vector<MonitorCommand>> Commands() override {
+    auto nice_stop = [this]() { return Stop(); };
+
+    Command cmd(CvdallocBinary(), KillSubprocessFallback(nice_stop));
+    cmd.AddParameter("--id=", instance_.id());
+    cmd.AddParameter("--socket=", their_socket_);
+    std::vector<MonitorCommand> commands;
+    commands.emplace_back(std::move(cmd));
+    return commands;
+  }
+
+  std::string Name() const override { return "Cvdalloc"; }
+  bool Enabled() const override { return instance_.use_cvdalloc(); }
+  std::unordered_set<SetupFeature *> Dependencies() const override {
+    return {};
+  }
+
+  // StatusCheckCommandSource
+  Result<void> WaitForAvailability() const override {
+    LOG(INFO) << "cvdalloc (run_cvd): waiting to finish allocation.";
+    CF_EXPECT(cvdalloc::Wait(socket_, kCvdAllocateTimeout),
+              "cvdalloc (run_cvd): Wait failed");
+    LOG(INFO) << "cvdalloc (run_cvd): allocation is done.";
+
+    return {};
+  }
+
+ private:
+  Result<void> ResultSetup() override {
+    std::pair<SharedFD, SharedFD> p =
+        CF_EXPECT(SharedFD::SocketPair(AF_LOCAL, SOCK_STREAM, 0));
+    socket_ = std::move(p.first);
+    their_socket_ = std::move(p.second);
+    return {};
+  }
+
+  StopperResult Stop() {
+    LOG(INFO) << "cvdalloc (run_cvd): stop requested; teardown started";
+    if (!cvdalloc::Post(socket_).ok()) {
+      LOG(INFO) << "cvdalloc (run_cvd): stop failed: couldn't Post";
+      return StopperResult::kStopFailure;
+    }
+
+    if (!cvdalloc::Wait(socket_, kCvdTeardownTimeout).ok()) {
+      LOG(INFO) << "cvdalloc (run_cvd): stop failed: couldn't Wait";
+      return StopperResult::kStopFailure;
+    }
+
+    LOG(INFO) << "cvdalloc (run_cvd): teardown completed";
+    return StopperResult::kStopSuccess;
+  }
+
+  const CuttlefishConfig::InstanceSpecific &instance_;
+  SharedFD socket_, their_socket_;
+};
+
+}  // namespace
+
+fruit::Component<fruit::Required<const CuttlefishConfig::InstanceSpecific>>
+CvdallocComponent() {
+  return fruit::createComponent()
+      .addMultibinding<vm_manager::VmmDependencyCommand, Cvdalloc>()
+      .addMultibinding<CommandSource, Cvdalloc>()
+      .addMultibinding<SetupFeature, Cvdalloc>();
+}
+
+}  // namespace cuttlefish
--- a/base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
@@ -125,14 +125,13 @@ Result<int> CvdallocMain(int argc, char *argv[]) {
   };
 
   CF_EXPECT(Allocate(id, bridge_name));
-  CF_EXPECT(Post(sock));
+  CF_EXPECT(cvdalloc::Post(sock));
 
   LOG(INFO) << "cvdalloc: waiting to teardown";
 
-  CF_EXPECT(Wait(sock, kSemNoTimeout));
+  CF_EXPECT(cvdalloc::Wait(sock, cvdalloc::kSemNoTimeout));
   std::move(teardown).Invoke();
-  CF_EXPECT(Post(sock));
-
+  CF_EXPECT(cvdalloc::Post(sock));
 
   return 0;
 }
--- a/base/cvd/cuttlefish/host/commands/cvdalloc/sem.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvdalloc/sem.cpp
@@ -17,7 +17,7 @@
 
 #include "cuttlefish/host/libs/command_util/util.h"
 
-namespace cuttlefish {
+namespace cuttlefish::cvdalloc {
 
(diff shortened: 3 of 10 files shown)


base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
```diff
@@ -0,0 +1,103 @@
...
+constexpr std::chrono::seconds kCvdTeardownTimeout = std::chrono::seconds(2);
+
+class Cvdalloc : public vm_manager::VmmDependencyCommand {
+ public:
+  INJECT(Cvdalloc(const CuttlefishConfig::InstanceSpecific &instance))
+      : instance_(instance) {}
+
+  // CommandSource
+  Result<std::vector<MonitorCommand>> Commands() override {
+    auto nice_stop = [this]() {
```
Databean: nit: can this be another method in the class? so that the callback looks like
```c++
auto nice_stop = [this]() { return NiceStop(); }
```

Will make it easier to read if it has less nesting.

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
```diff
@@ -0,0 +1,103 @@
...
+    LOG(INFO) << "cvdalloc (run_cvd): allocation is done.";
+
+    return {};
+  }
+
+ private:
+  Result<void> ResultSetup() override {
+    bool r = SharedFD::SocketPair(AF_LOCAL, SOCK_STREAM, 0, &socket_,
+                                  &their_socket_);
+    CF_EXPECT(r == true, "Unable to create socketpair: " << strerror(errno));
```
jemoreira: Please use the `SharedFD::SocketPair` version that returns a result of pair of sockets. You could even use structured binding for the assignment:
```
auto [socket_, their_socket_] = CF_EXPECT(SharedFD::SocketPair(AF_LOCAL, SOCK_STREAM, 0), ...);
```

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
```diff
@@ -0,0 +1,103 @@
...
+       LOG(INFO) << "cvdalloc (run_cvd): teardown completed";
+       return wait_ok? StopperResult::kStopSuccess : StopperResult::kStopFailure;
+    };
+    auto cmd = Command(CvdallocBinary(), KillSubprocessFallback(nice_stop));
+    cmd.AddParameter("--id=", instance_.id());
+    cmd.AddParameter("--socket=", their_socket_);
+
+    std::vector<MonitorCommand> commands;
+    commands.emplace_back(std::move(cmd));
+    return commands;
```
jemoreira, line 60: nit: `return {MonitorCommand(std::move(cmd))};`

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
```diff
@@ -0,0 +1,103 @@
...
+    return {};
+  }
+
+  // StatusCheckCommandSource
+  Result<void> WaitForAvailability() const override {
+    LOG(INFO) << "cvdalloc (run_cvd): waiting to finish allocation.";
+    auto result = Wait(socket_, kCvdAllocateTimeout);
+    if (!result.ok()) {
+      socket_->Close();
+      return result;
```
jemoreira: Do `return CF_EXPECT(std::move(result));` to add this function to the error stacktrace.

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
```diff
@@ -0,0 +1,103 @@
...
+  std::unordered_set<SetupFeature *> Dependencies() const override {
+    return {};
+  }
+
+  // StatusCheckCommandSource
+  Result<void> WaitForAvailability() const override {
+    LOG(INFO) << "cvdalloc (run_cvd): waiting to finish allocation.";
+    auto result = Wait(socket_, kCvdAllocateTimeout);
+    if (!result.ok()) {
+      socket_->Close();
```
jemoreira: Is this really necessary? In case of failure run_cvd is likely to try to stop the subprocess using the nice_stop procedure from `Commands()`, which is going to try to write to the socket first, shouldn't the socket be open until then?

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
```diff
@@ -0,0 +1,103 @@
...
+class Cvdalloc : public vm_manager::VmmDependencyCommand {
+ public:
+  INJECT(Cvdalloc(const CuttlefishConfig::InstanceSpecific &instance))
+      : instance_(instance) {}
+
+  // CommandSource
+  Result<std::vector<MonitorCommand>> Commands() override {
+    auto nice_stop = [this]() {
+       LOG(INFO) << "cvdalloc (run_cvd): stop requested; teardown started";
+       if (!Post(socket_).ok()) {
```
jemoreira: It would be nice to log the error message from the result of `Post`

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include <android-base/logging.h>
+
+#include "cuttlefish/common/libs/fs/shared_fd.h"
+#include "cuttlefish/common/libs/utils/result.h"
+#include "cuttlefish/common/libs/utils/subprocess.h"
+#include "cuttlefish/host/libs/config/cuttlefish_config.h"
+#include "cuttlefish/host/libs/config/known_paths.h"
+#include "cuttlefish/host/libs/vm_manager/vm_manager.h"
+#include "cuttlefish/host/libs/command_util/util.h"
+#include "cuttlefish/host/commands/cvdalloc/sem.h"
```
Databean: If `commands/cvdalloc/sem.h` is going to be used outside of `commands/cvdalloc` source files, the functions could use some more distinctive naming to avoid conflicts and confusion. `Wait` and `Post` define a protocol over a socket that involve sending a one-byte `char`, but the names are sufficiently generic that it's more confusing in `run_cvd` code that interacts with other processes using different protocols. `Wait` in a broader context could just as easily be purely a select(2) or poll(2) that doesn't actually mutate the socket, but this implementation additionally does a blocking read(2) that consumes a byte.

Alternatively, creating a C++ namespace like `cuttlefish::cvdalloc` would also make it clear that this is part of the cvdalloc protocol and not part of any other utility set.

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+  // CommandSource
+  Result<std::vector<MonitorCommand>> Commands() override {
+    auto nice_stop = [this]() {
+       LOG(INFO) << "cvdalloc (run_cvd): stop requested; teardown started";
+       if (!Post(socket_).ok()) {
+         return StopperResult::kStopFailure;
+       }
+
+       bool wait_ok = Wait(socket_, kCvdTeardownTimeout).ok();
```
jemoreira: same: log the error from `Wait` if it fails.

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
```diff
@@ -0,0 +1,103 @@
...
+       LOG(INFO) << "cvdalloc (run_cvd): teardown completed";
+       return wait_ok? StopperResult::kStopSuccess : StopperResult::kStopFailure;
+    };
+    auto cmd = Command(CvdallocBinary(), KillSubprocessFallback(nice_stop));
```
Databean: nit: can avoid `auto` here by writing

```c++
Command cmd(CvdallocBinary(), KillSubprocessFallback(nice_stop));
```

Alternatively, the setters on `Command` return the object back, so it's possible to write
```c++
std::vector<MonitorCommand> commands;
commands.emplace_back(
    Command(CvdallocBinary(), KillSubprocessFallback(nice_stop))
        .AddParameter("--id=", instance_.id())
        .AddParameter("--socket=", their_socket_));
return commands;
```

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
Databean: `socket_` and `their_socket_` will also [capture](https://github.com/google/android-cuttlefish/blob/025c51320154ff31491dfa3cacceee97abba558a/base/cvd/cuttlefish/common/libs/fs/shared_fd.cpp#L445) and then report the failure in `socket_->StrError()`, which offers some more security that `errno` hasn't been tampered with in the meantime.

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
Databean: Unfortunately `{}` initializer lists can't be used in this way for move-only types.

See https://stackoverflow.com/questions/8468774/can-i-list-initialize-a-vector-of-move-only-type

Databean: We discussed this two years ago: https://android-review.git.corp.google.com/c/device/google/cuttlefish/+/2527199/comment/9f5ca005_4b8ad24d/

jemoreira: Never mind then. Though I'm sure I'll forget and you'll have to remind me again in 2 years :)

Databean: single_element_emplace

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
3405691582: That declaration shadows those fields; you'd have to have temporaries and reassign them. Would you like me to do it like that anyway?

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
3405691582: I think this is just a remnant, maybe. I just removed it.

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
3405691582: Done.

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
3405691582: Done.

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
3405691582: Done.

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
3405691582: I had a mind to more generic usage, but you're right about exposing this in this way. I just added it to a sub-namespace for now.

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
3405691582: I did the former.

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
3405691582: Done.

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
3405691582: (Moot.)

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
jemoreira: no, don't use structured binding (what was I thinking?), but do use the SockerPair overhaul I suggested.

jemoreira: The errno value is only captured in case of success though, which is exactly when we don't need it. To be able to use StrError() you have to use the other overload of the SocketPair function.

3405691582: Done.

3405691582: The other SocketPair CF_EXPECT's with StrError, which should be sufficient, right?

3405691582: I'm going to resolve this; I think this is correct/as intended for the suggestion.

PR #1577, opened 2025-08-25
Return build information from fetches

cjreynol:
Include the branch+target, even when given a build ID. 

Refactored to condense duplicate AB API calls into one, when all the information was available at once.

Bug: 440398826


base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
```diff
@@ -219,11 +184,66 @@ Result<SeekableZipSource> AndroidBuildApi::FileReader(
...
+  while (retry_period_ != std::chrono::seconds::zero() &&
+         !StatusIsTerminal(status)) {
+    LOG(DEBUG) << "Status is \"" << status << "\". Waiting for "
+               << retry_period_.count() << " seconds and checking again.";
+    std::this_thread::sleep_for(retry_period_);
+    auto response =
+        CF_EXPECT(HttpGetToJson(http_client_, url, CF_EXPECT(Headers())));
+    Json::Value json = CF_EXPECT(GetResponseJson(response),
+                                 "Error retrying build status retrieval");
+    status = CF_EXPECT(GetValue<std::string>(json, {"buildAttemptStatus"}));
```
jemoreira, line 237: Why can't this use `CF_EXPECT(GetBuildInfo()).status`?

jemoreira (approved):

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
cjreynol: I did not reuse it because of the contextual error checking in the middle of that method based on the input from `LatestBuildId`.  Those messages do not make sense once we have successfully used the `proposed_build_id` and know it returns information from the API, so I did not want them to pop up here.  Will take another look and see about a way to re-use.

cjreynol: After trying a few ideas, I was unable to think of a cleaner way to separate out the error checking from the rest of the `GetBuildInfo` logic and make it re-usable.  The possible errors we cannot detect until the first call after `LatestBuildId` makes that logic so tightly coupled.  I would not want those error cases triggering in subsequent calls, so the logic is not really reusable.  I could split the URL building and response processing into a helper, but IMO it does not improve the situation.

I am going to leave it as is, but happy to discuss any ideas you have and make changes in a follow-up PR if we come up with something.

PR #1579, opened 2025-08-26
Sync ExtractArtifact between Host Orchestrator and its golang client

0405ysj:
Context: b/441169111

`UserArtifactsManager` of Host Orchestrator is taking responsibility of replying status 409(conflict) before extracting user artifacts. So status 409 can be replied with `/operations/{name}/:wait`, instead of `/v1/userartifacts/{checksum}/:extract`. However, its golang client tries to handle status 409 from `/v1/userartifacts/{checksum}/:extract`. This PR aims to fix difference between Host Orchestrator and its golang client.

### DIFF
--- a/frontend/src/libhoclient/host_orchestrator_client.go
+++ b/frontend/src/libhoclient/host_orchestrator_client.go
@@ -593,16 +593,18 @@ func (c *HostOrchestratorClientImpl) ExtractArtifact(filename string) error {
 		return err
 	}
 	op := &hoapi.Operation{}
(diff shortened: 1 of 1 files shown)


ikicha (approved):

jemoreira (approved):
I'm approving this because the end result is better than what we have, but the best approach is to return 409 on the :extract call, not on the operation.

0405ysj:
> I'm approving this because the end result is better than what we have, but the best approach is to return 409 on the :extract call, not on the operation.

Agree. Filed a bug for it. b/441375129

PR #1580, opened 2025-08-26
gce_install_cuttlefish_packages tool.

ser-io:
- Given a base image, create an amended one installing cuttlefish debian packages.

Bug: b/441328183

### DIFF
--- a/tools/baseimage/cmd/gce_install_cuttlefish_packages/main.go
+++ b/tools/baseimage/cmd/gce_install_cuttlefish_packages/main.go
@@ -1,0 +1,183 @@
+// Copyright (C) 2025 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"log"
+	"path/filepath"
+	"time"
+
+	"github.com/google/android-cuttlefish/tools/baseimage/pkg/gce"
+	"github.com/google/android-cuttlefish/tools/baseimage/pkg/gce/scripts"
+)
+
+const (
+	outImageName = "amended-image"
+)
+
+var (
+	project                    = flag.String("project", "", "GCP project whose resources will be used for creating the amended image")
+	zone                       = flag.String("zone", "us-central1-a", "GCP zone used for creating relevant resources")
+	source_image_project       = flag.String("source-image-project", "", "Source image GCP project")
+	source_image               = flag.String("source-image", "", "Source image name")
+	cuttlefish_debs_zip_source = flag.String("cuttlefish-debs-zip-source", "", "Local path to zip file containing cuttlefish debian packages")
+)
+
+type amendImageOpts struct {
+	SourceImageProject      string
+	SourceImage             string
+	CuttlefishDebsZipSource string
+}
+
+func installCuttlefishDebs(project, zone, insName, zipSrc string) error {
+	dstSrc := "/tmp/" + filepath.Base(zipSrc)
+	if err := gce.UploadFile(project, zone, insName, zipSrc, dstSrc); err != nil {
+		return fmt.Errorf("error uploading %s: %v", zipSrc, err)
+	}
+	if err := gce.UploadBashScript(project, zone, insName, "install.sh", scripts.InstallCuttlefishPackages); err != nil {
+		return fmt.Errorf("error uploading script: %v", err)
+	}
+	if err := gce.RunCmd(project, zone, insName, "./install.sh "+dstSrc); err != nil {
+		return err
+	}
+	return nil
+}
+
+func cleanupDeleteDisk(h *gce.GceHelper, disk string) {
+	log.Printf("cleanup: deleting disk %q...", disk)
+	if err := h.DeleteDisk(disk); err != nil {
+		log.Printf("cleanup: error deleting disk: %v", err)
+	} else {
+		log.Println("cleanup: disk deleted")
+	}
+}
+
+func cleanupDeleteInstance(h *gce.GceHelper, ins string) {
+	log.Printf("cleanup: deleting instance %q...", ins)
+	if err := h.DeleteInstance(ins); err != nil {
+		log.Printf("cleanup: error deleting instance: %v", err)
+	} else {
+		log.Println("cleanup: instance deleted")
+	}
+}
+
+func cleanupDetachDisk(h *gce.GceHelper, ins, disk string) {
+	log.Printf("cleanup: detaching disk %q from instance %q...", ins, disk)
+	if err := h.DetachDisk(ins, disk); err != nil {
+		log.Printf("cleanup: error detaching disk: %v", err)
+	} else {
+		log.Println("cleanup: disk detached")
+	}
+}
+
+func amendImageMain(project, zone string, opts amendImageOpts) error {
+	h, err := gce.NewGceHelper(project, zone)
+	if err != nil {
+		return fmt.Errorf("failed to create GCE helper: %w", err)
+	}
+	insName := outImageName
+	attachedDiskName := fmt.Sprintf("%s-attached-disk", insName)
+
+	log.Println("creating disk...")
+	_, err = h.CreateDisk(opts.SourceImageProject, opts.SourceImage, attachedDiskName)
+	if err != nil {
+		return fmt.Errorf("failed to create disk: %w", err)
+	}
+	log.Printf("disk created: %q", attachedDiskName)
+	defer cleanupDeleteDisk(h, attachedDiskName)
+
+	log.Println("creating instance...")
+	_, err = h.CreateInstance(insName)
+	if err != nil {
+		return fmt.Errorf("failed to create instance: %w", err)
+	}
+	log.Printf("instance created: %q", insName)
+	defer cleanupDeleteInstance(h, insName)
+
+	log.Println("attaching disk...")
+	if err := h.AttachDisk(insName, attachedDiskName); err != nil {
+		log.Fatalf("failed to attach disk %q to instance %q: %v", attachedDiskName, insName, err)
+	}
+	log.Println("disk attached")
+	defer cleanupDetachDisk(h, insName, attachedDiskName)
+
+	if err := gce.WaitForInstance(project, zone, insName); err != nil {
+		return fmt.Errorf("waiting for instance error: %v", err)
+	}
+
+	installCuttlefishDebs(project, zone, insName, opts.CuttlefishDebsZipSource)
+
+	// Reboot the instance to force a clean umount of the attached disk's file system.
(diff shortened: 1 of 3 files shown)


tools/baseimage/cmd/amend_gce_base_image/main.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func ammendImageMain(project, zone string, opts amendImageOpts) error {
+	h, err := gce.NewGceHelper(project, zone)
+	if err != nil {
+		return fmt.Errorf("failed to create GCE helper: %w", err)
+	}
+	insName := outImageName
+	attachedDiskName := fmt.Sprintf("%s-attached-disk", insName)
+	defer func() {
```
jemoreira: Should this be done after checking `CreateDisk` succeeded? As it is now it will run even if `CreateDisk` fails.

tools/baseimage/cmd/amend_gce_base_image/main.go
jemoreira, line 1: Can this share more code with the create_gce_x86_64_image script?

I'd expect creating the VM, setting up the disks and installing the packages to be roughly the same code for these scripts (with only the actual base image changing). The main difference would be in the extra steps taken around installing drivers and updating the host kernel.

jemoreira, line 1: nit: Consider renaming the script as `upgrade_cuttlefish_packages` instead. It's in under a directory  named `baseimage` so there is no doubt that this will be done in an image and not in the host where the script runs.

ser-io: Agree, there's room for refactoring here. However, I don't have a clear picture of what the abstractions should be at the moment resorting into duplication for now.

ser-io: renamed to `gce_install_cuttlefish_packages`.

tools/baseimage/cmd/amend_gce_base_image/main.go
ser-io: Done.

jemoreira (approved):

tools/baseimage/cmd/gce_install_cuttlefish_packages/main.go
```diff
@@ -0,0 +1,183 @@
...
+func cleanupDetachDisk(h *gce.GceHelper, ins, disk string) {
+	log.Printf("cleanup: detaching disk %q from instance %q...", ins, disk)
+	if err := h.DetachDisk(ins, disk); err != nil {
+		log.Printf("cleanup: error detaching disk: %v", err)
+	} else {
+		log.Println("cleanup: disk detached")
+	}
+}
+
+func ammendImageMain(project, zone string, opts amendImageOpts) error {
```
0405ysj: typo: amendImageMain

tools/baseimage/cmd/gce_install_cuttlefish_packages/main.go
```diff
@@ -0,0 +1,183 @@
...
+	// Reboot the instance to force a clean umount of the attached disk's file system.
+	if err := gce.RunCmd(project, zone, insName, "sudo reboot"); err != nil {
+		return err
+	}
+	time.Sleep(2 * time.Minute)
+	if err := gce.WaitForInstance(project, zone, insName); err != nil {
+		return fmt.Errorf("waiting for instance error: %v", err)
+	}
+	log.Printf("deleting instance %q...", insName)
+	if err := h.DeleteInstance(insName); err != nil {
```
0405ysj: Consider `StopInstance` rather than `DeleteInstance`. I speculate `cleanupDeleteInstance` always print an error.

tools/baseimage/cmd/gce_install_cuttlefish_packages/main.go
0405ysj, line 1: What was the consideration for creating another script for amending image, instead of installing CF debs when creating the base image?

tools/baseimage/cmd/gce_install_cuttlefish_packages/main.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func cleanupDetachDisk(h *gce.GceHelper, ins, disk string) {
+	log.Printf("cleanup: detaching disk %q from instance %q...", ins, disk)
+	if err := h.DetachDisk(ins, disk); err != nil {
+		log.Printf("cleanup: error detaching disk: %v", err)
+	} else {
+		log.Println("cleanup: disk detached")
+	}
+}
```
0405ysj, line 85: Why don't we place these functions into helpers.go? I wish to use it in https://github.com/google/android-cuttlefish/pull/1581 or following PR. Existing code would definitely want it.

tools/baseimage/cmd/gce_install_cuttlefish_packages/main.go
```diff
@@ -0,0 +1,183 @@
...
+	dstSrc := "/tmp/" + filepath.Base(zipSrc)
+	if err := gce.UploadFile(project, zone, insName, zipSrc, dstSrc); err != nil {
+		return fmt.Errorf("error uploading %s: %v", zipSrc, err)
+	}
+	if err := gce.UploadBashScript(project, zone, insName, "install.sh", scripts.InstallCuttlefishPackages); err != nil {
+		return fmt.Errorf("error uploading script: %v", err)
+	}
+	if err := gce.RunCmd(project, zone, insName, "./install.sh "+dstSrc); err != nil {
+		return err
+	}
```
0405ysj, line 56: I feel this logic can be repeated, calling `RunCmd` after `UploadBashScript`. Do you mind to refactor it?

tools/baseimage/cmd/gce_install_cuttlefish_packages/main.go
```diff
@@ -0,0 +1,183 @@
...
+const (
+	outImageName = "amended-image"
+)
+
+var (
+	project                    = flag.String("project", "", "GCP project whose resources will be used for creating the amended image")
+	zone                       = flag.String("zone", "us-central1-a", "GCP zone used for creating relevant resources")
+	source_image_project       = flag.String("source-image-project", "", "Source image GCP project")
+	source_image               = flag.String("source-image", "", "Source image name")
+	cuttlefish_debs_zip_source = flag.String("cuttlefish-debs-zip-source", "", "Local path to zip file containing cuttlefish debian packages")
```
0405ysj, line 37: Could you leave a short document around how to use this binary? Since default values of those flags are empty, it's tricky for understanding how to use it.

tools/baseimage/cmd/gce_install_cuttlefish_packages/main.go
```diff
@@ -0,0 +1,183 @@
...
+	ins, err := h.CreateInstance(insName)
+	if err != nil {
+		return fmt.Errorf("failed to create instance: %w", err)
+	}
+	log.Printf("instance created: %q", insName)
+	defer cleanupDeleteInstance(h, insName)
+
+	log.Println("attaching disk...")
+	if err := h.AttachDisk(insName, attachedDiskName); err != nil {
+		log.Fatalf("failed to attach disk %q to instance %q: %v", disk.Name, ins.Name, err)
```
0405ysj: We already have `insName` and `attachedDiskName`. Why don't we use those, rather than dealing with return values?

tools/baseimage/cmd/gce_install_cuttlefish_packages/main.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	if *source_image_project == "" {
+		log.Fatal("usage: `-source-image-project` must not be empty")
+	}
+	if *source_image == "" {
+		log.Fatal("usage: `-source_image` must not be empty")
+	}
+	if *cuttlefish_debs_zip_source == "" {
+		log.Fatal("usage: `-cuttlefish-debs-zip-source` must not be empty")
+	}
```
0405ysj, line 163: Is this `-<arg>`, rather than `--<arg>`?

tools/baseimage/cmd/gce_install_cuttlefish_packages/main.go
```diff
@@ -0,0 +1,183 @@
...
+	installCuttlefishDebs(project, zone, insName, opts.CuttlefishDebsZipSource)
+
+	// Reboot the instance to force a clean umount of the attached disk's file system.
+	if err := gce.RunCmd(project, zone, insName, "sudo reboot"); err != nil {
+		return err
+	}
+	time.Sleep(2 * time.Minute)
+	if err := gce.WaitForInstance(project, zone, insName); err != nil {
+		return fmt.Errorf("waiting for instance error: %v", err)
+	}
```
0405ysj, line 131: This is repeated either. Can't we place it into helper.go?

0405ysj (commented):
Could you add me as a reviewer or cc for PRs under `tools/baseimage` next time? I want to get notification at least, since decisions(e.g. code style) made here can affect my work including https://github.com/google/android-cuttlefish/pull/1581.

tools/baseimage/cmd/gce_install_cuttlefish_packages/main.go
ser-io: Done

tools/baseimage/cmd/gce_install_cuttlefish_packages/main.go
ser-io: Done

ser-io:
> Could you add me as a reviewer or cc for PRs under `tools/baseimage` next time? I want to get notification at least, since decisions(e.g. code style) made here can affect my work including #1581.

Sounds good. I'll cc'd you on changes modifying common code. From my understanding, you don't use the `create_gce_x86_64_image` and `gce_install_cuttlefish_packages` binaries, so I won't bother you with changes specific to these binaries as they don't affect the other binaries behavior.

tools/baseimage/cmd/gce_install_cuttlefish_packages/main.go
ser-io: Separation of concern. We want to install different version of cuttlefish packages in the same already created base image.

tools/baseimage/cmd/gce_install_cuttlefish_packages/main.go
ser-io: It could be done in a follow up PR.

tools/baseimage/cmd/gce_install_cuttlefish_packages/main.go
ser-io: Done.

tools/baseimage/cmd/gce_install_cuttlefish_packages/main.go
ser-io: `gce_install_cuttlefish_packages -h` should be good enough. More extensive documentation could be added when needed, otherwise it becomes stale pretty quick.

tools/baseimage/cmd/gce_install_cuttlefish_packages/main.go
ser-io: both works. however `gce_install_cuttlefish_packages -h` uses `-`, hence `-` is used for consistency purposes.

tools/baseimage/cmd/gce_install_cuttlefish_packages/main.go
ser-io: I'd like to keep it repeated for now. These are small and simple scripts and don't think there's a big need for refactoring and add new abstractions. The clearer the better.

tools/baseimage/cmd/gce_install_cuttlefish_packages/main.go
ser-io: I'd like to keep it repeated for now.

PR #1581, opened 2025-08-27
Create GCE image for cloud orchestration

0405ysj:
This PR aims for obtaining a valid script for creating GCE image containing Cloud Orchestrator inside.

Context: b/433533301, b/436999931

### DIFF
--- a/tools/baseimage/pkg/gce/scripts/cloud_orchestration.go
+++ b/tools/baseimage/pkg/gce/scripts/cloud_orchestration.go
@@ -1,0 +1,46 @@
+// Copyright (C) 2025 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package scripts
+
+const InstallCODockerContainer = `#!/usr/bin/env bash
+set -o errexit -o nounset -o pipefail
+
+ARCH=$(uname -m)
+if [ "$ARCH" == "x86_64" ]; then
+    ARCH="amd64"
+elif [ "$ARCH" == "aarch64" ]; then
+    ARCH="arm64"
+fi
+
+# TODO(b/442924024): Apply chroot
+sudo install -m 0755 -d /etc/apt/keyrings
+sudo curl -fsSL https://download.docker.com/linux/debian/gpg -o /etc/apt/keyrings/docker.asc
+sudo chmod a+r /etc/apt/keyrings/docker.asc
+echo "deb [arch=$ARCH signed-by=/etc/apt/keyrings/docker.asc] https://download.docker.com/linux/debian bookworm stable" |\
+    sudo tee /etc/apt/sources.list.d/docker.list
+sudo apt update
+sudo apt install -y containerd.io docker-buildx-plugin docker-ce docker-ce-cli docker-compose-plugin
+
+sudo docker pull us-docker.pkg.dev/android-cuttlefish-artifacts/cuttlefish-orchestration/cuttlefish-cloud-orchestrator
+sudo docker pull us-docker.pkg.dev/android-cuttlefish-artifacts/cuttlefish-orchestration/cuttlefish-orchestration
+sudo mkdir -p /etc/cloud_orchestrator
+sudo wget -O /etc/cloud_orchestrator/conf.toml \
+    https://artifactregistry.googleapis.com/download/v1/projects/android-cuttlefish-artifacts/locations/us/repositories/cloud-orchestrator-config/files/on-premise-single-server:main:conf.toml:download?alt=media
+
+sudo docker run --restart unless-stopped -d -p 8080:8080 -e CONFIG_FILE="/conf.toml" \
+    -v /etc/cloud_orchestrator/conf.toml:/conf.toml \
+    -v /var/run/docker.sock:/var/run/docker.sock \
+    -t us-docker.pkg.dev/android-cuttlefish-artifacts/cuttlefish-orchestration/cuttlefish-cloud-orchestrator:latest
+`
--- a/tools/baseimage/pkg/gce/scripts/orchestration.go
+++ b/tools/baseimage/pkg/gce/scripts/orchestration.go
@@ -14,23 +14,6 @@
 
 package scripts
 
-// https://cs.android.com/android/platform/superproject/main/+/main:device/google/cuttlefish/tools/update_gce_kernel.sh;drc=7f601ad9132960b58ee3d7fe8f8b382d20720a22
-const UpdateKernel = `#!/usr/bin/env bash
-set -o errexit -o nounset -o pipefail
-
-sudo apt update
-sudo apt install -t bookworm -y linux-image-cloud-amd64
-sudo reboot
-`
-
-// https://cs.android.com/android/platform/superproject/main/+/main:device/google/cuttlefish/tools/update_gce_kernel.sh;drc=7f601ad9132960b58ee3d7fe8f8b382d20720a22
-const RemoveOldKernel = `#!/usr/bin/env bash
-set -o errexit -o nounset -o pipefail
-
-dpkg --list | grep -v $(uname -r) | grep -E 'linux-image-[0-9]|linux-headers-[0-9]' | awk '{print $2" "$3}' | sort -k2,2 | awk '{print $1}' | xargs sudo apt-get -y purge
-sudo update-grub2
-`
-
 // https://cs.android.com/android/platform/superproject/+/android15-qpr2-release:device/google/cuttlefish/tools/create_base_image_gce.sh;drc=5480406e8fff1706a3901e2d7729b60ba8897aab
 // This script is based on existing create_base_image_gce.sh removing the part of
 // building and installing the cuttlefish debian packages.
--- a/.gitignore
+++ b/.gitignore
@@ -28,4 +28,5 @@ frontend/src/operator/webui/src/environments/version.ts
 /**/bazel-*
 MODULE.bazel.lock
 
+/tools/baseimage/create_gce_co_container_image
 /tools/baseimage/create_gce_x86_64_image
--- a/tools/baseimage/cmd/create_gce_co_container_image/main.go
+++ b/tools/baseimage/cmd/create_gce_co_container_image/main.go
@@ -1,0 +1,135 @@
+// Copyright (C) 2025 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"log"
+	"time"
+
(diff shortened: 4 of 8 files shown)


tools/baseimage/cmd/create_gce_cloud_orchestration_image/main.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	if err := rebootInstance(project, zone, instance); err != nil {
+		return err
+	}
+
+	log.Print("Creating GCE image from GCE image builder instance...")
+	if err := h.StopInstance(instance); err != nil {
+		return fmt.Errorf("error stopping instance: %v", err)
+	}
+	if err := h.CreateImage(instance, outImageName, []string{"IDPF"}); err != nil {
```
ser-io: Creating the image out of the boot disk is wrong as it leaks information. Boot disk image gets modified with stuff from GCP startup scripts we shouldn't port into the new image.

The solution for this is to work with an attached disk and chroot environment. This is the way existing scripts https://github.com/google/android-cuttlefish/blob/main/tools/baseimage/cmd/create_gce_x86_64_image/main.go work.

ser-io (changes requested):

tools/baseimage/cmd/create_gce_orchestration_image/main.go
ser-io, line 1: Don't rename this binary beforehand. Let's first introduce the new binary then rename the existing if needed.

tools/baseimage/cmd/create_gce_orchestration_image/main.go
```diff
@@ -97,8 +97,8 @@ func createImageMain(project, zone string) error {
 	if err := gce.UploadBashScript(project, zone, insName, "remove_old_kernel.sh", scripts.RemoveOldKernel); err != nil {
 		return fmt.Errorf("error uploading remove_old_kernel.sh: %v", err)
 	}
-	if err := gce.UploadBashScript(project, zone, insName, "create_base_image.sh", scripts.CreateBaseImage); err != nil {
-		return fmt.Errorf("error uploading create_base_image.sh: %v", err)
+	if err := gce.UploadBashScript(project, zone, insName, "create_orchestration_image.sh", scripts.CreateOrchestrationImage); err != nil {
```
ser-io: create_orchestration_image.sh name is incorrect in this context, keep it as `create_base_image.sh`.  The concept of `orchestration` is an advanced concept not relevant in the context of the base image.

tools/baseimage/pkg/gce/scripts/orchestration.go
```diff
@@ -14,27 +14,10 @@
...
-
-dpkg --list | grep -v $(uname -r) | grep -E 'linux-image-[0-9]|linux-headers-[0-9]' | awk '{print $2" "$3}' | sort -k2,2 | awk '{print $1}' | xargs sudo apt-get -y purge
-sudo update-grub2
-`
-
 // https://cs.android.com/android/platform/superproject/+/android15-qpr2-release:device/google/cuttlefish/tools/create_base_image_gce.sh;drc=5480406e8fff1706a3901e2d7729b60ba8897aab
 // This script is based on existing create_base_image_gce.sh removing the part of
 // building and installing the cuttlefish debian packages.
-const CreateBaseImage = `#!/usr/bin/env bash
+const CreateOrchestrationImage = `#!/usr/bin/env bash
```
ser-io: Same as before. CreateOrchestrationImage name is incorrect in this context, keep it as CreateBaseImage.

tools/baseimage/cmd/create_gce_cloud_orchestration_image/main.go
ser-io, line 1: Following discussion: https://github.com/google/android-cuttlefish/pull/1580#discussion_r2302148677, rename this binary as: `gce_install_cloud_orchestrator_container` or `gce_install_co_container`.

Let's be clear and direct with the name here, this script installs docker and install the cloud orchestrator container. The name `create_gce_cloud_orchestration_image` is misleading because this image can only be used in a very specific deployment of the cloud orchestration solution (single-server)

tools/baseimage/pkg/gce/scripts/cloud_orchestration.go
```diff
@@ -0,0 +1,45 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package scripts
+
+const CreateCloudOrchestrationImage = `#!/usr/bin/env bash
```
ser-io: Use a more specific name: InstallCODockerContainer

ser-io (changes requested):

tools/baseimage/cmd/create_gce_cloud_orchestration_image/main.go
0405ysj: I tried it before, but it didn't work well straightforward while dealing with docker daemon service. I will leave a TODO comment for this PR.

I'm not against on attaching disks, but it's not my highest priorities to do. Especially, I should have further iteration with this image such as setting up TURN server or alternative. Please understand I can't be stranded long time for this issue.

> Creating the image out of the boot disk is wrong as it leaks information. Boot disk image gets modified with stuff from GCP startup scripts we shouldn't port into the new image.

I missed to ask about this before. Which information is leaked? Is it critical? How can I check whether information is leaked or not?

tools/baseimage/pkg/gce/scripts/orchestration.go
0405ysj: After reading https://github.com/google/android-cuttlefish/pull/1580, I agree `CreateOrchestrationImage` can't be a good name since it doesn't contain any CF deb packages. OTOH I'm afraid of the naming `CreateBaseImage`, because it gives reader impression this script is commonly used for all kinds of images defined under `tools/baseimage` but actually not. We won't need JDK or tradefed for GCE image with CO. Thus, `CreateBaseImage` isn't appropriate either.

What about `CreateOrchestrationBaseImage` instead?

tools/baseimage/cmd/create_gce_cloud_orchestration_image/main.go
0405ysj: Combining discussion here and https://github.com/google/android-cuttlefish/pull/1581#discussion_r2308354586. Assuming I've accepted both comments, subdirs of `tools/baseimage/cmd` will be shown as below.

- create_gce_x86_64_image
- gce_install_cuttlefish_packages
- gce_install_co_container

However it doesn't make a clear understanding to future users. It gives impressions to users as `gce_install_cuttlefish_packages` and `gce_install_co_container` are in the same level and `create_gce_x86_64_image` is a common image used for both, which is different from our sight.

- `create_gce_x86_64_image` won't be a good name anymore, with introducing `gce_install_co_container`.
- `gce_install_cuttlefish_packages` is still a good name to represent its role.
- `gce_install_co_container` can't be a good name. `gce_install_co_container` installing backport kernel and appending IDPF feature sounds weird...

So, what about
`create_gce_x86_64_image` -> `create_orchestration_base_image`
`gce_install_co_container` -> `create_co_container_image`
?

tools/baseimage/cmd/create_gce_orchestration_image/main.go
0405ysj: We can discuss at https://github.com/google/android-cuttlefish/pull/1581#discussion_r2308697480. I close this comment, to avoid separated discussion channel.

tools/baseimage/cmd/create_gce_orchestration_image/main.go
0405ysj: https://github.com/google/android-cuttlefish/pull/1581#discussion_r2308388016

tools/baseimage/pkg/gce/scripts/orchestration.go
0405ysj: https://github.com/google/android-cuttlefish/pull/1581#discussion_r2308388016

0405ysj:
@ser-io I need your comment as the circumstance is changed with https://github.com/google/android-cuttlefish/pull/1580. Especially, I need to sync with your thoughts around naming. Could you leave your thoughts again?

0405ysj:
@ser-io Gentle ping.

tools/baseimage/cmd/create_gce_cloud_orchestration_image/main.go
ser-io: Keep the existing binaries and variable names as is, don't make any name change in this PR.  You can pick the name you think it's best in your new tool.

Bring your new names proposal into the team meeting so we can have a bigger forum to decide the names we want to have here. Discussing these names in this PR is not productive as you don't have the full context of the existing names  and I don't have any context of why we need this new binary.

tools/baseimage/cmd/create_gce_cloud_orchestration_image/main.go
ser-io: At least, you'll be leaking the `/home/` directory containing your `ldap` and your teammates `ldaps` (members in the same GCP project) and their respective ssh keys. You can verify what else is leaked by attaching two disks one created with an original image and the other one with an image created after boot, mount both disks and check the diff:

```
sudo mount /dev/sdb1 /mnt/image_after_boot/
sudo mount /dev/sdc1 /mnt/original/
sudo diff --brief --recursive  /mnt/image_after_boot /mnt/original | less
```

I understand you might be having issues having the docker container running when you were creating the image using a chroot environment, perhaps capturing the state that allows the container to start on boot isn't straightforward. 

An alternative here is to add a new systemd service definition into the image and leverage systemd to spin the container for you. 

`/etc/systemd/system/co_docker_startup.service`

```
[Unit]
Description=co_docker_startup

[Service]
ExecStart=docker run ...

[Install]
WantedBy=multi-user.target
```

ser-io (changes requested):

tools/baseimage/cmd/create_gce_cloud_orchestration_image/main.go
0405ysj: > Keep the existing binaries and variable names as is, don't make any name change in this PR.

Why? There's no script or code referring this naming AFAIK. Also, we can update the procedure to build and run tool via `tools/baseimage/README.md`. Naming can be changed when it needs, and there's no strong cons for changing naming.

> you don't have the full context of the existing names

I already have sufficient context to discuss about naming. IIUC, `create_gce_x86_64_image` is for creating GCE base image used for GCE VM instances, and `gce_install_cuttlefish_packages` is for creating amended GCE image with installing CF debs(e.g. `cuttlefish-base`). The purpose is to separate frequency between running other base script(e.g. kernel update, JDK installation, ...) and running CF debs installation, no? CWIIW.

> I don't have any context of why we need this new binary.

You should know the context why it needs.. I thought we already discussed via both meeting and chat. It's searchable via meeting note or chat history.

> Bring your new names proposal into the team meeting so we can have a bigger forum to decide the names we want to have here.

I left an agenda on upcoming meeting. However I still prefer to get agreement here because we even have a delay to wait for the meeting and I feel naming is a thing we can make agreement via text.

tools/baseimage/cmd/create_gce_cloud_orchestration_image/main.go
0405ysj: Thanks for letting me know. Depending on the progress of naming discussion, I'd resolve this on this or the following PR.

tools/baseimage/cmd/create_gce_cloud_orchestration_image/main.go
ikicha: I think we'd better rename commands when we all agree that. If convos about renaming blocks that, I prefer to do that in follow-up meeting and CL, not this CL.. I just want to check in the new command creating a GCE instance with CO docker image which is necessary for our new use case sooner.

BTW I think the naming for new tool from @0405ysj looks good to me

 @ser-io WDYT?

ser-io: The new tool name: `create_co_container_image` works for me. Let's not make this CL about renaming existing stuff and focus this CL on the code of the new tool only.

0405ysj: Thanks for the reply. I'd change the naming as current conclusion, and close the conversation here.

tools/baseimage/cmd/create_gce_cloud_orchestration_image/main.go
0405ysj: Left TODO comment for a while, and should be tracked with other bugs. b/442924024

PR #1584, opened 2025-08-28
Vectorize assemble_cvd's --config flag

jemoreira:
The --config flag is used to figure out the default values of other flags based on the type of device. When launching multiple devices it needs to support different configs per device instead of applying the first's device's default values to all.

Bug: b/441595948

### DIFF
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/flags/system_image_dir.h
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/flags/system_image_dir.h
@@ -32,6 +32,8 @@ class SystemImageDirFlag {
   // flags, it is accessed directly by index instead.
   std::string ForIndex(size_t argument_index) const;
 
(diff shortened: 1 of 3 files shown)


base/cvd/cuttlefish/host/commands/assemble_cvd/flags/system_image_dir.h
```diff
@@ -32,6 +32,8 @@ class SystemImageDirFlag {
   // flags, it is accessed directly by index instead.
   std::string ForIndex(size_t argument_index) const;
 
+  size_t size() const { return system_image_dirs_.size(); }
```
Databean: nit: `Size`

Databean (approved):

base/cvd/cuttlefish/host/commands/assemble_cvd/flags/system_image_dir.h
jemoreira: Done

PR #1589, opened 2025-08-29
Minor cvd display fixes

jmacnak:


base/cvd/cuttlefish/host/libs/vm_manager/crosvm_display_controller.cpp
```diff
@@ -107,7 +107,9 @@ Result<int> CrosvmDisplayController::RunCrosvmDisplayCommand(
 
   Result<std::string> res = RunAndCaptureStdout(std::move(command));
   if (res.ok()) {
-    *stdout_str = CF_EXPECT(std::move(res));
+    if (stdout_str != nullptr) {
+      *stdout_str = CF_EXPECT(std::move(res));
```
Databean: nit: since this is already in a `res.ok()` block, it could also be `*stdout_str = std::move(*res);`

Databean (approved):

base/cvd/cuttlefish/host/libs/vm_manager/crosvm_display_controller.cpp
jmacnak: Done.

PR #1592, opened 2025-08-29
Create a type to represent --blank_data_image_mb

cjreynol:
To allow reuse of the parsing logic for metrics data gathering.

Bug: 442024462


base/cvd/cuttlefish/host/commands/assemble_cvd/flags/blank_data_image_mb.cc
```diff
@@ -0,0 +1,95 @@
...
+
+Result<int> GetDefaultValue(
+    const std::map<std::string, std::string>& default_value_lookup) {
+  auto lookup = default_value_lookup.find(kFlagName);
+  CF_EXPECT(lookup != default_value_lookup.end());
+  int result;
+  CF_EXPECTF(android::base::ParseInt(lookup->second, &result),
+             "Failed to parse value as integer: \"{}\"", lookup->second);
+  return result;
+}
```
Databean: As far as I know `default_value_lookup` will only pick up values that come from other `SetCommandLineOptionWithMode` calls. Since there aren't any for `blank_data_image_mb`, can this part be left out?

base/cvd/cuttlefish/host/commands/assemble_cvd/flags/blank_data_image_mb.cc
```diff
@@ -0,0 +1,95 @@
...
+  gflags::CommandLineFlagInfo flag_info =
+      gflags::GetCommandLineFlagInfoOrDie(kFlagName);
+  std::vector<std::string> string_values =
+      android::base::Split(flag_info.current_value, ",");
+  std::vector<int> values(string_values.size());
+
+  for (int i = 0; i < string_values.size(); i++) {
+    if (string_values[i] != "unset" && string_values[i] != "\"unset\"") {
+      int result;
+      CF_EXPECTF(android::base::ParseInt(string_values[i], &result),
```
Databean: nit: this could use `&values[i]` rather than the intermediate `result`.

base/cvd/cuttlefish/host/commands/assemble_cvd/flags/blank_data_image_mb.h
```diff
@@ -0,0 +1,46 @@
...
+      const std::map<std::string, std::string>& default_value_lookup,
+      const std::vector<GuestConfig>& guest_configs);
+
+  int ForIndex(std::size_t index) const;
+
+ private:
+  BlankDataImageMbFlag(const int, const std::vector<GuestConfig>&,
+                       std::vector<int>);
+
+  int default_value_;
```
cjreynol: set default to 0

base/cvd/cuttlefish/host/commands/assemble_cvd/flags/blank_data_image_mb.cc
```diff
@@ -0,0 +1,95 @@
...
+
+  return BlankDataImageMbFlag(default_value, guest_configs, std::move(values));
+}
+
+int BlankDataImageMbFlag::ForIndex(std::size_t index) const {
+  if (index < blank_data_image_mb_values_.size()) {
+    return blank_data_image_mb_values_[index];
+  } else if (index < guest_configs_.size() &&
+             guest_configs_[index].blank_data_image_mb != 0) {
+    return guest_configs_[index].blank_data_image_mb;
```
Databean: Rather than store `guest_configs`, can the value prioritization happen during `FromGlobalGflags`?

base/cvd/cuttlefish/host/commands/assemble_cvd/flags/blank_data_image_mb.cc
Databean: Actually even other `SetCommandLineOptionWithMode` calls should be covered by reading `FLAGS_blank_data_image_mb`.

I think this was used to write a generic `GET_FLAG_STR_VALUE` value that worked for any flag, and have it be able to retrieve the default value for that flag. But since this class is specific to one flag and is not generic over all flags, we know what the default value should be for this flag, and can use that value directly.

base/cvd/cuttlefish/host/commands/assemble_cvd/flags/blank_data_image_mb.cc
cjreynol: Done.

base/cvd/cuttlefish/host/commands/assemble_cvd/flags/blank_data_image_mb.cc
cjreynol: Thanks for looking into that, I was simply adapting the existing logic.

base/cvd/cuttlefish/host/commands/assemble_cvd/flags/blank_data_image_mb.cc
cjreynol: Done.

base/cvd/cuttlefish/host/commands/assemble_cvd/flags/blank_data_image_mb.cc
```diff
@@ -0,0 +1,81 @@
...
+
+  for (int i = 0; i < size; i++) {
+    if (i < string_values.size()) {
+      if (string_values[i] != "unset" && string_values[i] != "\"unset\"") {
+        CF_EXPECTF(android::base::ParseInt(string_values[i], &values[i]),
+                   "Failed to parse value as integer: \"{}\"",
+                   string_values[i]);
+      } else {
+        values[i] = default_value;
+      }
```
Databean: nit: can this conditional be reversed? Related to go/tott/685 I think this would be more readable as
```c++
if (string_values[i] == "unset" || string_values[i] == "\"unset\"") {
  values[i] = default_value;
} else {
  CF_EXPECT(android::base::ParseInt(...));
}
```

cjreynol: Done.

PR #1604, opened 2025-09-02
Create type wrapper for `--daemon` flag

cjreynol:
The second commit aims to write a wrapper-type equivalent of `GET_FLAG_BOOL_VALUE` used in `flags.cc`.  There are a number of desired flags for metrics with no unique logic and it seemed worthwhile to condense some of the boilerplate.  I thought about making it generic to work with both `bool` and `int` types, but the added complexity of bringing in templates and passing in parsing methods did not seem worth it for 2-3 types.  Happy to hear other opinions on the right balance.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/flags/bool_flag.cc
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/flags/bool_flag.cc
@@ -1,0 +1,62 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cuttlefish/host/commands/assemble_cvd/flags/bool_flag.h"
+
+#include <cstdlib>
+#include <string>
+#include <vector>
+
+#include <android-base/strings.h>
+#include <gflags/gflags.h>
+
+#include "cuttlefish/common/libs/utils/flag_parser.h"
+#include "cuttlefish/common/libs/utils/result.h"
+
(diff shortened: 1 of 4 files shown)


base/cvd/cuttlefish/host/commands/assemble_cvd/flags/daemon.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include <cstdlib>
+#include <vector>
+
+#include "cuttlefish/common/libs/utils/result.h"
+#include "cuttlefish/host/commands/assemble_cvd/flags/bool_flag.h"
+
+namespace cuttlefish {
+
+class DaemonFlag : public BoolFlag {
```
Databean, line 27: Can this use [composition rather than inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance)?

base/cvd/cuttlefish/host/commands/assemble_cvd/flags/bool_flag.cc
```diff
@@ -0,0 +1,62 @@
...
+
+#include "cuttlefish/common/libs/utils/flag_parser.h"
+#include "cuttlefish/common/libs/utils/result.h"
+
+namespace cuttlefish {
+
+Result<BoolFlag> BoolFlag::FromGlobalGflagsAndName(
+    const std::string& flag_name) {
+  gflags::CommandLineFlagInfo flag_info =
+      gflags::GetCommandLineFlagInfoOrDie(flag_name.c_str());
```
Databean, line 34: Can the `gflags::CommandLineFlagInfo` instance the argument rather than `flag_name`? That makes the interaction with gflags globals and possible failing assertion (if the caller uses an `...OrDie` function) more obvious.

base/cvd/cuttlefish/host/commands/assemble_cvd/flags/bool_flag.h
```diff
@@ -0,0 +1,40 @@
...
+#include <string>
+#include <vector>
+
+#include "cuttlefish/common/libs/utils/result.h"
+
+namespace cuttlefish {
+
+class BoolFlag {
+ public:
+  BoolFlag(BoolFlag&&) = default;
```
Databean, line 29: This class can safely be both copyable and movable. Declaring the move constructor makes it only movable. Following [Copyable and Movable types](https://google.github.io/styleguide/cppguide.html#Copyable_Movable_Types) can this also be a copyable type?

cjreynol:
Closing this PR in favor of #1680, which is based on a new source branch.

PR #1612, opened 2025-09-04
Ensure the instance RIL is set up properly.

3405691582:
assemble_cvd introspects the on-device tap interfaces to populate flags so that the instance RIL gets the right network information. However, when using cvdalloc, the tap devices aren't available at assemble_cvd time, so defaults are used.

Unfortunately, these defaults aren't correct. Instead, we do the same thing that cuttlefish-host-resources does: allocates subnetwork parameters based on instance ID. The RIL flags are populated anticipating the network parameters we will give to the tap device when cvdalloc runs.

While we are here, we introduce new address ranges to assign to cvdalloc's interfaces, so they can't conflict with addresses assigned to interfaces created by cuttlefish-host-resources. Additionally, alloc_utils now maps instance number to the relevant address in the same way that cuttlefish-host-resources does.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvdalloc/interface.h
+++ b/base/cvd/cuttlefish/host/commands/cvdalloc/interface.h
@@ -20,7 +20,13 @@
 namespace cuttlefish {
 
 constexpr std::string_view kCvdallocInterfacePrefix = "cvd-pi";
+constexpr char kCvdallocMobileIp[] = "192.168.144";
+constexpr char kCvdallocWirelessIp[] = "192.168.160";
+constexpr char kCvdallocthernetIp[] = "192.168.176";
 
 std::string CvdallocInterfaceName(const std::string &name, int num);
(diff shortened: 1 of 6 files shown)


Databean (approved):

base/cvd/cuttlefish/host/commands/cvdalloc/interface.h
```diff
@@ -20,7 +20,14 @@
 namespace cuttlefish {
 
 constexpr std::string_view kCvdallocInterfacePrefix = "cvd-pi";
+constexpr char kCvdallocMobileIp[] = "192.168.144";
```
jemoreira: nit: This is not an IP address, it's a prefix of one. Consider renaming as `kCvdallocMobileIpPrefix`.

3405691582: Done.

PR #1617, opened 2025-09-08
Unstable/stable channel can build and deploy docker image

0405ysj:
Management around `latest` tag should be considered later.


0405ysj:
Bypassing presubmit since this PR only affects GH workflow of postsubmit.

PR #1622, opened 2025-09-09
[HO] Log kernel and cvd versions.

ser-io:


frontend/src/host_orchestrator/main.go
```diff
@@ -53,6 +55,34 @@ func startHttpServer(addr string, port int) error {
...
+	}{
+		{
+			name: "kernel",
+			cmd:  "uname",
+			args: []string{"-r"},
+		},
+		{
+			name: "cvd",
+			cmd:  "cvd",
+			args: []string{"version"},
```
jemoreira, line 73: Newer cvds print the version on every command invocation as long as stdio is not an interactive terminal.

jemoreira (approved):

frontend/src/host_orchestrator/main.go
ser-io: Good to know, I wasn't aware of it. Let's have it printed at the beginning of the logs for discover-ability purposes as well.

PR #1626, opened 2025-09-10
Use fetcher_config.json from the system image dir

jemoreira:
cvd fetch saves the fetcher_config.json file in the same directory where it downloads the images. assemble_cvd was searching for this file only under $HOME and current directory, which are the default values for the system_image_dir flag, but not guaranteed to have fetcher_config.json in all cases.

Bug: 440789112


base/cvd/cuttlefish/host/libs/config/fetcher_config.cpp
```diff
@@ -274,4 +276,15 @@ Result<void> FetcherConfig::AddFilesToConfig(
...
 
+void FetcherConfigs::Append(FetcherConfig&& config) {
+  fetcher_configs_.emplace_back(std::move(config));
+}
+
+const FetcherConfig& FetcherConfigs::ForInstance(size_t instance_index) const {
+  if (instance_index < fetcher_configs_.size()) {
+    return fetcher_configs_[instance_index];
+  }
+  return fetcher_configs_[0];
```
Databean, line 297: Can this validate that `fetcher_configs_` has at least one member?

My preference here would be that rather than starting with an empty `std::vector` in a no-argument constructor, it's only created with a function `static Result<FetcherConfigs> Create(std::vector<FetcherConfig>)` that validates the vector is non-empty, and there's no `Append` method or public no-argument constructor.

base/cvd/cuttlefish/host/libs/config/fetcher_config.h
```diff
@@ -97,4 +98,22 @@ class FetcherConfig {
                                 bool override_entry = false);
 };
 
+class FetcherConfigs{
```
Databean: nit: `git-clang-format HEAD~ --style=file --cached`

base/cvd/cuttlefish/host/libs/config/fetcher_config.h
```diff
@@ -97,4 +98,22 @@ class FetcherConfig {
...
+  public:
+  FetcherConfigs() = default;
+  FetcherConfigs(FetcherConfigs&&) = default;
+  ~FetcherConfigs() = default;
```
Databean: Can this class be copyable in addition to being movable? I don't see any downside to making it copyable besides performance.

base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd.cc
```diff
@@ -65,27 +64,20 @@ namespace {
...
-      if (fetcher_config.LoadFromFile(fetcher_file)) {
-        return fetcher_config;
-      }
-      LOG(ERROR) << "Could not load fetcher config file.";
+FetcherConfigs FindFetcherConfigs(
+    const SystemImageDirFlag& system_image_dir) {
+  FetcherConfigs fetcher_configs;
+  for (size_t i = 0; i < system_image_dir.Size(); ++i) {
+    std::string fetcher_file =
+        fmt::format("{}/{}", system_image_dir.ForIndex(i), kFetcherConfigFile);
```
Databean, line 72: Not related to this change, but do you think at some point it's worth making the `SystemImageDirFlag` iterable, i.e. provide `begin()` and `end()` methods so that users can do something like

```c++
for (std::string_view device_system_image_dir : system_image_dir) {
  // ...
}
```

jemoreira: Maybe, but implementing the iterators sounds like a lot of work just to avoid a couple of index based loops.

base/cvd/cuttlefish/host/libs/config/fetcher_config.h
jemoreira: Done

base/cvd/cuttlefish/host/libs/config/fetcher_config.h
jemoreira: I don't see any downside either, but `FetcherConfig` is not copyable because it holds the underlying JsonValue in a unique_ptr. That looked intentional enough that I didn't want to change it, do you want me to?

base/cvd/cuttlefish/host/libs/config/fetcher_config.cpp
jemoreira: Done.

Instead of returning result just made it add a default one when the vector is empty.

base/cvd/cuttlefish/host/libs/config/fetcher_config.h
Databean: Ah didn't realize that, not important.

PR #1628, opened 2025-09-10
Mark v1.23.0 stable.

Databean:
Bug: b/443322658


jemoreira (approved):

Databean:
Thanks for reviewing!

2chiefk:
> Thanks for reviewing!

I'm already noticing one problem. Remember everybody? Cuttlefish is supposed to say under 16 kb and you have it currently at what 4.6 and a zip file? That's not good. Megabytes was supposed to be gigabytes you guys need to work on your compression compress that a lot better cuz no reason why not

PR #1632, opened 2025-09-11
Use vectorized flags correctly

jemoreira:
The --uuid and --serial_number flags were declared as vectors and accepting comma separated lists, but their raw values were being used without splitting by commas first.

The changes in this PR were originally submitted in android in ag/If4c841a85d9544653b567fe756c3d146836547fa and ag/I5511f15fd9bcaf149e72609bfa83799505624cbc.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
@@ -485,8 +485,10 @@ Result<CuttlefishConfig> InitializeCuttlefishConfiguration(
       CF_EXPECT(GET_FLAG_STR_VALUE(userdata_format));
   std::vector<bool> guest_enforce_security_vec = CF_EXPECT(GET_FLAG_BOOL_VALUE(
       guest_enforce_security));
-  std::vector<bool> use_random_serial_vec = CF_EXPECT(GET_FLAG_BOOL_VALUE(
-      use_random_serial));
+  std::vector<std::string> serial_number_vec =
+      CF_EXPECT(GET_FLAG_STR_VALUE(serial_number));
+  std::vector<bool> use_random_serial_vec =
+      CF_EXPECT(GET_FLAG_BOOL_VALUE(use_random_serial));
   std::vector<bool> use_allocd_vec = CF_EXPECT(GET_FLAG_BOOL_VALUE(use_allocd));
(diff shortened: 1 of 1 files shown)


base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
```diff
@@ -741,6 +744,20 @@ Result<CuttlefishConfig> InitializeCuttlefishConfiguration(
...
+  CF_EXPECTF(
+      provided_serials_cnt == 1 || provided_serials_cnt == instances_size,
+      "Must have a single serial number prefix or one serial number per "
+      "instance, have {} but expectected {}",
+      provided_serials_cnt, instances_size);
+  if (provided_serials_cnt == 1 && instances_size > 1) {
+    // Make sure the serial numbers are different when running multiple
+    // instances and using the default value for the flag
+    for (size_t i = 0; i < instance_nums.size(); ++i) {
+      serial_number_vec[i] += std::to_string(instance_nums[i]);
```
ser-io, line 758: nit: should we add a separator, like '-' or '-`, between prefix and instance number in the final serial number string.

base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
```diff
@@ -741,6 +744,20 @@ Result<CuttlefishConfig> InitializeCuttlefishConfiguration(
   auto num_to_webrtc_device_id_flag_map =
       CF_EXPECT(CreateNumToWebrtcDeviceIdMap(tmp_config_obj, instance_nums,
                                              FLAGS_webrtc_device_id));
+  size_t provided_serials_cnt =
```
ser-io, line 747: nit: wondering whether `serials_count` name would be enough.

base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
jemoreira: This follows the patter that was there before, which generated strings like `CUTTLEFISHCVD01`. I don't know what characters are allowed in these values, but I know for a fact that some characters are not (commas), so I think it's safest to not introduce new ones.

base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
jemoreira: `serials_count` sounded like the number of serials in `serial_number_vec`, I wanted to make it more obvious that these are the ones provided by the caller in the flag and not the entire set available at this point.

PR #1633, opened 2025-09-11
Import changes from cl/804447160

Databean:
Bug: b/444508878


Databean:
Thanks for reviewing!

PR #1636, opened 2025-09-12
Mark release 1.24.0 as unstable

Databean:
Bug: b/444731475


Databean:
Thanks for reviewing!

PR #1640, opened 2025-09-16
Update sandboxed-api dependency and import internal changes

Databean:
- cl/806270901
- cl/764006885
- cl/767014600

Bug: b/445259151


Databean:
Thanks for reviewing!

2chiefk (commented):
Try using project hook jump without using studio. Use the official sudo don't use that. Use the official hook jump program. I can certainly see a repair at this end so I keep to reminding you

PR #1646, opened 2025-09-17
openwrt's vmm waits for cvdalloc.

3405691582:
openwrt uses a vmm to run a Wi-Fi access point, using a 802.11 hardware simulator connected to a tap device. Without cvdalloc, it assumes this device was already created by cuttlefish-host-resources. With cvdalloc, openwrt's vmm will need to wait for cvdalloc to complete in order to use the tap device it creates.

The obvious way to do this is to use WaitForAvailability on cvdalloc's launcher. However, cvdalloc's launcher is already a VmmDependencyCommand for the instance's vmm, and if both openwrt's and cvdalloc's launcher are executing at the same time, we have two contending reads on cvdalloc's socketpair, which complicates matters.

We could Wait and Post twice on the socket for notification, but that is rather ugly since that means we couple the number of socket Wait'ers to the cvdalloc binary unnecessarily. Instead, we guard WaitForAvailability by a mutex, so that there is always only one Wait'er at a time, and make WaitForAvailability idempotent, so that the second Wait need not actually attempt to read on the underlying socket.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
+++ b/base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
@@ -21,6 +21,7 @@
 #include <sys/stat.h>
 
 #include <chrono>
+#include <mutex>
 #include <string>
 #include <string_view>
 #include <unordered_set>
@@ -46,8 +47,14 @@ namespace cuttlefish {
 constexpr std::chrono::seconds kCvdAllocateTimeout = std::chrono::seconds(30);
 constexpr std::chrono::seconds kCvdTeardownTimeout = std::chrono::seconds(2);
 
+enum class CvdallocStatus {
+  kUnknown = 0,
+  kAvailable,
+  kFailed
+};
+
 Cvdalloc::Cvdalloc(const CuttlefishConfig::InstanceSpecific& instance)
-    : instance_(instance) {}
+    : instance_(instance), status_(CvdallocStatus::kUnknown) {}
 
 Result<std::vector<MonitorCommand>> Cvdalloc::Commands() {
   std::string path = CvdallocBinary();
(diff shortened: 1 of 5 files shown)


base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
jemoreira, line 1: The tests are failing because it's missing an `#include <mutex>`

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
```diff
@@ -67,10 +67,15 @@ bool Cvdalloc::Enabled() const { return instance_.use_cvdalloc(); }
...
 
 Result<void> Cvdalloc::WaitForAvailability() {
-  LOG(INFO) << "cvdalloc (run_cvd): waiting to finish allocation.";
-  CF_EXPECT(cvdalloc::Wait(socket_, kCvdAllocateTimeout),
-            "cvdalloc (run_cvd): Wait failed");
-  LOG(INFO) << "cvdalloc (run_cvd): allocation is done.";
+  std::lock_guard<std::mutex> lock(availability_mutex_);
+  if (!available_) {
+    LOG(INFO) << "cvdalloc (run_cvd): waiting to finish allocation.";
+    CF_EXPECT(cvdalloc::Wait(socket_, kCvdAllocateTimeout),
```
jemoreira, line 82: If two threads call this at roughly the same time the second one could wait for twice the `kCvdAllocateTimeout`: it will wait for the first thread to release the lock and then wait for the socket. If this happens one vm will fail in 30s while the other one will fail in 60s.

That can be avoided by making `available_` an enum with 3 possible values: `Pending`, `Done` or `Failed`. The first thread to wait for the socket changes it from `Pending` to the correct value and every subsequent call just reads that value.

3405691582: Good catch. I've tweaked the logic accordingly, and adjusted naming to read a little better.

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.h
```diff
@@ -29,6 +29,12 @@
 
 namespace cuttlefish {
 
+enum class CvdallocStatus {
```
jemoreira: nit: This isn't needed outside here, make it a private subclass of `Cvdalloc`

base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp
```diff
@@ -67,10 +68,17 @@ bool Cvdalloc::Enabled() const { return instance_.use_cvdalloc(); }
...
+  CF_EXPECT(status_ != CvdallocStatus::kFailed);
+  if (status_ == CvdallocStatus::kUnknown) {
+    LOG(INFO) << "cvdalloc (run_cvd): waiting to finish allocation.";
+    status_ = CvdallocStatus::kFailed;
+    CF_EXPECT(cvdalloc::Wait(socket_, kCvdAllocateTimeout),
+              "cvdalloc (run_cvd): Wait failed");
+    LOG(INFO) << "cvdalloc (run_cvd): allocation is done.";
+    status_ = CvdallocStatus::kAvailable;
+  }
+  LOG(INFO) << "cvdalloc (run_cvd): available.";
```
jemoreira: nit: This line will appear twice in the logs, maybe don't log here and do it from the caller instead so it's clear who's doing the waiting?

PR #1647, opened 2025-09-17
Small `cvd/cli/commands/start.cpp` refactor

Databean:
Inline some methods and put most of the file into an anonymous namespace.


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #1648, opened 2025-09-17
Add disabled code for minimal metrics reporting in `cvd start`

Databean:
This is disabled by a compile-time constant. It uses the same `CuttlefishLogEvent` type as the `host/commands/metrics` code.

The metrics messages sent are currently empty.

Bug: b/445773158


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #1649, opened 2025-09-18
Use the HttpClient wrapper over curl in host/commands/metrics

Databean:
Bug: b/445779396
Test: `cvd create --host_substitutions=bin/metrics` and check for `Metrics posted to ClearCut`


jmacnak (approved):

Databean:
Thanks for reviewing!

PR #1651, opened 2025-09-18
Delete `cvd/metrics`.

Databean:
Most of the code under this directory is a subset of one of the following:

- Copied from host/commands/metrics
- Not being compiled
- Targeting the acloud metrics proto type rather than the Cuttlefish metrics proto type.

The only part being used is an informational notice on logging, which is not the same as the latest notice we were approved to use.

Bug: b/445786863


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #1652, opened 2025-09-18
The server loop ends if the process monitor does.

3405691582:
The server loop runs StartAndMonitorProcesses, which then forks a routine to wait on the monitored processes. StartAndMonitorProcesses returns success immediately to the parent server loop, which then goes on to immediately accept on a socket which a monitored processes will connect on.

The problem occurs if the one of the monitored processes dies, causing the process monitor itself to die. This leaves the parent blocked on the accept indefinitely, which has the end-user effect of the cvd start operation hanging indefinitely.

To alleviate this effect, use the pipe trick between the process monitor's parent and child routines, then select between one end of the pipe and the server socket. If the process monitor dies, then the pipe becomes readable, and we can get out of the server loop and terminate. Otherwise, the accept can happen normally.

### DIFF
--- a/base/cvd/cuttlefish/common/libs/utils/subprocess.cpp
+++ b/base/cvd/cuttlefish/common/libs/utils/subprocess.cpp
@@ -418,6 +418,7 @@ Subprocess Command::Start(SubprocessOptions options) const {
     if (!prerequisiteResult.ok()) {
       LOG(ERROR) << "Failed to check prerequisites: "
                  << prerequisiteResult.error().FormatForEnv();
+      return Subprocess(-1, {});
     }
   }
 
--- a/base/cvd/cuttlefish/host/commands/run_cvd/server_loop_impl.cpp
+++ b/base/cvd/cuttlefish/host/commands/run_cvd/server_loop_impl.cpp
@@ -28,6 +28,7 @@
 
 #include "cuttlefish/common/libs/fs/shared_buf.h"
 #include "cuttlefish/common/libs/fs/shared_fd.h"
+#include "cuttlefish/common/libs/fs/shared_select.h"
 #include "cuttlefish/common/libs/utils/files.h"
 #include "cuttlefish/common/libs/utils/result.h"
 #include "cuttlefish/common/libs/utils/subprocess.h"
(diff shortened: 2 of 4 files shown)


base/cvd/cuttlefish/common/libs/utils/subprocess.cpp
```diff
@@ -418,6 +418,7 @@ Subprocess Command::Start(SubprocessOptions options) const {
     if (!prerequisiteResult.ok()) {
       LOG(ERROR) << "Failed to check prerequisites: "
                  << prerequisiteResult.error().FormatForEnv();
+      return Subprocess(-1, {});
```
jemoreira, line 421: nit: Should this function return `Result<Subprocess>` instead? I know this PR doesn't introduce this, this function actually predates the Result type, but now could be a good time to fix it.

base/cvd/cuttlefish/host/commands/run_cvd/server_loop_impl.cpp
```diff
@@ -103,6 +104,17 @@ Result<void> ServerLoopImpl::Run() {
 
   while (true) {
     // TODO: use select to handle simultaneous connections.
```
jemoreira, line 106: Is this comment still relevant?

jemoreira (approved):

base/cvd/cuttlefish/host/commands/run_cvd/server_loop_impl.cpp
3405691582: Yes, I think this covers using select to not have to process clients synchronously. The use of select here is just to cover waiting for a client *or* the monitor dies, so more fulsome use of select is still necessary.

base/cvd/cuttlefish/common/libs/utils/subprocess.cpp
3405691582: The refactor appears to be rather broad. I should probably do this in a separate pr.

jemoreira: SG

PR #1654, opened 2025-09-19
Homogenize group directory structure

jemoreira:
Regardless of how a user creates a new group and whether they provide artifacts or fetch them, the group directory structure is the same.

There is a group base directory under `/tmp/cvd/{uid}` that contains the `home` directory and the `artifacts` directory, the latter contains `host_tools` and `0`, `1`, etc for the host and per-instance target artifacts respectively.

Some of these directories will be symbolic links instead, for example when the user overrides the HOME variable or provides their own build artifacts. `cvd remove` deletes the directories under `/tmp/cvd` but doesn't follow links and won't remove user provided artifacts.


2chiefk:
Wasn't saying you guys weren't doing good. I'm just saying you know they're
trying to get this and keep this under 16 kb it's for a fast compression so
it's basically going to be taking the whole operating system and making it
so it's compatible for our one page on a browser look they kind of already
do but they want to do it for a phone. Just so figured I'd give you guys
all the heads up of what that's all about. God bless Peace out

base/cvd/cuttlefish/host/commands/cvd/utils/common.cpp
```diff
@@ -168,5 +170,38 @@ std::string AssemblyDirFromHome(const std::string& group_home_dir) {
   return group_home_dir + "/cuttlefish/assembly";
 }
 
+Result<void> LinkOrMakeDir(const std::string& path,
```
Databean: nit: should this be `static` or in an anonymous namespace? It's not mentioned in the header.

base/cvd/cuttlefish/host/commands/cvd/utils/common.cpp
```diff
@@ -168,5 +170,38 @@ std::string AssemblyDirFromHome(const std::string& group_home_dir) {
...
+
+Result<GroupDirectories> GenerateGroupDirectories(
+    std::optional<std::string> base, std::optional<std::string> home,
+    std::optional<std::string> host_tools,
+    std::vector<std::optional<std::string>> targets) {
+  GroupDirectories ret;
+
+  ret.base = DefaultBaseDir();
+  CF_EXPECT(LinkOrMakeDir(ret.base, std::move(base)));
+  ret.home = ret.base + "/home";
```
Databean: Rather than member variables, can `home`, `artifacts`, etc. be functions that produce values derived from `home`? Are there any setters besides this function?

base/cvd/cuttlefish/host/commands/cvd/utils/common.cpp
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/utils/common.cpp
jemoreira: Done

PR #1655, opened 2025-09-19
Parity between public and internal metrics protos

Databean:
This will require another round of import script updates, but the import script part will be a more standard substitution of header files and build targets rather than the patches we're carrying now.

Bug: b/446052689


jmacnak (approved):

jmacnak (approved):

Databean:
Thanks for reviewing!

PR #1661, opened 2025-09-19
Move metrics protos outside of the `cuttlefish` directory

Databean:
This will simplify the internal import script.

Bug: b/446052689


Databean:
Thanks for reviewing!

PR #1662, opened 2025-09-20
Refactor `metrics/events.cc` and reuse legacy `launch_cvd` metrics in `cvd`

Databean:
Bug: b/446266666


Databean:
Thanks for reviewing!

PR #1669, opened 2025-09-23
Add formatted output for the `LauncherResponse` type

Databean:
This changes an error message from
```
Expected "(int)response" == "(int)LauncherResponse::kSuccess" but was 69 vs 83.
```
to
```
Expected "response" == "LauncherResponse::kSuccess" but was LauncherResponse::kError vs LauncherResponse::kSuccess.
```

Bug: b/446765997

### DIFF
--- a/base/cvd/cuttlefish/host/libs/command_util/runner/defs.cc
+++ b/base/cvd/cuttlefish/host/libs/command_util/runner/defs.cc
@@ -1,0 +1,36 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


base/cvd/cuttlefish/host/libs/command_util/runner/defs.cc
```diff
@@ -0,0 +1,35 @@
...
+std::ostream& operator<<(std::ostream& out, LauncherResponse response) {
+  switch (response) {
+    case LauncherResponse::kSuccess:
+      return out << "LauncherResponse::kSuccess";
+    case LauncherResponse::kError:
+      return out << "LauncherResponse::kError";
+    case LauncherResponse::kUnknownAction:
+      return out << "LauncherResponse::kUnknownAction";
+    default:
+      return out << "LauncherResponse::(unknown)";
```
3405691582: Perhaps output the integer value in the default case?

Databean: Done.

Databean:
Thanks for reviewing!

PR #1670, opened 2025-09-23
cvdalloc supports bridged Wi-Fi.

3405691582:
Bridged Wi-Fi necessitates ensuring cvdalloc can handle setting up bridges and dnsmasq(8) properly. The existing alloc_utils does this, but assumes it gets run as root. If cvdalloc were setuid root, then this isn't an issue. However, we use capabilities and thus need more permissions: one to set up the raw DHCP socket, another to allow binding to low ports, again for DHCP.

alloc_utils also defaults to putting the logs and lease files in /var/run which would otherwise require root, but we can just put those in CvdDir instead.

The rest of the support necessary for bridged Wi-Fi is much like non-bridged Wi-Fi.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
@@ -41,23 +41,26 @@ void Usage() {
   LOG(ERROR) << "Should only be invoked from run_cvd.";
 }
 
-Result<void> Allocate(int id, const std::string &bridge_name) {
+Result<void> Allocate(int id, const std::string &ethernet_bridge_name,
+                      const std::string &wireless_bridge_name) {
   LOG(INFO) << "cvdalloc: allocating network resources";
 
-  CreateBridge(bridge_name);
   CF_EXPECT(CreateMobileIface(CvdallocInterfaceName("mtap", id), id,
                               kCvdallocMobileIpPrefix));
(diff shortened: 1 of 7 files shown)


base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -41,23 +41,26 @@ void Usage() {
...
+Result<void> Allocate(int id, const std::string &ethernet_bridge_name,
+                      const std::string &wireless_bridge_name) {
   LOG(INFO) << "cvdalloc: allocating network resources";
 
-  CreateBridge(bridge_name);
   CF_EXPECT(CreateMobileIface(CvdallocInterfaceName("mtap", id), id,
                               kCvdallocMobileIpPrefix));
-  CF_EXPECT(CreateMobileIface(CvdallocInterfaceName("wtap", id), id,
-                              kCvdallocWirelessIpPrefix));
+  CreateEthernetBridgeIface(wireless_bridge_name, kCvdallocWirelessIpPrefix);
```
Databean, line 50: Can the `CreateEthernetBridgeIface` calls use `CF_EXPECT` as well?

3405691582: The same issue applies where we can't do that because the bridge may already exist and we'd fail unnecessarily. 

I better get to finally doing this.

3405691582: I'll have a pr out today to fix this.

PR #1676, opened 2025-09-24
Pass source image in create base image tool.

ser-io:


tools/baseimage/cmd/create_gce_x86_64_image/main.go
```diff
@@ -24,22 +24,23 @@ import (
 	"github.com/google/android-cuttlefish/tools/baseimage/pkg/gce/scripts"
 )
 
-// Cuttlefish base images are based on debian images.
-const (
-	debianSourceImageProject = "debian-cloud"
```
jmacnak, line 29: Can we set these as the flag defaults or can you also update `tools/baseimage/README.md`  so that there is some baseline guidance on what images to use?

ser-io: Good call, updated tools/baseimage/README.md. 

For the `create_gce_x86_64_image` tool I want the user to be aware and acknowledge the kernel they'll be using by passing the new `source-image-project` and `source-image` flags. We default  to `debian-cloud` and `debian-12-bookworm-v20250610` in the `create_gce_x86_64_fixed_kernel` tool.

PR #1677, opened 2025-09-24
Make bridge creation idempotent.

3405691582:
This allows us to properly and accurately CF_EXPECT on bridge creation as prior to this change, attempting to create a bridge after the first instance would fail, though it would not actually be a problem since the first instance already set up the bridge.


Databean (approved):
Thanks!

PR #1678, opened 2025-09-24
Report the correct bridge name when use_cvdalloc.

3405691582:
This enables fastboot to work correctly.

This should be a symbolic reference, but we need to make some refactors in order to do this properly, so let's get this working first.

While we are here, fix the bridge name for Wi-Fi, though this may be unused.


base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
```diff
@@ -994,6 +994,10 @@ Result<CuttlefishConfig> InitializeCuttlefishConfiguration(
     instance.set_mobile_bridge_name(StrForInstance("cvd-mbr-", num));
     instance.set_wifi_bridge_name("cvd-wbr");
     instance.set_ethernet_bridge_name("cvd-ebr");
+    if (const_instance.use_cvdalloc()) {
```
jemoreira, line 995: nit: Please put the other two possible values in the `else` clause.

3405691582: Done.

PR #1680, opened 2025-09-25
Add and use flag wrapper template class

cjreynol:
This PR is a new branch based on feedback from #1604.

The goal of the new `FlagBase` is to speed the conversion of flags into a form where the parsing can be re-used in different locations within the codebases.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/flags/from_gflags.cc
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/flags/from_gflags.cc
@@ -1,0 +1,81 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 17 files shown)


base/cvd/cuttlefish/host/commands/assemble_cvd/flags/from_gflags.cc
```diff
@@ -0,0 +1,106 @@
...
+template <typename T>
+Result<std::vector<T>> FromGlobalGflagsWithBackup(
+    const gflags::CommandLineFlagInfo& flag_info, const std::string& flag_name,
+    std::function<Result<T>(const std::string& value, const std::string& name)>
+        parse_func,
+    const std::vector<T>& backup_values) {
+  std::vector<T> values =
+      CF_EXPECT(FromGlobalGflags(flag_info, flag_name, parse_func));
+  if (backup_values.size() > values.size()) {
+    // backup_values is used to pad out the result
```
Databean: `IntFromGlobalGflagsWithBackup` only uses `guest_config_values` to "pad out the result", by adding more members at the end, but the implementation in `BlankDataImageMbFlag` additionally supported `unset` values mixed throughout which fall back to the default.

Another option here is to leave `BlankDataImageMbFlag` as-is until there's another flag that needs similar handling. That avoids creating a "common" utility that really only exists for the benefit of one caller.

Databean (approved):

base/cvd/cuttlefish/host/commands/assemble_cvd/flags/from_gflags.cc
cjreynol: I moved the implementation into `BlankDataImageMbFlag::FromGlobalGflags`.  I went the way I did for the sake of discoverability, but I am just as happy keeping `from_gflags.h/cc` slimmer for now.

PR #1681, opened 2025-09-25
alloc_utils and cvdalloc use string_view.

3405691582:
Some of the idioms are starting to clash, which suggests a refactor. All uses of `const std::string &` in alloc_utils and cvdalloc are now forklifted wholesale to `std::string_view`, which is the more recent and preferred idiomatic mechanism for passing bits of strings like this.

### DIFF
--- a/base/cvd/allocd/alloc_utils.cpp
+++ b/base/cvd/allocd/alloc_utils.cpp
@@ -17,8 +17,10 @@
 
 #include <cstdint>
 #include <fstream>
+#include <string_view>
 
 #include "android-base/logging.h"
+#include "absl/strings/str_format.h"
 
 #include "cuttlefish/host/commands/cvd/utils/common.h"
 
@@ -53,7 +55,7 @@ int RunExternalCommand(const std::string& command) {
   return ret;
 }
 
-bool AddTapIface(const std::string& name) {
+bool AddTapIface(std::string_view name) {
   std::stringstream ss;
   ss << "ip tuntap add dev " << name << " mode tap group cvdnetwork vnet_hdr";
   auto add_command = ss.str();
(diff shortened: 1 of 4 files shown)


base/cvd/allocd/alloc_utils.cpp
```diff
@@ -17,17 +17,19 @@
...
 #include "cuttlefish/host/commands/cvd/utils/common.h"
 
 namespace cuttlefish {
 
-int RunExternalCommand(const std::string& command) {
+int RunExternalCommand(std::string_view command) {
   FILE* fp;
   LOG(INFO) << "Running external command: " << command;
-  fp = popen(command.c_str(), "r");
+  fp = popen(command.data(), "r");
```
Databean: This is not safe. `std::string_view` does not guarantee null termination. If the argument represents a fragment of a larger `std::string`, `popen` will consume the rest of the string up until the null terminator.

If this function needs to accept a `std::string_view`, it has to create a `std::string` copy so that it can guarantee there is a null terminator in the right place. As far as I can tell most of the callers create a `std::stringstream` and produce a string with `.str()`, so it seems more appropriate to still accept a `const std::string&`.

Databean (approved):

base/cvd/allocd/alloc_utils.cpp
3405691582: True, but in this case, we're only feeding null-terminated strings into this. But you're right -- we should be safer about this. Would you prefer the copy and changing the callers, or the signature change?

Databean: In this case I prefer the signature change to `string` since it lines up with how the callers are using the function, but it's not a strong opinion. I see the value as well in standardizing around `string_view` in APIs as well and only converting to `string` at the moment something needs to be null-terminated.

3405691582: I was going to add a copy here -- but then realized it'd be two copies, which feels a bit silly, so I put it back to `const std::string &` for just this function.

PR #1685, opened 2025-09-26
Initial metrics gather

cjreynol:
First version of metrics gathering.  Does not transmit, merely writes the collected information out to a file


base/cvd/cuttlefish/host/libs/metrics/host_metrics.cc
```diff
@@ -0,0 +1,65 @@
...
+  }
+  return HostOs::Unknown;
+}
+
+}  // namespace
+
+Result<HostMetrics> GetHostMetrics() {
+  utsname out;
+  CF_EXPECTF(uname(&out) == 0, "uname() call failed with error: {}",
+             strerror(errno));
```
jemoreira: According to the manual `uname` can only fail if the `out` buffer is not valid, which this code makes sure doesn't happen. A `CHECK` makes sense here because of that, so you should probably reuse (and maybe expand) the code from `architecture.h`.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
```diff
@@ -232,6 +233,23 @@ Result<void> EnsureSymlink(const std::string& target, const std::string link) {
   return {};
 }
 
+void GatherMetrics(const LocalInstanceGroup& instance_group) {
+  const std::string metrics_directory =
+      GenerateMetricsDirectoryFilepath(instance_group);
+  Result<void> metrics_setup_result = SetUpMetrics(metrics_directory);
+  if (!metrics_setup_result.ok()) {
+    LOG(INFO) << fmt::format("Failed to initialize metrics.  Error: {}",
```
jemoreira: LOG(ERROR) for error messages.

base/cvd/cuttlefish/host/libs/metrics/metrics_orchestration.h
```diff
@@ -0,0 +1,33 @@
...
+#pragma once
+
+#include <string>
+
+#include "cuttlefish/common/libs/utils/result.h"
+#include "cuttlefish/host/commands/cvd/instances/instance_group_record.h"
+
+namespace cuttlefish {
+
+std::string GenerateMetricsDirectoryFilepath(
```
jemoreira: Please add comments explaining what these functions do.

base/cvd/cuttlefish/host/libs/metrics/metrics_orchestration.cc
```diff
@@ -0,0 +1,59 @@
...
+    "The existence of records in this directory does"
+    " not mean metrics are being transmitted, the data is always gathered and "
+    "written out for debugging purposes.  To enable metrics transmission "
+    "<TODO: chadreynolds - metrics transmission not connected, add triggering "
+    "step"
+    " when it does>";
+
+}  // namespace
+
+std::string GenerateMetricsDirectoryFilepath(
```
jemoreira: nit: rename as `GetMetricsDirectoryFilePath`. "generate" sounds like there is some random element in the path.

base/cvd/cuttlefish/host/libs/metrics/metrics_orchestration.h
```diff
@@ -0,0 +1,33 @@
...
+    const LocalInstanceGroup& instance_group);
+
+Result<void> SetUpMetrics(const std::string& metrics_directory);
+
+Result<void> GatherVmInstantiationMetrics(const std::string& metrics_directory);
```
jemoreira: nit: Instead of having a function return a directory that the callers never use and simply pass to other functions of this header, consider an object oriented approach like this:
```
class MyBetterNamedMetricsClass {
 public:
  static Result<MyBetterNamedMetricsClass> Create(); 

  Result<void> GatherVmInstantiationMetrics();

 private:
  MyBetterNamedMetrics(std::string metrics_directory) : metrics_directory_(std::move(metrics_directory)) {}

  std::string metrics_directory_;
};
```

base/cvd/cuttlefish/host/libs/metrics/metrics_writer.cc
```diff
@@ -0,0 +1,51 @@
...
+
+#include <fmt/format.h>
+
+#include "cuttlefish/common/libs/utils/files.h"
+#include "cuttlefish/common/libs/utils/result.h"
+
+namespace cuttlefish {
+namespace {
+
+std::string GenerateRandomString() {
```
jemoreira: This will be the third different implementation of random string we have in this code base (there is one in host/libs/config/config_utils.cpp and another in host/commands/process_sandboxer/sandbox_manager.cpp). Could you add a common implementation under common/libs/utils/random.h that's used from everywhere? I'm guessing the new function would need to accept a length and an alphabet string.

base/cvd/cuttlefish/host/libs/metrics/metrics_writer.cc
```diff
@@ -0,0 +1,51 @@
...
+  std::string str("0123456789");
+  std::random_device rd;
+  std::mt19937 generator(rd());
+  std::shuffle(str.begin(), str.end(), generator);
```
jemoreira: This isn't what you'd normally expect from a "random string" as it only generates permutations of the input string.

The answer to the stackoverflow question you linked has comments pointin this out, IMHO that answer is wrong.

base/cvd/cuttlefish/host/libs/metrics/host_metrics.cc
cjreynol: Instead of the new `host/libs/metrics/host_metrics.h/cc`, I expanded the scope of `common/libs/utils/architecture.h/cpp` and renamed the files `host_info.h/cpp`.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
cjreynol: Done, though this logic now lives in `metrics_orchestration.cc`.

base/cvd/cuttlefish/host/libs/metrics/metrics_orchestration.h
cjreynol: There is now only a single export from this header (`GatherVmInstantiationMetrics`), which I think is self-explanatory.  Let me know if you would prefer a comment.

base/cvd/cuttlefish/host/libs/metrics/metrics_orchestration.cc
cjreynol: Done.

base/cvd/cuttlefish/host/libs/metrics/metrics_orchestration.h
cjreynol: Considered and agreed with the idea of not passing information within the header scope, but decided instead to shift the boundary of the metrics code to include the logic previously in `GatherMetrics` from `create.cpp`.  I think the spirit of the request is still satisfied.

The current version will make re-use for other events across the follow-up PRs easier (for example, pulling in the metrics transmission code currently in `start.cpp`).

base/cvd/cuttlefish/host/libs/metrics/metrics_writer.cc
cjreynol: Somewhat done.  I added a commit to factor out the `config_utils.cpp` implementation which does not permute the string but pulls random characters from the given alphabet.  Then I used that for `metrics_writer.cc`.

I am hesitant to replace the `SandboxManager` `absl::BitGen`-based implementation of random generation with the new `random.h` pseudorandom logic due to potentially more requirements beyond "a random-ish string" needed in `config_utils` and `metrics_writer`.

base/cvd/cuttlefish/host/libs/metrics/metrics_writer.cc
cjreynol: See other comment, but I adopted the implementation from `config_utils.cpp`.

base/cvd/cuttlefish/host/libs/config/config_utils.cpp
```diff
@@ -100,13 +100,9 @@ std::string ForCurrentInstance(const char* prefix) {
...
 std::string RandomSerialNumber(const std::string& prefix) {
-  const char hex_characters[] = "0123456789ABCDEF";
-  std::srand(time(0));
-  char str[10];
-  for(int i=0; i<10; i++){
-    str[i] = hex_characters[rand() % strlen(hex_characters)];
-  }
-  return prefix + str;
+  const std::string hex_characters = "0123456789ABCDEF";
+  std::string result = GenerateRandomString(hex_characters, 10);
```
jemoreira: nit: s/result/random_suffix/ or avoid the extra variable altogether `prefix + GenerateRandomString(...)` is clear enough.

cjreynol: Done.

PR #1689, opened 2025-09-26
Add screen recordings endpoint to host orchestrator

jemoreira:
POST /cvds/{group}/{instance}/:start_screen_recording - start recording POST /cvds/{group}/{instance}/:stop_screen_recording - stop recording GET /cvds/{group}/{instance}/screen_recordings - list recordings GET /cvds/{group}/{instance}/screen_recordings/{recording} - download

It's not possible to implement a truly RESTful api at the moment because the cvd command to start recording doesn't return the recording file name(s) and it's not trivial to implement because of the current state of the migration of the run_cvd binary. When it becomes possible we could extend :start_screen_recording's response to return the recording file name as recording id, make it an alias of a POST to .../screen_recordings and make :stop_screen_recording an alias of a POST to .../screen_recordings/{id}/:stop.

Bug: 447155201


frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -88,6 +88,10 @@ func (c *Controller) AddRoutes(router *mux.Router) {
 		httpHandler(newExecCVDInstanceCommandHandler(c.Config, c.OperationManager, &powerwashCvdCommand{}))).Methods("POST")
 	router.Handle("/cvds/{group}/{name}/:powerbtn",
 		httpHandler(newExecCVDInstanceCommandHandler(c.Config, c.OperationManager, &powerbtnCvdCommand{}))).Methods("POST")
+	router.Handle("/cvds/{group}/{name}/:start_screen_recording",
+		httpHandler(newExecCVDInstanceCommandHandler(c.Config, c.OperationManager, &startScreenRecordingCvdCommand{}))).Methods("POST")
```
Databean, line 92: The commit message says

> It's not possible to implement a truly RESTful api at the moment because the cvd command to start recording doesn't return the recording file name(s)

If it's not conformant anyway, can the API least start with the HTTP paths we want, and the responses can be updated / made more complete later?

jemoreira: I think that would be misleading: I would expect `POST /resource` to return something that identifies the created resource. Having the `:action` URL type instead indicates that an action was done on the device and it may not return much more than a success or failure flag.

frontend/src/host_orchestrator/api/v1/messages.go
```diff
@@ -208,3 +208,5 @@ type DisplayScreenshotResponse struct {
 	ScreenshotBytesBase64 string `json:"screenshot_bytes"`
 	ScreenshotMimeType    string `json:"screenshot_mime_type"`
 }
+
+type ScreenRecordingListResponse []string
```
ser-io: Use struct to represent top-level request and response objects: https://google.aip.dev/132#response-message

ser-io: use name `ListScreenRecordingResponse`. I noticed the naming pattern was not followed with Display* object names, I'll send a PR fixing that, however let's keep the original pattern with the new screen recording methods.

frontend/src/host_orchestrator/orchestrator/BUILD.bazel
```diff
@@ -9,9 +9,11 @@ go_library(
         "createsnapshotaction.go",
         "execcvdcommandaction.go",
         "fetchartifactsaction.go",
+        "getscreenrecordingaction.go",
         "instancemanager.go",
         "listcvdsaction.go",
         "operation.go",
+        "screenrecordinglistaction.go",
```
ser-io: `listscreenrecordingaction`

ser-io (changes requested):
Let's add an e2e test for this.

frontend/src/host_orchestrator/api/v1/messages.go
jemoreira: Done

jemoreira: Done

frontend/src/host_orchestrator/orchestrator/BUILD.bazel
jemoreira: Done

jemoreira:
> Let's add an e2e test for this.

Done

e2etests/orchestration/screen_recording_test/main_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	recordings, err := srv.ListScreenRecordings(cvd.Group, cvd.Name)
+	if err != nil {
+		t.Fatalf("Failed to list screen recordings: %v", err)
+	}
+	if len(recordings) != 1 {
+		t.Fatalf("Expected list of recordings with single element, got: %v", recordings)
+	}
+
+	if err := srv.DownloadScreenRecording(cvd.Group, cvd.Name, recordings[0], io.Discard); err != nil {
```
ser-io: nit: wondering whether there could be an assertion based on part of the downloaded content making sure it represents a recording content. For example, the server could return a 200 with some json object which will pass this assertion.

jemoreira: The server is returning a file as part of this request, making it return JSON at the same time is not trivial. The simplest option is to encode the file contents as part of the JSON response, but then it requires holding the entire video file in memory to parse the JSON, which could be prohibitively expensive depending on how long the recording is.
There are no constraints on what format the downloaded file is. The current implementation uses webm IIRC, but that could change in the future and break this test.
Checking that the file is actually a video is of little help too, since the screen recording feature could be broken and still return a video file with the wrong contents, but there would be no way to determine that.

I think the cost+risk of being able to detect when the device produces a file that is not a video outweighs its benefits.

ser-io: Got it. Agree, json encoding is not the way to go. My concern is the HO layer mishandling the endpoints and then when the user opens the file, it finds something like the following:

```
HTTP/1.1 200 OK
Server: nginx/1.22.1
Date: Fri, 03 Oct 2025 12:30:16 GMT
Content-Type: application/json
Content-Length: 12
Connection: keep-alive

{"cvds":[]}
```

I understand digging into the file content outweighs its benefits. Can we at least verify the file is not empty?

jemoreira: Done

PR #1692, opened 2025-09-30
Put locks in CvdDir when invoked as cvd.

3405691582:
When invoked as cvd, lock files should be placed in a subdirectory of CvdDir. This means that the InstanceLockFileManager should have the lock directory as a parameter; when invoked as acloud, we use the old subdirectory as the parameter instead.

(This contains the locking changes we had discussed briefly, but I feel like I might be forgetting something. You might also have a better idea if this leads to acloud incompatability or brokenness; please point me in the right direction if so.)

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.cpp
@@ -63,11 +63,13 @@ bool InstanceLockFile::operator<(const InstanceLockFile& other) const {
   return lock_file_ < other.lock_file_;
 }
 
-InstanceLockFileManager::InstanceLockFileManager() {}
+InstanceLockFileManager::InstanceLockFileManager(
+    std::string_view instance_locks_path)
+    : instance_locks_path_(instance_locks_path) {};
 
 Result<std::string> InstanceLockFileManager::LockFilePath(int instance_num) {
   std::stringstream path;
-  path << InstanceLocksPath();
+  path << instance_locks_path_;
   CF_EXPECT(EnsureDirectoryExists(path.str()));
   path << "local-instance-" << instance_num << ".lock";
   return path.str();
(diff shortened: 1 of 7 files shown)


base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.cpp
```diff
@@ -63,11 +63,13 @@ bool InstanceLockFile::operator<(const InstanceLockFile& other) const {
   return lock_file_ < other.lock_file_;
 }
 
-InstanceLockFileManager::InstanceLockFileManager() {}
+InstanceLockFileManager::InstanceLockFileManager(
+    std::string_view instance_locks_path)
+    : instance_locks_path_(instance_locks_path) {};
```
jemoreira, line 68: nit: Creating `std::string` from `std::string_view`, requires a copy. If the constructor accepts an `std::string` parameter instead and `std::move` is used where necessary it's possible to avoid all extra copies.

jemoreira (approved):

jemoreira:
> (This contains the locking changes we had discussed briefly, but I feel like I might be forgetting something. You might also have a better idea if this leads to acloud incompatability or brokenness; please point me in the right direction if so.)

This could cause problems if `acloud --local-instance` and `cvd` are used in the same machine, but we don't need to worry about that case.

base/cvd/cuttlefish/host/commands/cvd/instances/lock/instance_lock.cpp
3405691582: Ah, I forgot I made the ivar a string. I'll just make it a string_view too; I don't think it really needs to own the lock path string.

jemoreira: That sounds even scarier. Can't the `InstanceLockFileManager` outlive the string from which the string_view was created?

3405691582: Ugh -- yes, I forgot this wasn't initialized from a constant. I'll fix this in the next change.

PR #1693, opened 2025-09-30
Generate and gather metrics session ID 

cjreynol:
Bug: 448398451


cjreynol:
First four commits are from #1685 and this PR should be rebased before leaving draft status.

base/cvd/cuttlefish/host/libs/metrics/session_id.cc
```diff
@@ -0,0 +1,52 @@
...
+
+namespace cuttlefish {
+namespace {
+
+constexpr char kSessionIdFileName[] = "metrics_session_id.txt";
+
+std::string GenerateUuid() {
+  uuid_t uuid;
+  uuid_generate_random(uuid);
+  std::string uuid_str = "xxxxxxxx-xxxx-Mxxx-Nxxx-xxxxxxxxxxxx";
```
jemoreira: According to `UUID_UNPARSE(3)` the only requirement on this string is that it is 36 chars long plus the `\0` at the end. The pattern you've included here guarantees that, but it's error prone as someone could inadvertently delete part of it and not notice it. Consider building the string with something like `std::string(36, 'x')`, even better if you make that 36 a constant.

jemoreira (approved):

base/cvd/cuttlefish/host/libs/metrics/session_id.cc
cjreynol: Great idea, made the change and referenced the man page in the comment to provide context in the future.

PR #1694, opened 2025-09-30
Gather initial metrics for remaining events

cjreynol:
`cvd start`, `VIRTUAL_DEVICE_BOOT_COMPLETE`, `cvd stop`

Bug: 448398451

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
@@ -73,6 +73,7 @@
 #include "cuttlefish/host/libs/config/config_constants.h"
 #include "cuttlefish/host/libs/config/cuttlefish_config.h"
 #include "cuttlefish/host/libs/metrics/metrics_defs.h"
+#include "cuttlefish/host/libs/metrics/metrics_orchestration.h"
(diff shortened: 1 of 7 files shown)


cjreynol:
The initial commits are from #1685 and #1693 and this PR will need to be rebased after those merge.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
```diff
@@ -668,6 +670,7 @@ Result<void> CvdStartCommandHandler::LaunchDevice(
     CF_EXPECT(CvdResetGroup(group));
   }
   CF_EXPECT(CheckProcessExitedNormally(infop));
+  GatherVmBootCompleteMetrics(group);
```
jemoreira, line 673: This won't run if any of the above CF_EXPECT calls produces an error. Consider moving it to the function calling this one or use a ScopeGuard if reporting errors is expected.

cjreynol: Intended behavior, since we have no notion of capturing the difference between a boot success or failure in the metrics protobuf.  Perhaps that would be a good thing to capture, but we would need more updates than here.

PR #1697, opened 2025-10-01
Remove acloud compat code.

3405691582:
As discussed elsewhere, this should no longer be needed.
    
Substitution file parsing will need to change to handle the fact that we aren't vending `acloud_translator` any more and existing substitution files may still refer to it. We could address that here, but instead it might be better to do this more generally in #1716 before this commit.


jemoreira:
Can the code under `base/cvd/cuttlefish/host/commands/cvd/acloud` be deleted too? And `base/cvd/cuttlefish/host/commands/cvd/cli/commands/acloud*`?

3405691582:
@jemoreira I've torn out just about everything acloud-related.

3405691582:
kokoro will fail until #1716 lands now.

base/cvd/cuttlefish/host/commands/cvd/instances/cvd_persistent_data.proto
```diff
@@ -52,5 +52,5 @@ message InstanceGroup {
 
 message PersistentData {
   repeated InstanceGroup instance_groups = 1;
-  bool acloud_translator_optout = 2;
+  reserved 2;
```
jemoreira: nit: leave a comment with the field name

3405691582: Done.

PR #1702, opened 2025-10-07
Connect new metrics gathering with metrics transmission

cjreynol:
For all event types, still conditional on the `kEnableCvdMetrics` flag, which is `false` without the import rewrite.


cjreynol:
Needs to be rebased once #1694 merges.

PR #2041, opened 2026-01-22
Migrate e2etests to bazel 9

Databean:
This is a distinct bazel repository from base/cvd with fewer dependencies, so it is easier to migrate.

The tools/bazel script is updated to not resolve symlinks when finding the `.bazelversion` path, so that the e2etests directory can still symlink `base/cvd/tools/bazel` and reuse it, while maintaining a separate `.bazelversion` file.

Bug: b/477396584


jemoreira (approved):

ser-io (approved):

rmuthiah:
Please reopen if needed

PR #2088, opened 2026-01-29
Added host orchestrator endpoint for cvd restart

SuperStrongDinosaur:
b/478814714

### DIFF
--- a/e2etests/orchestration/restart_test/main_test.go
+++ b/e2etests/orchestration/restart_test/main_test.go
@@ -1,0 +1,83 @@
+// Copyright (C) 2026 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"log"
+	"testing"
+
+	"github.com/google/android-cuttlefish/e2etests/orchestration/common"
+
+	hoclient "github.com/google/android-cuttlefish/frontend/src/libhoclient"
+)
+
+const baseURL = "http://0.0.0.0:2080"
+
+func TestRestart(t *testing.T) {
+	adbH := common.NewAdbHelper()
+	if err := adbH.StartServer(); err != nil {
+		t.Fatal(err)
+	}
+	srv := hoclient.NewHostOrchestratorClient(baseURL)
+	t.Cleanup(func() {
+		if err := common.CollectHOLogs(baseURL); err != nil {
+			log.Printf("failed to collect HO logs: %s", err)
+		}
+	})
+	imageDir, err := common.PrepareArtifact(srv, "../artifacts/images.zip")
+	if err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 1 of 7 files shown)


ser-io (changes requested):
Please add e2e test to this PR. Example: https://github.com/google/android-cuttlefish/blob/main/e2etests/orchestration/powerbtn_test

e2etests/orchestration/restart_test/main_test.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	if err := adbH.Connect(cvd.ADBSerial); err != nil {
+		t.Fatal(err)
+	}
+	if err := adbH.WaitForDevice(cvd.ADBSerial); err != nil {
+		t.Fatal(err)
+	}
+
+	// Restart the device
+	if err := srv.Restart(cvd.Group, cvd.Name); err != nil {
```
ser-io, line 65: If we were to remove this line calling `srv.Restart(cvd.Group, cvd.Name);` would the test still pass? If yes, can we add another assertion to verify the device was restarted?

frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -489,6 +491,12 @@ func (c *HostOrchestratorClientImpl) Stop(groupName, instanceName string) error
 	return c.doEmptyResponseRequest(rb)
 }
 
+func (c *HostOrchestratorClientImpl) Restart(groupName, instanceName string) error {
```
ser-io, line 494: Thanks for adding the e2e test.

Lets add the client library changes in its own commit having the PR looking like:

commit A: backend (HO) change
commit B: client change
commit C: e2e test.

ser-io (changes requested):

e2etests/orchestration/restart_test/main_test.go
ser-io: I think we can work with `adb shell uptime -s` https://man7.org/linux/man-pages/man1/uptime.1.html. We can get the first timestamp, wait X seconds, restart, get second timestamp, verify the difference is >= than X seconds.

SuperStrongDinosaur: yes, that works, added the check

frontend/src/libhoclient/host_orchestrator_client.go
SuperStrongDinosaur: splitted commits as requsted

e2etests/orchestration/restart_test/main_test.go
```diff
@@ -0,0 +1,82 @@
...
+	// Verify the device was actually restarted by checking its uptime
+	uptimeStr, err := adbH.ExecShellCommand(cvd.ADBSerial, []string{"cat", "/proc/uptime"})
+	if err != nil {
+		t.Fatal(err)
+	}
+	uptime, err := strconv.ParseFloat(strings.Fields(uptimeStr)[0], 64)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if uptime > 120 {
```
ser-io: If we didn't restart the device `uptime > 120` will still be true.

`adb shell uptime -s` gives you a timestamp for when the system was started.


```
cvd create

adb shell uptime -s  --> 2026-03-19 15:34:46

wait 1 second

cvd restart

adb shell uptime -s  --> 2026-03-19 15:34:59
```

Now we can assert that if the difference between the 1st and 2nd timestamp is more than 1 second the device was indeed restarted.

ser-io (changes requested):

ser-io (approved):
Thanks for your contribution!

jemoreira (changes requested):
Avoid merge commits, use a rebase strategy instead.

SuperStrongDinosaur:
> Avoid merge commits, use a rebase strategy instead.

removed merge commit

jemoreira (approved):

PR #2215, opened 2026-03-04
Allow omitting connection to light server

jemoreira:
The light server runs guest side and listens on a vsock port. The webRTC process connects to this server. This change makes this connection optional, configurable via the android-info.txt or cuttlefish-guest-config.txtpb file. This allows android targets that don't have a lights HAL (like some auto targets) to indicate so and avoid spurious error messages in the logs.

Bug: b/489145090

### DIFF
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/proto/guest_config.proto
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/proto/guest_config.proto
@@ -119,6 +119,9 @@ enum DeviceType {
   Unknown = 7;
 }
 
+// The presence of this field in the proto indicates the lights HAL is enabled.
+message Lights {}
+
 // GuestConfigFile represents the configuration for a Cuttlefish guest device,
 // typically parsed from an cuttlefish-guest-config.txtpb file. It defines the
 // device's hardware capabilities, software features, and virtualization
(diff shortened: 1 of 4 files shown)


base/cvd/cuttlefish/host/commands/assemble_cvd/proto/guest_config.proto
```diff
@@ -119,6 +119,11 @@ enum DeviceType {
   Unknown = 7;
 }
 
+message Lights {
+  // Whether the guest runs the lights server.
+  bool guest_server_enabled = 1;
+}
```
Databean: If this structure ever gains other fields, would any of the other fields have meaning in the case that `guest_server_enabled` is `false`?

Another way to represent this information would be to leave the `Lights` structure empty, and instead use field presence for the `Lights lights = 8;` as the boolean value, i.e. `has_lights()`.

- https://protobuf.dev/programming-guides/field_presence/#presence-editions
- https://protobuf.dev/programming-guides/editions/#field-labels
- https://protobuf.dev/reference/cpp/cpp-generated/#embeddedmessage

Databean (approved):

base/cvd/cuttlefish/host/commands/assemble_cvd/proto/guest_config.proto
jemoreira: Done

PR #2238, opened 2026-03-09
Add URL-based build sources (GCS/HTTPS) to cvd fetch

larsers:
Add support for GCS (gs://), HTTPS, and HTTP URLs as build sources in cvd fetch, alongside the existing Android CI and local directory build sources.

  - Add UrlBuildString and UrlBuild variant types to the build type system
    (BuildString/Build std::variant)
  - Add UrlDownloader class for GCS (via JSON API) and HTTP/HTTPS downloads,
    reusing existing HttpClient infrastructure
  - Scope OAuth2 auth headers to Google hosts only (*.googleapis.com,
    *.google.com) to prevent credential leakage to third-party servers
  - Handle automatic archive extraction (.zip, .tar.gz) and selective file
    extraction via {filepath} suffix syntax (e.g.,
    gs://bucket/images.zip{boot.img})
  - Add unit tests for URL parsing, filename extraction, and Google host validation

This change allows direct use of URLs in --default_build and other --*_build flags, e.g:

```
  cvd fetch --default_build=gs://my-bucket/aosp-images.zip
  cvd fetch --default_build=https://example.com/cvd-host_package.tar.gz
```


cjreynol:
This update does not seem to align with the goal of `cvd fetch`, which is to take an Android build and retrieve all of the necessary artifacts necessary to launch a Cuttlefish device without the user needing to know anything about those artifacts.  This PR is introducing more generic downloading of specified artifacts the user is already aware of, rather than actual Android build output artifacts, which seems like a task better suited to something like `cURL` or the `gcloud` CLI tool.

Thank you for the PR, but this does not match the current design plans.

base/cvd/cuttlefish/host/libs/web/android_build_api.h
```diff
@@ -86,6 +86,10 @@ class AndroidBuildApi : public BuildApi {
       const DirectoryBuild& build,
       const std::vector<std::string>& artifact_filenames);
 
+  Result<std::unordered_set<std::string>> Artifacts(
```
jemoreira, line 91: This doesn't belong here. `AndroidBuildApi` is meant for builds downloaded from the android build api. While the URL could point to a build from that api, it's technically more general and doesn't fit the narrower scope of this class.

base/cvd/cuttlefish/host/libs/web/url_downloader.h
```diff
@@ -0,0 +1,95 @@
...
+  UrlDownloader(HttpClient& http_client, CredentialSource* credential_source);
+
+  // Downloads a URL to the specified directory.
+  // Returns the path to the downloaded file.
+  Result<std::string> Download(const std::string& url,
+                               const std::string& target_directory);
+
+  // Downloads multiple URLs to the specified directory.
+  // Returns paths to all downloaded files in the same order as input URLs.
+  Result<std::vector<std::string>> DownloadAll(
```
jemoreira: This isn't used, delete it. The implementation is trivial enough that if needed the caller can do it.

base/cvd/cuttlefish/host/libs/web/url_downloader.cc
```diff
@@ -0,0 +1,183 @@
...
+}
+
+Result<void> UrlDownloader::DownloadFromHttp(const std::string& url,
+                                             const std::string& dest_path) {
+  // Only send auth headers to Google hosts to avoid leaking credentials
+  // to third-party servers.
+  std::vector<std::string> headers;
+  if (IsGoogleHost(url)) {
+    headers = CF_EXPECT(AuthHeaders());
+  }
```
jemoreira: This needs a more robust design than just sending credentials to google servers. A possible approach could be having two different build types represented by urls: authenticated and unauthenticated. That means there will have to be a way to differentiate the two when parsing the build string from `--default_build` and the like.

base/cvd/cuttlefish/host/libs/web/url_downloader.cc
```diff
@@ -0,0 +1,183 @@
...
+
+    case UrlScheme::kHttps:
+    case UrlScheme::kHttp:
+      return DownloadFromHttp(url, dest_path);
+  }
+
+  return CF_ERR("Unknown URL scheme");
+}
+
+Result<void> UrlDownloader::DownloadFromGcs(const std::string& bucket,
```
jemoreira: Downloading from GCS is specific enough that it deserves its own build type, with its own separate "downloader" class.

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_context.cc
```diff
@@ -91,9 +102,19 @@ Result<void> FetchArtifact::DownloadTo(std::string local_path) {
 
 Result<ReadableZip*> FetchArtifact::AsZip() {
   if (!zip_) {
-    zip_ = CF_EXPECT(
-        ::cuttlefish::OpenZip(fetch_build_context_.fetch_context_.build_api_,
-                              fetch_build_context_.build_, artifact_name_));
+    if (std::holds_alternative<UrlBuild>(fetch_build_context_.build_)) {
```
jemoreira: The `::cuttlefish::OpenZip` class should handle http and https urls and probaly gs:// too. The full download should be avoided.

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_context.cc
```diff
@@ -104,6 +125,32 @@ Result<void> FetchArtifact::ExtractAll() {
 }
 
 Result<void> FetchArtifact::ExtractAll(const std::string& local_path) {
+  // Non-zip archives (tar.gz, tgz) require downloading first and use the
+  // system archive utility for extraction.
+  if (absl::EndsWith(artifact_name_, ".tar.gz") ||
+      absl::EndsWith(artifact_name_, ".tgz")) {
```
jemoreira: nit: IIUC this adds supports for downloading tar.gz and tgz files in addition to zip files. This is somewhat unrelated to the general url downloading feature, so it could be in its own commit to make reviewing easier.

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_context.h
```diff
@@ -60,21 +65,20 @@ class FetchArtifact {
...
+// - Desparsing images.
+//
+// File paths for return values and argument values are relative to the target
+// directory.
+//
+// By hiding the target directory from direct access, IO operations are
+// funneled through an instance of this type, which guarantees none of the
+// cross-cutting concerns are missed. Additionally, this could be replaced
+// with a fake implementation later to support unit testing the business
+// logic.
```
jemoreira: Please keep the old comment format, changing it only adds noise to the review.

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
```diff
@@ -398,43 +430,77 @@ Result<void> FetchChromeOsTarget(
...
+// component-specific fetch function (for CI/directory builds).
 Result<void> FetchTarget(FetchContext& fetch_context,
                          const DownloadFlags& flags,
                          const bool keep_downloaded_archives) {
-  if (std::optional<FetchBuildContext> context = fetch_context.DefaultBuild()) {
-    bool has_system_build = fetch_context.SystemBuild().has_value();
-    CF_EXPECT(FetchDefaultTarget(*context, keep_downloaded_archives, flags,
-                                 has_system_build));
+  if (auto ctx = fetch_context.DefaultBuild()) {
+    if (std::holds_alternative<UrlBuild>(ctx->Build())) {
```
jemoreira: This `if` should happen in `FetchDefaultTarget`. The same applies to similar conditionals below.

jemoreira (commented):
Left a few preliminary comments, but getting this to an acceptable state that can be merged will probably take a few iterations.

I'm particularly concerned with the number of places where an `if` is needed to differentiate between url builds and others. That should be solved more elegantly, with something like polymorphism or the visitor pattern on the variant.

base/cvd/cuttlefish/host/libs/web/android_build_api.h
larsers: Done. Moved URL handling out of AndroidBuildApi. URL builds now go through their own classes, `GcsBuildApi` and `HttpBuildApi`. `AndroidBuildApi` keeps a few defensive stubs for the variant cases it does not handle, since `CachingBuildApi` still wraps a `BuildApi&` and the variant methods need to compile. The stubs return `CF_ERR` and are unreachable in correct usage, since the new Downloaders dispatch routes URL builds to the right specialist class directly.

base/cvd/cuttlefish/host/libs/web/url_downloader.h
larsers: Done.

base/cvd/cuttlefish/host/libs/web/url_downloader.cc
larsers: Yeah, you're right. Changed so that the auth model is now encoded in the type. There are two distinct build types, `GcsBuild` and `HttpBuild`. `GcsBuildApi` takes a `CredentialSource&` so it cannot be constructed without credentials. `HttpBuildApi` has no credential source field at all, so it physically cannot send auth headers to a third party. The` IsGoogleHost` heuristic is gone.` ParseBuildString` differentiates by scheme prefix, `gs://` goes to `GcsBuildString` and https:// or http:// goes to `HttpBuildString`.

base/cvd/cuttlefish/host/libs/web/url_downloader.cc
larsers: Fixed. `GcsBuild` is now a distinct build type with its own `GcsBuildApi` class. I've tried to follow the `LuciBuildApi` pattern, a standalone class that does not inherit from `BuildApi`.

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_context.cc
larsers: OK, fixed. . `FetchArtifact::AsZip` now calls `Downloaders::FileReader`, which dispatches to the per-type class. Both `GcsBuildApi::FileReader` and `HttpBuildApi::FileReader` return `ZipSourceFromUrl` from `remote_zip.h`, so the archive is read as a streaming `SeekableZipSource` via HTTP Range requests. The full download fallback in `AsZip` is removed.

One caveat. `FetchUrlTarget` still calls `Download()` before extraction to preserve the existing `AddFileToConfig` behavior. The streaming primitive is in place but the main URL fetch flow does not yet exercise it. I plan to file a follow-up that skips the explicit `Download()` for zip URLs so the streaming path actually gets used.

`OpenZip` itself is now unused after this change, so I deleted `build_api_zip.{h,cc}` along with the rest of the cleanup.
````

File: base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_context.cc (hunk 6 above)
Comment (larsers) [reply base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_context.cc]: OK, done. The tar.gz and tgz block in `FetchArtifact::ExtractAll` has been removed from this PR

File: base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_context.h (hunk 7 above)
Comment (larsers) [reply base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_context.h]: OK...reverted to the `/** */` form.

File: base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc (hunk 8 above)
Comment (larsers) [reply base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc]: OK, fixed it so that `FetchTarget` is now a clean orchestrator that just lists the per-target fetches. Each `Fetch*Target` has an `IsUrlBuild` guard at the top that delegates to `FetchUrlTarget` when needed. The 8 `holds_alternative` checks in `FetchTarget` are gone. 

To make this work, `FetchKernelTarget`, `FetchBootloaderTarget`, and `FetchAndroidEfiLoaderTarget` gained a `keep_downloaded_archives` parameter so they can pass it through to `FetchUrlTarget`.

**Comment by larsers** [comment]
I've made several updates and changes based on feedback...

**Build types.** Replaced `UrlBuild` with two distinct types. `GcsBuild { bucket, object, filepath }` always uses Google credentials. `HttpBuild { url, filepath }` never sends auth headers. The auth model is encoded in the type system rather than in a hostname heuristic.

**Specialist classes.** `GcsBuildApi` and `HttpBuildApi` follow the `LuciBuildApi` pattern, standalone classes that do not inherit from `BuildApi`. `GcsBuildApi` requires a `CredentialSource&`. `HttpBuildApi` has no credential field at all. Both implement a `FileReader` equivalent that returns `ZipSourceFromUrl` from `remote_zip.h`, so URL-backed archives can be read as streaming `SeekableZipSource`s via HTTP Range requests.

**Centralized dispatch.** Added three methods to `Downloaders`, `GetBuild`, `DownloadFile`, and `FileReader`. Each is a single `std::visit` over the variant, routing to the right specialist class. Callers (`FetchContext`, `FetchHostPackage`, the build resolution helpers in `fetch_cvd.cc`) take `Downloaders&` and use these methods. No more variant inspection at every call site.

**Per-target dispatch.** `FetchTarget` is now a clean orchestrator. Each `Fetch*Target` has a single `IsUrlBuild` guard at the top that delegates to `FetchUrlTarget` when needed. The 8 if-chains in `FetchTarget` are gone.

File: base/cvd/cuttlefish/host/commands/cvd/fetch/downloaders.cc [hunk 9]
```diff
@@ -127,4 +144,86 @@ BuildApi& Downloaders::AndroidBuild() {
...
+
+    Result<Build> operator()(const DeviceBuildString&) {
+      return self.AndroidBuild().GetBuild(build_string);
+    }
+    Result<Build> operator()(const DirectoryBuildString&) {
+      return self.AndroidBuild().GetBuild(build_string);
+    }
+    Result<Build> operator()(const GcsBuildString& bs) {
+      CF_EXPECT(self.Gcs() != nullptr,
+                "GCS build requested but no GCS credentials are configured");
```
jemoreira: nit: This code sees "no GCS downloader" and prints "no credentials configured". Consider using a `Result<GcsBuildApi>` instead of a pointer, so when this code doesn't see the object it can get the reason from the error message. That way, the "no credentials" message is written next to the check for credentials.

Alternatively, the object could have a pointer to or an optional of the credentials and always be constructed. Then each of the methods it implements would fail if it doesn't have credentials with the appropriate message. This allows Downloaders::Gcs to return a reference like the other accessors.

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
```diff
@@ -69,6 +69,56 @@ bool IsSignedBuild(const Build& build) {
...
 }
 
+bool IsUrlBuild(const Build& build) {
+  return std::holds_alternative<GcsBuild>(build) ||
+         std::holds_alternative<HttpBuild>(build);
+}
+
+// Returns the local filename to use when downloading a URL-style build.
+// Derived from the GCS object path or the HTTP URL path (sans query
+// parameters). Returns empty string for non-URL builds.
```
jemoreira: nit: Return Result<std::string>. It's an error to call this method on a non-URL build.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
```diff
@@ -448,10 +491,29 @@ Result<std::string> AndroidBuildApi::DownloadTargetFile(
...
+        const DeviceBuild& b) const {
+      return {b.id, b.target};
+    }
+    std::tuple<std::string, std::string> operator()(
+        const DirectoryBuild& b) const {
+      return {b.id, b.target};
+    }
+    std::tuple<std::string, std::string> operator()(
+        const GcsBuild& b) const {
+      return {"gs://" + b.bucket + "/" + b.object, "url"};
```
jemoreira: nit: use absl::StrCat

base/cvd/cuttlefish/host/libs/web/android_build_string.cpp
```diff
@@ -163,6 +193,15 @@ Result<BuildString> ParseBuildString(const std::string& build_string) {
   CF_EXPECT(!build_string.empty(), "The given build string cannot be empty");
   auto [remaining_build_string, filepath] =
       CF_EXPECT(ParseFilepath(build_string));
+  // URL detection must come before the ':' check, because URLs contain
+  // '://' which would otherwise be misinterpreted as a directory separator.
+  if (android::base::StartsWith(remaining_build_string, "gs://")) {
```
jemoreira: nit: `remaining_build_string.starts_with("gs://")`, avoid using android::base when something else is available.

base/cvd/cuttlefish/host/libs/web/gcs_build_api.cc
```diff
@@ -0,0 +1,98 @@
...
+std::string GcsBuildApi::BuildGcsApiUrl(const std::string& bucket,
+                                        const std::string& object) {
+  return fmt::format(
+      "https://storage.googleapis.com/storage/v1/b/{}/o/{}?alt=media",
+      UrlEscape(bucket), UrlEscape(object));
+}
+
+Result<Build> GcsBuildApi::GetBuild(const GcsBuildString& build_string) {
+  std::string_view url_view = build_string.url;
+  CF_EXPECT(android::base::ConsumePrefix(&url_view, "gs://"),
```
jemoreira: nit: absl::ConsumePrefix

base/cvd/cuttlefish/host/libs/web/gcs_build_api.cc
```diff
@@ -0,0 +1,98 @@
...
+            "GCS URL must start with gs://: " << build_string.url);
+  auto parts = android::base::Split(std::string(url_view), "/");
```
jemoreira: nit: absl::StrSplit

jemoreira (commented):
LGTM
I've asked Cody to review it too as he has worked with this more reently.

base/cvd/cuttlefish/host/commands/cvd/fetch/downloaders.cc
larsers: Done. `Gcs()` now returns `Result<GcsBuildApi*>`, so the missing-credentials reason lives in one place (the accessor) and the three variant visitors just do `GcsBuildApi* gcs = CF_EXPECT(self.Gcs());`.

I went with `Result<GcsBuildApi*>` rather than `Result<GcsBuildApi>` by value because `GcsBuildApi` holds reference members (`HttpClient&`, `CredentialSource&`) and so isn't default-constructible — it can't be stored as a bare `Result<>` in the pimpl `Impl`. The pointer form achieves the same
goal. The failure reason is carried in the error instead of being re-stated at each call site.

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
larsers: Done. `UrlBuildFilename` now returns `Result<std::string>` and returns `CF_ERR(...)` for the non-URL case; the caller in `FetchUrlTarget` uses `CF_EXPECT`.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
larsers: Done.

base/cvd/cuttlefish/host/libs/web/android_build_string.cpp
larsers: Done. All three prefix checks (`gs://`, `https://`, `http://`) now use `std::string::starts_with`.

base/cvd/cuttlefish/host/libs/web/gcs_build_api.cc
larsers: Done.

base/cvd/cuttlefish/host/libs/web/gcs_build_api.cc
larsers: Done, and I also switched the adjacent `android::base::Join` to `absl::StrJoin`, which let me drop the `//libbase` dependency from this file entirely.

larsers:
Closing this PR in favor of a upcoming redesigned successor.

PR #2239, opened 2026-03-09
Add WebSocket signaling for WebRTC connections

larsers:
Add WebSocket-based signaling as a faster alternative to HTTP long-polling for WebRTC connections between the browser client and the Cuttlefish operator.

Key changes:
- Add WsClient type implementing the Client interface with goroutine-based read/write pumps and ping/pong keepalive
- Add GET /devices/{deviceId}/connect WebSocket upgrade endpoint
- Add WebSocketConnector JavaScript class that tries WebSocket first and falls back to HTTP polling if unavailable
- Full backward compatibility: existing polling connections continue to work unchanged

Uses the existing gorilla/websocket dependency (already used for the ADB WebSocket proxy).

### DIFF
--- a/frontend/src/liboperator/operator/clients.go
+++ b/frontend/src/liboperator/operator/clients.go
@@ -15,10 +15,27 @@
 package operator
 
 import (
+	"encoding/json"
+	"fmt"
+	"log"
 	"math/rand"
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/gorilla/websocket"
+)
+
+// WebSocket timing constants
+const (
+	// Time allowed to write a message to the peer
+	wsWriteWait = 10 * time.Second
+	// Time allowed to read the next pong message from the peer
+	wsPongWait = 60 * time.Second
+	// Send pings to peer with this period (must be less than wsPongWait)
+	wsPingPeriod = 30 * time.Second
+	// Maximum message size allowed from peer
+	wsMaxMessageSize = 64 * 1024
 )
 
 type Client interface {
@@ -138,4 +155,159 @@ func randStr(l int, r *rand.Rand) string {
 		s.WriteByte(runes[r.Intn(len(runes))])
 	}
 	return s.String()
+}
+
+// WsClient implements Client for WebSocket connections.
+type WsClient struct {
+	// The id given to this client by the device
+	clientId int
+	conn     *websocket.Conn
+	device   *Device
+	// Device id for logging
+	deviceId string
+	// Buffered channel of outbound messages
+	send chan interface{}
+	// Signals that the connection is closed
+	done chan struct{}
+	// Ensures Close() only runs once
+	closeOnce sync.Once
+}
+
+// NewWsClient creates a new WebSocket client and starts its read/write pumps.
+func NewWsClient(conn *websocket.Conn, device *Device, deviceId string) *WsClient {
+	c := &WsClient{
+		conn:     conn,
+		device:   device,
+		deviceId: deviceId,
+		send:     make(chan interface{}, 256),
+		done:     make(chan struct{}),
+	}
+	go c.writePump()
+	go c.readPump()
+	return c
+}
+
+// Send queues a message for delivery to the WebSocket client.
+func (c *WsClient) Send(msg interface{}) error {
+	// Check if already closed before attempting to send. Without this,
+	// select would pick randomly between send and done when both are ready.
+	select {
+	case <-c.done:
+		return fmt.Errorf("WebSocket connection closed")
+	default:
+	}
+	select {
+	case c.send <- msg:
+		return nil
+	case <-c.done:
+		return fmt.Errorf("WebSocket connection closed")
+	}
+}
+
+// OnDeviceDisconnected notifies the client that the device disconnected.
+func (c *WsClient) OnDeviceDisconnected() {
+	// Send an error message to the client before closing
+	c.Send(map[string]interface{}{
+		"error": "Device disconnected",
+	})
+	c.Close()
+}
+
+// Close signals shutdown of the WebSocket connection.
+// The actual connection close is handled by writePump's defer.
+func (c *WsClient) Close() {
+	c.closeOnce.Do(func() {
+		close(c.done)
+	})
+}
+
+// Done returns a channel that is closed when the client disconnects.
+func (c *WsClient) Done() <-chan struct{} {
+	return c.done
+}
+
+// ClientId returns the id assigned by the device.
+func (c *WsClient) ClientId() int {
+	return c.clientId
+}
+
+// readPump reads messages from the WebSocket and forwards them to the device.
+func (c *WsClient) readPump() {
+	defer func() {
+		log.Printf("WebSocket client %d disconnected from device %s",
+			c.clientId, c.deviceId)
+		c.device.Unregister(c.clientId)
+		c.Close()
+	}()
+
+	c.conn.SetReadLimit(wsMaxMessageSize)
+	c.conn.SetReadDeadline(time.Now().Add(wsPongWait))
+	c.conn.SetPongHandler(func(string) error {
+		c.conn.SetReadDeadline(time.Now().Add(wsPongWait))
+		return nil
+	})
+
+	for {
+		_, message, err := c.conn.ReadMessage()
+		if err != nil {
+			if !websocket.IsCloseError(err,
+				websocket.CloseNormalClosure,
+				websocket.CloseGoingAway) {
+				log.Printf("WebSocket read error: %v", err)
+			}
+			return
+		}
+
+		var msg map[string]interface{}
+		if err := json.Unmarshal(message, &msg); err != nil {
+			log.Printf("Invalid JSON from WebSocket client: %v", err)
+			return
+		}
+
+		// Forward to device as client_msg
+		clientMsg := map[string]interface{}{
+			"message_type": "client_msg",
+			"client_id":    c.clientId,
+			"payload":      msg,
+		}
+		if err := c.device.Send(clientMsg); err != nil {
+			log.Printf("Failed to forward message to device: %v", err)
+			return
+		}
+	}
+}
+
+// writePump writes messages from the send channel to the WebSocket.
+func (c *WsClient) writePump() {
+	ticker := time.NewTicker(wsPingPeriod)
+	defer func() {
+		ticker.Stop()
+		c.conn.Close()
+	}()
+
+	for {
+		select {
+		case msg, ok := <-c.send:
+			c.conn.SetWriteDeadline(time.Now().Add(wsWriteWait))
+			if !ok {
+				// Channel closed
+				c.conn.WriteMessage(websocket.CloseMessage, []byte{})
+				return
+			}
+
+			if err := c.conn.WriteJSON(msg); err != nil {
+				log.Printf("WebSocket write error: %v", err)
+				return
+			}
+
+		case <-ticker.C:
+			c.conn.SetWriteDeadline(time.Now().Add(wsWriteWait))
+			if err := c.conn.WriteMessage(websocket.PingMessage, nil); err != nil {
+				return
+			}
+
+		case <-c.done:
+			return
+		}
+	}
 }
--- a/frontend/src/liboperator/operator/clients_test.go
+++ b/frontend/src/liboperator/operator/clients_test.go
@@ -100,3 +100,79 @@ func TestRandStr(t *testing.T) {
 		t.Error("randStr returned two equal strings of len 100")
 	}
 }
+
+// TestWsClientImplementsClient verifies WsClient implements the Client interface.
+func TestWsClientImplementsClient(t *testing.T) {
+	var _ Client = (*WsClient)(nil)
+}
+
+// TestWsClientClose verifies that Close() can be called multiple times safely.
+func TestWsClientClose(t *testing.T) {
+	// Create a minimal WsClient without a real connection
+	// to test the Close() sync.Once behavior
+	c := &WsClient{
+		send: make(chan interface{}, 1),
+		done: make(chan struct{}),
+	}
+
+	// First close should work
+	c.Close()
+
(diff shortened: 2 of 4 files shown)


frontend/src/operator/intercept/js/server_connector.js
```diff
@@ -42,7 +42,19 @@ export function deviceId() {
...
+// Tries WebSocket first, falls back to HTTP polling if WebSocket is unavailable.
 export async function createConnector() {
+  // Check if WebSocket is supported
+  if (typeof WebSocket !== 'undefined') {
+    try {
+      const connector = new WebSocketConnector();
+      console.log('Using WebSocket signaling');
+      return connector;
+    } catch (e) {
+      console.warn('WebSocket connector creation failed, falling back to polling:', e);
```
jemoreira: IIUC this only falls back to HTTP polling if this browser doesn't support websockets, not if the websocket connection fails. The PollingConnector was added alongside the old Websocket based connector in https://android-review.git.corp.google.com/c/device/google/cuttlefish/+/1811217, you can use the fallback logic in that cl as guidance for this one.

frontend/src/operator/intercept/js/server_connector.js
```diff
@@ -274,6 +286,116 @@ class PollingConnector extends Connector {
...
+          }
+          return;
+        }
+
+        // Normal message - forward to callback
+        // Messages from device come with payload wrapper
+        if (message.payload !== undefined) {
+          this.#onDeviceMsgCb(message.payload);
+        } else {
+          this.#onDeviceMsgCb(message);
```
jemoreira: IIRC messages without a payload don't need to be delivered to the message callback. I don't remember any use case of messages without payload, have you seen any such messages?

frontend/src/operator/intercept/js/server_connector.js
```diff
@@ -274,6 +286,116 @@ class PollingConnector extends Connector {
...
+      };
+    });
+  }
+
+  async sendToDevice(msg) {
+    if (this.#ws && this.#ws.readyState === WebSocket.OPEN) {
+      this.#ws.send(JSON.stringify(msg));
+    } else {
+      throw new Error('WebSocket not connected');
+    }
```
jemoreira: nit: reverse this condition and implement like this
```
if (!this.#ws || this.#ws.readyState !== Websocket.OPEN) {
  throw ...
}
this.#ws.send(...);
```

frontend/src/liboperator/operator/clients.go
```diff
@@ -139,3 +155,151 @@ func randStr(l int, r *rand.Rand) string {
...
+	})
+	c.Close()
+}
+
+// Close terminates the WebSocket connection.
+func (c *WsClient) Close() {
+	c.closeOnce.Do(func() {
+		close(c.done)
+		if c.conn != nil {
+			c.conn.Close()
```
jemoreira: It doesn't seem necessary to call `conn.Close()` here as it will be called at the end of the write pump.

frontend/src/liboperator/operator/clients.go
```diff
@@ -139,3 +155,151 @@ func randStr(l int, r *rand.Rand) string {
...
+	return c
+}
+
+// Send queues a message for delivery to the WebSocket client.
+func (c *WsClient) Send(msg interface{}) error {
+	select {
+	case c.send <- msg:
+		return nil
+	case <-c.done:
+		return nil
```
jemoreira: Shouldn't `Send` return an error when the connection is closed while sending?

frontend/src/liboperator/operator/clients.go
```diff
@@ -139,3 +155,151 @@ func randStr(l int, r *rand.Rand) string {
...
+		c.conn.SetReadDeadline(time.Now().Add(wsPongWait))
+		return nil
+	})
+
+	for {
+		_, message, err := c.conn.ReadMessage()
+		if err != nil {
+			if websocket.IsUnexpectedCloseError(err,
+				websocket.CloseGoingAway,
+				websocket.CloseNormalClosure) {
```
jemoreira: Why log only these kinds of error? Shouldn't errors of all types be logged?

frontend/src/liboperator/operator/clients.go
```diff
@@ -139,3 +155,151 @@ func randStr(l int, r *rand.Rand) string {
...
+				log.Printf("WebSocket read error: %v", err)
+			}
+			return
+		}
+
+		var msg map[string]interface{}
+		if err := json.Unmarshal(message, &msg); err != nil {
+			log.Printf("Invalid JSON from WebSocket client: %v", err)
+			continue
```
jemoreira: Close the connection if the client sends invalid data. It's unlikely to send anything useful later.

frontend/src/liboperator/operator/clients_test.go
```diff
@@ -100,3 +100,81 @@ func TestRandStr(t *testing.T) {
 		t.Error("randStr returned two equal strings of len 100")
 	}
 }
+
+// TestWsClientImplementsClient verifies WsClient implements the Client interface.
+// This is a compile-time check, but we include it as a test for documentation.
+func TestWsClientImplementsClient(t *testing.T) {
+	// This is checked at compile time via: var _ Client = (*WsClient)(nil)
+	// If WsClient doesn't implement Client, the code won't compile.
```
jemoreira: Would it make sense to move that check to this function?

frontend/src/liboperator/operator/clients_test.go
```diff
@@ -100,3 +100,81 @@ func TestRandStr(t *testing.T) {
...
+	c := &WsClient{
+		send: make(chan interface{}, 1),
+		done: make(chan struct{}),
+	}
+
+	c.Close()
+
+	// Send should return immediately without blocking
+	err := c.Send("test message")
+	if err != nil {
```
jemoreira: Send should return an error if called after closed.

frontend/src/liboperator/operator/operator.go
```diff
@@ -550,6 +555,50 @@ func adbProxy(w http.ResponseWriter, r *http.Request, pool *DevicePool) {
...
+
+	log.Printf("WebSocket client %d connected to device %s", clientId, deviceId)
+
+	// Send device info to the client
+	client.Send(map[string]interface{}{
+		"device_info": device.privateData,
+	})
+
+	// Block until the client disconnects
+	<-client.Done()
```
jemoreira: This will block the handler until the websocket disconnects. Should this be done in a goroutine instead? I'm worried that enough websocket connections could cause the server to stop accepting connections if it uses a goroutine pool to handle requests.

It seems the only reason you're waiting for the connection to close is to write the log. If you move that log to the WsClient there won't be a need to wait here at all.

frontend/src/operator/intercept/js/server_connector.js
larsers: Yeah, you're right...This only catches the case where the browser does not support WebSocket, not when the actual connection fails. I will restructure this with a `FallbackConnector` that wraps both transports. It tries WebSocket in `requestDevice()` and falls back to `PollingConnector` if the connection fails (timeout, 404, upgrade rejected, etc.).

```js
export async function createConnector() {
  return new FallbackConnector();
}
```

The `FallbackConnector.requestDevice()` tries WebSocket first, and the existing 3 second `WS_CONNECT_TIMEOUT` bounds how long we wait before falling back to polling. This way the fallback is driven by actual connectivity, not just browser capability.

frontend/src/operator/intercept/js/server_connector.js
larsers: Nah, you're right... Looking at `deviceEndpoint()` in operator.go, all device to client messages are wrapped with `"payload": payload`. The `device_info` and `error` types are already handled earlier in `onmessage`.

I'll remove the `else` branch and only deliver `message.payload`:

```js
if (message.payload !== undefined) {
  this.#onDeviceMsgCb(message.payload);
}
```

frontend/src/operator/intercept/js/server_connector.js
larsers: OK, I'll fix.

frontend/src/liboperator/operator/clients.go
larsers: Yeah, good catch... When the connection is closed `Send()` should return an error so that `Device.ToClient()` can detect the disconnected client and notify the device... I'll do it like this instead:

```go
func (c *WsClient) Send(msg interface{}) error {
	select {
	case c.send <- msg:
		return nil
	case <-c.done:
		return fmt.Errorf("WebSocket connection closed")
	}
}
```

frontend/src/liboperator/operator/clients.go
larsers: Yep, you're right... The `writePump` defer already closes the connection. I'll remove `conn.Close()` from `Close()` so it only signals shutdown via the `done` channel:

```go
func (c *WsClient) Close() {
	c.closeOnce.Do(func() {
		close(c.done)
	})
}
```

frontend/src/liboperator/operator/clients.go
larsers: It's an artifact from when I debugged the implementation. So good point, the `IsUnexpectedCloseError` filter is too narrow. It only catches close errors with unexpected codes but silently drops other error types like timeouts and connection resets. What about switching to `IsCloseError` to log everything except expected closures, like this?

```go
if err != nil {
	if !websocket.IsCloseError(err,
		websocket.CloseNormalClosure,
		websocket.CloseGoingAway) {
		log.Printf("WebSocket read error: %v", err)
	}
	return
}
```

frontend/src/liboperator/operator/clients.go
larsers: Agreed... A client sending invalid JSON is not going to recover. I'll change the `continue` to `return` which lets the `readPump` defer handle cleanup (Unregister + Close):

```go
if err := json.Unmarshal(message, &msg); err != nil {
	log.Printf("Invalid JSON from WebSocket client: %v", err)
	return
}
```

frontend/src/liboperator/operator/clients_test.go
larsers: Sure... I can move the compile time check directly into the test body and remove it from the package level, like this:

```go
func TestWsClientImplementsClient(t *testing.T) {
	var _ Client = (*WsClient)(nil)
}
```

frontend/src/liboperator/operator/clients_test.go
larsers: OK, as this follows from the `Send()` fix. After `Close()`, `Send()` now returns an error so the test should expect that:

```go
err := c.Send("test message")
if err == nil {
	t.Error("Send after close should return an error")
}
```

frontend/src/liboperator/operator/operator.go
larsers: Yeah, you're right... The handler only blocks here to log the disconnect which is not worth holding the goroutine imo. I'll add `deviceId` to `WsClient` and move the disconnect log into `readPump()`'s defer:

```go
func (c *WsClient) readPump() {
	defer func() {
		log.Printf("WebSocket client %d disconnected from device %s",
			c.clientId, c.deviceId)
		c.device.Unregister(c.clientId)
		c.Close()
	}()
	// ...
}
```

The handler can then return right after setup since `Upgrade()` hijacks the connection and the readPump/writePump goroutines manage the lifecycle from there...

frontend/src/liboperator/operator/operator.go
```diff
@@ -592,11 +592,7 @@ func connectWebSocket(w http.ResponseWriter, r *http.Request, pool *DevicePool)
 	client.Send(map[string]interface{}{
 		"device_info": device.privateData,
 	})
-
-	// Block until the client disconnects
-	<-client.Done()
-
-	log.Printf("WebSocket client %d disconnected from device %s", clientId, deviceId)
+	// readPump and writePump goroutines manage the connection lifecycle
```
jemoreira: nit: This code doesn't know about the pumps, rewrite the comment to say "the client manages ..."

frontend/src/operator/intercept/js/server_connector.js
```diff
@@ -286,6 +275,56 @@ class PollingConnector extends Connector {
...
+    super();
+  }
+
+  async requestDevice(deviceId) {
+    if (typeof WebSocket !== 'undefined') {
+      try {
+        const ws = new WebSocketConnector();
+        ws.onDeviceMsg(this.#onDeviceMsgCb);
+        const result = await ws.requestDevice(deviceId);
+        this.#delegate = ws;
```
jemoreira, line 301: Maybe it doesn't matter, but please move the assignment to the `#delegate` field to before calling `ws.requestDevice` and do the same below for the fallback logic.

`requestDevice` has `async` calls, during its execution it may (this is the part I'm not sure about) call the `onDeviceMsgCb` callback, which in turn could call `sendToDevice` on this object which would result in attempting to access `this.#delegate` which is still undefined.

frontend/src/liboperator/operator/clients.go
jemoreira: SG

jemoreira:
Thanks for addressing my feedback in a new commit, it made the second pass a lot easier. I just left two very minor suggestions this time. When you fix those please squash everything into a single commit and I'll merge it then.

frontend/src/liboperator/operator/operator.go
larsers: Done, reworded to "The client manages the connection lifecycle."

frontend/src/operator/intercept/js/server_connector.js
larsers: Yeah, you're right that it could happen. Moved the assignment before the await for both paths. If the WebSocket attempt fails, `#delegate` is reset to `null` before falling through to polling.

larsers:
Thanks for the review! Both fixes added and I've squashed it into a single commit.

jemoreira (approved):

PR #2249, opened 2026-03-10
cuttlefish-integration-gigabyte-arm64: install openjdk by default

grandpaul:
We've already migrate Gigabyte Ampere server installer to Trixie. Thus we can install openjdk-21 from the repo. Install it by default.

We also add modify the tests to check that the openjdk is installed and is the version we want.

### DIFF
--- a/.github/workflows/gigabyte-ampere-cuttlefish-installer.yaml
+++ b/.github/workflows/gigabyte-ampere-cuttlefish-installer.yaml
@@ -280,7 +280,15 @@ jobs:
       run: |
         sshpass -p cuttlefish ssh -o "StrictHostKeyChecking no" -o "UserKnownHostsFile /dev/null" -p 33322 vsoc-01@localhost 'ulimit -a'
         test $(sshpass -p cuttlefish ssh -o "StrictHostKeyChecking no" -o "UserKnownHostsFile /dev/null" -p 33322 vsoc-01@localhost 'ulimit -n') -ge 2048
+    - name: Check iptables points to iptables-legacy
+      run: |
         sshpass -p cuttlefish ssh -o "StrictHostKeyChecking no" -o "UserKnownHostsFile /dev/null" -p 33322 vsoc-01@localhost 'update-alternatives --display iptables' | grep "link currently points to /usr/sbin/iptables-legacy"
+    - name: Test Java version
+      run: |
+        #sshpass -p cuttlefish ssh -o "StrictHostKeyChecking no" -o "UserKnownHostsFile /dev/null" -p 33322 vsoc-01@localhost 'java --version'
+        #JAVA_VERSION=$(sshpass -p cuttlefish ssh -o "StrictHostKeyChecking no" -o "UserKnownHostsFile /dev/null" vsoc-01@localhost 'java -version 2>&1' | sed -n ';s/.* version "\([^"]*\)".*/\1/p;')
+        JAVA_REQUIRED_VERSION="21.0"
+        #test "$(printf '%s\n%s' "${JAVA_REQUIRED_VERSION}" "${JAVA_VERSION}" | sort -V | head -n1)" = "${JAVA_REQUIRED_VERSION}"
     - name: Shutdown qemu
       run: |
         sshpass -p cuttlefish ssh -o "StrictHostKeyChecking no" -o "UserKnownHostsFile /dev/null" -p 33322 vsoc-01@localhost 'echo cuttlefish | sudo -S -k shutdown -h 1'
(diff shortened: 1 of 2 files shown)


rmuthiah (approved):

.github/workflows/gigabyte-ampere-cuttlefish-installer.yaml
```diff
@@ -279,7 +279,13 @@ jobs:
       run: |
         sshpass -p cuttlefish ssh -o "StrictHostKeyChecking no" -o "UserKnownHostsFile /dev/null" -p 33322 vsoc-01@localhost 'ulimit -a'
         test $(sshpass -p cuttlefish ssh -o "StrictHostKeyChecking no" -o "UserKnownHostsFile /dev/null" -p 33322 vsoc-01@localhost 'ulimit -n') -ge 2048
+    - name: Check iptables points to iptables-legacy
+      run: |
         sshpass -p cuttlefish ssh -o "StrictHostKeyChecking no" -o "UserKnownHostsFile /dev/null" -p 33322 vsoc-01@localhost 'update-alternatives --display iptables' | grep "link currently points to /usr/sbin/iptables-legacy"
+    - name: Test Java version
+      run: |
+        sshpass -p cuttlefish ssh -o "StrictHostKeyChecking no" -o "UserKnownHostsFile /dev/null" -p 33322 vsoc-01@localhost 'java --version'
+        sshpass -p cuttlefish ssh -o "StrictHostKeyChecking no" -o "UserKnownHostsFile /dev/null" -p 33322 vsoc-01@localhost 'java --version' | grep 21.0
```
0405ysj: What if downloaded version of `default-jdk` from apt repository is higher than `21.0`? If it's okay, what about adding `default-jdk (>=2:1.21)` on `debian/control`?

grandpaul: OK. Done. Thanks. I've changed the test to compare for the minimal version. 
Also we will need openjdk-25 for the future so I've installed it too.

.github/workflows/gigabyte-ampere-cuttlefish-installer.yaml
```diff
@@ -280,7 +280,15 @@ jobs:
       run: |
         sshpass -p cuttlefish ssh -o "StrictHostKeyChecking no" -o "UserKnownHostsFile /dev/null" -p 33322 vsoc-01@localhost 'ulimit -a'
         test $(sshpass -p cuttlefish ssh -o "StrictHostKeyChecking no" -o "UserKnownHostsFile /dev/null" -p 33322 vsoc-01@localhost 'ulimit -n') -ge 2048
+    - name: Check iptables points to iptables-legacy
+      run: |
         sshpass -p cuttlefish ssh -o "StrictHostKeyChecking no" -o "UserKnownHostsFile /dev/null" -p 33322 vsoc-01@localhost 'update-alternatives --display iptables' | grep "link currently points to /usr/sbin/iptables-legacy"
+    - name: Test Java version
+      run: |
+        sshpass -p cuttlefish ssh -o "StrictHostKeyChecking no" -o "UserKnownHostsFile /dev/null" -p 33322 vsoc-01@localhost 'java --version'
```
0405ysj: Presubmit check have shown `bash: line 1: java: command not found`.

0405ysj (approved):

.github/workflows/gigabyte-ampere-cuttlefish-installer.yaml
grandpaul: Yeah. We have a chicken and egg problem here. The test install packages from stable release. And the modification doesn't release yet. Any suggestions?

0405ysj: The best way would be passing debian package from GitHub [presubmit](https://github.com/google/android-cuttlefish/blob/main/.github/workflows/presubmit.yaml#L246-L249) check, but I'm not certain if passing files is a feasible option on the server image.

Or, I think we can [point CF deb package apt repo](https://github.com/google/android-cuttlefish/blob/main/gigabyte-ampere-cuttlefish-installer/preseed/after_install_1.sh#L67-L70) into nightly channel(android-cuttlefish-nightly). This works for postsubmit check, but have no clear mechanism for presubmit.

0405ysj: Depending on its urgency, adding test around java version later seems good to me as well.

grandpaul: OK. I've temporarily disabled the test in this commit. And I will open the test when we have a stable release.

PR #2290, opened 2026-03-19
Gather data for fetch metrics (but do not transmit)

cjreynol:
Gather the data for fetch metrics and make it available in `metrics_conversion.cc`, ready to be added to the transmitted proto AFTER the new fields are approved.

Also includes refactors to the existing `metrics_orchestration` logic, due to the differences in how the necessary data is collected when fetching and for device group operations.

Bug: 491965176


cjreynol:
Closing in favor of #2391 on a new branch.

PR #2300, opened 2026-03-23
Introduce E2E tests on podcvd

0405ysj:
`podcvd` aims to provide identical interface as `cvd` though, it doesn't ensure its functionality as `cvd` ensures as of today. `podcvd` has a lot of unimplemented features still, and there's a risk to break working features while implementing new. To improve `podcvd` more, leaning on E2E test is necessary and E2E test of `cvd` is reusable for validating `podcvd`, as their interface should be identical.

Today, `podcvd` only passes a part of `cvd` E2E tests. I set `display_test` as a minimal set of testing `podcvd` in advance though, I'd enlarge the test coverage of `podcvd` with enabling more tests on it.

On the other hand, I chose GitHub workflow rather than Kokoro for `podcvd` E2E testing because Kokoro check is already a critical path of the presubmit check, and I believe adding `podcvd` on it could be burdening on our DevX. And I expect adding `podcvd` E2E test won't charge more as it uses GitHub standard(free tier) runner. Probably we can switch `podcvd` E2E tests into Kokoro instances afterwards, when we desires.

Context: b/495601167

### DIFF
--- a/.github/workflows/presubmit.yaml
+++ b/.github/workflows/presubmit.yaml
@@ -520,7 +520,10 @@ jobs:
         # Grant permission on devices as it's hard to grant on test env
         sudo chmod 666 /dev/kvm /dev/vhost-net /dev/vhost-vsock
         cd e2etests
-        bazel test //cvd/display_tests \
+        bazel test \
+          //cvd/cvd_powerwash_tests \
+          //cvd/display_tests \
+          //cvd/env_tests \
           --test_env=HOME=$HOME \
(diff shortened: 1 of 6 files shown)


k311093 (approved):

0405ysj:
Note: https://github.com/google/android-cuttlefish/pull/2465 can be the fix of https://github.com/google/android-cuttlefish/pull/2463. However, probably I should seek for the better solution with getting help from this E2E test.

PR #2366, opened 2026-04-09
Add udev rules for reverse ncm device interface service

zenczykowski:
Pixel 6+ phones in "Reverse Ncm Device Interface Service" mode (without/with adb debug interface) show up with USB VID:PID 18d1:4ef{2,3}.

This is effectively the reverse of ncm tethering.
In normal tethering the phone provides a DHCPv4 server (and broadcasts IPv6 RAs). In reverse tethering the phone runs a DHCP client, and expects the other end of the connection to operate the DHCP server (and/or provide IPv6 RAs).

Cuttlefish already provides the required services (cvd-ebr bridge, dnsmasq, iptables MASQUERADE rules), so all we actually need to do is add the appropriate usb cdc ncm interface into the bridge and things magically just work.

This mode can be enabled via:
  adb shell svc usb setFunctions 'rndis'
and disabled via:
  adb shell svc usb setFunctions ''

Bug: 500283166

### DIFF
--- a/base/debian/cuttlefish-base.udev
+++ b/base/debian/cuttlefish-base.udev
@@ -1,2 +1,5 @@
 ACTION=="add", KERNEL=="vhost-net", SUBSYSTEM=="misc", MODE="0660", GROUP="cvdnetwork"
 ACTION=="add", KERNEL=="vhost-vsock", SUBSYSTEM=="misc", MODE="0660", GROUP="cvdnetwork"
+
+ACTION=="add", SUBSYSTEM=="net", ENV{ID_NET_DRIVER}=="cdc_ncm", ENV{ID_VENDOR_ID}=="18d1", ENV{ID_MODEL_ID}=="4ef2", ENV{NM_UNMANAGED}="1", RUN+="/sbin/ip link set dev $env{ID_NET_NAME} master cvd-ebr up"
+ACTION=="add", SUBSYSTEM=="net", ENV{ID_NET_DRIVER}=="cdc_ncm", ENV{ID_VENDOR_ID}=="18d1", ENV{ID_MODEL_ID}=="4ef3", ENV{NM_UNMANAGED}="1", RUN+="/sbin/ip link set dev $env{ID_NET_NAME} master cvd-ebr up"


jemoreira (approved):

zenczykowski:
Somehow due to this being at priority 60 or in the cf directory [ /usr/lib/udev/rules.d/60-cuttlefish-base.rules ]  instead of were I had it locally [ /usr/lib/udev/rules.d/99-reverse-tether.rules ] causes this to not work.  :-(

There's also a /usr/lib/udev/rules.d/60-cuttlefihs-base.rules [hs not sh] which I don't know where it is coming from.

zenczykowski:
further experimentation seems to show this works as
/usr/lib/udev/rules.d/80-reverse-tether.rules
but not as
/usr/lib/udev/rules.d/79-reverse-tether.rules

based on

/usr/lib/udev/rules.d/78-sound-card.rules
/usr/lib/udev/rules.d/80-debian-compat.rules
/usr/lib/udev/rules.d/80-drivers.rules
/usr/lib/udev/rules.d/80-ifupdown.rules
/usr/lib/udev/rules.d/80-iio-sensor-proxy.rules
/usr/lib/udev/rules.d/80-libinput-device-groups.rules
/usr/lib/udev/rules.d/80-net-setup-link.rules
/usr/lib/udev/rules.d/80-reverse-tether.rules
/usr/lib/udev/rules.d/80-udisks2.rules
/usr/lib/udev/rules.d/80-uvcdynctrl.rules
/usr/lib/udev/rules.d/81-net-bridge.rules
/usr/lib/udev/rules.d/81-net-dhcp.rules
/usr/lib/udev/rules.d/82-net-auto-link-local.rules
/usr/lib/udev/rules.d/84-nm-drivers.rules
/usr/lib/udev/rules.d/85-hdparm.rules
/usr/lib/udev/rules.d/85-nm-unmanaged.rules

I'm guessing we rely on net-setup-link or something around there

zenczykowski:
```
$ cat /usr/lib/udev/rules.d/84-nm-drivers.rules
# Do not modify this file, it will get overwritten on updates.
# To override or extend the rules place a file in /etc/udev/rules.d

SUBSYSTEM!="net", GOTO="nm_drivers_end"
ACTION!="add|change|move", GOTO="nm_drivers_end"

# Determine ID_NET_DRIVER if there's no ID_NET_DRIVER or DRIVERS (old udev?)
ENV{ID_NET_DRIVER}=="?*", GOTO="nm_drivers_end"
DRIVERS=="?*", GOTO="nm_drivers_end"
PROGRAM="/bin/sh -c '/usr/sbin/ethtool -i $$1 |/usr/bin/sed -n s/^driver:\ //p' -- $env{INTERFACE}", ENV{ID_NET_DRIVER}="%c"
```

so probably want to be after that -- since we match on ENV{ID_NET_DRIVER} == "cdc_ncm"...

maybe 88 would be a good priority???

PR #2374, opened 2026-04-10
aspect_bazel_lib: add riscv64 expand_template support

monkey-jsun:
Patch aspect_bazel_lib to add linux_riscv64 platform support for the
expand_template toolchain.  Download prebuilt expand_template binary
from GitHub releases instead of requiring a local host build.

Note this PR depends on #2365

PR #2375, opened 2026-04-10
openssl: add riscv64 support with perlasm assembly generation

monkey-jsun:
Add riscv64 to openssl build using a patched overlay tarball with
constants-linux-riscv64.bzl.  Enable perlasm assembly generation
for riscv64 (same approach as x86_64/aarch64) using a relocatable
perl toolchain.  Post-process perlasm output to fix non-PIC JAL
instructions for shared library linking.

Note this PR depends on #2374


monkey-jsun:

PR #2365, opened 2026-04-09
toolchains_llvm: fix %workspace% prefix stripping in system_modulemap

monkey-jsun:
_is_hermetic_or_exists needs to strip the %workspace% prefix before
checking path existence, so include dirs like lib64/ or arch-specific
c++/v1 are filtered out when they don't exist on the host. This
is a generic bug affecting any toolchain_root usage with absolute paths.

PR #2376, opened 2026-04-10
rules_foreign_cc: add riscv64 platform support and fix pkg-config config.guess

monkey-jsun:
Add riscv64 to SUPPORTED_CPU and triplet_name in platform.bzl.
Pass --build explicitly to pkg-config configure when not cross-
compiling, so outdated config.guess (2015 vintage, bundled in
pkg-config 0.29.2) does not need to detect the host architecture.

Note this PR depends on #2365


monkey-jsun:

PR #2379, opened 2026-04-10
swiftshader: add riscv64 default condition to select() blocks

monkey-jsun:
The LLVM backend sources select() in BUILD.swiftshader.bazel only lists
x86_64 and aarch64 entries. On riscv64 Bazel fails with 'no matching
condition'. Add //conditions:default: [] so riscv64 falls through to
pure C/software rendering (SwiftShader's CPU renderer works without
arch-specific LLVM backends).

### DIFF
--- a/base/cvd/build_external/swiftshader/BUILD.swiftshader.bazel
+++ b/base/cvd/build_external/swiftshader/BUILD.swiftshader.bazel
@@ -1447,6 +1447,10 @@ cc_library(
             "third_party/llvm-16.0/llvm/lib/Target/X86/X86VZeroUpper.cpp",
             "third_party/llvm-16.0/llvm/lib/Target/X86/X86WinEHState.cpp",
         ],
+        "@platforms//cpu:riscv64": [
+            # No arch-specific LLVM JIT backend sources for riscv64.
+            # SwiftShader will build but JIT compilation will not be available.
+        ],
     }),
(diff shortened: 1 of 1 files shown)


base/cvd/build_external/swiftshader/BUILD.swiftshader.bazel
```diff
@@ -1447,6 +1447,11 @@ cc_library(
             "third_party/llvm-16.0/llvm/lib/Target/X86/X86VZeroUpper.cpp",
             "third_party/llvm-16.0/llvm/lib/Target/X86/X86WinEHState.cpp",
         ],
+        "//conditions:default": [
```
jmacnak: Could you make the condition `"@platforms//cpu:riscv64"`? I think it's nice to be explicit so that we are forced to check and update this as new hosts are used.

3405691582: Alternatively, make the comment not riscv specific (with the argument that a buildbreak should happen when there is a change that _must_ be made).

monkey-jsun: Changed to "@platforms//cpu:riscv64" per your suggestion.

jmacnak (approved):

Databean (approved):

PR #2380, opened 2026-04-10
gfxstream: include sys/syscall.h for memfd_create on riscv64

monkey-jsun:
SharedMemory_posix.cpp has hardcoded __NR_memfd_create syscall numbers
for x86/arm but not riscv64.  Including sys/syscall.h provides the
correct kernel-defined value, so the hardcoded fallback is skipped.

### DIFF
--- a/base/cvd/build_external/gfxstream/gfxstream.MODULE.bazel
+++ b/base/cvd/build_external/gfxstream/gfxstream.MODULE.bazel
@@ -8,5 +8,6 @@ git_repository(
     patches = [
         "@//build_external/gfxstream:PATCH.gfxstream.build_variables.patch",
         "@//build_external/gfxstream:bazel9.patch",
+        "@//build_external/gfxstream:riscv64_memfd_create.patch",
     ],
 )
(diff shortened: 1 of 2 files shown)


jmacnak:
@monkey-jsun can you confirm if https://github.com/google/gfxstream/pull/117 and https://github.com/google/android-cuttlefish/pull/2393 addresses this correctly?

monkey-jsun:
Yes.  That should take care of RISC-V build.

Now on a second thought, a more thorough fix is probably to remove all existing syscall num defs, which share the same root cause as riscv.  I will leave it to you to make the call.  See below.  Feel free to close the PR afterwards.

```
--- a/common/base/SharedMemory_posix.cpp
+++ b/common/base/SharedMemory_posix.cpp
@@ -23,21 +23,9 @@
 #include <fcntl.h>
 #include <sys/mman.h>
 #include <sys/stat.h>
+#if !defined(HAVE_MEMFD_CREATE)
+#include <sys/syscall.h>
+#endif
 #ifndef _MSC_VER
 #include <unistd.h>
 #endif

-#ifndef __NR_memfd_create
-#if __aarch64__
-#define __NR_memfd_create 279
-#elif __arm__
-#define __NR_memfd_create 279
-#elif __powerpc64__
-#define __NR_memfd_create 360
-#elif __i386__
-#define         __NR_memfd_create 356
-#elif __x86_64__
-#define __NR_memfd_create 319
-#endif
-#endif
-
 namespace gfxstream {
```

jmacnak:
Ack, will have to dig through some history and run some internal builds to double check why that was needed in the first place... will close this PR out.

PR #2382, opened 2026-04-10
xz: add riscv64 config to disable x86 intrinsics

monkey-jsun:
xz BUILD.bazel only has config settings for x86_64 and arm64.  On
riscv64 it falls through to //conditions:default which uses the x86_64
config defining HAVE_IMMINTRIN_H — but immintrin.h is x86-only.
Add a linux_riscv64 config_setting and expand_template that undefs
x86 intrinsics, mirroring the arm64 approach.


monkey-jsun:

PR #2387, opened 2026-04-10
Metrics refactors in preparation for incoming fetch metrics gathering

cjreynol:
Meant to separate and expand upon the refactors initially introduced in the current draft of #2290.  The expanded refactors are planned to support both:

- an improved design over that first draft of the fetch metrics gathering PR
- preparing for the remaining device metrics field gathering that will require more inputs from the various `cvd` operations

Bug: 491965176


base/cvd/cuttlefish/host/libs/metrics/metrics_orchestration.cc
```diff
@@ -110,26 +112,29 @@ Result<MetricsData> GatherMetrics(const MetricsPaths& metrics_paths,
...
-  CF_EXPECT_EQ(result.guest_metrics.size(), result.flag_metrics.size(),
-               "The gathered guest and flag metrics vectors must be equal, as "
-               "flags are per guest.");
+  if (!metrics_paths.guests.IsEmpty()) {
+    result.guest_metrics = CF_EXPECT(GetGuestMetrics(metrics_paths.guests));
+    result.flag_metrics =
+        CF_EXPECT(GetFlagMetrics(metrics_paths.guests.guest_infos.size()));
+    CF_EXPECT_EQ(
+        result.guest_metrics.size(), result.flag_metrics.size(),
+        "The gathered guest and flag metrics vectors must be equal, as "
```
jemoreira: nit: the vectors sizes are equal, not the vectors themselves.

base/cvd/cuttlefish/host/libs/metrics/metrics_writer.cc
```diff
@@ -41,7 +42,7 @@ std::string GenerateFilenameSuffix() {
 }  // namespace
 
 Result<void> WriteMetricsEvent(EventType event_type,
-                               const std::string& metrics_directory,
+                               const std::string_view& metrics_directory,
```
jemoreira: Pass std::string_view by value. `const` is almost always unnecessary too since the underlying string can't be modified through the string_view object.

base/cvd/cuttlefish/host/libs/metrics/metrics_writer.h
```diff
@@ -25,7 +25,7 @@
 namespace cuttlefish {
 
 Result<void> WriteMetricsEvent(
-    EventType event_type, const std::string& metrics_directory,
+    EventType event_type, const std::string_view& metrics_directory,
```
jemoreira: pass string_view by value

base/cvd/cuttlefish/host/libs/metrics/metrics_orchestration.h
```diff
@@ -28,7 +29,7 @@ namespace cuttlefish {
 
 struct MetricsInput {
   std::string metrics_directory;
-  Guests guests;
+  std::optional<Guests> guests;
```
jemoreira, line 32: Would it make sense to use different types instead of a single type with multiple `std::optional` fields? `GatherMetrics` could have overloads for each of the different types instead of a bunch of `if (field.is_present())` calls in the same body.

base/cvd/cuttlefish/host/libs/metrics/metrics_orchestration.cc
cjreynol: This was removed in a later commit when moving the `GetFlagMetrics` call into `GetGuestMetrics`, no longer needed.

But you were correct about the message leaving off "sizes" :)

base/cvd/cuttlefish/host/libs/metrics/metrics_writer.cc
cjreynol: Bad find-and-replacing, thanks for the catch.  Fixed.

base/cvd/cuttlefish/host/libs/metrics/metrics_writer.h
cjreynol: Bad find-and-replacing, thanks for the catch.  Fixed.

base/cvd/cuttlefish/host/libs/metrics/metrics_orchestration.h
cjreynol: That certainly makes sense.  The current plan is only to have one more conditional for the fetch input data.  Would you be okay if I just drop the `std::optional` and try the separate types/overloads when I introduce the fetch data in the follow-up PR to this?

jemoreira: Sounds good, but please rewrite the commit message to not say this change is to enable having multiple optionals in that struct.

jemoreira (approved):

base/cvd/cuttlefish/host/libs/metrics/metrics_orchestration.h
cjreynol: Done.

PR #2391, opened 2026-04-14
Gather data for fetch metrics (but do not transmit) V2

cjreynol:
(Updated version of #2290 after #2387)

Gather the data for fetch metrics and make it available in metrics_conversion.cc, ready to be added to the transmitted proto AFTER the new fields are approved.

Also, includes some changes to the existing device metrics gathering to account for a the new fetch metrics logic paths.

Bug: 491965176


base/cvd/cuttlefish/host/libs/metrics/fetch_metrics_orchestration.h
```diff
@@ -0,0 +1,33 @@
...
+#include "cuttlefish/host/commands/cvd/fetch/fetch_cvd_parser.h"
+
+namespace cuttlefish {
+
+void GatherFetchStartMetrics(const FetchFlags& fetch_flags);
+
+void GatherFetchCompleteMetrics(const std::string& target_directory,
+                                const FetchResult& fetch_result);
+
+void GatherFetchFailedMetrics(const std::string& target_directory);
```
jemoreira, line 31: The implementation of these two functions is nearly identical. Merge them into one and accept a `Result<FetchResult>` or `std::optional<FetchResult>` as parameter.

base/cvd/cuttlefish/host/libs/metrics/fetch_metrics.cc
```diff
@@ -0,0 +1,78 @@
...
+std::string FetchCompleteMetrics::ToEventLabel() const {
+  return "fetch_complete";
+}
+
+std::string FetchFailedMetrics::ToEventLabel() const { return "fetch_failed"; }
+
+std::string ToEventLabel(const FetchMetrics& fetch_metrics) {
+  return std::visit([](auto&& arg) { return arg.ToEventLabel(); },
+                    fetch_metrics);
+}
```
Databean: An "easier" way to write this is
```c++
struct ToEventLabel {
  std::string operator()(const FetchStartMetrics&) { return "fetch_start"; }
  std::string operator()(const FetchCompleteMetrics&) { return "fetch_complete"; }
  std::string operator()(const FetchFailedMetrics&) { return "fetch_failed"; }
};
```
Which can be invoked as

```c++
std::visit(ToEventLabel(), fetch_metrics);
```

([Compiler explorer](https://godbolt.org/z/sPW4Yodzf))

There's also the "overload pattern", shown in the example on [`std::variant::visit`](https://en.cppreference.com/w/cpp/utility/variant/visit) and provided [in abseil](https://github.com/abseil/abseil-cpp/blob/b9536c952c6f6bbcb37a5edb4fa88ee62d27cd7e/absl/functional/overload.h#L19). See also [go/totw/155](http://go/totw/155).

base/cvd/cuttlefish/host/libs/metrics/fetch_metrics.cc
```diff
@@ -0,0 +1,78 @@
...
+                      [](bool x) { return x; }),
+  };
+}
+
+Result<FetchMetrics> GetFetchMetrics(const FetchResult& fetch_result) {
+  std::vector<Builds> builds;
+  std::transform(fetch_result.fetch_artifacts.cbegin(),
+                 fetch_result.fetch_artifacts.cend(),
+                 std::back_inserter(builds),
+                 [](const FetchArtifacts& x) { return x.builds; });
```
Databean: It would be simpler to write
```c++
for (const FetchArtifacts& artifact : fetch_result.fetch_artifacts) {
  builds.push_back(artifact);
}
```

base/cvd/cuttlefish/host/libs/metrics/fetch_metrics.h
```diff
@@ -0,0 +1,60 @@
...
+#include "cuttlefish/host/commands/cvd/fetch/fetch_cvd_parser.h"
+#include "cuttlefish/result/result.h"
+
+namespace cuttlefish {
+
+using FetchInput = std::variant<FetchFlags, FetchResult>;
+
+struct FetchStartMetrics {
+  bool enable_local_caching = false;
+  bool dynamic_super_image_mixing = false;
```
Databean, line 36: Nothing actually uses the fields of the various `FetchMetrics` alternatives, which makes it difficult to judge if the `std::variant` type will later be necessary.

As written now, it would also be possible to define a superclass
```c++
class FetchMetrics {
public:
  virtual ~FetchMetrics() = default;

  virtual std::string ToEventLabel() const = 0;
};
```

and produce a `std::unique_ptr<FetchMetrics>` rather than a `std::variant` of the subtypes.

base/cvd/cuttlefish/host/libs/metrics/fetch_metrics_orchestration.h
Databean: Maybe something like
```c++
void ProcessMetricsInput(const MetricsInput& metrics_input) {
  Result<void> run_metrics_result = RunMetrics(metrics_input);
  if (!run_metrics_result.ok()) {
    VLOG(0) << fmt::format(
        "Failed during metrics gathering and (possible) outputting.  Error: {}",
        run_metrics_result.error());
    return;
  }
}
```
could be reused.

base/cvd/cuttlefish/host/libs/metrics/fetch_metrics_orchestration.cc
```diff
@@ -0,0 +1,141 @@
...
+#include "cuttlefish/host/libs/metrics/metrics_orchestration.h"
+#include "cuttlefish/result/result.h"
+
+namespace cuttlefish {
+namespace {
+
+struct FetchMetricsInput {
+  MetricsInput base_input;
+  std::optional<FetchInput> fetch_input;
+};
```
Databean: I'm not sure this type provides much value. What do you think of changing `GetFetchMetricsInput` to produce the `MetricsInput` type directly, and having `GatherFetchMetrics` and `RunMetrics` accept two parameters instead?

base/cvd/cuttlefish/host/libs/metrics/fetch_metrics.cc
cjreynol: Done.

base/cvd/cuttlefish/host/libs/metrics/fetch_metrics.h
cjreynol: From our discussion, the goal of the variant is to pass to `metrics_conversion` and have it generate different `oneOf` proto messages based on which variant version is provided.  The goal is to keep the proto generating logic contained in `metrics_conversion`, and this accomplishes that.  Leaving it as is.

base/cvd/cuttlefish/host/libs/metrics/fetch_metrics.cc
cjreynol: Done.  I think this has more mental overhead and requires more variant knowledge to read, but it is nice and condensed.

base/cvd/cuttlefish/host/libs/metrics/fetch_metrics_orchestration.h
cjreynol: From Cody's comment, b/503481886 for simplifying these in a way that works generically for other cases in the codebase as well.

I would prefer to not combine them, because I want the public interface to be explicit about what case is occurring at the callsites.  My concern for having a single external interface for both "Complete" and "Failed" would be in a future update we start accidentally sending the wrong data because of an argument change.  For example, it would be unclear if the existence of a `FetchResult` was the determining factor between the metrics logic transmitting the success or failure type.

If we merge them, make that a private helper function, and then create two public interface functions, we are not really saving any lines of code or simplifying after the above linked bug is handled to reduce the conditional into one call.

base/cvd/cuttlefish/host/libs/metrics/fetch_metrics_orchestration.cc
cjreynol: Done.

base/cvd/cuttlefish/host/libs/metrics/metrics_orchestration.h
```diff
@@ -16,26 +16,28 @@
...
 #include <string_view>
 
 #include "cuttlefish/common/libs/utils/tee_logging.h"
-#include "cuttlefish/host/libs/metrics/device_event_type.h"
 #include "cuttlefish/host/libs/metrics/metrics_conversion.h"
 #include "cuttlefish/result/result.h"
 
 namespace cuttlefish {
 
+inline std::string_view kMetricsEnabledNotice =
```
Databean: nit: `inline constexpr`: https://abseil.io/tips/168

base/cvd/cuttlefish/host/libs/metrics/device_metrics_orchestration.cc
```diff
@@ -61,11 +66,14 @@ std::vector<GuestInfo> GetGuestInfos(
   return result;
 }
 
-Result<MetricsInput> GetInstanceGroupMetricsInput(
+Result<DeviceMetricsInput> GetDeviceMetricsInput(
```
Databean, line 69: Does this need to return a `Result`? It doesn't look like it can ever fail.

Databean (approved):

base/cvd/cuttlefish/host/libs/metrics/device_metrics_orchestration.cc
cjreynol: Yes, line 81 calls `CF_EXPECT(GetParsedFlags())` to call all of the `FlagBase` flag instances can fail.

base/cvd/cuttlefish/host/libs/metrics/metrics_orchestration.h
cjreynol: Done.

PR #2392, opened 2026-04-14
Fix redundant get() call on smart pointer in oauth2_consent.cpp

Databean:
Remove redundant get() call on smart pointer, as recommended by ClangTidyReadability.

Port of cl/898040821

 #gemini


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #2395, opened 2026-04-14
Added host orchestrator endpoint for cvd status

SuperStrongDinosaur:
b/502498047

### DIFF
--- a/frontend/src/host_orchestrator/api/v1/messages.go
+++ b/frontend/src/host_orchestrator/api/v1/messages.go
@@ -214,3 +214,21 @@ type DisplayScreenshotResponse struct {
 type ListScreenRecordingsResponse struct {
 	ScreenRecordings []string `json:"screen_recordings"`
 }
+
+type CVDInstanceStatus struct {
+	InstanceName string   `json:"instance_name"`
+	Status       string   `json:"status"`
+	Displays     []string `json:"displays,omitempty"`
+	AssemblyDir  string   `json:"assembly_dir"`
+	InstanceDir  string   `json:"instance_dir"`
+	// Note: The web_access field returned by `cvd status` is intentionally omitted here.
+	// It uses 'localhost' as the address, which is only valid on the machine where cvd ran.
+	// Since the host orchestrator is accessed remotely, that URL is not useful to clients.
+	WebRTCDeviceID string `json:"webrtc_device_id,omitempty"`
+	ADBPort        int    `json:"adb_port,omitempty"`
+	ADBSerial      string `json:"adb_serial,omitempty"`
+}
+
+type CVDStatusResponse struct {
+	Status []*CVDInstanceStatus `json:"status"`
+}
--- a/frontend/src/host_orchestrator/orchestrator/controller.go
+++ b/frontend/src/host_orchestrator/orchestrator/controller.go
@@ -79,6 +79,8 @@ func (c *Controller) AddRoutes(router *mux.Router) {
 	router.Handle("/cvds/{group}/:bugreport",
 		httpHandler(newCreateCVDBugReportHandler(c.Config, c.OperationManager))).Methods("POST")
 	router.Handle("/cvds/{group}/{name}",
+		httpHandler(newGetCVDInstanceStatusHandler(c.Config, c.OperationManager))).Methods("GET")
+	router.Handle("/cvds/{group}/{name}",
 		httpHandler(newExecCVDGroupCommandHandler(c.Config, c.OperationManager, &removeCvdCommand{}))).Methods("DELETE")
 	router.Handle("/cvds/{group}/{name}/:start",
 		httpHandler(newStartCVDHandler(c.Config, c.OperationManager))).Methods("POST")
@@ -936,6 +938,25 @@ func okHandler() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusOK)
 	})
+}
+
+type getCVDInstanceStatusHandler struct {
+	Config Config
+	OM     OperationManager
+}
+
+func newGetCVDInstanceStatusHandler(c Config, om OperationManager) *getCVDInstanceStatusHandler {
+	return &getCVDInstanceStatusHandler{Config: c, OM: om}
+}
+
+func (h *getCVDInstanceStatusHandler) Handle(r *http.Request) (interface{}, error) {
+	vars := mux.Vars(r)
+	opts := CVDInstanceStatusActionOpts{
+		Selector:         cvd.InstanceSelector{GroupName: vars["group"], Name: vars["name"]},
+		OperationManager: h.OM,
+		ExecContext:      exec.CommandContext,
+	}
+	return NewCVDInstanceStatusAction(opts).Run()
 }
 
 func getFetchCredentials(config BuildAPICredentialsConfig, r *http.Request) cvd.FetchCredentials {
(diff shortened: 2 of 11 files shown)


frontend/src/host_orchestrator/api/v1/messages.go
```diff
@@ -212,3 +212,7 @@ type DisplayScreenshotResponse struct {
 type ListScreenRecordingsResponse struct {
 	ScreenRecordings []string `json:"screen_recordings"`
 }
+
+type CVDStatusResponse struct {
+	Status map[string]interface{} `json:"status"`
```
jemoreira, line 217: Please use a more specific field definition than `interface{}`.

frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -68,6 +68,10 @@ func (c *Controller) AddRoutes(router *mux.Router) {
 		httpHandler(newCreateCVDHandler(c.Config, c.OperationManager, c.UserArtifactsManager))).Methods("POST")
 	router.Handle("/cvds", httpHandler(&listCVDsHandlerAll{Config: c.Config})).Methods("GET")
 	router.Handle("/cvds/{group}", httpHandler(&listCVDsHandler{Config: c.Config})).Methods("GET")
+	router.Handle("/cvds/{group}/:status",
+		httpHandler(newGetCVDGroupStatusHandler(c.Config, c.OperationManager))).Methods("GET")
+	router.Handle("/cvds/{group}/{name}/:status",
+		httpHandler(newGetCVDInstanceStatusHandler(c.Config, c.OperationManager))).Methods("GET")
```
jemoreira, line 74: The information returned by `GET /cvds/{group}/status/:status` is the same as that returned by `GET /cvds/{group}`. The only difference is that the latter (inconveniently) wraps it inside an array with a single element in a property named `cvds`.

Similarly, a `GET /cvds/{group}/{name}` endpoint could exist that returns the information related to a single instance, but it doesn't need the `/:status` part.

frontend/src/host_orchestrator/orchestrator/cvd/cvd.go
```diff
@@ -370,14 +370,14 @@ func (i *Instance) ADBSerial() string {
 	return i.instance.ADBSerial
 }
 
-func (i *Instance) CVDStatus() (map[string]interface{}, error) {
+func (i *Instance) CVDStatus() (interface{}, error) {
```
jemoreira, line 373: Don't create a new commit to fix the previous two, fix the corresponding commit instead. Unless you're addressing code review feedback, in which case new commits are actually desired, but you're still expected to combine those with the original ones before merging.

SuperStrongDinosaur: Acknowledged. I've addressed the feedback in my latest changes and will be sure to squash all these fixups into the original commits before we merge this to keep the commit history clean

frontend/src/host_orchestrator/api/v1/messages.go
SuperStrongDinosaur: Done. I've replaced interface{} with a yped CVDInstanceStatus struct in messages.go to explicitly define the expected fields

frontend/src/host_orchestrator/orchestrator/controller.go
SuperStrongDinosaur: Addressed. I've removed the redundant GET /cvds/{group}/status/:status endpoint entirely. I have also renamed the instance-specific endpoint to GET /cvds/{group}/{name} as suggested

frontend/src/host_orchestrator/orchestrator/controller_test.go
```diff
@@ -53,7 +53,7 @@ func TestGetCVDLogsIsHandled(t *testing.T) {
 	dir := orchtesting.TempDir(t)
 	defer orchtesting.RemoveDir(t, dir)
 	rr := httptest.NewRecorder()
-	req, err := http.NewRequest("GET", "/cvds/cvd-1/logs", strings.NewReader("{}"))
+	req, err := http.NewRequest("GET", "/cvds/cvd-1/cvd-1/logs", strings.NewReader("{}"))
```
SuperStrongDinosaur, line 56: Changing the test didn't fix a backward compatibility break. It fixed a typo in the test to match the API contract

frontend/src/host_orchestrator/orchestrator/cvd/cvd.go
```diff
@@ -383,6 +383,12 @@ func (i *Instance) ADBPort() uint32 {
 	return i.instance.ADBPort
 }
 
+func (i *Instance) CVDStatus() ([]byte, error) {
```
jemoreira: The function needs to parse the command output into an object. Similar to how `ListDisplays` does it.

frontend/src/host_orchestrator/orchestrator/cvd/cvd.go
```diff
@@ -551,6 +557,12 @@ func (g *Group) Remove() error {
 	return err
 }
 
+func (g *Group) CVDStatus() ([]byte, error) {
```
jemoreira: Same comment about returning an object, not `[]byte`

frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -69,6 +69,8 @@ func (c *Controller) AddRoutes(router *mux.Router) {
 	router.Handle("/cvds", httpHandler(&listCVDsHandlerAll{Config: c.Config})).Methods("GET")
 	router.Handle("/cvds/{group}", httpHandler(&listCVDsHandler{Config: c.Config})).Methods("GET")
 	router.PathPrefix("/cvds/{group}/{name}/logs").Handler(&getCVDLogsHandler{Config: c.Config}).Methods("GET")
+	router.Handle("/cvds/{group}/{name}",
```
jemoreira: nit: Move this to line 83, just before the DELETE with the same url path as this one.

frontend/src/host_orchestrator/orchestrator/cvd/cvd.go
SuperStrongDinosaur: CVDStatus() now parses the command output and returns an output.Status object rather than a raw []byte

frontend/src/host_orchestrator/orchestrator/controller.go
SuperStrongDinosaur: moved

frontend/src/host_orchestrator/orchestrator/cvd/cvd.go
SuperStrongDinosaur: same fix

frontend/src/host_orchestrator/api/v1/messages.go
```diff
@@ -214,3 +214,19 @@ type DisplayScreenshotResponse struct {
...
 	ScreenRecordings []string `json:"screen_recordings"`
 }
+
+type CVDInstanceStatus struct {
+	InstanceName   string   `json:"instance_name"`
+	Status         string   `json:"status"`
+	Displays       []string `json:"displays,omitempty"`
+	AssemblyDir    string   `json:"assembly_dir"`
+	InstanceDir    string   `json:"instance_dir"`
+	WebAccess      string   `json:"web_access,omitempty"`
```
jemoreira: The `web_access` field returned by `cvd status` is only useful from the same machine where `cvd` ran (uses localhost as the address). The host orchestrator is intended to be accessed remotely, so that URL is not valid.
Remove the field from this struct and keep it `Instance` only. Leave a comment here explaining why the field is ommitted.

jemoreira (commented):
Only one minor comment left. Please squash the extra commits into the ones you actually want to be submitted and I'll approve and merge the PR.

frontend/src/host_orchestrator/api/v1/messages.go
SuperStrongDinosaur: I have removed the WebAccess field from CVDInstanceStatus and added a comment explaining why it is omitted

SuperStrongDinosaur:
> Only one minor comment left. Please squash the extra commits into the ones you actually want to be submitted and I'll approve and merge the PR.

Reorganized commits into a logical functional split

jemoreira (approved):

PR #2398, opened 2026-04-14
Replace libbase's ReadFully and WriteFully

jemoreira:
android::base::ReadFully -> cuttlefish::ReadExact
android::base::WriteFully -> cuttlefish::WriteAll

Bug: 502682796

### DIFF
--- a/base/cvd/cuttlefish/host/libs/image_aggregator/image_from_file.cc
+++ b/base/cvd/cuttlefish/host/libs/image_aggregator/image_from_file.cc
@@ -21,24 +21,24 @@
 #include <string>
 #include <utility>
 
-#include <android-base/file.h>
 #include "absl/strings/match.h"
 
+#include "cuttlefish/common/libs/fs/shared_fd.h"
+#include "cuttlefish/common/libs/fs/shared_buf.h"
 #include "cuttlefish/common/libs/utils/files.h"
 #include "cuttlefish/host/libs/image_aggregator/composite_disk.h"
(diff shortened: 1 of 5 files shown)


base/cvd/cuttlefish/host/libs/image_aggregator/image_from_file.cc
```diff
@@ -47,11 +47,11 @@ Result<std::unique_ptr<DiskImage>> ImageFromFile(const std::string& file_path) {
   constexpr uint64_t MAGIC_BLOCK_SIZE = 4096;
   std::string magic(std::min(file_size, MAGIC_BLOCK_SIZE), '\0');
 
-  CF_EXPECTF(android::base::ReadFully(fd, magic.data(), magic.size()),
-             "Failed to read '{}': {}'", file_path, StrError(errno));
+  CF_EXPECTF(ReadExact(fd, &magic) > 0,
```
Databean, line 50: There is a behavioral difference here that if `ReadExact` hits an EOF it will return the number of bytes read, while `ReadFully` considers EOF to be an error case.

Databean (approved):

base/cvd/cuttlefish/host/libs/image_aggregator/image_from_file.cc
jemoreira: The code already takes that into account by making `magic` at most `file_size` bytes long, apparently it's not supposed to be an error.

PR #2401, opened 2026-04-15
Arm64 Server 1.51 hotfix

rmuthiah:
The arm64 servers don't boot with the GPU enabled if Nvidia backports are used. Disable backports.


ser-io (approved):

grandpaul (approved):
I've checked the ISO file generated by the workflow. The preseed.cfg does disable the backports usage.

jemoreira (approved):

0405ysj:
Do we need to submit this hotfix either, based on conversation thread at https://github.com/google/android-cuttlefish/pull/2402? We observed an issue for installing on-premise server image(versioned 1.51.0) on our-side server, and are speculating this PR is required.

PR #2402, opened 2026-04-15
Arm64 1.50 hotfix

rmuthiah:


ser-io (approved):

grandpaul (approved):
I've checked the ISO file generated by the workflow. The preseed.cfg does disable the backports usage.

jemoreira (approved):
Do these need to be backported to 1.51 too?

rmuthiah:
> Do these need to be backported to 1.51 too?

https://github.com/google/android-cuttlefish/pull/2401 yup.

PR #2403, opened 2026-04-15
Use absl logging instead of android-base logging

jmacnak:
Bug: b/502947484
Test: bazel build cuttlefish/host/frontend/webrtc/libcommon/...


jemoreira (approved):

jmacnak:
(need to revisit after b/502947484)

PR #2405, opened 2026-04-15
Add some utilities for PATH searching.

3405691582:
A lot of our tooling relies on executing binaries found on the host system. The subprocess library ends up calling execvpe on Linux, which means that PATH resolution happens for the executable.

Two issues present themselves. Firstly, execvpe is a Linuxism -- it is not in the SUS; our tentative non-Linux (macOS) implementation calles execve instead.

Secondly and more importantly, if the executable that we expect to be able to run is not in PATH, then we will fail to exec it, even if the actual executable is installed; this can happen on some distributions where nonprivileged users are configured by default to not have executables where root privileges are required in PATH.

Therefore, we need to be able to perform PATH resolution ourselves, so that we can do things like test if a binary truly isn't available or whether it merely isn't in the current PATH.

We add a reminder for ourselves too that while the usual colon-separated PATH splitting makes perfect sense for SUS systems, it is not the case on platforms like Windows -- if in the future we want to go down that direction.

(Also, perhaps we should just migrate to using posix_spawn, but that's not important right now.)

Bug: 488432653

### DIFF
--- a/base/cvd/cuttlefish/common/libs/utils/files_test.cpp
+++ b/base/cvd/cuttlefish/common/libs/utils/files_test.cpp
@@ -21,6 +21,8 @@
 #include <android-base/file.h>
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
+#include "absl/cleanup/cleanup.h"
+#include "absl/strings/str_cat.h"
 
 #include "cuttlefish/common/libs/utils/files_test_helper.h"
 #include "cuttlefish/result/result.h"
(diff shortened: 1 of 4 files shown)


base/cvd/cuttlefish/common/libs/utils/files_test.cpp
```diff
@@ -180,4 +182,50 @@ INSTANTIATE_TEST_SUITE_P(
...
 
+TEST(FilesTest, PathWithCustomEnv) {
+  const std::string env_name = "TEST_PATH";
+  const std::string dir1 = "/foo/bar";
+  const std::string dir2 = "/baz";
+  const std::string env_value = absl::StrCat(dir1, ":", dir2);
+  setenv(env_name.c_str(), env_value.c_str(), 1);
+  absl::Cleanup cleanup = [&env_name]() { unsetenv(env_name.c_str()); };
+
+  auto result = Path(env_name);
```
Databean: nit: `std::vector<std::string>` following https://abseil.io/tips/232

base/cvd/cuttlefish/common/libs/utils/files_test.cpp
```diff
@@ -180,4 +182,50 @@ INSTANTIATE_TEST_SUITE_P(
...
+  const std::string temp_dir = testing::TempDir();
+  const std::string file_name = "search_test_file";
+  const std::string full_path = absl::StrCat(temp_dir, "/", file_name);
+
+  EXPECT_TRUE(android::base::WriteStringToFile("", full_path));
+  absl::Cleanup cleanup = [full_path]() { unlink(full_path.c_str()); };
+
+  auto result = Search({temp_dir}, file_name);
+  ASSERT_TRUE(result.ok());
+  EXPECT_EQ(*result, full_path);
```
Databean: [`result_matchers.h`](https://github.com/google/android-cuttlefish/blob/01ac0a549e031fd272dc1b04bcc31cc6e8fff710/base/cvd/cuttlefish/result/result_matchers.h#L4) provides some utilities that can be used like

```c++
EXPECT_THAT(result, IsOkAndValue(full_path));
```

Databean (approved):
I'm not sure we can use `posix_spawn` because we do a bunch of work between the `fork` and `exec` steps, though some of that is also linux-specific.

Definitely regret the change to use `execvpe` over `execve`, which came in [aosp/1584808](https://android-review.googlesource.com/c/device/google/cuttlefish/+/1584808), and would prefer to use `execve` again though I don't know how much other fallout it would cause.

base/cvd/cuttlefish/common/libs/utils/files_test.cpp
3405691582: Done.

base/cvd/cuttlefish/common/libs/utils/files_test.cpp
3405691582: Done.

PR #2423, opened 2026-04-18
Add CpioReader for reading CPIO archives

Databean:
Bug: b/503906646
Assisted-by: Gemini Next:latest

### DIFF
--- a/base/cvd/cuttlefish/io/cpio.cc
+++ b/base/cvd/cuttlefish/io/cpio.cc
@@ -1,0 +1,290 @@
+//
+// Copyright (C) 2026 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "cuttlefish/io/cpio.h"
+
+#include <endian.h>
+#include <string.h>
+
+#include <charconv>
+#include <string_view>
+
+#include "cuttlefish/common/libs/utils/size_utils.h"
+#include "cuttlefish/io/read_exact.h"
+#include "cuttlefish/io/read_window_view.h"
+#include "cuttlefish/result/expect.h"
+
+// For the CPIO file format specification, see:
+// https://github.com/libyal/dtformats/blob/c55446a4369149bb109bf783f29a2838c7328718/documentation/Copy%20in%20and%20out%20(CPIO)%20archive%20format.asciidoc
+
+namespace cuttlefish {
+
+namespace {
+
+constexpr std::string_view kMagicNewc1 = "070701";
+constexpr std::string_view kMagicNewc2 = "070702";
+constexpr std::string_view kMagicOdc = "070707";
+constexpr std::string_view kTrailerName = "TRAILER!!!";
+constexpr std::string_view kBinaryLeMagic = "\xC7\x71";
+constexpr std::string_view kBinaryBeMagic = "\x71\xC7";
+
+// None of these members are null-terminated.
+struct CpioNewcHeader {
+  char magic[6];
+  char ino[8];
+  char mode[8];
+  char uid[8];
+  char gid[8];
+  char nlink[8];
+  char mtime[8];
+  char filesize[8];
+  char maj[8];
+  char min[8];
+  char rmaj[8];
+  char rmin[8];
+  char namesize[8];
+  char chksum[8];
+  char filename[];
+};
+
+// None of these members are null-terminated.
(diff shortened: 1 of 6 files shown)


base/cvd/cuttlefish/io/cpio.cc
```diff
@@ -0,0 +1,237 @@
...
+      CF_EXPECT(Parse(*reader));
+  return std::unique_ptr<CpioReader>(
+      new CpioReader(std::move(reader), std::move(entries)));
+}
+
+CpioReader::CpioReader(std::unique_ptr<ReaderSeeker> reader,
+                       std::map<std::string, FileEntry, std::less<>> entries)
+    : reader_(std::move(reader)), entries_(std::move(entries)) {}
+
+Result<std::map<std::string, CpioReader::FileEntry, std::less<>>>
```
3405691582: Worth a typedef of the map type?

base/cvd/cuttlefish/io/cpio.cc
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  std::string_view magic_view(magic, 6);
+  if (magic_view == kMagicNewc1 || magic_view == kMagicNewc2) {
+    return ParseNewc(reader);
+  } else if (magic_view == kMagicOdc) {
+    return ParseOdc(reader);
+  } else {
+    return CF_ERRF("Unknown cpio magic: {}", magic_view);
+  }
```
3405691582: I think the flow of this is a little difficult to reason about on first glance. What about something like this. First:

```
   constexpr std::string_view kBinaryLEMagic = "\xc7\x71";
   constexpr std::string_view kBinaryBEMagic = "\x71\xc7";
```

```suggestion
  std::string_view magic_view(magic, 6);
  if (magic_view.starts_with(kBinaryLEMagic)) {
    return ParseBin(reader, /* file_is_big_endian= */ false);
  } else if (magic_view.starts_with(kBinaryBEMagic)) {
    return ParseBin(reader, /* file_is_big_endian= */ true);
  } else if (magic_view == kMagicNewc1 || magic_view == kMagicNewc2) {
    return ParseNewc(reader);
  } else if (magic_view == kMagicOdc) {
    return ParseOdc(reader);
  } else {
    return CF_ERRF("Unknown cpio magic: {}", magic_view);
  }
```

We do abuse the string_view a little for non-ASCII content, though, but I think it's fine?

base/cvd/cuttlefish/io/cpio.cc
```diff
@@ -0,0 +1,237 @@
...
+    offset = AlignToPowerOf2(offset, 2);
+
+    char header[110];
+    CF_EXPECT(PReadExact(reader, header, sizeof(header), offset));
+
+    std::string_view magic(header, 6);
+    CF_EXPECT(magic == kMagicNewc1 || magic == kMagicNewc2,
+              "Invalid magic in header");
+
+    uint32_t filesize = CF_EXPECT(ParseHex(header + 54, 8));
```
3405691582: (I don't love the direct offset indexing like this. It might be better to declare an updating pointer or even declare a struct to overlay here, but those are significant changes.)

base/cvd/cuttlefish/io/cpio.cc
```diff
@@ -0,0 +1,237 @@
...
+}
+
+Result<uint32_t> ParseOctal(const char* buf, size_t size) {
+  uint32_t val = 0;
+  std::from_chars_result result = std::from_chars(buf, buf + size, val, 8);
+  CF_EXPECT(result.ec == std::errc(), "Failed to parse octal value");
+  return val;
+}
+
+uint16_t Read16(const char* ptr, bool file_is_big_endian) {
```
3405691582: I haven't checked, but would this work if we used a `const char[4]` instead of a `const char *`? The other point is whether to just use `uint8_t` instead of `char`, to emphasize that we don't expect a const null-terminated C string here. Same for `Read32`.

base/cvd/cuttlefish/io/cpio.cc
Databean: I misremembered part of the style guide:

https://google.github.io/styleguide/cppguide.html#Aliases

> Don't put an alias in your public API just to save typing in the implementation; do so only if you intend it to be used by your clients.

... but this wouldn't be a public alias, so

> local convenience aliases are fine in function definitions, private sections of classes, explicitly-marked internal namespaces, and in .cc files:

so added a private alias.

base/cvd/cuttlefish/io/cpio.cc
Databean: Thanks, this looks better. Used `std::string` since it needs to read the data into somewhere.

base/cvd/cuttlefish/io/cpio.cc
Databean: Declared structs for all of the `Parse` variants, and used `offsetof` to reference the members to do unaligned reads from the packed structs.

base/cvd/cuttlefish/io/cpio.cc
Databean: Changed to use `const uint8_t[2]` and `const uint8_t[4]`.

Databean:
Thanks for reviewing!

3405691582 (approved):

PR #2428, opened 2026-04-21
Metrics gathering for cvd invocation sources

cjreynol:
Gathering, but not yet converting and transmitting, metrics field for `cvd` invocation sources and environments devices are run in.

Bug: 456483667


cjreynol:
Closed in favor of #2471

PR #2429, opened 2026-04-21
Hide irrelevant commands in cvd help by default

Databean:
Only show relevant commands and selector options when no devices are running. All commands are shown if devices are running, with device-specific commands displayed first.

Output without a running device:
================================
```
Cuttlefish Virtual Device (CVD) CLI.

usage: cvd <selector/driver options> <command> <args>

Driver Options:
  -help                  Print this message
  -verbosity=<LEVEL>     Adjust Cvd verbosity level. LEVEL is Android log
                         severity. (Required: cvd >= v1.3)

Commands (cvd help <command> for more information):

        cache - Manage the files cached by cvd

        create - Create a Cuttlefish virtual device or environment

        fetch, fetch_cvd - Retrieve build artifacts based on branch and target names

        help - Used to display help information for other commands

        lint - error checks the input virtual device json config file
        load - Loads the given JSON configuration file and launches devices based on the options provided

        login - Acquire credentials

        reset - Used to stop devices, optionally clean up instance files, and shut down the deprecated cvd server process

        version - Prints version of cvd client and cvd server
```

Output with a running device:
=============================
```
Cuttlefish Virtual Device (CVD) CLI.

usage: cvd <selector/driver options> <command> <args>

Driver Options:
  -help                  Print this message
  -verbosity=<LEVEL>     Adjust Cvd verbosity level. LEVEL is Android log
                         severity. (Required: cvd >= v1.3)

Commands (cvd help <command> for more information):

        cache - Manage the files cached by cvd

        create - Create a Cuttlefish virtual device or environment

        fetch, fetch_cvd - Retrieve build artifacts based on branch and target names

        help - Used to display help information for other commands

        lint - error checks the input virtual device json config file

        load - Loads the given JSON configuration file and launches devices based on the options provided

        login - Acquire credentials

        reset - Used to stop devices, optionally clean up instance files, and shut down the deprecated cvd server process

        version - Prints version of cvd client and cvd server

Selector Options:
  -group_name <name>     Specify the name of the instance group created
                         or selected.
  -instance_name <name>  Selects the device of the given name to perform the
                         commands for.
  -instance_name <names> Takes the names of the devices to create within an
                         instance group. The 'names' is comma-separated.

Device-Specific Commands (cvd help <command> for more information):
        display - Enables hotplug/unplug of displays from running cuttlefish virtual devices

        env - Enumerate + Query APIs for all gRPC services made available by this virtual device instance

        fleet - lists active devices with relevant information

        clear - Clears the instance database, stopping any running instances first.

        bugreport, host_bugreport, cvd_host_bugreport - Run cvd bugreport --help for command description

        stop, stop_cvd - Stop all instances in a group

        powerbtn - Trigger power button event on the device

        powerwash - Reset device to first boot state

        restart - Restart device

        remove, rm - Remove devices and artifacts from the system.

        screen_recording - Record screen contents

        suspend, resume, snapshot_take - Suspend/resume the cuttlefish device, or take snapshot of the device

        start, launch_cvd - Start a Cuttlefish virtual device or environment

        status, cvd_status - Query status of a single instance group.  Use `cvd fleet` for all devices
```

Assisted-by: Gemini:Next
Bug: b/504758273


base/cvd/cuttlefish/host/commands/cvd/cli/commands/help.h
```diff
@@ -22,7 +22,9 @@
 #include "cuttlefish/host/commands/cvd/cli/commands/command_handler.h"
 
 namespace cuttlefish {
+class InstanceManager;
```
cjreynol: Should be an include instead:

```cpp
#include "cuttlefish/host/commands/cvd/instances/instance_manager.h"
```

base/cvd/cuttlefish/host/commands/cvd/cli/commands/help.cpp
```diff
@@ -109,20 +120,45 @@ class CvdHelpHandler : public CvdCommandHandler {
...
+    if (devices_running) {
+      help_message << "\nDevice-Specific Commands (cvd help <command> for more information):\n";
+      for (const auto& handler : request_handlers_) {
+        if (handler->RequiresRunningDevice()) {
+          CF_EXPECT(PrintHandler(help_message, *handler));
+        }
+      }
+    }
+
+    help_message << kDriverOptionsText << std::endl;
```
cjreynol: I get splitting the previous `kHelpIntroText` variable into components, but can we still output this after the selector text (or before, if that is easier because of the `devices_running` check)?  In my opinion guiding users towards `cvd help <command>` in particular is highly valuable, if they are unaware that option exists.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/help.cpp
```diff
@@ -109,20 +120,45 @@ class CvdHelpHandler : public CvdCommandHandler {
 
   Result<std::string> TopLevelHelp() {
     std::stringstream help_message;
-    help_message << kHelpIntroText << std::endl;
+    help_message << kHelpIntroText;
+    bool devices_running = CF_EXPECT(instance_manager_.HasInstanceGroups());
```
cjreynol: NIT: `const`

base/cvd/cuttlefish/host/commands/cvd/cli/commands/help.cpp
```diff
@@ -109,20 +120,45 @@ class CvdHelpHandler : public CvdCommandHandler {
...
+    if (devices_running) {
+      help_message << kSelectorOptionsText;
+    }
+
+    if (devices_running) {
```
cjreynol: Combine these identical conditionals on 125 and 129.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/help.cpp
```diff
@@ -109,20 +120,45 @@ class CvdHelpHandler : public CvdCommandHandler {
...
+      for (const auto& handler : request_handlers_) {
+        if (handler->RequiresRunningDevice()) {
+          CF_EXPECT(PrintHandler(help_message, *handler));
+        }
+      }
+    }
+
+    help_message << kDriverOptionsText << std::endl;
+
     for (const auto& handler : request_handlers_) {
```
cjreynol: If we move the `kDriverOptionsText` output, there should be a horizontal rule or header inserted before these commands to visually separate the two groups.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/help.cpp
```diff
@@ -66,15 +72,20 @@ Example usage:
   cvd help <command> - displays more detailed help for the specific command
 )";
 
-constexpr char kIgnorableHandlerCommand[] = "experimental";
+const std::vector<std::string> kIgnoredCommandLists = {
+    "experimental",
```
cjreynol: These were deleted in https://github.com/google/android-cuttlefish/pull/714, so we can remove this.

I think it is worth outputting `kill-server` (or deleting it) and removing the logic around `kIgnoredCommandLists` for simplicity.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/help.cpp
```diff
@@ -109,20 +120,45 @@ class CvdHelpHandler : public CvdCommandHandler {
...
     }
+
     return help_message.str();
   }
 
+  Result<void> PrintHandler(
+      std::stringstream& help_message,
+      const CvdCommandHandler& handler) const {
+    std::string command_list = absl::StrJoin(handler.CmdList(), ", ");
+    if (!command_list.empty() &&
```
cjreynol: Manually looking, it appears that `CvdCmdListHandler` is the only command that has an empty command list.  And even that handler actually has a value, because its `CanHandle` implementation is checking if the subcommand value is `cmd-list`.

Like the `kIgnoredCommandsList` comment, I think we should update that command to appear and simplify our logic here a bit.

Then go forward with the expectation of no longer having hidden `cvd` APIs.  We can always mark that command as "unstable" or "not for users" in the help text, but if it is accessible we might as well output it.  Nothing stops users from becoming dependent on it as-is, at least if we are explicit about it and we make a change we have something to point to saying "we told you not to use this".

base/cvd/cuttlefish/host/commands/cvd/cli/commands/command_handler.h
```diff
@@ -35,6 +35,7 @@ class CvdCommandHandler {
   virtual cvd_common::Args CmdList() const = 0;
   // used for command help text
   virtual Result<std::string> SummaryHelp() const = 0;
+  virtual bool RequiresRunningDevice() const;
```
cjreynol: These commands do not require a running device, for instance `cvd start` works as long as devices exist.

Should be renamed, maybe `RequiresExistingDevice`?  Your choice.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/help.h
Databean: Done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/help.cpp
Databean: Done.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/help.cpp
Databean: I couldn't find any users for `cmd-list` so I deleted it.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/help.cpp
Databean: Removed this variable.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/help.cpp
Databean: Done.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/help.cpp
Databean: Switched it back to print non-device-specific help and commands before the device-specific ones.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/help.cpp
Databean: Rearranged the sections so there is an implicit divider.

Databean (commented):
Thanks for reviewing!

base/cvd/cuttlefish/host/commands/cvd/cli/commands/command_handler.h
Databean: Renamed to `RequiresDeviceExists`.

cjreynol (approved):

jemoreira:
I don't think it's a good idea to show a subset of the commands on `cvd -help`. I sometimes run man/help on commands in one terminal and leave it open for reference while I use it elsewhere, that flow would be broken by this behavior.

I like the grouping of subcommands shown when devices are running. I would additionally change the order of the subcommands to put the more useful ones first or just sort them alphabetically.

Also, I know it's not related to this change, but what is a "driver" in the context of cvd? We should probably change that word for something else, like "global options".

PR #2431, opened 2026-04-21
Add a handler to the WebRTC operator for context.

3405691582:
We may be running in environments where only the WebRTC operator is exposed. This can make it difficult to verify information about that environment; for example, is the operator (and cvd, etc.) running in a container, or not? One simple way to do this is to expose an endpoint on the operator for debugging purposes, and one simple way to get information about the environment is to expose the hostname.

For now, we just expose the hostname, but it may be worthwhile adding extra fields in the future.

Bug: 503333785

### DIFF
--- a/frontend/src/liboperator/operator/operator.go
+++ b/frontend/src/liboperator/operator/operator.go
@@ -174,7 +174,21 @@ func CreateHttpHandlers(
 	router.HandleFunc("/infra_config", func(w http.ResponseWriter, r *http.Request) {
 		ReplyJSONOK(w, config)
 	}).Methods("GET")
+	router.HandleFunc("/context", func(w http.ResponseWriter, r *http.Request) {
+		getContext(w, r)
+	}).Methods("GET")
 	return router
+}
+
+func getContext(w http.ResponseWriter, r *http.Request) {
+	hostname, err := os.Hostname()
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	ctx := make(map[string]interface{})
+	ctx["hostname"] = hostname
+	ReplyJSONOK(w, ctx)
 }
 
 // Control endpoint


jemoreira:
> for example, is the operator (and cvd, etc.) running in a container, or not?

You should know this from the environment you're running on, not from the operator. It should be possible for the owners of said environments to hide those details from the user, so the operator should not expose them.

It would only be ok to expose this info the Cloud Orchestrator, when requesting details about a host.

frontend/src/liboperator/operator/operator.go
```diff
@@ -174,9 +174,23 @@ func CreateHttpHandlers(
 	router.HandleFunc("/infra_config", func(w http.ResponseWriter, r *http.Request) {
 		ReplyJSONOK(w, config)
 	}).Methods("GET")
+	router.HandleFunc("/context", func(w http.ResponseWriter, r *http.Request) {
```
ser-io, line 177: There's a  endpoint  [/_debug/varz](https://github.com/google/android-cuttlefish/blob/main/frontend/src/host_orchestrator/orchestrator/controller.go#L138) in HO and I think they aim to achieve the same, if so, let's use the same path for alignment.

3405691582:
> > for example, is the operator (and cvd, etc.) running in a container, or not?
> 
> You should know this from the environment you're running on, not from the operator. It should be possible for the owners of said environments to hide those details from the user, so the operator should not expose them.

It's useful to know _something_ about the environment for debugging purposes. 

My other approach would be to use something like the EchoServer as a Schelling point for this information. Is that better, is that also problematic?

frontend/src/liboperator/operator/operator.go
```diff
@@ -174,9 +174,23 @@ func CreateHttpHandlers(
...
 	return router
 }
 
+func getContext(w http.ResponseWriter, r *http.Request) {
+	hostname, err := os.Hostname()
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	ctx := make(map[string]interface{})
```
ser-io, line 189: Let's use a defined `struct` rather than a free form map to keep track of what's being returned.

ser-io (changes requested):

frontend/src/liboperator/operator/operator.go
```diff
@@ -174,9 +174,23 @@ func CreateHttpHandlers(
...
 		ReplyJSONOK(w, config)
 	}).Methods("GET")
+	router.HandleFunc("/context", func(w http.ResponseWriter, r *http.Request) {
+		getContext(w, r)
+	}).Methods("GET")
 	return router
 }
 
+func getContext(w http.ResponseWriter, r *http.Request) {
+	hostname, err := os.Hostname()
```
ser-io, line 184: Right now we are only querying the hostname. In the future `context` as a whole should have more information. Wondering whether you can populate context fields that won't change during the lifetime of the operator when the operator starts and the return the already populated struct on every GET request rather than querying the system everytime?

ser-io (changes requested):

PR #2432, opened 2026-04-21
Update e2e tests containers documentation to using rootless podman.

ser-io:
Bug: b/8154620


jemoreira (approved):
The PR title needs changing since `Containerfile` also changed, it's not just documentation.

PR #2436, opened 2026-04-22
Refine GH postsubmit workflow to calculate version schema earlier

0405ysj:
Currently podcvd is relying on the tag `nightly`, where underlying container image is mutable. GH postsubmit check should pass information around proper image/tag name of fixed container image when building `cuttlefish-podcvd` deb package. However, its calculation is handled by each job or action today. This PR aims for calculating it earlier, so that GH postsubmit workflow is getting more clear and also being usable by more jobs in the workflow

Context: b/504803548

### DIFF
--- a/.github/actions/build-debian-packages/action.yaml
+++ b/.github/actions/build-debian-packages/action.yaml
@@ -1,49 +1,26 @@
 name: 'Build debian packages for Cuttlefish host'
 inputs:
-  deploy-channel:
-    required: false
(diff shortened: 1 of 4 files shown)


0405ysj:
Note that I will check if merge queue is empty in advance and merge this PR with bypassing merge queue, as most validation of this PR is available by checking GH workflow after merging it.

k311093 (approved):

PR #2439, opened 2026-04-22
Podcvd uses proper container image version, based on version itself

0405ysj:
Context: b/504803548

### DIFF
--- a/container/src/podcvd/internal/const.go
+++ b/container/src/podcvd/internal/const.go
@@ -14,9 +14,9 @@
 
 package internal
 
-// TODO(seungjaeyoo): Propagate these constant values from the debian package wrapping podcvd.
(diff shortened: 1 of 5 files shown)


container/src/podcvd/internal/const.go
```diff
@@ -14,9 +14,9 @@
 
 package internal
 
-// TODO(seungjaeyoo): Propagate these constant values from the debian package wrapping podcvd.
+var imageName = "us-docker.pkg.dev/android-cuttlefish-artifacts/cuttlefish-orchestration/cuttlefish-orchestration:nightly"
```
k311093, line 17: Not sure, but might be const (not var)?

k311093 (approved):

container/src/podcvd/internal/const.go
0405ysj: That's intended to inject customized value into `imageName` while building debian package.

PR #2442, opened 2026-04-22
gigabyte-ampere-cuttlefish-installer: modify actions and workflow to support pinned version

grandpaul:
Let the installer reproducible. Previously the installer always install
the latest stable version. We change it to only install the version
when the image is being built. This makes the image reproducible if
the newer image has issues.


0405ysj (approved):

0405ysj:
Sorry, I missed one thing to mention after stamping approval. However as it could be submitted in parallel, I introduced https://github.com/google/android-cuttlefish/pull/2447. Just FYI that on-premise server image will be built and deployed after deb packages.

PR #2443, opened 2026-04-22
Media e2e test

ser-io:


jemoreira (approved):
Are all 3 commits in the series needed? I seems to me they could be squashed.

ser-io:
> Are all 3 commits in the series needed? I seems to me they could be squashed.

Not needed, good catch, thanks!

Thanks for reviewing!

PR #2444, opened 2026-04-22
Protbuf best practice: Do Include an Unspecified Value in an Enum

ser-io:
https://protobuf.dev/best-practices/dos-donts/


rmuthiah (approved):

jemoreira (approved):
Please mention in the commit description that this change is only safe because the proto is not in use anywhere yet.

ser-io:
> Please mention in the commit description that this change is only safe because the proto is not in use anywhere yet.

Done

PR #2446, opened 2026-04-22
Replace usages of android::base::Realpath

jemoreira:
* Adds our own implementation of RealPath, returning Result and handling EINTR errors.
* Don't call realpath in SystemWideHome to break a circular dependency and because it's not correct in most situations.
* Adds tilde expansion to cuttlefish::AbsolutePath instead of returning empty string (it still does if SystemWideHome fails, but at least that's less likely)
* 

Bug: b/505481472
Assisted-by: Gemini:Next

### DIFF
--- a/base/cvd/cuttlefish/common/libs/utils/files.cpp
+++ b/base/cvd/cuttlefish/common/libs/utils/files.cpp
@@ -432,10 +432,6 @@ std::string AbsolutePath(std::string_view path) {
   if (path[0] == '/') {
     return std::string(path);
   }
-  if (path[0] == '~') {
-    LOG(WARNING) << "Tilde expansion in path " << path <<" is not supported";
-    return {};
-  }
 
   std::array<char, PATH_MAX> buffer{};
   if (!realpath(".", buffer.data())) {
--- a/base/cvd/cuttlefish/common/libs/utils/files.cpp
+++ b/base/cvd/cuttlefish/common/libs/utils/files.cpp
@@ -23,11 +23,15 @@
 #endif
 
 #include <dirent.h>
+#include <errno.h>
 #include <fcntl.h>
 #include <ftw.h>
 #include <libgen.h>
 #include <sched.h>
+#include <stddef.h>
+#include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
 #include <sys/ioctl.h>
 #include <sys/select.h>
 #include <sys/socket.h>
@@ -37,12 +41,7 @@
 #include <unistd.h>
 
 #include <array>
-#include <cerrno>
 #include <chrono>
-#include <cstddef>
-#include <cstdio>
-#include <cstdlib>
-#include <cstring>
 #include <fstream>
 #include <ios>
 #include <iosfwd>
--- a/base/cvd/cuttlefish/common/libs/utils/files.cpp
+++ b/base/cvd/cuttlefish/common/libs/utils/files.cpp
@@ -433,13 +433,24 @@ std::string AbsolutePath(std::string_view path) {
     return std::string(path);
   }
 
-  std::array<char, PATH_MAX> buffer{};
-  if (!realpath(".", buffer.data())) {
-    LOG(WARNING) << "Could not get real path for current directory \".\""
-                 << ": " << strerror(errno);
+  Result<std::string> real_cwd = RealPath(".");
+  if (!real_cwd.ok()) {
+    LOG(WARNING) << "Could not get real path for current directory \".\": "
+                 << real_cwd.error();
     return {};
   }
-  return absl::StrCat(buffer.data(), "/", path);
+  return absl::StrCat(*real_cwd, "/", path);
+}
+
+Result<std::string> RealPath(const std::string& path) {
+  std::array<char, PATH_MAX> buffer{};
+  char* res;
+  do {
+    res = realpath(path.c_str(), buffer.data());
+  } while (res == nullptr && errno == EINTR);
+  CF_EXPECTF(res != nullptr, "Could not get real path for path \"{}\": {}",
+             path, StrError(errno));
+  return std::string(buffer.data());
 }
 
 off_t FileSize(const std::string& path) {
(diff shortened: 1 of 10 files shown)


base/cvd/cuttlefish/host/libs/command_util/snapshot_utils.cc
```diff
@@ -143,9 +143,9 @@ Result<void> CopyDirectoryImpl(
  * If emulating absolute path fails, "path" is returned as is.
  */
 std::string RealpathOrSelf(const std::string& path) {
-  std::string output;
-  if (android::base::Realpath(path, &output)) {
-    return output;
+  auto output = RealPath(path);
```
Databean: nit: `Result<std::string>`, not `auto`

base/cvd/cuttlefish/host/libs/config/fetcher_configs.cc
```diff
@@ -40,9 +40,12 @@ FetcherConfigs FetcherConfigs::ReadFromDirectories(
 
   for (const std::string& dir : directories) {
     std::string real;
-    if (!android::base::Realpath(dir, &real)) {
-      LOG(WARNING) << "Failed to resolve real path for '" << dir << "'";
+    auto real_res = cuttlefish::RealPath(dir);
```
Databean: nit: `Result<std::string>`, not `auto`

base/cvd/cuttlefish/host/libs/config/fetcher_configs.cc
```diff
@@ -40,9 +40,12 @@ FetcherConfigs FetcherConfigs::ReadFromDirectories(
...
+    if (!real_res.ok()) {
+      LOG(WARNING) << "Failed to resolve real path for '" << dir << "': " << real_res.error().Message();
```
Databean: This can print `error()` rather than `error().Message()`.

base/cvd/cuttlefish/host/libs/image_aggregator/sparse_image.cc
```diff
@@ -38,8 +40,7 @@ namespace {
 constexpr std::string_view kAndroidSparseImageMagic = "\x3A\xFF\x26\xED";
 
 Result<SharedFD> AcquireLockForImage(const std::string& image_path) {
-  std::string image_realpath;
-  CF_EXPECT(android::base::Realpath(image_path, &image_realpath));
+  std::string image_realpath = CF_EXPECT(cuttlefish::RealPath(image_path));
```
Databean: This is already in `namespace cuttlefish`, it doesn't need to use `cuttlefish::`.

base/cvd/cuttlefish/host/commands/cvd/instances/config_path.cpp
```diff
@@ -25,7 +25,7 @@ namespace cuttlefish {
 Result<std::string> GetCuttlefishConfigPath(const std::string& home) {
   std::string home_realpath;
   CF_EXPECT(DirectoryExists(home), "Invalid Home Directory");
-  CF_EXPECT(android::base::Realpath(home, &home_realpath));
+  home_realpath = CF_EXPECT(RealPath(home));
```
Databean: The declaration of `std::string home_realpath` can be combined with this line.

base/cvd/cuttlefish/common/libs/utils/files.cpp
```diff
@@ -433,18 +432,33 @@ std::string AbsolutePath(std::string_view path) {
   if (path[0] == '/') {
     return std::string(path);
   }
-  if (path[0] == '~') {
-    LOG(WARNING) << "Tilde expansion in path " << path <<" is not supported";
-    return {};
+  if (path == "~" || absl::StartsWith(path, "~/")) {
+    auto home = SystemWideUserHome();
```
Databean: This is not how `~` expansion works in the shell. For comparison, look at
```
HOME=$PWD bash -c 'echo ~'
```
This will print `$PWD`. So really `~` expansion should look at `HOME`. This also breaks the circular dependency, and `SystemWideUserHome` can still use `AbsolutePath`.

jemoreira: I mentioned it in the commit message, but forgot to copy to the PR description. I didn't want to use the HOME var because we frequently modify that value. We'd only get ~ in a path given by the user, the user is not expecting cvd to replace that with the group home instead of their home. I can drop this commit, since neither the variable or the system wide home are correct and it worked fine without it before. It just bothered me that it was returning and empty string, but that wasn't fixed by this either.

SystemWideHome doesn't need AbsolutePath. It's getting the value from getpwuid, which should be absolute or that user is going to have a different home directory every time they `cd`.

jemoreira: Removed the special handling of `~`. This should be handled by the shell only, commands should allow it in paths since it's a valid file name.

base/cvd/cuttlefish/host/commands/cvd/instances/config_path.cpp
jemoreira: Done

base/cvd/cuttlefish/host/libs/command_util/snapshot_utils.cc
jemoreira: Done

base/cvd/cuttlefish/host/libs/config/fetcher_configs.cc
jemoreira: Done

base/cvd/cuttlefish/host/libs/config/fetcher_configs.cc
jemoreira: Done and fixed line length

base/cvd/cuttlefish/host/libs/image_aggregator/sparse_image.cc
jemoreira: Done

Databean (approved):

PR #2457, opened 2026-04-24
assemble_cvd: Support named/A-B partitions in custom_partition_path flag

pjj1005:
1. Extended custom_partition_path flag format (disk_flags.cc)

Previously, the --custom_partition_path flag only accepted a list of image file paths separated by semicolons, and partition names were hard-coded as "custom", "custom_1", "custom_2", etc. This made it impossible to assign OEM-specific partition names required by the bootloader (e.g. "oem_ext", "confptstatic", etc.).

The entry format is now extended to 'name:path[:ab]':
  - name  : partition label visible to the bootloader/kernel
  - path  : image file path
  - :ab   : optional suffix; if present, both _a and _b A/B slot
            partitions are created from the same image file

Legacy format (path only) is still accepted for backward compatibility, in which case partition names fall back to the original auto-generated "custom" / "custom_N" naming.

A LOG(INFO) message is emitted for each partition that is added to aid debugging of partition configuration issues.

Example:
  --custom_partition_path="oem_ext:./oem_ext.img:ab;confcal:./confcal.img"

2. Include partition label in TextConfig() (disk_builder.cpp)

TextConfig() is used to detect whether the composite disk needs to be rebuilt on the next run. Previously it stored only the image file path for each partition. As a result, changing only a partition label (while keeping the same image file) would not trigger a rebuild, leaving the stale composite disk in use.

Now the format is 'label:path' per line, so any change to either the partition name or the image path correctly invalidates the cached composite disk and forces a rebuild.

Change-Id: Ic953c84ce2c07bea2a1f12111726b366a43e33c1

### DIFF
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd_flags.cpp
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd_flags.cpp
@@ -499,13 +499,16 @@ DEFINE_string(
     custom_partition_path, CF_DEFAULTS_CUSTOM_PARTITION_PATH,
     "Location of custom image that will be passed as a custom partition "
     "to rootfs. Multiple entries can be passed, separated by semicolons. "
-    "Each entry format is 'name:path[:ab]'. If ':ab' suffix is present, "
-    "A/B partition slots (name_a, name_b) are created. "
-    "If no name is provided (legacy), defaults to 'custom', 'custom_1', etc. "
+    "Each entry uses key=value pairs separated by commas. "
+    "Supported keys: name (partition label), path (image file path), "
+    "ab (true/false for A/B slots). 'path' is required; 'name' defaults to "
+    "'custom', 'custom_1', etc.; 'ab' defaults to false. "
     "Example with A/B: "
-    "--custom_partition_path=\"oem:./oem.img:ab;config:./config.img\" "
+    "--custom_partition_path=\"name=oem1,path=./oem1.img,ab=true;"
+    "name=oem2,path=./oem2.img\" "
     "Example without names (legacy): "
-    "--custom_partition_path=\"/path/to/custom.img;/path/to/other.img\"");
+    "--custom_partition_path=\"/path/to/custom.img;/path/to/other.img\" "
+    "For multi-instance, separate per-instance values with '|'.");
 
 DEFINE_string(
     blank_sdcard_image_mb, CF_DEFAULTS_BLANK_SDCARD_IMAGE_MB,
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd_flags.cpp
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd_flags.cpp
@@ -497,10 +497,14 @@ DEFINE_string(fuchsia_root_image, CF_DEFAULTS_FUCHSIA_ROOT_IMAGE,
 
 DEFINE_string(
     custom_partition_path, CF_DEFAULTS_CUSTOM_PARTITION_PATH,
-    "Location of custom image that will be passed as a \"custom\" partition"
-    "to rootfs and can be used by /dev/block/by-name/custom. Multiple images "
-    "can be passed, separated by semicolons and can be used as "
-    "/dev/block/by-name/custom_1, /dev/block/by-name/custom_2, etc. Example: "
+    "Location of custom image that will be passed as a custom partition "
+    "to rootfs. Multiple entries can be passed, separated by semicolons. "
+    "Each entry format is 'name:path[:ab]'. If ':ab' suffix is present, "
+    "A/B partition slots (name_a, name_b) are created. "
+    "If no name is provided (legacy), defaults to 'custom', 'custom_1', etc. "
+    "Example with A/B: "
+    "--custom_partition_path=\"oem:./oem.img:ab;config:./config.img\" "
+    "Example without names (legacy): "
     "--custom_partition_path=\"/path/to/custom.img;/path/to/other.img\"");
 
 DEFINE_string(
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/disk/android_composite_disk_config.cc
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/disk/android_composite_disk_config.cc
@@ -26,6 +26,7 @@
 #include <string_view>
 #include <vector>
 
+#include "absl/log/log.h"
 #include "absl/strings/str_cat.h"
 #include "absl/strings/str_split.h"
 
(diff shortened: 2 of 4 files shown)


base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd_flags.cpp
```diff
@@ -497,10 +497,14 @@ DEFINE_string(fuchsia_root_image, CF_DEFAULTS_FUCHSIA_ROOT_IMAGE,
 
 DEFINE_string(
     custom_partition_path, CF_DEFAULTS_CUSTOM_PARTITION_PATH,
-    "Location of custom image that will be passed as a \"custom\" partition"
-    "to rootfs and can be used by /dev/block/by-name/custom. Multiple images "
-    "can be passed, separated by semicolons and can be used as "
-    "/dev/block/by-name/custom_1, /dev/block/by-name/custom_2, etc. Example: "
+    "Location of custom image that will be passed as a custom partition "
```
dimorinny, line 500: The idea is both valid and useful. My only concern is that the new syntax relies on segment ordering and specific positioning, which will be difficult to manage as we introduce additional segments in the future.

What do you think about using a `key=value` syntax separated by commas instead? It would look something like this: `name=oem,path=./oem.img,ab=true;name=oem2,path=./oem2.img;....` Utilizing the `absl` library should make the parsing code simple to implement and easy to maintain.

dimorinny (changes requested):

base/cvd/cuttlefish/host/commands/assemble_cvd/disk/android_composite_disk_config.cc
```diff
@@ -214,10 +215,39 @@ Result<std::vector<ImagePartition>> AndroidCompositeDiskConfig(
...
+      std::string image_path;
+      bool ab_enabled = false;
+      std::vector<std::string_view> parts =
+          absl::StrSplit(custom_partition_paths[i], ':');
+      if (parts.size() >= 2) {
+        label = std::string(parts[0]);
+        image_path = std::string(parts[1]);
+        if (parts.size() >= 3 && parts[2] == "ab") {
+          ab_enabled = true;
+        }
```
jemoreira: Validate the input and return errors if needed. For example, if the third parameter is not "ab" or if there are more than 3 parameters total.

base/cvd/cuttlefish/host/commands/assemble_cvd/disk/android_composite_disk_config.cc
```diff
@@ -214,10 +215,39 @@ Result<std::vector<ImagePartition>> AndroidCompositeDiskConfig(
...
+      } else {
+        label = i > 0 ? "custom_" + std::to_string(i) : "custom";
+        image_path = std::string(custom_partition_paths[i]);
+      }
```
jemoreira: Encapsulate the string parsing into a function like this:
```
struct CustomPartitionSpec {
  std::string label;
  std::string path;
  bool ab_enabled;
};
Result<CustomPartitionSpec> ParseCustomPartitionSpec(std::string_view spec) {...}
```

base/cvd/cuttlefish/host/commands/assemble_cvd/disk/android_composite_disk_config.cc
```diff
@@ -214,10 +215,39 @@ Result<std::vector<ImagePartition>> AndroidCompositeDiskConfig(
...
+      if (ab_enabled) {
+        LOG(INFO) << "Adding custom partition: " << label + "_a";
```
jemoreira: These logs are too noisy. Use verbose logging (`VLOG(1)`) or no logging at all (no other partitions are logged in this function).

pjj1005:
Thanks, @dimorinny and @jemoreira, for the review.
I've updated the code based on your comments.
I’d appreciate another look when you have time.

[assemble_cvd: Support named/A-B partitions in custom_partition_path flag]
(https://github.com/google/android-cuttlefish/pull/2457/commits/a999770e1f4d39a1abefd41f69aa5bc1d7f22f29)

Databean:
It's also possible to define additional partitions through the android build system, without needing to specify them as command-line arguments. If extra partitions are added to the [`BOARD_CUSTOMIMAGES_PARTITION_LIST`](https://cs.android.com/android/platform/superproject/+/android-latest-release:build/make/core/Makefile;l=6976;drc=0e426f646784e2d233098f6aa0f7444070b1049f) variable, then the Cuttlefish will pick these up automatically from the `img.zip` or the `target_files.zip` files produced by the Android build system. These [make it into the partition table](https://github.com/google/android-cuttlefish/blob/01f86c417f9de2ecd1cd796d4234546adcb304df/base/cvd/cuttlefish/host/commands/assemble_cvd/disk/android_composite_disk_config.cc#L120) via the [`AndroidBuild::PhysicalPartitions`](https://github.com/google/android-cuttlefish/blob/01f86c417f9de2ecd1cd796d4234546adcb304df/base/cvd/cuttlefish/host/commands/assemble_cvd/android_build/physical_partitions.cc#L43) implementation.

pjj1005:
@Databean, Thanks for the comment. 
I believe these two approaches are complementary.

1. **PhysicalPartitions (automatic)** automatically discovers partitions that are already present in the Android build outputs.

2. **--custom_partition_path (explicit)** allows arbitrary image files to be injected at runtime, independent of the Android build system outputs. It can be particularly useful in development and testing scenarios, such as attaching locally built OEM images or images taken from a different build.

yeunbok2g:
Thank you for suggesting the BOARD_CUSTOMIMAGES_PARTITION_LIST approach. We attempted to implement it, but found that the current host tools (v1.57) do not fully support this path out of the box. The following host-side code modifications were required to make it work:

### Required Host Code Modifications
**1. misc_info_metadata.cc — Add AbPartitions() implementation**

**Before**: MetadataFromMiscInfo only implements SystemPartitions(), VendorPartitions(), and LogicalPartitions(). The AbPartitions() method is left as the base class default which returns "Unimplemented".

**After**: Added AbPartitions() override that returns logical partitions merged with partitions listed in avb_custom_images_partition_list from misc_info.txt. Without this, there is no way for the composite disk builder to know which custom partitions require A/B slot suffixes.

**2. android_composite_disk_config.cc — Dynamic A/B set extension + partition ordering fix**

**Before**: The A/B partition set is hardcoded to 7 standard partitions (boot, init_boot, vbmeta, etc.). Also, PhysicalPartitions() results are inserted into a std::map (alphabetically sorted), causing custom partitions like confcal, confptstatic to be placed between boot and init_boot, shifting all subsequent partition device numbers (vdaN).

**After**: 
(a) The A/B set is dynamically extended by querying android_build.AbPartitions(), enabling custom partitions like oem_ext to receive _a/_b suffixes. 
(b) Custom partitions are collected separately and appended at the end of the composite disk, preserving standard partition numbering and avoiding SELinux file_contexts/fstab mismatches (which previously caused userdata to be labeled as swap_block_device).


### Android Build System Changes Also Required
Beyond host code, the Android platform tree also requires non-trivial additions:

- BoardConfig.mk: BOARD_CUSTOMIMAGES_PARTITION_LIST, individual BOARD_AVB_<partition> signing configs, AB_OTA_PARTITIONS entries for A/B partitions
- Makefile rules to generate each custom partition image
- Proper avb_custom_images_partition_list propagation into misc_info.txt

### Alternative Proposal
Given the complexity of these modifications — spanning both the host tools and the Android build system — we would like to request reconsideration of a simpler runtime approach:

Extending the existing --custom_partition_path flag to support partition-specific metadata (e.g., A/B designation, filesystem type) would allow custom partitions to be added at launch time without any build system integration or host code restructuring. This would provide equivalent functionality with significantly less invasive changes and better maintainability across host tool version upgrades.

We would appreciate your review of this alternative direction.

PR #2459, opened 2026-04-24
Install linux-headers/linux-image although not having backports.

0405ysj:
As there was no `linux-headers-arm64`, I observed nvidia drivers weren't installed as intended, on the on-permise server. I expect `linux-image-arm64` would be installed as default though, I added it into `after_install_1.sh` to match parity between whether having backports or not.


k311093 (approved):

grandpaul (approved):
Looks good to me.

PR #2460, opened 2026-04-24
Pass through the path to the iptables binary.

3405691582:
Earlier in #2405 we added some functionality such that we could resolve names against PATH manually. We need to do this for the iptables command specifically, since some environments do not have iptables in PATH unless invoked with sudo, and it is the one utility that both the netlink driver and the iproute2 drivers still shell out to the system for.

Ahead of us using the new utilities to resolve the iptables binary, let us manually specify the path for now. While there is no guarantee that we'll find iptables there outside of Linux, both these drivers are all Linux specific anyway, and if iptables isn't found here, this still wouldn't work, so we aren't regressing in terms of portability or functionality.

### DIFF
--- a/base/cvd/allocd/alloc_utils.cpp
+++ b/base/cvd/allocd/alloc_utils.cpp
@@ -85,7 +85,7 @@ bool CreateMobileIface(std::string_view name, uint16_t id,
     DestroyIface(name);
   }
 
-  if (!IptableConfig(network, true).ok()) {
+  if (!IptableConfig("/sbin/iptables", network, true).ok()) {
     DestroyGateway(name, gateway, netmask);
     DestroyIface(name);
     return false;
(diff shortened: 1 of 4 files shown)


base/cvd/allocd/alloc_utils.cpp
```diff
@@ -85,7 +85,7 @@ bool CreateMobileIface(std::string_view name, uint16_t id,
     DestroyIface(name);
   }
 
-  if (!IptableConfig(network, true).ok()) {
+  if (!IptableConfig("/sbin/iptables", network, true).ok()) {
```
Databean, line 88: FYI it looks like `iptables` moved `/sbin/iptables` to `/usr/sbin/iptables` in ubuntu 20 (https://www.github.com/iiab/iiab/issues/2665) though with a symlink from `/sbin` to `/usr/sbin`, so it's possible this might be at different paths in different distros.

Databean (approved):

base/cvd/allocd/alloc_utils.cpp
3405691582: Yeah; we'll turn this into a PATH resolved thing next, though we'll see if there's time (it may be better to come back to that later).

PR #2467, opened 2026-04-27
Do not advertise TIMESTAMP_MONOTONIC for buffer timestamps.

ser-io:
- This behavior was taken from virtio-media reference device, however it's not relevant to this device implementation.

Bug: b/502639876


jemoreira (approved):

ser-io:
Fails v4l2-compliance

PR #2469, opened 2026-04-27
Branch tool to create GCE image with Docker.

3405691582:
We would eventually like a Debian image in GCE that we can use for use- cases to run containers and/or in bare-metal installations. For this, we just want Debian and Docker installed. We can borrow the existing implementation for create_gce_debian for this and make the necessary changes.

To do this properly necessitates some additional changes, which will be handled in subsequent commits.

Bug: 505513391

### DIFF
--- a/tools/baseimage/cmd/create_gce_debian_container/main.go
+++ b/tools/baseimage/cmd/create_gce_debian_container/main.go
@@ -1,0 +1,102 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"log"
+	"maps"
+	"slices"
+
+	"github.com/google/android-cuttlefish/tools/baseimage/pkg/cli"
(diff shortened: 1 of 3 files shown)


ser-io (changes requested):
Can we rather use existing tool gce_install_cuttlefish_container or this new tool aims for something different?

https://github.com/google/android-cuttlefish/blob/main/tools/baseimage/cmd/gce_install_cuttlefish_container/main.go

3405691582:
> Can we rather use existing tool gce_install_cuttlefish_container or this new tool aims for something different?
> 
> https://github.com/google/android-cuttlefish/blob/main/tools/baseimage/cmd/gce_install_cuttlefish_container/main.go

I still need to investigate and confirm this. In the meantime, I'll flip to draft, as we probably won't be able to get back to this for a while.

PR #2471, opened 2026-04-27
Metrics gathering for cvd invocation sources (version 2)

cjreynol:
Gathering, but not yet converting and transmitting, metrics field for cvd invocation sources and environments devices are run in.

Bug: 456483667

NOTE: This PR is a rebase of #2428 after the fetch gathering changes merged.

### DIFF
--- a/base/cvd/cuttlefish/host/libs/metrics/gce_environment.cc
+++ b/base/cvd/cuttlefish/host/libs/metrics/gce_environment.cc
@@ -1,0 +1,102 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cuttlefish/host/libs/metrics/gce_environment.h"
+
+#include <memory>
+#include <optional>
+#include <string>
+#include <string_view>
+#include <vector>
+
+#include <fmt/format.h>
+#include "absl/strings/str_split.h"
+
+#include "cuttlefish/host/libs/web/http_client/curl_global_init.h"
+#include "cuttlefish/host/libs/web/http_client/curl_http_client.h"
+#include "cuttlefish/host/libs/web/http_client/http_client.h"
+#include "cuttlefish/host/libs/web/http_client/http_string.h"
+#include "cuttlefish/result/result.h"
+
+namespace cuttlefish {
+namespace {
+
+constexpr std::string_view kDetectUrl = "metadata.google.internal";
+constexpr std::string_view kDetectHeader = "Metadata-Flavor";
+constexpr std::string_view kExpectedDetectHeaderValue = "Google";
+
+// https://docs.cloud.google.com/compute/docs/metadata/predefined-metadata-keys
+constexpr std::string_view kGceMetadataUrlBase =
+    "http://metadata.google.internal/computeMetadata/v1";
+constexpr std::string_view kNumericProjectId = "project/numeric-project-id";
+constexpr std::string_view kZone = "instance/zone";
+
+// https://docs.cloud.google.com/compute/docs/instances/detect-compute-engine
+Result<bool> IsGceEnvironment(HttpClient& http_client) {
+  const HttpResponse<std::string> response =
+      CF_EXPECT(HttpGetToString(http_client, std::string(kDetectUrl)));
+  CF_EXPECTF(response.HttpSuccess(),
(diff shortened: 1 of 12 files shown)


base/cvd/cuttlefish/host/libs/metrics/gce_environment.cc
```diff
@@ -0,0 +1,101 @@
...
+             response.http_code, response.StatusDescription(), response.data);
+  return response.data;
+}
+
+Result<std::string> GetZoneValue(std::string_view response_string) {
+  std::vector<std::string> components = absl::StrSplit(response_string, "/");
+  CF_EXPECTF(
+      components.size() == 4 && components[2] == "zones",
+      "Unexpected format in response string for zone metadata, expected "
+      "the form \"projects/<project_num>/zones/<zone>\" and received: {}",
```
Databean, line 80: Should this also check `components[0] == "projects"`?

base/cvd/cuttlefish/host/libs/metrics/gce_environment.cc
```diff
@@ -0,0 +1,101 @@
...
+  CurlGlobalInit curl_init;
+  std::unique_ptr<HttpClient> http_client = CurlHttpClient();
+  if (!CF_EXPECT(IsGceEnvironment(*http_client))) {
+    return std::nullopt;
+  }
+  const std::string zone_response =
+      CF_EXPECT(QueryGceMetadata(*http_client, kZone));
+  return GceEnvironment{
+      .numeric_project_id =
+          CF_EXPECT(QueryGceMetadata(*http_client, kNumericProjectId)),
```
Databean, line 97: If the zone metadata response is `projects/<project_num>/zones/<zone>`, can it use `<project_num>` from the zone response rather than issuing another request?

base/cvd/cuttlefish/host/libs/metrics/github_environment.cc
```diff
@@ -0,0 +1,58 @@
...
+
+// https://docs.github.com/en/actions/reference/workflows-and-actions/variables
+// set to "true" when GitHub Actions is running the workflow
+constexpr std::string_view kGitHubActions = "GITHUB_ACTIONS";
+// holds the owner and repository name
+constexpr std::string_view kGitHubRepository = "GITHUB_REPOSITORY";
+
+constexpr std::string_view kAndroidCuttlefish = "google/android-cuttlefish";
+constexpr std::string_view kCloudAndroidOrchestration =
+    "google/cloud-android-orchestration";
```
Databean, line 36: Should we detect forks of these repositories as well?

base/cvd/cuttlefish/host/libs/metrics/host_metrics.h
```diff
@@ -0,0 +1,40 @@
...
+#include "cuttlefish/result/result.h"
+
+namespace cuttlefish {
+
+using Environment = std::variant<GceEnvironment, GitHubRepository>;
+
+struct HostMetrics {
+  HostInfo os;
+  Invoker invoker;
+  std::optional<Environment> environment;
```
Databean: If there was a third variant alternative `UnknownEnvironment`, this wouldn't have to be a `std::optional`.

base/cvd/cuttlefish/host/libs/metrics/gce_environment.cc
cjreynol: Done.

base/cvd/cuttlefish/host/libs/metrics/gce_environment.cc
cjreynol: We could, but that detail just happens to be true of the response right now and I would rather not rely on it.  Obviously if they drastically change the format of the response it would break `zones` gathering and require a fix, but they could potentially return the "project-id" instead of the "numeric-project-id" for some reason and we would quietly break.  It felt more accurate to explicitly query the fields we wanted to gather and match their metadata key API.

base/cvd/cuttlefish/host/libs/metrics/github_environment.cc
cjreynol: We would not want to detect forks until we eventually get external metrics approval.

base/cvd/cuttlefish/host/libs/metrics/host_metrics.h
cjreynol: Done.  I felt like `std::optional` represented "detected an environment we care about gathering for metrics or not", but this will better match what we need for the proto message, since this field will be a `oneOf` and we want to plug some value in there.

cjreynol: Apparently `oneOf` can always be unset, so we do not need an empty message.  But still, this accomplishes the same goal.

base/cvd/cuttlefish/host/libs/metrics/gce_environment.cc
Databean: The https://docs.cloud.google.com/compute/docs/metadata/predefined-metadata-keys link mentioned above writes

> This value has the following format: `projects/PROJECT_NUM/zones/ZONE`

Databean (approved):

base/cvd/cuttlefish/host/libs/metrics/gce_environment.cc
cjreynol: True, but if we make the calls separately we have more likelihood that only one ever breaks at a time.

PR #2472, opened 2026-04-27
Add const char* and std::string_view overloads for StringFromEnv

Databean:
Bug: b/507161864

### DIFF
--- a/base/cvd/cuttlefish/common/libs/utils/environment.h
+++ b/base/cvd/cuttlefish/common/libs/utils/environment.h
@@ -17,12 +17,17 @@
 
 #include <optional>
 #include <string>
+#include <string_view>
 
 namespace cuttlefish {
 
 std::optional<std::string> StringFromEnv(const std::string& varname);
+std::optional<std::string> StringFromEnv(const char* varname);
+std::optional<std::string> StringFromEnv(std::string_view varname);
 
 std::string StringFromEnv(const std::string& varname,
                           const std::string& defval);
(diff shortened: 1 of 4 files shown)


base/cvd/cuttlefish/common/libs/utils/environment.h
```diff
@@ -17,12 +17,17 @@
 
 #include <optional>
 #include <string>
+#include <string_view>
 
 namespace cuttlefish {
 
 std::optional<std::string> StringFromEnv(const std::string& varname);
+std::optional<std::string> StringFromEnv(const char* varname);
```
cjreynol, line 25: Do we need or should we encourage using this type in the API?

A brief search looks like the vast majority of our uses are string values defined and passed directly to the `StringFromEnv` call.  The instances we pass a constant `char[]` it gets converted implicitly anyways.  We could probably update those constants to `std::string_view` if we were concerned with the conversion.

cjreynol (approved):

base/cvd/cuttlefish/common/libs/utils/environment.h
Databean: The implicit conversion technically has worse performance since it needs to call `strlen` to create the `std::string_view` with the length and drops the knowledge that it is null terminated, then `StringFromEnv` needs to make an allocation to produce a `std::string` because `getenv` needs to know it is null terminated.

It seems more straightforward to expose the types used by `getenv` in case they can be passed through with no conversion.

Databean:
Thanks for reviewing!

PR #2474, opened 2026-04-28
Lift requirement that devices have bootconfig

Databean:
Bug: b/507191700


cjreynol (approved):

Databean:
Thanks for reviewing!

PR #2477, opened 2026-04-28
Avoid using user-wide global podman socket on podcvd

0405ysj:
Context: b/503144016


0405ysj:
Discard this PR, as it turned out that it was not a right one to solve b/503144016.

PR #2478, opened 2026-04-28
Detect stockouts on ARM machine types in GCE.

3405691582:
The default ARM machine type can potentially stockout. We will handle this properly soon, but for now, detect and handle this gracefully.

Bug: 505513391

### DIFF
--- a/tools/baseimage/pkg/gce/gce.go
+++ b/tools/baseimage/pkg/gce/gce.go
@@ -101,20 +101,22 @@ func (h *GceHelper) DetachDisk(ins, disk string) error {
 }
 
 func (h *GceHelper) CreateInstance(name string, arch Arch) (*compute.Instance, error) {
-	var machineType, sourceImage string
+	var machineTypes []string
+	var sourceImage string
 	switch arch {
 	case ArchX86:
-		machineType = "n1-standard-16"
+		machineTypes = []string{"n1-standard-16"}
 		sourceImage = "debian-12-bookworm-v20250415"
 	case ArchArm:
-		machineType = "t2a-standard-16"
+		machineTypes = []string{"t2a-standard-16", "c4a-standard-16"}
 		sourceImage = "debian-12-bookworm-arm64-v20250415"
 	default:
 		return nil, errors.New("unsupported arch")
 	}
+
 	payload := &compute.Instance{
 		Name:        name,
-		MachineType: fmt.Sprintf("zones/%s/machineTypes/%s", h.Zone, machineType),
+		MachineType: "",
 		Disks: []*compute.AttachedDisk{
 			{
 				InitializeParams: &compute.AttachedDiskInitializeParams{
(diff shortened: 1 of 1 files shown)


tools/baseimage/pkg/gce/gce.go
```diff
@@ -101,20 +101,22 @@ func (h *GceHelper) DetachDisk(ins, disk string) error {
...
+	var machineTypes []string
+	var sourceImage string
 	switch arch {
 	case ArchX86:
-		machineType = "n1-standard-16"
+		machineTypes = []string{"n1-standard-16"}
 		sourceImage = "debian-12-bookworm-v20250415"
 	case ArchArm:
-		machineType = "t2a-standard-16"
+		machineTypes = []string{"t2a-standard-16", "c4a-standard-16"}
```
ser-io, line 111: Suggest the have only one default ( it could be the new name `c4a-standard-16`) and let the user select the machine type when running the script through a new flag.

Iterating through machine types internally would result into adding new machine types to this list every time users have a stock out error. What happens if the list gets to have 5 types and I want to try the 5th type every time?

Users of this tool are aware of the project they are using to create images, they are aware  of their resource usage and quotas, maybe they can just delete a couple of stale instance to give room for this script to be successful or try an specific type they know they have quota for.

ser-io (changes requested):

tools/baseimage/pkg/gce/gce.go
3405691582: >  maybe they can just delete a couple of stale instance to give room for this script to be successful or try an specific type they know they have quota for.

Stockouts aren't about the user's resource usage or quota, it's about the availability of the machine type. Deleting existing instances won't fix that. Stockouts are a failure of GCE to vend a machine with the given machine type. It is reasonable to either retry with the same machine type or try a different machine type in this circumstance.

ser-io: Good to know. I think we can retry the same type. However we should give the user the ability to select the machine type rather than having the tool trying different hardcoded machine types behind scenes.

3405691582: I don't think it makes sense to allow the user the ability to select the machine type, because the instance isn't actually what's the significant output of this process, it's the disk image. We don't actually care about the instance: we just need it to have a good enough geometry to run the scripts. We only care about the machine type inasmuch to be able to reliably schedule the VM.

We could make this decision elsewhere, but the user shouldn't have to think too deeply about this.

ser-io: Good Point. Agree.

tools/baseimage/pkg/gce/gce.go
```diff
@@ -405,3 +421,10 @@ func runCmd(opts RunCmdOpts, name string, args ...string) error {
 	log.Printf("Executing command: `%s`\n", cmd.String())
 	return cmd.Run()
 }
+
+func IsStockoutError(err error) bool {
```
ser-io: nit: make it package private `isStockoutError`

tools/baseimage/pkg/gce/gce.go
```diff
@@ -135,15 +137,29 @@ func (h *GceHelper) CreateInstance(name string, arch Arch) (*compute.Instance, e
...
+
+		op, err := h.Service.Instances.Insert(h.Project, h.Zone, payload).Do()
+		if err != nil {
+			return nil, err
+		}
+
+		if err := h.waitForOperation(op); err != nil {
+			lastErr = err
+			if IsStockoutError(err) {
+				log.Printf("Stockout detected for machine type %s when creating instance %s", mt, name)
```
ser-io: nit: let's keep all logs message lowercase

ser-io (approved):

tools/baseimage/pkg/gce/gce.go
3405691582: Done.

tools/baseimage/pkg/gce/gce.go
3405691582: Done.

PR #2480, opened 2026-04-28
Cleanup disk if it can't be attached.

3405691582:
We use log.Fatalf to report attachment failures, then exit immediately. However, defer doesn't happen in this case, and so we don't clean up the disk resource we just created.

We could just return an error here and let callers handle this, but here we preserve behavior, keep the overall change scope small, remind ourselves to address this behavior with a TODO, and clean up the disk manually.

Bug: 505513391

### DIFF
--- a/tools/baseimage/pkg/gce/gce.go
+++ b/tools/baseimage/pkg/gce/gce.go
@@ -255,7 +255,7 @@ func (h *GceHelper) BuildImage(project, zone string, opts BuildImageOpts) error 
 
 	log.Println("attaching disk...")
 	if err := h.AttachDisk(insName, attachedDiskName); err != nil {
-		log.Fatalf("failed to attach disk %q to instance %q: %v", attachedDiskName, insName, err)
+		return fmt.Errorf("failed to attach disk %q to instance %q: %v", attachedDiskName, insName, err)
 	}
 	defer h.cleanupDetachDisk(insName, attachedDiskName)
 	log.Println("disk attached")


tools/baseimage/pkg/gce/gce.go
```diff
@@ -255,6 +255,12 @@ func (h *GceHelper) BuildImage(project, zone string, opts BuildImageOpts) error
 
 	log.Println("attaching disk...")
 	if err := h.AttachDisk(insName, attachedDiskName); err != nil {
+		// TODO: It would be better to return the error and let the caller handle it,
```
ser-io: why not return error with this change as you suggest and have rather a TODO?

ser-io (changes requested):

tools/baseimage/pkg/gce/gce.go
3405691582: Because all callers of BuildImage will need to change to handle this, which is a more disruptive change. It is better to make a smaller scoped change to fix the behavior and then a larger change to change the callers. 

If we make the small change correctly and made an error in the large change, reverting the large change still means we have made progress. If we only make the large change incorrectly, reverting it gets us to existing, incorrect behavior.

ser-io: I see.  We have 5 `BuildImage` calls, one per tool. `BuildImage` calls all happen at the end of the relevant main.go files, all looking alike:

```
	if err := h.BuildImage(project, zone, buildImageOpts); err != nil {
		log.Fatal(err)
	}
```

I think it's safe to just return error in this case, as no BuildImage call would need to handle the error differently.

ser-io (changes requested):

tools/baseimage/pkg/gce/gce.go
3405691582: Done.

ser-io (approved):

PR #2481, opened 2026-04-28
Remove default for -arch and make it mandatory.

3405691582:
The -arch flag is typed as cli.Arch. When this is empty, it defaults unconditionally to "x86_64", even when using a source image that doesn't match that architecture. We should probably change this default in cli.Arch, but that would increase the scope of the change. For now, it is more logical to make the arch flag mandatory.

Bug: 505513391

### DIFF
--- a/tools/baseimage/cmd/create_gce_base_image/main.go
+++ b/tools/baseimage/cmd/create_gce_base_image/main.go
@@ -46,6 +46,9 @@ func init() {
 func main() {
 	flag.Parse()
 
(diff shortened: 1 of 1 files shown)


ser-io (changes requested):
let's make it mandatory for all the tools.

3405691582:
> let's make it mandatory for all the tools.

Done this in #2485.

ser-io (approved):

PR #2482, opened 2026-04-28
Install image with docker pull as well as load.

3405691582:
The docker documentation doesn't suggest that, for example, docker load can accept an OCI container image which may be emitted by podman push. The implication is that only a tarball emitted by docker save can be used in a subsequent docker load. This requires docker to be installed.

Even if we can construct a tarball that can be loaded without docker, it is a tedious part of the workflow considering we could just docker pull the image. Allow the possibility.

Bug: 505513391


0405ysj (approved):
Approved. However I think it's okay to remove `containerImageSrc`, if that flag won't be used for making GCE image with container. I'm believing the purpose of `containerImageRef` is to replace `containerImageSrc`.

PR #2487, opened 2026-04-29
Auto login for cvd fetch

cjreynol:
Bug: 470040267

### DIFF
--- a/base/cvd/cuttlefish/common/libs/utils/BUILD.bazel
+++ b/base/cvd/cuttlefish/common/libs/utils/BUILD.bazel
@@ -228,6 +228,15 @@ cf_cc_library(
 )
 
 cf_cc_library(
+    name = "is_google_corp",
+    srcs = ["is_google_corp.cpp"],
+    hdrs = ["is_google_corp.h"],
+    deps = [
+        "//cuttlefish/common/libs/utils:files",
+    ],
+)
+
+cf_cc_library(
     name = "json",
     srcs = ["json.cpp"],
     hdrs = ["json.h"],
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/fetch.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/fetch.cpp
@@ -31,6 +31,8 @@
 #include "cuttlefish/host/commands/cvd/cli/command_request.h"
 #include "cuttlefish/host/commands/cvd/cli/commands/command_handler.h"
 #include "cuttlefish/host/commands/cvd/cli/types.h"
+#include "cuttlefish/host/commands/cvd/fetch/auto_login.h"
+#include "cuttlefish/host/commands/cvd/fetch/build_api_flags.h"
 #include "cuttlefish/host/commands/cvd/fetch/fetch_cvd.h"
 #include "cuttlefish/host/commands/cvd/fetch/fetch_cvd_parser.h"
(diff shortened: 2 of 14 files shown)


base/cvd/cuttlefish/common/libs/utils/BUILD.bazel
```diff
@@ -227,6 +227,15 @@ cf_cc_library(
     ],
 )
 
+cf_cc_library(
+    name = "is_corp",
+    srcs = ["is_corp.cpp"],
```
jemoreira: nit: Mention google, like `is_google_corp`. This applies to the target name, file names and function name.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/fetch.cpp
```diff
@@ -50,14 +52,38 @@ class CvdFetchCommandHandler : public CvdCommandHandler {
   Result<std::string> DetailedHelp(std::vector<std::string>&) const override;
 };
 
+Result<void> RunAutoLogin(const BuildApiFlags& build_api_flags) {
```
jemoreira, line 55: nit: `RunAutoLogin` may not run anything if credentials are available. Either move the check for available credentials to the call site like
```
if (!CF_EXPECT(AreCredentialsAvailable(...))) {
  CF_EXPECT(RunAutoLogin(...));
}
```
or rename the function as something like `EnsureCredentialsAreAvailable`.

base/cvd/cuttlefish/host/commands/cvd/fetch/auto_login.cc
```diff
@@ -0,0 +1,70 @@
...
+}
+
+Result<bool> IsCredentialFilePresent() {
+  return FileExists(GetAcloudOauthFilepath()) ||
+         !CF_EXPECT(GetCvdCredentialFilepaths()).empty();
+}
+
+}  // namespace
+
+Result<bool> AreCredentialsAvailable(const BuildApiFlags& flags) {
```
jemoreira: The naming of this function is throwing me off: If credential file is present, wouldn't there be available credentials?

Is this function supposed to check whether auto login is needed?

base/cvd/cuttlefish/common/libs/utils/BUILD.bazel
cjreynol: Done.

base/cvd/cuttlefish/host/commands/cvd/fetch/auto_login.cc
cjreynol: Renamed to `ShouldAutoLogin`

base/cvd/cuttlefish/host/commands/cvd/cli/commands/fetch.cpp
cjreynol: Renamed the function.  Part of the reason for the helper is to wrap all of this logic so there is a single `Result` to deal with in `Handle`.  I wanted to make sure failures attempting to auto-login (or detect if we should auto-login) do not stop the attempt to fetch, but only deal with manually handling the `Result` once.

jemoreira (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/fetch.cpp
cjreynol: Re-renaming back to `RunAutoLogin`

PR #2489, opened 2026-04-29
[assemble_cvd] Replace slow lsof for file cleanup

bridadan:
CleanPriorFiles called lsof to ensure files weren't in use before deletion, but this was taking ~5 seconds, significantly slowing down startup. lsof was slow because it scanned all processes on the system and performed name resolutions by default.

Replacing it with a targeted C++ scan of /proc/*/fd for the current user's processes avoids these overheads and reduces the execution time to ~132 ms.

Bug: 401259067
Test: cvd create
      check launcher.log for ~5s speed up to UBOOT start


Databean (approved):

bridadan:
@Databean I updated this PR to match the Android source tree CL

Databean:
Thanks!

PR #2491, opened 2026-04-29
Substitute all artifacts when the build has the debian_substitution_marker present.

ser-io:
debian_substitution_marker present.

Bug: b/507524719

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
+++ b/base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
@@ -161,17 +161,6 @@ Result<void> SubstituteWithMarker(const std::string& target_dir,
 Result<void> HostPackageSubstitution(
     const std::string& target_dir,
     const std::vector<std::string>& host_substitutions) {
-  std::string marker_file = target_dir + "/etc/debian_substitution_marker";
-  // Use a local debian_substitution_marker file for development purposes.
-  std::optional<std::string> local_marker_file =
-      StringFromEnv("LOCAL_DEBIAN_SUBSTITUTION_MARKER_FILE");
-  if (local_marker_file.has_value()) {
-    marker_file = local_marker_file.value();
-    CF_EXPECTF(FileExists(marker_file),
-               "local debian substitution marker file does not exist: {}",
-               marker_file);
-    LOG(INFO) << "using local debian substitution marker file: " << marker_file;
-  }
 
   // TODO: remove when we can safely add this to the marker file.
   std::string cvdalloc_src =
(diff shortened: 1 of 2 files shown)


base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
```diff
@@ -156,33 +215,47 @@ Result<void> SubstituteWithMarker(const std::string& target_dir,
   return {};
 }
 
+Result<void> SubstituteAll(const std::string& target_dir) {
```
jemoreira: You don't need a new function with the artifact names constant. Just pass `{"all"}` to `SubstituteWithFlag`.

base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
```diff
@@ -156,33 +215,47 @@ Result<void> SubstituteWithMarker(const std::string& target_dir,
...
+      fmt::format("{}/bin/{}", CF_EXPECT(GetCuttlefishCommonDir()), "cvdalloc");
+  std::string cvdalloc_name = fmt::format("{}/bin/{}", target_dir, "cvdalloc");
+  CF_EXPECT(Substitute(cvdalloc_src, cvdalloc_name));
+
+  // [DEVELOPMENT ONLY] Substitute when `--host_substitutions` flag is used.
+  if (!host_substitutions.empty()) {
+    return SubstituteWithFlag(target_dir, host_substitutions);
+  }
+
+  // [DEVELOPMENT ONLY] Substitute when LOCAL_DEBIAN_SUBSTITUTION_MARKER_FILE is set.
```
jemoreira, line 176: nit: While not introduced by this PR it'd make sense to enforce that only one of `--host_substitutions` flag and the `LOCAL_DEBIAN_SUBSTITUTION_MARKER_FILE` variable is set. If both are it's most likely unintentionally and it may lead to unexpected results.

base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
```diff
@@ -40,6 +40,65 @@
 namespace cuttlefish {
 namespace {
 
+static constexpr std::string_view kArtifactNames[] = {
```
Databean: There's already special logic for `--host_substitutions=all`:
https://github.com/google/android-cuttlefish/blob/311ff45aa2625265178005cc4072e876cba39ef1/base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc#L83

That implementation scans the bazel/debian package artifacts for files that it can substitute, and aggressively replaces as many as possible. what do you think of using this logic rather than adding a new hardcoded list?

Databean: Based on the discussion in the meeting today it sounds like the preferred strategy is to go with a curated list rather than replacing everything possible.

jemoreira: The curated list could be "everything in the cvd package". Executables will still be built even if we don't add them there.

base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
ser-io: Agree. I'll handle this in a follow-up PR.

base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
ser-io: Done.

base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
ser-io: Used {"all"}. As we have already logic in place to substitute elements listed in the cvd package I agree with using the package as the source of truth rather than having a duplicated logic based on a new hardcoded list.

jemoreira (approved):

PR #2492, opened 2026-04-29
`cvd create` detects and fails when creating already existing group

jemoreira:
It couldn't do that before because `cvd load` was calling `cvd create` after creating the group. Making `cvd load` call `cvd start` instead allows `create` to enforce the constraint while at the same time breaking the circular dependency between the two commands (`create` calls `load` when given `--config_file`).

Bug: b/507595144


base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
```diff
@@ -243,20 +243,11 @@ Result<LocalInstanceGroup> CvdCreateCommandHandler::GetOrCreateGroup(
...
-        std::move(creation_info.group_directories));
-  }
-  auto& group = groups[0];
-  CF_EXPECTF((size_t)group.Instances().size() ==
-                 creation_info.group_creation_params.instances.size(),
-             "Mismatch in number of instances from analisys: {} vs {}",
-             group.Instances().size(),
-             creation_info.group_creation_params.instances.size());
-  return group;
+  CF_EXPECTF(groups.empty(), "Group named '{}' already exists",
```
cjreynol, line 241: This is stricter than the check immediately above, so that one can be removed.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/load_configs.cpp
```diff
@@ -178,8 +179,14 @@ class LoadConfigsCommand : public CvdCommandHandler {
                  GetFetchLogsFileName(cvd_flags.target_directory));
     }
 
-    auto launch_cmd = CF_EXPECT(BuildLaunchCmd(request, cvd_flags, group));
-    CF_EXPECT(executor_.ExecuteOne(launch_cmd, std::cerr));
+    // Instances go from preparing to stopped state after fetching is done.
+    group.SetAllStates(cvd::INSTANCE_STATE_STOPPED);
```
cjreynol, line 182: Could this be shared logic with the identical step in `create.cpp::342`, to avoid these steps falling out of sync?

For instance, right now that logic also calls `group.SetStartTime`.  Is that intentionally inconsistent, or just because of two sites to update?

cjreynol (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/load_configs.cpp
jemoreira: They should be the same.

the start time is actually being used as "creation time" instead. I was looking for a different place to put it for `cvd load` but apparently forgot. Good catch!

jemoreira: I could move setting the start time to the creation function function in the instance manager.

I suspect gathering metrics must happen after fetch, but if that's not the case then it could also move there.

Finally, setting the stopped state should still happen in the command, but that's just two lines.

WDYT?

cjreynol: I am fine with the proposed solution to move the start time and keep it to the two lines for state setting here and in `create`.

For the instantiation metrics- I think it would be better to move it there as well.  I think "creation time", like the timestamp, makes sense as the point where we put the metrics gathering event.  That, and it is less likely to be missed if we have it somewhere shared.

The timestamps we care about if we are looking at performance are for start and boot complete.  I cannot think of a reason we are going to be looking at the instantiation timestamps in comparison to any other events.  If anything, having it located before the new fetch events will give us a way to distinguish `cvd load` calls from `cvd fetch && cvd create` runs.

jemoreira: Done:

Moved the instantiation metrics to the instance manager and kept the two lines to set stopped state repeated.

I was wrong about start_time. It was in fact being set at creation and start times. Made it so it's only set at start time now to be consistent with its name, it's not really used for anything important at the moment anyways.

PR #2493, opened 2026-04-29
Transmit new metrics data

cjreynol:
Add the conversion of the newly gathering metrics data for fetches and invocation source information.  This will transmit the new data, if metrics are enabled.

Bug: 491968279
Bug: 491975427
Bug: 491971187


base/cvd/cuttlefish/host/libs/metrics/metrics_conversion.cc
```diff
@@ -144,6 +172,68 @@ CuttlefishHost_OsType ConvertHostOs(const HostInfo& host_info) {
...
+
+CuttlefishGitHubActionsEnvironment_Repository ConvertGitHubRepository(
+    const GitHubRepository repository) {
+  switch (repository) {
+    case GitHubRepository::AndroidCuttlefish:
+      return CuttlefishGitHubActionsEnvironment_Repository::
+          CuttlefishGitHubActionsEnvironment_Repository_CUTTLEFISH_GITHUB_ACTIONS_ENVIRONMENT_REPOSITORY_GOOGLE_ANDROID_CUTTLEFISH;
+    case GitHubRepository::CloudAndroidOrchestration:
+      return CuttlefishGitHubActionsEnvironment_Repository::
+          CuttlefishGitHubActionsEnvironment_Repository_CUTTLEFISH_GITHUB_ACTIONS_ENVIRONMENT_REPOSITORY_GOOGLE_CLOUD_ANDROID_ORCHESTRATION;
```
jmacnak, line 200: full line enum name xD

base/cvd/cuttlefish/host/libs/metrics/metrics_conversion.cc
```diff
@@ -144,6 +172,68 @@ CuttlefishHost_OsType ConvertHostOs(const HostInfo& host_info) {
...
+    CuttlefishGitHubActionsEnvironment& github = *host.mutable_github();
+    github.set_repository(ConvertGitHubRepository(repository));
+  }
+
+  void operator()(const UnknownEnvironment&) {}
+};
+
+void PopulateCuttlefishHostEnvironment(CuttlefishHost& host,
+                                       const Environment& environment) {
+  std::visit(PopulateEnvironment{.host = host}, environment);
```
jmacnak, line 226: optional: i personally find the `if constexpr` pattern:

```
std::visit([](auto&& arg) {
    using T = std::decay_t<decltype(arg)>;
    if constexpr (std::is_same_v<T, GceEnvironment>) {
        CuttlefishGceEnvironment& gce = *host.mutable_gce();
        gce.set_numeric_project_id(environment.numeric_project_id);
        gce.set_zone(environment.zone);
    } else if constexpr (std::is_same_v<T, GitHubRepository>) {
        CuttlefishGitHubActionsEnvironment& github = *host.mutable_github();
        github.set_repository(ConvertGitHubRepository(repository));
    } else if constexpr (std::is_same_v<T, UnknownEnvironment>) {
    }
}, environment);
```

more readable ("if it is a GceEnvironemnt, do this" vs "let me wrap this in a struct") but looks like there is already a struct precedent in this file so feel free to ignore.

jmacnak (approved):

base/cvd/cuttlefish/host/libs/metrics/metrics_conversion.cc
cjreynol: I tried something like this in the past and was struggling to get it to work, so thanks for the example!  I 100% agree this is more readable, which is already a problem when using `std::variant`.  

The upside with the struct (or using the similar method with method overloading arguments over each type in the variant) is getting a compiler check that every possible case is covered.  If a new type is added to the `Environment` definition, this logic throws an error at build time stating the `std::visit` usage is non-exhaustive.  With the `constexpr` conditional version, the best we can do is add a default case to error at runtime.

I think that improvement in maintainability outweighs the lost readability, so I am going to stick with the current version.  Happy to discuss more for the future, since we also do the method overloading version in the `android_build_api.h/cc` code.

PR #2501, opened 2026-05-01
Add a MCP endpoint to the host orchestrator with create_cvd and list_cvds tools

Databean:


frontend/src/host_orchestrator/orchestrator/controller.go
```diff
@@ -136,6 +136,7 @@ func (c *Controller) AddRoutes(router *mux.Router) {
 	router.Handle("/cvd_imgs_dirs/{id}", httpHandler(&updateImageDirectoryHandler{c.ImageDirectoriesManager, c.OperationManager, c.UserArtifactsManager})).Methods("PUT")
 	router.Handle("/cvd_imgs_dirs/{id}", httpHandler(&deleteImageDirectoryHandler{c.ImageDirectoriesManager, c.OperationManager})).Methods("DELETE")
 	router.Handle("/reset", httpHandler(&resetCVDHandler{c.OperationManager})).Methods("POST")
+	router.PathPrefix("/sse").Handler(NewMCPHandler(c.Config, c.OperationManager, c.UserArtifactsManager)).Methods("DELETE", "GET", "POST")
```
ser-io: let's version this endpoint: `/v0/sse`

ser-io (changes requested):

frontend/src/host_orchestrator/orchestrator/controller.go
Databean: Done.

Databean:
Created https://github.com/google/android-cuttlefish/pull/2504 to update the go version first

ikicha:
If we want to run in several environments like local directly, inside each pods, remote ARM servers, it means we need to set up remote(even local) mcp server for each.(not sure mcp server is dynamically loadable by an agent or skill) Maybe skill for each HO API might be universal(and let a user specify ip addr with the same skill)

0405ysj:
Sorry, I don't have clear context about this MCP endpoint yet. I'm curious whether this will replace the role of CF MCP server on top of `podcvd` or not, and also it'll be when in case of replacement. Besides, I want to know how this MCP server can be integrated with CF containers, in perspective of multiple agents in a machine with MCP server described here. I may need a design or a discussion to understand it.

Context: b/502711134

Databean:
This was intended as a proof of concept for what it would look like to add new MCP server endpoints into the existing host orchestrator HTTP service.

> I want to know how this MCP server can be integrated with CF containers, in perspective of multiple agents in a machine with MCP server described here. I may need a design or a discussion to understand it.

I do not have answers to these questions and will not be able to focus on this area during the current demo sprint so I will close this PR for now.

@ser-io also raised the question in chat whether this should be added to the current HTTP service or made a new HTTP service with routing between the two handled by nginx.

PR #2504, opened 2026-05-01
Update frontend build to go 1.25

Databean:
This is a prerequisite for using the go-sdk mcp library: https://github.com/modelcontextprotocol/go-sdk/blob/f5f2015413f17f578fbacc839d2c05d0fdb3f1cc/go.mod#L3

Bug: b/508728217

### DIFF
--- a/frontend/src/host_orchestrator/go.mod
+++ b/frontend/src/host_orchestrator/go.mod
@@ -8,7 +8,6 @@ require (
 	github.com/google/go-cmp v0.5.9
 	github.com/google/uuid v1.6.0
 	github.com/gorilla/mux v1.8.0
-	github.com/hashicorp/go-multierror v1.1.1
 )
 
 require (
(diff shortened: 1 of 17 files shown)


frontend/src/host_orchestrator/go.mod
```diff
@@ -1,14 +1,13 @@
 module github.com/google/android-cuttlefish/frontend/src/host_orchestrator
 
-go 1.17
+go 1.25
```
jemoreira, line 3: This has been a problem in the past because external users are in older OS versions that don't have this version available. For example, ubuntu 24.04 comes with go 1.22 by default while ubuntu 22.04 comes with 1.18; the newest version that can be installed in both is 1.23.

Databean: This seems to work on my machine where `go version` reports `go version go1.24.13 linux/amd64`, I validated it using `tools/buildutils/build_package.sh frontend`. I think older versions of go are capable of downloading newer versions, but updating the `debian/control` file will probably make it fail.

jemoreira (approved):

Databean:
I tried PR #2501 with go 1.18 and it seems to build, so I'll drop this PR for now.

PR #2506, opened 2026-05-02
cvd: Add support for logging to a file

Databean:
Implement timestamped log file creation for the cvd command. Logs are saved in CvdDir() + "/logs/" with a name like
cvd_YYYYMMDD_HHMMSS.ms.log. This ensures logs are preserved across runs.

Bug: b/507600500

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/log_files.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/log_files.cpp
@@ -1,0 +1,57 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cuttlefish/host/commands/cvd/cli/log_files.h"
+
+#include <chrono>
+#include <iomanip>
+#include <iostream>
+#include <sstream>
+#include <string>
+#include <vector>
+
+#include "absl/strings/str_cat.h"
+
+#include "cuttlefish/common/libs/utils/files.h"
+#include "cuttlefish/host/commands/cvd/utils/common.h"
+
+namespace cuttlefish {
+
+std::optional<std::string> GetCvdLogFile() {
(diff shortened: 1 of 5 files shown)


base/cvd/cuttlefish/host/commands/cvd/cli/log_files.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include "absl/strings/str_cat.h"
+
+#include "cuttlefish/common/libs/utils/files.h"
+#include "cuttlefish/host/commands/cvd/utils/common.h"
+
+namespace cuttlefish {
+
+std::vector<std::string> GetLogFiles() {
+  std::string log_dir = CvdDir() + "/logs";
```
jemoreira: This should use `PerUserDir()` instead.

base/cvd/cuttlefish/host/commands/cvd/cli/log_files.cpp
```diff
@@ -0,0 +1,57 @@
...
+          now.time_since_epoch()) %
+      1000;
+
+  std::tm tm_now;
+  localtime_r(&now_time_t, &tm_now);
+
+  std::stringstream ss;
+  ss << std::put_time(&tm_now, "%Y%m%d_%H%M%S") << "." << std::setfill('0')
+     << std::setw(3) << now_ms.count();
+  return {absl::StrCat(log_dir, "/cvd_", ss.str(), ".log")};
```
jemoreira: Why return a vector if it never returns more than one? Would an optional make more sense?

base/cvd/cuttlefish/host/commands/cvd/main.cc
```diff
@@ -20,23 +20,28 @@
 #include <time.h>
 #include <unistd.h>
 
+#include <chrono>
```
jemoreira: `chrono` ins't used in this file. Some of the other new headers may not be needed either.

base/cvd/cuttlefish/host/commands/cvd/cli/log_files.cpp
Databean: `std::optional` makes more sense for the function, `std::vector` was more convenient to match the interface of `LogToStderrAndFiles`. Updated to return a `std::optional`.

Unfortunately adding an overload of `LogToStderrAndFiles` that takes a `std::optional` breaks several invocations that pass `{}` as the first argument because it becomes ambiguous between `std::optional` and `std::vector` even though the semantics of an empty optional and empty vector are the same.

base/cvd/cuttlefish/host/commands/cvd/cli/log_files.cpp
Databean: Done.

base/cvd/cuttlefish/host/commands/cvd/main.cc
Databean: Good point, done.

jemoreira (approved):

PR #2508, opened 2026-05-04
Tidy up cURL error messages in metrics cases

cjreynol:
During the GCE environment detection logic, the HTTP request to make the detection would print out error lines if GCE was not detected.  This update makes all of the cURL outputs go into the logs, instead.

It makes the V2 `metrics/metrics.log` a bit noisier, but that is better than cluttering up the command-line output with the messages.

Bug: 508689956

### DIFF
--- a/base/cvd/cuttlefish/host/libs/metrics/gce_environment.cc
+++ b/base/cvd/cuttlefish/host/libs/metrics/gce_environment.cc
@@ -86,7 +86,9 @@ Result<std::string> GetZoneValue(std::string_view response_string) {
 
 Result<std::optional<GceEnvironment>> DetectGceEnvironment() {
   CurlGlobalInit curl_init;
-  std::unique_ptr<HttpClient> http_client = CurlHttpClient();
+  const bool use_logging_debug_function = true;
+  std::unique_ptr<HttpClient> http_client =
+      CurlHttpClient(use_logging_debug_function);
   if (!CF_EXPECT(IsGceEnvironment(*http_client))) {
     return std::nullopt;
   }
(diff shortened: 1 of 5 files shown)


jemoreira (approved):

base/cvd/cuttlefish/host/libs/metrics/gce_environment.cc
```diff
@@ -86,7 +86,9 @@ Result<std::string> GetZoneValue(std::string_view response_string) {
 
 Result<std::optional<GceEnvironment>> DetectGceEnvironment() {
   CurlGlobalInit curl_init;
-  std::unique_ptr<HttpClient> http_client = CurlHttpClient();
+  const bool use_logging_debug_function = true;
+  std::unique_ptr<HttpClient> http_client =
+      CurlHttpClient(use_logging_debug_function);
```
jmacnak: maybe `CurlHttpClient(/*use_logging_debug_function=*/true);` to avoid the extra variable

jmacnak (approved):

base/cvd/cuttlefish/host/libs/metrics/gce_environment.cc
cjreynol: Done, and for the rest of the `CurlHttpClient` calls.  Added the [bugprone-argument-comment](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/argument-comment.html#bugprone-argument-comment) `clang-tidy` check to verify the comments match the parameter name.

jmacnak (approved):

PR #2509, opened 2026-05-04
Stop using EmulatedAbsolutePath

jemoreira:
EmulateAbsolutePath was needed for the cvd server as the client may have
been executing in a different directory and with a different $HOME.

The function handled "~" at the begining of the path, which the cvd
command shouldn't do as that's the shell's responsibility.

Without the special handling for "~" and with all callers just using the
current working directory, calls to the function were doing exactly the
same as `AbsolutePath` or `RealPath` (with `follow_symlinks` as true or
false respectively).


base/cvd/cuttlefish/common/libs/utils/files.cpp
```diff
@@ -638,100 +627,5 @@ Result<void> WalkDirectory(const std::string& dir,
   return {};
 }
 
-namespace {
```
cjreynol, line 641: These were added recently, maybe there was an upcoming use planned?

https://github.com/google/android-cuttlefish/pull/2405

Worth checking before the merge.

cjreynol (approved):

base/cvd/cuttlefish/common/libs/utils/files.cpp
jemoreira: Removed that commit from the PR.

PR #2512, opened 2026-05-05
Remove CommandSequenceExeuctor

Databean:
Commands that invoke other commands now directly create the handlers for those commands, rather than going through parsing. This makes it possible to drop the `CommandSequenceExecutor` implementation.

This also drops some of the CLI output that prints intermediate commands, which are not particularly informative for most users.

Assisted-by: Jetski
Bug: b/324653441


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #2515, opened 2026-05-05
Resolve path to the iptables binary.

3405691582:
From work done in #2405 and #2460, now wire up PATH resolution for the iptables binary.

### DIFF
--- a/base/cvd/allocd/alloc_utils.h
+++ b/base/cvd/allocd/alloc_utils.h
@@ -22,7 +22,9 @@
 #include <unistd.h>
 
 #include <optional>
+#include <string>
 #include <string_view>
+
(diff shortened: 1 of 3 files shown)


base/cvd/allocd/alloc_utils.h
```diff
@@ -22,9 +22,12 @@
 #include <unistd.h>
 
 #include <optional>
+#include <string>
 #include <string_view>
 #include "cuttlefish/result/result.h"
 
+#include "cuttlefish/result/result.h"
```
Databean, line 28: nit: duplicate include, maybe remove the existing one since it's not grouped correctly?

Databean (approved):

PR #2517, opened 2026-05-05
cvd: start and bugreport help works everywhere

jemoreira:
Both commands rely on `cvd_internal_start` and `cvd_internal_host_bugreport` for their help output. Those executables are searched for in `$ANDROID_HOST_OUT/bin`, `$HOME/bin` and `./bin`, which means the command's help only works after building cuttlefish from source or downloading an android build from the server.

This PR makes cvd use the internal binaries it uses for executable substitution, which are located in the same directory where cvd is.

Bug: b/510047667

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/bugreport.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/bugreport.cpp
@@ -19,6 +19,7 @@
 #include <stdlib.h>
 
 #include <functional>
+#include <iostream>
 #include <memory>
 #include <string>
 #include <utility>
@@ -168,10 +169,13 @@ Result<std::string> CvdBugreportCommandHandler::DetailedHelp(
   Command command = CF_EXPECT(ConstructSiblingHelpCommand(
       kHostBugreportBin, request.Env(), request.SubcommandArguments()));
   std::string stdout;
-  int res = RunWithManagedStdio(std::move(command), nullptr, &stdout, nullptr);
+  std::string stderr;
+  int res = RunWithManagedStdio(std::move(command), nullptr, &stdout, &stderr);
   // gflags returns exit code 1 when --help is given
-  CF_EXPECTF(res == 0 || res == 1,
-             "Failed to execute bugreport binary, exit code: {}", res);
+  if (res != 0 && res != 1) {
+    std::cerr << stderr;
+    return CF_ERRF("Failed to execute bugreport binary, exit code: {}", res);
+  }
   return stdout;
 }
 
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/bugreport.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/bugreport.cpp
@@ -165,10 +165,8 @@ bool CvdBugreportCommandHandler::RequiresDeviceExists() const { return true; }
 
 Result<std::string> CvdBugreportCommandHandler::DetailedHelp(
     const CommandRequest& request) const {
-  std::string android_host_out = CF_EXPECT(AndroidHostPath(request.Env()));
-  Command command = CF_EXPECT(
-      ConstructCvdHelpCommand(kHostBugreportBin, request.Env(),
-                              request.SubcommandArguments(), request));
+  Command command = CF_EXPECT(ConstructSiblingHelpCommand(
+      kHostBugreportBin, request.Env(), request.SubcommandArguments()));
   std::string stdout;
   int res = RunWithManagedStdio(std::move(command), nullptr, &stdout, nullptr);
   // gflags returns exit code 1 when --help is given
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
@@ -41,7 +41,6 @@
 #include "absl/log/log.h"
 #include "absl/strings/match.h"
 
-#include "cuttlefish/common/libs/utils/contains.h"
 #include "cuttlefish/common/libs/utils/files.h"
 #include "cuttlefish/common/libs/utils/flag_parser.h"
 #include "cuttlefish/common/libs/utils/json.h"
@@ -536,22 +535,9 @@ Result<std::string> CvdStartCommandHandler::DetailedHelp(
       selector::SelectGroup(instance_manager_, request);
   if (group_res.ok()) {
     envs["HOME"] = group_res.value().HomeDir();
-  } else if (Contains(envs, "HOME")) {
-    if (envs["HOME"].empty()) {
-      envs.erase("HOME");
-    } else {
-      envs["HOME"] = AbsolutePath(envs.at("HOME"));
-    }
   }
-  auto android_host_out =
-      CF_EXPECT(AndroidHostPath(envs),
-                "\nTry running this command from the same directory as the "
-                "downloaded or fetched host tools.");
-  const auto bin = CF_EXPECT(FindStartBin(android_host_out));
-
-  Command command = CF_EXPECT(ConstructCvdHelpCommand(
-      bin, envs, request.SubcommandArguments(), request));
-  LOG(INFO) << "help command: " << command;
+  Command command = CF_EXPECT(ConstructSiblingHelpCommand(
+      "cvd_internal_start", envs, request.SubcommandArguments()));
   std::string stdout;
   int res = RunWithManagedStdio(std::move(command), nullptr, &stdout, nullptr);
   // gflags returns exit code 1 when --help is given
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
@@ -539,10 +539,13 @@ Result<std::string> CvdStartCommandHandler::DetailedHelp(
   Command command = CF_EXPECT(ConstructSiblingHelpCommand(
       "cvd_internal_start", envs, request.SubcommandArguments()));
   std::string stdout;
-  int res = RunWithManagedStdio(std::move(command), nullptr, &stdout, nullptr);
+  std::string stderr;
+  int res = RunWithManagedStdio(std::move(command), nullptr, &stdout, &stderr);
   // gflags returns exit code 1 when --help is given
-  CF_EXPECTF(res == 0 || res == 1,
-             "Failed to execute start binary, exit code: {}", res);
+  if (res != 0 && res != 1) {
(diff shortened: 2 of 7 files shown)


base/cvd/cuttlefish/host/libs/config/config_flag.cpp
```diff
@@ -87,12 +87,15 @@ class ConfigReader : public FlagFeature {
   std::unordered_set<FlagFeature*> Dependencies() const override { return {}; }
   Result<void> Process(std::vector<std::string>&) override {
     auto config_path = DefaultHostArtifactsPath("etc/cvd_config");
-    auto dir_contents = CF_EXPECT(DirectoryContents(config_path));
-    for (const std::string& file : dir_contents) {
-      std::string_view local_file(file);
-      if (absl::ConsumePrefix(&local_file, "cvd_config_") &&
-          absl::ConsumeSuffix(&local_file, ".json")) {
-        allowed_config_presets_.emplace(local_file);
+    if (DirectoryExists(config_path)) {
```
Databean: Can this be written as a [guard](https://en.wikipedia.org/wiki/Guard_(computer_science))?
```c++
if (!DirectoryExists(config_path)) {
  return {};
}
```

base/cvd/cuttlefish/host/libs/config/config_flag.cpp
```diff
@@ -144,8 +147,14 @@ class ConfigFlagImpl : public ConfigFlag {
     LOG(INFO) << "Launching CVD using --config='"
               << VectorizedFlagValue(configs_) << "'.";
     for (size_t i = 0; i < configs_.size(); ++i) {
-      Json::Value config_values =
-          CF_EXPECT(config_reader_.ReadConfig(configs_[i]));
+      Result<Json::Value> config_values_res =
+          config_reader_.ReadConfig(configs_[i]);
+      if (!config_values_res.ok()) {
+        LOG(WARNING) << "Config '" << configs_[i]
+                     << "' not found: " << config_values_res.error().Message();
```
Databean: Printing `error()` gives the full stack trace but `Message()` is reduced, is that intentional?

base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
```diff
@@ -105,6 +106,40 @@ Result<Command> ConstructCvdHelpCommand(
...
+  std::string exec_dir = android::base::GetExecutableDirectory();
+
+  std::string bin_path = exec_dir + "/" + bin_name;
+  CF_EXPECTF(FileExists(bin_path),
+             "Could not find {} in executable directory '{}'", bin_name,
+             exec_dir);
+
+  CF_EXPECTF(!bin_path.empty(),
+             "Could not find any of [{}] in executable directory '{}'",
+             bin_name, exec_dir);
```
Databean: Is there a scenario where `FileExists(bin_path)` is `true` but `bin_path` is empty?

base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
```diff
@@ -105,6 +106,40 @@ Result<Command> ConstructCvdHelpCommand(
...
+
+  Command command(bin_name);
+  command.SetExecutable(bin_path);
+  for (const auto& env_var: env) {
+    const std::string& var = env_var.first;
+    const std::string& value = env_var.second;
```
Databean: This can be a [structured binding](https://en.cppreference.com/cpp/language/structured_binding).

```c++
for (const auto& [var, value] : env) {
  ...
}
```

See also http://go/totw/169

base/cvd/cuttlefish/host/commands/cvd/cli/commands/bugreport.cpp
```diff
@@ -165,15 +166,16 @@ bool CvdBugreportCommandHandler::RequiresDeviceExists() const { return true; }
...
+      kHostBugreportBin, request.Env(), request.SubcommandArguments()));
   std::string stdout;
-  int res = RunWithManagedStdio(std::move(command), nullptr, &stdout, nullptr);
+  std::string stderr;
+  int res = RunWithManagedStdio(std::move(command), nullptr, &stdout, &stderr);
   // gflags returns exit code 1 when --help is given
-  CF_EXPECTF(res == 0 || res == 1,
-             "Failed to execute bugreport binary, exit code: {}", res);
+  if (res != 0 && res != 1) {
+    std::cerr << stderr;
```
Databean, line 176: iirc the default behavior of `RunWithManagedStdio` for a `nullptr` argument is to inherit the parent file descriptor. Does this capture `stderr` to intentionally hide it on a successful invocation?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
```diff
@@ -536,27 +535,17 @@ Result<std::string> CvdStartCommandHandler::DetailedHelp(
...
+      "cvd_internal_start", envs, request.SubcommandArguments()));
   std::string stdout;
-  int res = RunWithManagedStdio(std::move(command), nullptr, &stdout, nullptr);
+  std::string stderr;
+  int res = RunWithManagedStdio(std::move(command), nullptr, &stdout, &stderr);
   // gflags returns exit code 1 when --help is given
-  CF_EXPECTF(res == 0 || res == 1,
-             "Failed to execute start binary, exit code: {}", res);
+  if (res != 0 && res != 1) {
+    std::cerr << stderr;
```
Databean, line 546: iirc the default behavior of `RunWithManagedStdio` for a `nullptr` argument is to inherit the parent file descriptor. Does this capture `stderr` to intentionally hide it on a successful invocation?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
```diff
@@ -536,27 +535,17 @@ Result<std::string> CvdStartCommandHandler::DetailedHelp(
...
-      CF_EXPECT(AndroidHostPath(envs),
-                "\nTry running this command from the same directory as the "
-                "downloaded or fetched host tools.");
-  const auto bin = CF_EXPECT(FindStartBin(android_host_out));
-
-  Command command = CF_EXPECT(ConstructCvdHelpCommand(
-      bin, envs, request.SubcommandArguments(), request));
-  LOG(INFO) << "help command: " << command;
+  Command command = CF_EXPECT(ConstructSiblingHelpCommand(
+      "cvd_internal_start", envs, request.SubcommandArguments()));
```
Databean, line 540: This unconditionally uses the "sibling" version of `cvd_internal_start`, which might produce inaccurate flags for an older device build. Should it use the device version when it is present?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/bugreport.cpp
```diff
@@ -165,15 +166,16 @@ bool CvdBugreportCommandHandler::RequiresDeviceExists() const { return true; }
 
 Result<std::string> CvdBugreportCommandHandler::DetailedHelp(
     const CommandRequest& request) const {
-  std::string android_host_out = CF_EXPECT(AndroidHostPath(request.Env()));
-  Command command = CF_EXPECT(
-      ConstructCvdHelpCommand(kHostBugreportBin, request.Env(),
-                              request.SubcommandArguments(), request));
+  Command command = CF_EXPECT(ConstructSiblingHelpCommand(
+      kHostBugreportBin, request.Env(), request.SubcommandArguments()));
```
Databean, line 170: This unconditionally uses the "sibling" version of `cvd_internal_host_bugreport`, which might produce inaccurate flags for an older device build. Should it use the device version when it is present?

Databean (approved):

base/cvd/cuttlefish/host/libs/config/config_flag.cpp
jemoreira: It was intentional. Changed it to the full stack trace because it's only a benign error when handling the help flag, but stderr output is not shown in that case.

base/cvd/cuttlefish/host/libs/config/config_flag.cpp
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/bugreport.cpp
jemoreira: There are too many cases to take into account here. For example, the help command may run before substitutions are made by `cvd start` or the user may choose a different set of substitutions than the default as determined by debian_substitution_marker.

So, I chose the one that applies in most situations.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/bugreport.cpp
jemoreira: Yes.

It's not as important for bugreport, but start does print a lot of logs. When running with `--help` those logs are mostly irrelevant so they are only shown in case of failure.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
jemoreira: There are too many cases to take into account here. For example, the help command may run before substitutions are made by cvd start or the user may choose a different set of substitutions than the default as determined by debian_substitution_marker.

So, I chose the one that applies in most situations.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
jemoreira: Yes.

It prints a lot of logs to stderr. When running with --help those logs are mostly irrelevant so they are only shown in case of failure.

base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
jemoreira: Good point

base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
Databean: Fair enough.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/bugreport.cpp
Databean: Fair enough.

Databean (approved):

PR #2519, opened 2026-05-06
Load modules in postinst.

3405691582:
This alleviates the need to reboot after installation. We should also improve the module presence checks, but we'll tackle that separately.

Bug: 507983767


Databean:
Nice!

Databean (approved):

PR #2524, opened 2026-05-06
Refactor podcvd setup procedure

jemoreira:
Implements go/podcvd-setup

Introduces podcvd-setup script, /etc/podcvd.users config file for users
with podcvd access and IP range allocation, and the podcvd-acl-restore
system service.

Podcvd also ensures `podman.socket` is running before attempting to contact it.

Assisted-by: Gemini


container/debian/podcvd-setup
```diff
@@ -0,0 +1,128 @@
...
+    echo "Configuring subordinate UIDs/GIDs for '$username'..."
+    start_id=$(awk -F: '{print $2 + $3}' /etc/subuid /etc/subgid 2>/dev/null | sort -n | tail -1)
+    start_id="${start_id:-100000}"
+    id_range="$start_id-$((start_id + 65535))"
+    sudo usermod --add-subuids "$id_range" --add-subgids "$id_range" "$username"
+fi
+
+# 8. Pre-download Cuttlefish container image (Optional)
+echo "Pre-downloading Cuttlefish container image (stable)..."
+podman pull us-docker.pkg.dev/android-cuttlefish-artifacts/cuttlefish-orchestration/cuttlefish-orchestration:nightly || echo "Skip pre-downloading image. You might need to build it locally or pull it manually later."
```
0405ysj: Tag of CF container image is determined when building `cuttlefish-podcvd` debian package. If pre-downloading container image is desired, the tag name should be coming from there, instead of hardcoded one from the script.

container/debian/cuttlefish-podcvd.cuttlefish-podcvd.init
```diff
@@ -29,7 +29,7 @@
 if [ -f /etc/default/cuttlefish-podcvd ]; then
     . /etc/default/cuttlefish-podcvd
 fi
-podcvd_cidr=${podcvd_cidr:-192.168.80.0/24}
+podcvd_cidr=${podcvd_cidr:-192.168.80.0/20}
```
0405ysj: As I described from the design, we should choose another IP range if it's `/20`. From current status, both `cuttlefish-podcvd` and `cuttlefish-base` cannot be installed together. Of course those two don't need to be installed together, but it can be mattered during conversion period.

container/debian/cuttlefish-podcvd.install
```diff
@@ -1,4 +1,5 @@
-debian/cuttlefish-podcvd-prerequisites.sh /usr/lib/cuttlefish-podcvd/bin/
 src/podcvd/cuttlefish_mcp_server /usr/lib/cuttlefish-podcvd/cuttlefish-mcp-server/
 src/podcvd/gemini-extension.json /usr/lib/cuttlefish-podcvd/cuttlefish-mcp-server/
 src/podcvd/podcvd /usr/lib/cuttlefish-podcvd/bin/
+debian/podcvd-setup /usr/bin/
```
0405ysj: Current `podcvd` has binary under `/usr/lib/cuttlefish-podcvd/bin/`, and have a link under `/usr/bin`. I'm fine for both approaches, but I wish to have a unified approach between two.

container/debian/cuttlefish-podcvd.install
```diff
@@ -1,4 +1,5 @@
-debian/cuttlefish-podcvd-prerequisites.sh /usr/lib/cuttlefish-podcvd/bin/
```
0405ysj, line 1: It would need to modify e2e test script of podcvd as well.

container/debian/cuttlefish-podcvd.postinst
```diff
@@ -0,0 +1,47 @@
...
+    configure)
+        if [ ! -f "$USER_CONFIG" ]; then
+            echo "Initializing podcvd users registry at $USER_CONFIG..."
+            cat << 'EOF' > "$USER_CONFIG"
+# WARNING: Do not edit this file directly.
+# To configure users and allocate IP ranges, please run:
+# [USER=a_different_user] podcvd-setup
+#
+# Format: <username>:<allocated_cidr>
+EOF
```
0405ysj: Can we add this file into `container/debian/cuttlefish-podcvd.install` instead of creating a file via `cat`?

container/debian/podcvd-setup
```diff
@@ -0,0 +1,128 @@
...
+# 1. Read default CIDR from /etc/default/cuttlefish-podcvd or fallback
+podcvd_cidr="192.168.80.0/20"
+if [ -f /etc/default/cuttlefish-podcvd ]; then
+    # Source the config file to get podcvd_cidr if defined
+    # (Suppress errors if it's not a valid shell script)
+    . /etc/default/cuttlefish-podcvd 2>/dev/null || true
+fi
+
+# Parse the CIDR
+IFS='/ ' read -r ip mask <<< "$podcvd_cidr"
```
0405ysj, line 38: Probably `IFS='/'` instead of `IFS='/ '`?

container/debian/podcvd-setup
```diff
@@ -0,0 +1,128 @@
...
+IFS='.' read -r a b c d <<< "$ip"
+mask="${mask:-20}"
```
0405ysj: I think it's an error when either ip or mask is not set. As the default value is set as `192.168.80.0/20` now, probably default mask value isn't necessary.

container/debian/podcvd-setup
```diff
@@ -0,0 +1,128 @@
...
+
+if [ "$mask" -lt 16 ] || [ "$mask" -gt 24 ]; then
+    echo "Error: Unsupported subnet mask /$mask in $podcvd_cidr. Must be between 16 and 24."
+    exit 1
+fi
+
+# Calculate range of the third octet for allocation
+num_subnets=$(( 2 ** (24 - mask) ))
+start_subnet=$c
```
0405ysj: It can have an unexpected error when `podcvd_cidr` is set as `192.168.79.0/20`, for example. Probaly `AND` operator is required.

container/src/podcvd/internal/check.go
```diff
@@ -17,11 +17,29 @@ package internal
...
+		return fmt.Errorf("failed to get current user: %w", err)
+	}
+	username := u.Username
+
+	// Try to start Podman socket on-demand
+	ensurePodmanSocketRunning()
+
+	// Verify user is registered in /etc/podcvd.users
+	_, err = readUserCidrFromConfig(username)
+	if err != nil {
```
0405ysj: nit: 	`if _, err = readUserCidrFromConfig(username); err != nil {`?

container/src/podcvd/internal/check.go
```diff
@@ -32,13 +50,13 @@ func CheckDeviceAccessible() error {
...
-			return fmt.Errorf("device %q is not accessible: %w", device, err)
+			log.Printf("User setup is incomplete. Please execute `podcvd-setup` to configure necessary permissions.")
+			return fmt.Errorf("device %q is not accessible (permission denied): %w", device, err)
 		}
 	}
 	return nil
 }
+
+func ensurePodmanSocketRunning() {
+	_ = exec.Command("systemctl", "--user", "start", "podman.socket").Run()
```
0405ysj: Can we introduce a log when it failed?

container/src/podcvd/internal/main.go
```diff
@@ -267,7 +267,9 @@ func fleetAllCuttlefishHosts(ccm libcfcontainer.CuttlefishContainerManager) erro
 	if err := errors.Join(errs...); err != nil {
 		return err
 	}
-	combinedRes := cvdFleetResponse{}
+	combinedRes := cvdFleetResponse{
+		Groups: []any{},
+	}
```
0405ysj, line 272: Thanks for the fix!

container/debian/cuttlefish-podcvd.cuttlefish-podcvd.init
jemoreira: Chose 192.168.112.0/20

container/debian/podcvd-setup
jemoreira: `IFS='/ '` just allows extra white space, in case someone edits it carelessly. It doesn't make much difference.

container/debian/cuttlefish-podcvd.install
jemoreira: Done

container/debian/cuttlefish-podcvd.install
jemoreira: Used the symlink approach for podcvd-setup too.

container/debian/cuttlefish-podcvd.postinst
jemoreira: Done

container/debian/podcvd-setup
jemoreira: Done

container/debian/podcvd-setup
jemoreira: done

container/src/podcvd/internal/check.go
jemoreira: Done

container/src/podcvd/internal/check.go
jemoreira: Done

container/debian/podcvd-setup
jemoreira: Dropped it from podcvd-setup.

container/debian/podcvd-setup
0405ysj: That makes sense, thanks for the explanation!

container/debian/podcvd-setup
0405ysj, line 1: I could see `/usr/lib/cuttlefish-podcvd/bin/podcvd-setup: Permission denied` error from E2E test. This file should be executable.

0405ysj (approved):
LGTM, thanks for improving `podcvd`!

PR #2525, opened 2026-05-07
Expose operator into another port of the host to avoid port collision

0405ysj:
`podcvd` tries to bind `1443` port of specified IP address via pasta setup, but it fails to bind when operator installed via `cuttlefish-user` obtains same port. This change is for using another port rather than `1443`, when exposing operator running in containers.

Context: b/509403701


0405ysj:
@ikicha @k311093 Gentle ping, could I get review on this PR?

ikicha (approved):

PR #2527, opened 2026-05-07
Statically allocate the iptables path.

3405691582:
We resolve the path to iptables each time we call in to IptablesPath; we can make this slightly more efficient if we statically allocate the backing storage for the path string.

### DIFF
--- a/base/cvd/allocd/alloc_utils.cpp
+++ b/base/cvd/allocd/alloc_utils.cpp
@@ -22,6 +22,7 @@
 #include <string_view>
 #include <sstream>
 
+#include "absl/base/no_destructor.h"
 #include "absl/log/log.h"
 #include "absl/strings/str_cat.h"
 #include "absl/strings/str_format.h"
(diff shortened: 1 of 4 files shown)


base/cvd/allocd/alloc_utils.cpp
```diff
@@ -337,17 +356,10 @@ bool DestroyEthernetBridgeIface(std::string_view name,
...
-  Result<std::string> path = Search(Path(), "iptables");
-  if (path.ok()) {
-    return *path;
-  }
-
-  path = Search({"/usr/sbin", "/sbin"}, "iptables");
-  if (path.ok()) {
-    return *path;
-  }
+  static const std::string iptables_path = SearchForIptables().value_or("");
```
Databean: nit: following the [style guide](https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables), please use [`absl::NoDestructor`](https://github.com/abseil/abseil-cpp/blob/d0d4c599108be215585d7380d84a652ddd87607c/absl/base/no_destructor.h#L19) for variables with static storage duration that have types with nontrivial destructors.

base/cvd/allocd/alloc_utils.cpp
```diff
@@ -38,6 +38,25 @@
...
+  if (p.ok()) {
+    return p;
+  }
+
+  p = cuttlefish::Search({"/usr/sbin", "/sbin"}, "iptables");
+  if (p.ok()) {
+    return p;
+  }
+
+  return p;
```
Databean: nit: Since it's the same in both paths, can leave out the conditional?

Databean (approved):

base/cvd/allocd/alloc_utils.cpp
3405691582: Done.

base/cvd/allocd/alloc_utils.cpp
3405691582: Done.

PR #2529, opened 2026-05-07
Group-level logs

ser-io:


base/cvd/cuttlefish/host/commands/cvd/cli/commands/logs.cpp
```diff
@@ -54,12 +53,25 @@ Result<void> PrintLogsList(const std::string& dir) {
     }
     std::cout << filename;
     std::cout << std::endl;
-    return {};
   };
-  CF_EXPECT(WalkDirectory(dir, callback));
   return {};
 }
 
+Result<std::vector<std::string>> Filter(
```
jemoreira: Please choose a less generic name, one that more clearly conveys what this function does. Consider `RemoveInaccessibleFiles`.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/logs.cpp
```diff
@@ -54,12 +53,25 @@ Result<void> PrintLogsList(const std::string& dir) {
...
+    const std::vector<std::string>& filenames) {
```
jemoreira: Consider accepting the input by value instead of reference.  Then you can implement the function such that it modifies and returns the parameter. Something like this:
```
std::vector<std::string> Filter(std::vector<std::string> files) {
  auto removed_begin = std::remove_if(files.begin(), files.end(), [](const std::string& file){ return !IsAccessible(file);});
  files.erase(removed_begin, files.end());
  return files;
}
```
That implementation can avoid all copies of the vector if the parameter is moved into it (which is the case where this function is called as it's given a temporary object), otherwise it makes a single copy (at the call site) just like the original. See go/totw/117 for details.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/logs.cpp
```diff
@@ -54,12 +53,25 @@ Result<void> PrintLogsList(const std::string& dir) {
...
+  std::vector<std::string> result = {};
+  for (auto& filename : filenames) {
+    if (access(filename.data(), F_OK) == 0) {
```
jemoreira: nit: Consider using `c_str()` instead of `data()`. While `data` and `c_str` do the same thing since C++11, using `c_str` makes it clear that a null terminated string is required. `data` is more appropriate when you need to modify the contents of the string or don't care about the null character at the end.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/logs.cpp
jemoreira: Don't return `Result` if the function can't fail

base/cvd/cuttlefish/host/commands/cvd/cli/commands/logs.cpp
```diff
@@ -74,31 +86,41 @@ class CvdLogsHandler : public CvdCommandHandler {
...
+    logs_filenames.insert(logs_filenames.end(), group_names.begin(),
+                          group_names.end());
+    auto ins_names = CF_EXPECT(Filter(instance.LogsFilenames()));
+    logs_filenames.insert(logs_filenames.end(), ins_names.begin(),
+                          ins_names.end());
 
     if (opts.print_target.empty()) {
-      CF_EXPECT(PrintLogsList(logs_dir));
+      if (logs_filenames.empty()) {
+        LOG(INFO) << "There are no logs files available";
```
jemoreira: nit: "log files"

base/cvd/cuttlefish/host/commands/cvd/instances/local_instance.cpp
```diff
@@ -272,4 +272,16 @@ Result<SharedFD> LocalInstance::GetLauncherMonitor(
...
 
+std::vector<std::string> LocalInstance::LogsFilenames() const {
+  return {
+      instance_dir() + "/logs/" + "build_info.log",
+      instance_dir() + "/logs/" + "crosvm_openwrt_boot.log",
+      instance_dir() + "/logs/" + "crosvm_openwrt.log",
+      instance_dir() + "/logs/" + "kernel.log",
+      instance_dir() + "/logs/" + "launcher.log",
+      instance_dir() + "/logs/" + "logcat",
+      instance_dir() + "/logs/" + "modem_simulator.log",
```
jemoreira: Don't hardcode this list. Use every file from the directory that is not a group log file, otherwise this code will miss any new file added. It's OK to hardcode the group log list though.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/logs.cpp
ser-io: Done.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/logs.cpp
ser-io: Good reading, thanks. Done.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/logs.cpp
ser-io: Good to know. Done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/logs.cpp
ser-io: Done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/logs.cpp
ser-io: Done.

base/cvd/cuttlefish/host/commands/cvd/instances/local_instance.cpp
ser-io: This new method return the expected instance log names, not the existing log names. We should be able to know the expected log names an instance can have without having to launch the instance itself and then scan the logs directory.

I understand, that if you are creating a new LOG_NAME, we don't want the extra step of modifying this function local_instance.cpp, you should be able to declare your new LOG_NAME just once somewhere. I'm planning to solve this problem in a follow up PR aiming to localize LOG NAMES declarations in a single place. b/511186368

jemoreira: > We should be able to know the expected log names an instance can have without having to launch the instance itself and then scan the logs directory

Why?

> planning to solve this problem in a follow up PR aiming to localize LOG NAMES declarations in a single place

That sounds a lot more complex than just reading a directory. The logs command only needs to show the available logs, not alert that a log file is missing. A missing log file is usually an indication of a problem that is likely visible in one of the existing log files.

jemoreira: https://github.com/google/android-cuttlefish/pull/2537 is nice to have in the sense that it centralizes log names, but it doesn't force developers to use it for new logs nor add them to `cvd logs`.

ser-io: Changed the logic to return the list of files in the `/logs` directory screened by the list of defined log names. PTAL.

jemoreira: That doesn't solve the problem. It still won't return a file that is not in the list. Just drop that `kInstanceLogNames` list.

ser-io: Source of truth for log names is now log_names.h, not every file name under the `/logs` directory. If there's a filename under `logs` directory whose name hasn't been added to logs_names.h it shouldn't be considered a log file.

Relying on the logs directory for source of truth for log names have created the following problems:

1. Symlinks to group-level logs in the instance `/logs` that should be rather fetched from original place, cause the `/logs` directory won't be available if the instance fails to launch.
2.  Duplicated names: b/511264010, as there's no central place to see all the log names already used in cuttlefish.

I wanted to change this behavior and rely on a compiled list of log names.

jemoreira: > If there's a filename under `logs` directory whose name hasn't been added to logs_names.h it shouldn't be considered a log file.

Why? This is just making things harder for ourselves.

> 1. Symlinks to group-level logs in the instance `/logs` that should be rather fetched from original place, cause the `/logs` directory won't be available if the instance fails to launch.

I totally agree that there should be a hard-coded list of log file names for group level logs and that those could/should be removed from the instance logs obtained from the directory.

> 2. Duplicated names: b/511264010, as there's no central place to see all the log names already used in cuttlefish.
> 
> I wanted to change this behavior and rely on a compiled list of log names.

To summarize:
* It's ok to have a group-level list of log file names in the cvd code, as group level operations are conducted by cvd itself.
* It's great to have a centralized place for log file names.
* It's NOT ok to expect all log files will be added to that centralized list in the future. It's possible (even likely) that a new log file is added to that directory but not added to the list and then we'll get a bug report because the log is not available. We can avoid this very easily: Just assume that everything under the `logs` directory is a log file. Removing the duplicated group level logs is also simple enough.

ser-io: Removed the screening logic. PTAL

base/cvd/cuttlefish/host/commands/cvd/cli/commands/logs.cpp
```diff
@@ -55,10 +54,23 @@ Result<void> PrintLogsList(const std::string& dir) {
...
     std::cout << std::endl;
-    return {};
   };
-  CF_EXPECT(WalkDirectory(dir, callback));
-  return {};
+}
+
+std::vector<std::string> RemoveInaccessibleFilenames(
+    std::vector<std::string> filenames) {
+  auto removed_begin = std::remove_if(
```
jemoreira: nit: Use `std::erase_if`, which is the more modern version of remove_if.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/logs.cpp
```diff
@@ -73,31 +85,51 @@ CvdLogsHandler::CvdLogsHandler(InstanceManager& instance_manager)
...
 
-  std::string dir = instance.instance_dir();
-  std::string logs_dir = dir + "/logs";
-  if (!FileExists(logs_dir)) {
-    VLOG(0) << "Logs directory `" << logs_dir << "` does not exist.";
-    LOG(INFO) << "There are no logs files available";
-    return {};
+  std::vector<std::string> logs_filenames;
+  auto group_names = RemoveInaccessibleFilenames(group.LogsFilenames());
+  logs_filenames.insert(logs_filenames.end(), group_names.begin(),
```
jemoreira: nit: `logs_filenames` is still empty at this point. You can use the copy constructor at this line instead of inserting into an empty vector.

jemoreira (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/logs.cpp
ser-io: Done.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/logs.cpp
ser-io: Done.

PR #2531, opened 2026-05-07
Fix launch with kernel repacking

Databean:
At some point assemble_cvd expected `cpio` to return non-zero values on success, which is no longer true. Newer versions of cpio return 0 on success. Ignoring the return value of cpio seems to work, the kernel is replaced successfully.

Tested with:
```
bazel-bin/cuttlefish/package/cuttlefish-common/bin/cvd fetch --default_build=git_main/aosp_cf_x86_64_only_phone-trunk_staging-userdebug --target_directory=$HOME/dl_kernel --kernel_build=aosp_kernel-common-android-mainline/kernel_virt_x86_64
bazel-bin/cuttlefish/package/cuttlefish-common/bin/cvd create --host_path=$HOME/dl_kernel --product_path=$HOME/dl_kernel
```

Bug: b/427835411
Bug: b/510369069


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #2532, opened 2026-05-07
Create privileged executable for refreshing groups

Databean:
`acloud` relies on [`sg`](https://man7.org/linux/man-pages/man1/sg.1.html) to [refresh group membership](https://cs.android.com/android/platform/superproject/+/android-latest-release:tools/acloud/internal/lib/utils.py;l=1322;drc=02cef780efe034b789d0e3708d4929d2f236327b) if the user has joined the `cvdnetwork` / `kvm` / `render` groups but has not yet restarted or re-logged in.

Debian 13 changed providers of `sg` / `newgrp` from `shadow` to `util-linux`. In the `util-linux` implementation, `sg` only updates the primary group but does not update secondary groups.

- https://www.github.com/util-linux/util-linux/issues/4098
- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1130245

Since `sg` no longer works to refresh the apparent secondary group memberships, we'll need an alternative implementation.

Mechanism for raising privilege copied from https://www.github.com/google/android-cuttlefish/pull/1706

Sample invocation:
```
cvd_refresh_groups /usr/bin/groups groups
```

Bug: b/510899675

### DIFF
--- a/base/cvd/cuttlefish/host/commands/refresh_groups/cvd_refresh_groups.cc
+++ b/base/cvd/cuttlefish/host/commands/refresh_groups/cvd_refresh_groups.cc
@@ -1,0 +1,55 @@
+//
+// Copyright (C) 2026 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <grp.h>
+#include <limits.h>
+#include <stdio.h>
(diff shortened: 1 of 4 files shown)


base/cvd/cuttlefish/host/commands/refresh_groups/cvd_refresh_groups.cc
```diff
@@ -0,0 +1,38 @@
...
+ * cvd_refresh_groups /usr/bin/groups groups
+ * ```
+ */
+int main(int argc, char** argv) {
+  if (argc < 3) {
+    fprintf(stderr,
+            "Usage: cvd_refresh_groups <file> <argv[0]> [argv<N> ...]\n");
+    return 1;
+  }
+  char user[4096];
```
3405691582: and `#include <limits.h>`

```suggestion
  char user[LOGIN_NAME_MAX];
```

3405691582 (approved):

base/cvd/cuttlefish/host/commands/refresh_groups/cvd_refresh_groups.cc
Databean: Thanks, done.

Databean:
Thanks for reviewing!

PR #2533, opened 2026-05-08
Do not show full group JSON to users by default at the end of `cvd create`

cjreynol:
Show only the group and instance names instead, but leave the full JSON output for tools like the host orchestrator.

Bug: 511258310

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
@@ -20,6 +20,7 @@
 #include <stddef.h>
 #include <stdlib.h>
 #include <string.h>
+#include <unistd.h>
 
 #include <algorithm>
 #include <cstdlib>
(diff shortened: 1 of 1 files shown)


base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
```diff
@@ -449,8 +450,19 @@ Result<void> CvdStartCommandHandler::Handle(const CommandRequest& request) {
...
 
-  auto group_json = CF_EXPECT(group.FetchStatus());
-  std::cout << group_json.toStyledString();
+  // show user a more succinct output
+  if (isatty(0)) {
+    std::vector<std::string> instance_names;
+    for (const auto& instance : group.Instances()) {
+      instance_names.push_back(instance.name());
+    }
+    std::cout << fmt::format("group_name:{}|instance_name(s):{}\n",
```
jemoreira: nit: Drop `_name` from the messages.

jemoreira (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
cjreynol: Done.

PR #2534, opened 2026-05-08
Refresh groups if necessary when running `cvd` commands

Databean:
Check user groups in ConstructCommand and fallback to cvd_refresh_groups if necessary

`cvd_refresh_groups` is changed to use `getpwuid(getuid())` instead of `getlogin_r` to support the `sudo -u` flow used in the end-to-end tests.

Introduces a check in ConstructCommand to ensure the user is a member of cvdnetwork, kvm, and render groups. If not, it executes the command via cvd_refresh_groups.

Bug: b/510930737

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
@@ -25,14 +25,15 @@
 #include <utility>
 #include <vector>
 
-#include <android-base/file.h>
 #include "absl/strings/str_cat.h"
+#include "android-base/file.h"
 #include "fmt/format.h"
 #include "fmt/ranges.h"  // NOLINT(misc-include-cleaner): version difference
 
 #include "cuttlefish/common/libs/fs/shared_fd.h"
 #include "cuttlefish/common/libs/utils/contains.h"
 #include "cuttlefish/common/libs/utils/files.h"
+#include "cuttlefish/common/libs/utils/users.h"
 #include "cuttlefish/host/commands/cvd/instances/config_path.h"
 #include "cuttlefish/host/commands/cvd/utils/common.h"
 #include "cuttlefish/host/libs/config/config_constants.h"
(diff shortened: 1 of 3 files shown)


base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
```diff
@@ -55,9 +57,24 @@ Result<void> CheckProcessExitedNormally(siginfo_t infop,
   }
 }
 
+static Command FixGroupsIfNecessary(const std::string& command_name,
+                                    const std::string& bin_path) {
+  bool in_required_groups =
+      InGroup("cvdnetwork") && InGroup("kvm") && InGroup("render");
```
3405691582: As I understand it, render is only necessary when using qemu, right?

Also, I mentioned elsewhere about cvdnetwork. If I remember correctly we end up chowning the tun device this way. If the tun device ends up being owned by kvm, is there any point in using cvdnetwork? 

Maybe this is fine for now, if there isn't a clean way to address these points.

base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
```diff
@@ -55,9 +57,24 @@ Result<void> CheckProcessExitedNormally(siginfo_t infop,
...
+  if (in_required_groups) {
+    return Command(command_name).SetExecutable(bin_path);
+  } else {
```
3405691582: Don't need the else here; dropping it will make the early return pattern more self-evident.

3405691582 (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
Databean: I don't think `/dev/net/tun` is `chown`d anywhere, it belongs to the `root` user and group on my machine. You might be thinking of the udev rules that set the groups of `/dev/vhost-net` and `/dev/vhost-vsock` to be `cvdnetwork`.

https://github.com/google/android-cuttlefish/blob/018b6301387b455c35897fda8b481909ebdf8ec2/base/debian/cuttlefish-base.udev#L1-L2

This may be another issue with the pre-reboot flow, it depends if the udev rules are applied when the `modprobe` you added recently happens. There is a comparable `chown` for these devices now, but it's only applied in the docker case:

https://github.com/google/android-cuttlefish/blob/018b6301387b455c35897fda8b481909ebdf8ec2/base/debian/cuttlefish-base.cuttlefish-host-resources.init#L295-L302

@jemoreira suggested using `setfacl` as an alternative which was added to `podcvd` recently,

https://github.com/google/android-cuttlefish/blob/018b6301387b455c35897fda8b481909ebdf8ec2/container/debian/podcvd-setup#L102-L106

but this depends on knowing the username, which works better for something like `cvd setup` though it requires root.

Databean: Removed the check for `render`.

base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
Databean: Good point, rewrote it as a guard clause.

PR #2546, opened 2026-05-12
Some QoL improvements in cvd command handlers

jemoreira:
These changes are part of a larger effort to improve the `cvd help` command.

Bug: b/510057806


base/cvd/cuttlefish/host/commands/cvd/cli/commands/help.cpp
```diff
@@ -132,7 +130,7 @@ class CvdHelpHandler : public CvdCommandHandler {
   Result<void> PrintHandler(std::stringstream& help_message,
```
cjreynol: This no longer needs to return a `Result` after removing the `Result` from `SummaryHelp`.  It looks like that then applies to `TopLevelHelp` as well.

Since you are doing further cleanups, I am fine if you change this in a follow-up PR.

cjreynol (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/help.cpp
jemoreira: Done

Databean (approved):

PR #2553, opened 2026-05-12
Adds systemd unit file for cuttlefish-host-resources.service

ser-io:
Bug: b/473836983

### DIFF
--- a/base/debian/rules
+++ b/base/debian/rules
@@ -76,6 +76,12 @@ override_dh_installinit:
 	dh_installinit --name=cuttlefish-defaults
 	dh_installinit
 
+
+.PHONY: override_dh_installsystemd
+override_dh_installsystemd:
+	dh_installsystemd --name=cuttlefish-host-resources
+	dh_installsystemd
+
 # the `--workspace_status_command` flag path depends on the current working directory of base/cvd
 .PHONY: override_dh_auto_build
 override_dh_auto_build:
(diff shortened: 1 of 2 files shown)


Databean (approved):

base/debian/rules
```diff
@@ -76,6 +76,12 @@ override_dh_installinit:
 	dh_installinit --name=cuttlefish-defaults
 	dh_installinit
 
+
+.PHONY: override_dh_installsystemd
+override_dh_installsystemd:
+	dh_installsystemd --name=cuttlefish-host-resources
```
jemoreira, line 82: What happens if the package is installed in system that still has the init-to-systemd translator? Will it create two different instances of the same service?

jemoreira (approved):

base/debian/rules
ser-io: Good question. Only one instance of `cuttlefish-host-resources.service`, systemd uses the new systemd service unit file.

**With thew new systemd service unit.**

```
# systemctl status cuttlefish-host-resources.service
● cuttlefish-host-resources.service - Set up initial cuttlefish environment
     Loaded: loaded (/etc/init.d/cuttlefish-host-resources; enabled; preset: enabled)
     Active: active (exited) since Wed 2026-05-13 00:04:33 UTC; 13h ago
    Process: 24 ExecStart=/etc/init.d/cuttlefish-host-resources start (code=exited, status=0/SUCCESS)
        CPU: 368ms
```

**Without systemd service unit.**

```
● cuttlefish-host-resources.service - LSB: Set up initial cuttlefish environment
     Loaded: loaded (/etc/init.d/cuttlefish-host-resources; generated)
     Active: active (exited) since Wed 2026-05-13 14:09:26 UTC; 19s ago
       Docs: man:systemd-sysv-generator(8)
    Process: 24 ExecStart=/etc/init.d/cuttlefish-host-resources start (code=exited, status=0/SUCCESS)
        CPU: 349ms
```

PR #2561, opened 2026-05-13
Migrate iptables and ebtables usages in host resources script to nftables

dxapd:
Also adds a basic networking test and a workaround bazel arg to disable sandboxing to run locally (since some of the upstream images don't have the fix yet)

cvdalloc-related usages will be migrated in another PR.

b/500704721


base/debian/cuttlefish-base.cuttlefish-host-resources.init
```diff
@@ -114,6 +105,85 @@ destroy_tap() {
...
+    esac
+
+    if [ -z "$family" ] || [ -z "$table" ] || [ -z "$chain" ]; then
+        echo "Unknown rule type: $type"
+        return 1
+    fi
+
+    handle_file="${cuttlefish_run_dir}/${type}-${id}.handle"
+
+    if [ "$op" = "add" ]; then
```
Databean: Can this also use a `case` statement?

base/debian/cuttlefish-base.cuttlefish-host-resources.init
```diff
@@ -114,6 +105,85 @@ destroy_tap() {
...
+            echo "Contents required for add operation"
+            return 1
+        fi
+        # Add the rule and store the handle.
+        handle=$(nft -a -e add rule "$family" "$table" "$chain" "$contents" | awk '/# handle/ {print $NF}')
+        echo "$handle" > "$handle_file"
+    elif [ "$op" = "delete" ]; then
+        if [ -f "$handle_file" ]; then
+            # Find the handle file and use it to clean things up.
+            handle=$(cat "$handle_file")
```
Databean: nit: `$(< "$handle_file")`

base/debian/cuttlefish-base.cuttlefish-host-resources.init
```diff
@@ -114,6 +105,85 @@ destroy_tap() {
...
+
+    handle_file="${cuttlefish_run_dir}/${type}-${id}.handle"
+
+    if [ "$op" = "add" ]; then
+        if [ -z "$contents" ]; then
+            echo "Contents required for add operation"
+            return 1
+        fi
+        # Add the rule and store the handle.
+        handle=$(nft -a -e add rule "$family" "$table" "$chain" "$contents" | awk '/# handle/ {print $NF}')
```
Databean: Please add `Depends` entries for newly added tools, in this case `nft` and `awk`:

https://github.com/google/android-cuttlefish/blob/40a03a3f12e6218abfce72b8455d1d56a0dfd738/base/debian/control#L36-L40

base/debian/cuttlefish-base.cuttlefish-host-resources.init
```diff
@@ -114,6 +105,85 @@ destroy_tap() {
...
+    local op="$1"
+    local type="$2"
+    local id="$3"
+    local contents="$4"
+
+    local family=""
+    local table=""
+    local chain=""
+    local handle_file=""
+    local handle=""
```
Databean: Can the variable declarations for `handle` and `handle_file` be combined with the first assignment, or at least declared more closely?

https://google.github.io/styleguide/shellguide.html#s7.5-use-local-variables does mention though that

> Declaration and assignment must be separate statements when the assignment value is provided by a command substitution; as the `local` builtin does not propagate the exit code from the command substitution.

e2etests/cvd/networking_tests/main_test.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	if ruleRes, err := c.RunCmd("adb", "shell", "ip", "rule"); err == nil {
+		t.Logf("IP Rules:\n%s", ruleRes.Stdout)
+	} else {
+		t.Logf("failed to get ip rules: %v", err)
+	}
+}
+
+// Discover an IP address of the host that the guest can ping to verify default routing.
+func discoverPingableHostIP(t *testing.T) string {
```
Databean: What do you think of pinging a standard publicly available IP address like `8.8.8.8`?

e2etests/cvd/networking_tests/main_test.go
```diff
@@ -0,0 +1,151 @@
...
+func TestDeviceNetworking(t *testing.T) {
+	testcases := []struct {
+		branch string
+		target string
+	}{
+		{
+			branch: "aosp-android-latest-release",
+			target: "aosp_cf_x86_64_only_phone-userdebug",
+		},
+	}
```
Databean, line 39: Please add a test case for `git_main` as well.

e2etests/cvd/common/common.go
```diff
@@ -207,6 +207,9 @@ func (tc *TestContext) CVDCreate(args CreateArgs) error {
 	createCmd := []string{tc.TargetBin(), "--verbosity=DEBUG", "create"}
 	createCmd = append(createCmd, "--report_anonymous_usage_stats=y")
 	createCmd = append(createCmd, "--undefok=report_anonymous_usage_stats")
+	if os.Getenv("DISABLE_CROSVM_SANDBOX") == "true" {
+		createCmd = append(createCmd, "--enable_sandbox=false")
+	}
```
Databean, line 212: I don't see any reference to this environment variable in the PR or elsewhere in the repository, is this left over from testing?

base/debian/cuttlefish-base.cuttlefish-host-resources.init
Databean: An alternative to `awk` is using `nft -j` to output JSON and parsing it. This has the advantage of of relying on the [documented](https://manpages.debian.org/testing/libnftables1/libnftables-json.5.en.html) json output format, rather than the current output format, which is more of an implementation detail. JSON parsing is easier with a tool like `jq`, but using that has the downside of a new package depenency.

e2etests/cvd/networking_tests/main_test.go
dxapd: I chose this IP because it'd be pingable even if the host's network is down but tbh internet access is a prerequisite so 8.8.8.8 would let us skip this entire step and simplify the test case so i might as well do that

e2etests/cvd/common/common.go
dxapd: i actually did call it out in the PR description, but maybe I didn't write it clearly enough:

> a workaround bazel arg to disable sandboxing to run locally

So that I can run locally against Android images that don't have the fix in them yet.

Databean: Sorry, missed that part of the description.

base/debian/cuttlefish-base.cuttlefish-host-resources.init
dxapd: my first instinct says that if all of the possible variables are declared at the top of the function, in the case of a language like bash, it would make it easier to reason about the function body (since then you already know all of the variables you're going to see), what do you think?

base/debian/cuttlefish-base.cuttlefish-host-resources.init
dxapd: yeah it's a bit fragile like this. i'll weigh how to do that..

base/debian/cuttlefish-base.cuttlefish-host-resources.init
dxapd: yup

base/debian/cuttlefish-base.cuttlefish-host-resources.init
dxapd: TIL

e2etests/cvd/networking_tests/main_test.go
dxapd: okay

base/debian/cuttlefish-base.cuttlefish-host-resources.init
dxapd: okay

dxapd: i'll pull in jq yea

3405691582:
also see usages in base/cvd/allocd.

dxapd:
> also see usages in base/cvd/allocd.

yup - i was gonna tackle the cvdalloc usages in another PR (i should probably clarify that in the pr title)

base/debian/cuttlefish-base.cuttlefish-host-resources.init
```diff
@@ -44,21 +44,12 @@ ipv4_bridge=${ipv4_bridge:-1}
 ipv6_bridge=${ipv4_bridge:-1}
 dns_servers=${dns_servers:-8.8.8.8,8.8.4.4}
 dns6_servers=${dns6_servers:-2001:4860:4860::8888,2001:4860:4860::8844}
+cuttlefish_run_dir="/run/cuttlefish"
```
Databean: Can this be declared as `readonly`?

https://google.github.io/styleguide/shellguide.html#constants-environment-variables-and-readonly-variables

base/debian/cuttlefish-base.cuttlefish-host-resources.init
Databean: The [Google C++ style guide](https://google.github.io/styleguide/cppguide.html#Local_Variables) writes

> Place a function's variables in the narrowest scope possible, and initialize variables in the declaration.
>
> C++ allows you to declare variables anywhere in a function. We encourage you to declare them in a scope as local as possible, and as close to the first use as possible.

Though I can't find similar language in the [Google bash style guide](https://google.github.io/styleguide/shellguide.html) or the [Google Python style guide](https://google.github.io/styleguide/pyguide.html) which is relatively closer to bash.

I think seeing the `local` declaration closer to the assignment helps reassure the reader that it's not assigning to a global.

base/debian/cuttlefish-base.cuttlefish-host-resources.init
```diff
@@ -232,12 +306,12 @@ destroy_bridged_interfaces() {
...
-                $ebtables -t filter -D FORWARD  -p ipv4 --out-if "${tap}" -j DROP
-                $ebtables -t broute -D BROUTING -p ipv4 --in-if  "${tap}" -j DROP
+                manage_nft_rule delete forward4 "${tap}"
+                manage_nft_rule delete broute4 "${tap}"
             fi
             if [ "$ipv6_bridge" != "1" ]; then
-                $ebtables -t filter -D FORWARD  -p ipv6 --out-if "${tap}" -j DROP
-                $ebtables -t broute -D BROUTING -p ipv6 --in-if  "${tap}" -j DROP
+                manage_nft_rule delete forward6 "${tap}"
+                manage_nft_rule delete broute6 "${tap}"
```
Databean, line 316: Something that might be easier than making sure every `add` has a corresponding `delete` command is to make a dedicated directory below `$cuttlefish_run_dir` to hold the handles and iterate over the contents of the directory during shutdown, deleting both the rules and the files at cleanup.

Alternatively, the separate directory could still be useful as a validation step at cleanup to make sure every `add` was in fact covered by a `delete`.

Databean (approved):

base/debian/cuttlefish-base.cuttlefish-host-resources.init
dxapd: i see your point now, yeah

base/debian/cuttlefish-base.cuttlefish-host-resources.init
dxapd: hmm, i think i wanna spend some more time thinking about the safest way to handle that in follow-on, perhaps when i get around to adding host tests (since each option incl. keeping it like this has some more tradeoffs) - i'll leave it like this for now since i've verified it manually

base/debian/cuttlefish-base.cuttlefish-host-resources.init
dxapd: ya (til)

PR #2564, opened 2026-05-14
Centralize .clang-tidy configuration files

Databean:
The .clang-tidy configuration files throughout base/cvd now are all symlinks to files in base/cvd/clang_tidy_configs. This will make it easier to track the variations in our clang-tidy configurations and avoid out-of-sync versions.

Bug: b/513004576


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #2568, opened 2026-05-14
cvd: prune old log files in user log directory

Databean:
- Introduces PruneLogsDirectory to keep only the 30 most recent log files in PerUserDir() + "/logs".
- Refactors log initialization in main.cc to call this pruning function during startup.
- Refactors GetCvdLogFile into GetCvdLogFileName returning std::string.
- Adds a test for PruneLogsDirectory.

Bug: b/507600500

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/log_files.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/log_files.cpp
@@ -16,6 +16,7 @@
 
 #include "cuttlefish/host/commands/cvd/cli/log_files.h"
 
+#include <algorithm>
 #include <chrono>
 #include <iomanip>
 #include <iostream>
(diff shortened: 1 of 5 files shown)


base/cvd/cuttlefish/host/commands/cvd/cli/log_files.cpp
```diff
@@ -49,9 +58,35 @@ std::optional<std::string> GetCvdLogFile() {
...
+  if (non_log_files > 0) {
+    VLOG(0) << non_log_files << " non-log files found.";
+  }
+
+  std::sort(log_files.begin(), log_files.end());
+
+  if (log_files.size() <= retain) {
+    return {};
+  }
+  for (size_t i = 0; i < log_files.size() - retain; ++i) {
```
jemoreira: nit: You can avoid the extra `if` if you write the loop condition as `i + retain < log_files.size()`.

jemoreira (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/log_files.cpp
Databean: Thanks, was trying to figure out how to do this without underflow on the unsigned type but didn't see this solution.

PR #2569, opened 2026-05-15
Improve `cvd help start` output

jemoreira:
* Update the help summary to indicate that it always applies to groups and never individual instances.
* Add a detailed description of the command instead of depending on cvd_internal_start's (which doesn't have one).
* Filter flags from cvd_internal_start that the user is not supposed to set.
* Add its own flags (--host_substitutions only, since the selectors flags are handled elsewhere)

Bug: b/515472673


base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
```diff
@@ -297,6 +310,21 @@ Result<Command> ConstructCvdNonHelpCommand(const std::string& bin_file,
...
+  std::vector<Flag> flags =
+      CF_EXPECT(GetSiblingCommandFlags("cvd_internal_start", env, args));
+  // Remove flags set by cvd and intented to be exposed to the user
+  const std::vector<std::string> to_remove = {
+      "daemon", "instance_nums", "num_instances", "base_instance_num"};
+  flags.erase(std::remove_if(flags.begin(), flags.end(),
+                             [&to_remove](Flag f) {
+                               return Contains(to_remove, f.name());
+                             }),
+              flags.end());
```
Databean: There's [`std::erase_if`](https://en.cppreference.com/cpp/container/vector/erase2) in C++20.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.h
```diff
@@ -51,8 +51,14 @@ class CvdStartCommandHandler : public CvdCommandHandler {
...
                                          const CommandRequest& request);
 
+  // Flags handled by `cvd start` itself, not cvd_internal_start.
+  std::vector<Flag> BuildOwnFlags();
+
   InstanceManager& instance_manager_;
   SubprocessWaiter subprocess_waiter_;
+  struct {
+    std::vector<std::string> host_substitutions;
+  } own_flags_;
```
Databean, line 61: Is this an inner `struct` for organizational purposes?

base/cvd/cuttlefish/common/libs/utils/gflags_xml_parser_test.cpp
```diff
@@ -0,0 +1,86 @@
...
+                        Field(&GflagDescription::meaning, Eq("meaning1")),
+                        Field(&GflagDescription::default_value, Eq("def1")),
+                        Field(&GflagDescription::current_value, Eq("curr1")),
+                        Field(&GflagDescription::type, Eq("string"))),
+                  AllOf(Field(&GflagDescription::file, Eq("path/to/file2.cpp")),
+                        Field(&GflagDescription::name, Eq("flag2")),
+                        Field(&GflagDescription::meaning, Eq("meaning2")),
+                        Field(&GflagDescription::default_value, Eq("def2")),
+                        Field(&GflagDescription::current_value, Eq("curr2")),
+                        Field(&GflagDescription::type, Eq("int32")))));
```
Databean: Rather than this construction, what do you think of defining `operator==` on `GflagDescription` so that this test can create the expected `GflagDescription` instances normally for comparison?

base/cvd/cuttlefish/common/libs/utils/gflags_xml_parser_test.cpp
```diff
@@ -0,0 +1,86 @@
...
+using ::testing::Field;
+
+TEST(GflagsXmlParserTest, ParseValidXml) {
+  std::string xml = R"(<?xml version="1.0"?>
+<AllFlags>
+<program>some_program_name</program>
+<usage>Usage string, may contain a warning and parenthesis()</usage>
+<flag><file>path/to/file.cpp</file><name>flag1</name><meaning>meaning1</meaning><default>def1</default><current>curr1</current><type>string</type></flag>
+<flag><file>path/to/file2.cpp</file><name>flag2</name><meaning>meaning2</meaning><default>def2</default><current>curr2</current><type>int32</type></flag>
+</AllFlags>)";
```
Databean, line 55: nit: can this be formatted with indentation?

base/cvd/cuttlefish/common/libs/utils/gflags_xml_parser.h
```diff
@@ -0,0 +1,39 @@
...
+struct GflagDescription {
+  std::string file;
+  std::string name;
+  std::string meaning;
+  std::string default_value;
+  std::string current_value;
+  std::string type;
+};
+Result<std::vector<GflagDescription>> ParseGflagsXmlHelp(
+    const std::string& xml_help);
```
Databean: Can this accept a `std::string_view`? It looks like `xmlReadMemory` accepts both a pointer and a size, implying it doesn't need to be null-terminated.

base/cvd/cuttlefish/common/libs/utils/gflags_xml_parser.cpp
```diff
@@ -0,0 +1,82 @@
...
+        flag_desc.name = content;
+      } else if (tag_name == "meaning") {
+        flag_desc.meaning = content;
+      } else if (tag_name == "default") {
+        flag_desc.default_value = content;
+      } else if (tag_name == "current") {
+        flag_desc.current_value = content;
+      } else if (tag_name == "type") {
+        flag_desc.type = content;
+      }
```
Databean: This implementation is similar to `FlagsForSubprocess`:

https://github.com/google/android-cuttlefish/blob/e69253508f349ed09166c7f20dff9b8698527c60/base/cvd/cuttlefish/host/commands/start/flag_forwarder.cc#L226-L233

Is it possible to share anything here?

base/cvd/cuttlefish/common/libs/utils/flag_parser.cpp
```diff
@@ -300,19 +300,18 @@ bool Flag::WriteGflagsCompatXml(std::ostream& out) const {
 }
 
 std::ostream& operator<<(std::ostream& out, const Flag& flag) {
```
Databean, line 302: Should this print `name()` rather than everything in `aliases_`? It would probably be easier to read.

base/cvd/cuttlefish/common/libs/utils/gflags_xml_parser.cpp
```diff
@@ -0,0 +1,82 @@
...
+Result<std::vector<GflagDescription>> ParseGflagsXmlHelp(
+    const std::string& xml_help) {
+  std::unique_ptr<xmlDoc, void(*)(xmlDocPtr)> doc(
+      xmlReadMemory(xml_help.c_str(), xml_help.size(), nullptr, nullptr, 0),
+      xmlFreeDoc);
+  CF_EXPECT(doc != nullptr, "Failed to parse XML memory");
+
+  xmlNodePtr root_element = xmlDocGetRootElement(doc.get());
+  CF_EXPECT(root_element != nullptr, "XML document has no root element");
+  CF_EXPECT(std::string((char*)root_element->name) == "AllFlags",
```
Databean, line 38: Should this check if `root_element->name` is `nullptr` as well?

base/cvd/cuttlefish/common/libs/utils/gflags_xml_parser.cpp
```diff
@@ -0,0 +1,82 @@
...
+            "Root element is not AllFlags");
+
+  std::vector<GflagDescription> flags;
+  for (auto child = root_element->children; child != nullptr;
+       child = child->next) {
+    if (child->type != XML_ELEMENT_NODE) {
+      continue;
+    }
+    if (std::string((char*)child->name) != "flag") {
```
Databean, line 47: Should this check if `child->name` is `nullptr` as well?

base/cvd/cuttlefish/common/libs/utils/gflags_xml_parser.cpp
```diff
@@ -0,0 +1,82 @@
...
+      continue;
+    }
+
+    GflagDescription flag_desc;
+    for (auto flag_child = child->children; flag_child != nullptr;
+         flag_child = flag_child->next) {
+      if (flag_child->type != XML_ELEMENT_NODE) {
+        continue;
+      }
+      std::string tag_name((char*)flag_child->name);
```
Databean, line 57: Should this check if `flag_child->name` is `nullptr` as well?

base/cvd/cuttlefish/common/libs/utils/gflags_xml_parser.cpp
```diff
@@ -0,0 +1,82 @@
...
+      std::string content;
+      if (flag_child->children != nullptr &&
+          flag_child->children->type == XML_TEXT_NODE) {
+        content = (char*)flag_child->children->content;
```
Databean: Should this check if `flag_child->children->content` is `nullptr` as well?

base/cvd/cuttlefish/common/libs/utils/gflags_xml_parser.cpp
```diff
@@ -0,0 +1,82 @@
...
+  for (auto child = root_element->children; child != nullptr;
+       child = child->next) {
+    if (child->type != XML_ELEMENT_NODE) {
+      continue;
+    }
+    if (std::string((char*)child->name) != "flag") {
+      continue;
+    }
+
+    GflagDescription flag_desc;
```
Databean: This could be
```c++
GflagDescription& flag_desc = flags.emplace_back();
```
and then the functions below can modify it in place without needing to `push_back` at the end.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
```diff
@@ -501,24 +524,44 @@ Result<void> CvdStartCommandHandler::LaunchDeviceInterruptible(
...
+  ss << SummaryHelp() << "\n\n";
+  ss << kCommandDescription << "\n\n";
+  for (const Flag& flag : flags) {
+    ss << flag << std::endl;
   }
-  return stdout;
+  return ss.str();
+}
+
+std::vector<Flag> CvdStartCommandHandler::BuildOwnFlags() {
```
Databean, line 554: If this accepted a mutable reference to an instance of the `own_flags_` type, then:

- `own_flags_` doesn't have to be an instance member and
- `BuildOwnFlags()` doesn't have to be a class function and
- `DetailedHelp()` can still be `const`.

base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
```diff
@@ -186,6 +189,67 @@ Result<Command> ConstructCvdGenericNonHelpCommand(
...
+             args.end());
+  args.emplace_back("-helpxml");
+  Command command =
+      CF_EXPECT(ConstructSiblingHelpCommand(bin_name, env, args));
+  std::string stdout;
+  std::string stderr;
+  int res = RunWithManagedStdio(std::move(command), nullptr, &stdout, &stderr);
+  // gflags returns exit code 1 when --help is given
+  if (res != 0 && res != 1) {
+    std::cerr << stderr;
```
Databean: Should this be part of the `Result` rather than sent to the parent stderr?

base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
```diff
@@ -186,6 +189,67 @@ Result<Command> ConstructCvdGenericNonHelpCommand(
...
+  for (const GflagDescription& desc : gflag_descs) {
+    if (desc.name != "help" &&
+        android::base::Basename(desc.file).starts_with("gflags")) {
+      // Skip gflags-specific flags. Reporting these flags is quite noisy and
+      // adds little to no value in the context of cvd.
+      continue;
+    }
+    if (desc.type.starts_with("bool")) {
+      // A reference to bool variable is needed to make it generate the
+      // aliases for matching bool flags. The reference to the variable is
```
Databean: If `GflagsCompatBoolFlagBase` was exposed then this wouldn't need to create the dangling reference.

base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
```diff
@@ -186,6 +189,67 @@ Result<Command> ConstructCvdGenericNonHelpCommand(
...
+      CF_EXPECT(ParseGflagsXmlHelp(stdout));
+  std::vector<Flag> flags;
+  for (const GflagDescription& desc : gflag_descs) {
+    if (desc.name != "help" &&
+        android::base::Basename(desc.file).starts_with("gflags")) {
+      // Skip gflags-specific flags. Reporting these flags is quite noisy and
+      // adds little to no value in the context of cvd.
+      continue;
+    }
+    if (desc.type.starts_with("bool")) {
```
Databean: Instead of referencing `flags.back()` repeatedly below, these could store the reference returned by `emplace_back()`.

```c++
Flag& flag = flags.emplace_back(desc.type.starts_with("bool") ?GflagsCompatBoolFlagBase(desc.name) : GflagsCompatFlag(desc.name));
```

base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
```diff
@@ -186,6 +189,67 @@ Result<Command> ConstructCvdGenericNonHelpCommand(
...
+    // Add setter and getter that won't crash
+    flags.back().Setter(
+        [](const FlagMatch& match) -> Result<void> {
+        return {};
+        });
+    // The getter always returns the current value reported by the internal
+    // binary so the help message is accurate.
+    flags.back().Getter(
+        [value = desc.current_value]() { return value; });
+    flags.back().Help(desc.meaning);
```
Databean: These functions return the object being operated on, so it could be written as `Setter(...).Getter(...).Help(...)`.

base/cvd/cuttlefish/host/commands/cvd/cli/utils.h
```diff
@@ -65,6 +66,16 @@ struct ConstructNonHelpForm {
 Result<Command> ConstructCvdGenericNonHelpCommand(
     const ConstructNonHelpForm& request_form, const CommandRequest& request);
 
+// Returns the flags supported by a sibling command.
+// A sibling command is one who's executable resides in the same directory as
```
Databean: nit: "who's" is short for "who is" or "who has" so this should be "whose"

base/cvd/cuttlefish/host/commands/start/flag_forwarder.cc
```diff
@@ -27,7 +27,7 @@
 
 #include "absl/strings/strip.h"
 #include <gflags/gflags.h>
-#include <libxml/parser.h>
+#include "cuttlefish/common/libs/utils/gflags_xml_parser.h"
```
Databean: nit: include in the wrong block

base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
```diff
@@ -186,6 +189,67 @@ Result<Command> ConstructCvdGenericNonHelpCommand(
   return CF_EXPECT(ConstructCommand(construct_cmd_param));
 }
 
+Result<std::vector<Flag>> GetSiblingCommandFlags(const std::string& bin_name,
+                                                 const cvd_common::Envs& env,
+                                                 cvd_common::Args args) {
+  // Remove help-like flags to ensure --helpxml takes effect
+  args.erase(std::remove_if(args.begin(), args.end(),
```
Databean: This can use [`std::erase_if`](https://en.cppreference.com/cpp/container/vector/erase2).

base/cvd/cuttlefish/host/commands/start/flag_forwarder.cc
```diff
@@ -212,28 +182,23 @@ std::vector<gflags::CommandLineFlagInfo> FlagsForSubprocess(std::string helpxml_
...
 
-  xmlDocPtr doc = xmlReadMemory(helpxml_output.c_str(), helpxml_output.size(),
-                                NULL, NULL, 0);
-  if (doc == NULL) {
-    LOG(FATAL) << "Could not parse xml of subprocess `--helpxml`";
+  auto parse_result = ParseGflagsXmlHelp(helpxml_output);
+  if (!parse_result.ok()) {
+    LOG(FATAL) << "Could not parse xml of subprocess `--helpxml`: "
+               << parse_result.error();
   }
```
Databean: nit: `CHECK(parse_result.ok()) << ...`

base/cvd/cuttlefish/common/libs/utils/flag_parser.cpp
jemoreira: I considered it, but the problem is that it won't show other aliases and won't be correct for boolean flags. Plus, technically, you can name your flag anything you want irrespective of what it matches.

base/cvd/cuttlefish/common/libs/utils/gflags_xml_parser.cpp
jemoreira: No: libxml2 guarantees that element nodes have a valid name. It would be invalid xml otherwise.

base/cvd/cuttlefish/common/libs/utils/gflags_xml_parser.cpp
jemoreira: No: libxml2 guarantees that element nodes have a valid name. It would be invalid xml otherwise.

base/cvd/cuttlefish/common/libs/utils/gflags_xml_parser.cpp
jemoreira: No: libxml2 guarantees that element nodes have a valid name. It would be invalid xml otherwise.

base/cvd/cuttlefish/common/libs/utils/gflags_xml_parser.cpp
jemoreira: Done

base/cvd/cuttlefish/common/libs/utils/gflags_xml_parser.cpp
jemoreira: Done.

Text nodes produced by libxml2 parsing typically have non-null `content`, but it doesn't hurt to check.

base/cvd/cuttlefish/common/libs/utils/gflags_xml_parser.h
jemoreira: done

base/cvd/cuttlefish/common/libs/utils/gflags_xml_parser_test.cpp
jemoreira: Done

base/cvd/cuttlefish/common/libs/utils/gflags_xml_parser_test.cpp
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
jemoreira: My intention is to use the start flags from outside the handler (from the create command handler so it can print all the flags it accepts), not just from the help function. The OwnFlags structure is private to this class and I'd like to keep it that way. If I use a local variable in the function that builds all the flags then the returned flags will have dangling references.

The only alternative I found that would keep the help function `const` was to declare the `own_flags_` member `mutable`. That's what I did initially and then the last commit of this PR undoes it by making the functions non-const. Let me know if you prefer the `mutable` approach, otherwise I'd prefer to keep this one.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.h
jemoreira: Yes. TBH I was expecting more flags, but it's still nice to separate the flag from the other fields. Although I could have just called it `host_substitutions_flag`... WDYT?

base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
jemoreira: yes, done

base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/utils.h
jemoreira: done

base/cvd/cuttlefish/host/commands/start/flag_forwarder.cc
jemoreira: fixed

base/cvd/cuttlefish/host/commands/start/flag_forwarder.cc
jemoreira: done

base/cvd/cuttlefish/common/libs/utils/flag_parser.h
```diff
@@ -85,6 +87,8 @@ class Flag {
   Flag& Setter(std::function<Result<void>(const FlagMatch&)>) &;
   Flag Setter(std::function<Result<void>(const FlagMatch&)>) &&;
 
+  const std::string& name() const { return name_; }
```
Databean: nit: `Name()` for parity with other functions?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
```diff
@@ -73,6 +74,18 @@
...
+    R"(The `cvd start` command applies to the instance group, not specific instances
+because Cuttlefish instances in the same group must all be started (and stopped)
+in unisom.
+
+Flags that modify individual instances accept a comma separated list of values.
+If the number of values in one of these flags is less than the number of
+instances, then the last instances in the group will default to the first value
+in the list (as a result a single value provided applies to all instances). The
+empty string is interpreted as a valid value, to force a particular instance to
+use its intended default value the special value \"unset\" must be given.)";
```
Databean: Saying `\"unset\"` with backslashes may be confusing in output: this is already a raw string, so the backslashes are not stripped and a user will see them.

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
jemoreira: Done

base/cvd/cuttlefish/common/libs/utils/flag_parser.h
jemoreira: Done

The other functions modify the property, while this one only allows const reference access though.

PR #2571, opened 2026-05-15
feat: support dynamic bind mounting in podcvd

ikicha:
- Add heuristic parsing for dynamic host path resolution in cvd arguments.
- Handle edge-cases including symlinks and empty paths.
- Resolve duplicate mount bugs by tracking bind mapping with a composite "hostPath:containerPath" key.
- Optimize mount options by using Overlay (:O) for directories to allow Cuttlefish boot writes, and Read-Only (:ro) for files to bypass kernel/crun file overlay limits.
- Enhance parsing robustness to prevent panics on malformed bind strings and log unexpected path stat errors.

### DIFF
--- a/container/src/podcvd/internal/host.go
+++ b/container/src/podcvd/internal/host.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"log"
 	"math/rand"
 	"net"
 	"net/http"
@@ -37,11 +38,11 @@ import (
 	"github.com/docker/go-connections/nat"
 )
 
-func CreateCuttlefishHost(ccm libcfcontainer.CuttlefishContainerManager, commonArgs *CvdCommonArgs) error {
+func CreateCuttlefishHost(ccm libcfcontainer.CuttlefishContainerManager, cvdArgs *CvdArgs) error {
 	if err := pullContainerImage(ccm); err != nil {
 		return err
 	}
-	ip, err := createAndStartContainer(ccm, commonArgs)
+	ip, err := createAndStartContainer(ccm, cvdArgs)
 	if err != nil {
 		return err
 	}
@@ -245,7 +246,8 @@ func findAvailableGroupName(groupNameIpAddrMap map[string]string) string {
 	return fmt.Sprintf("%s%d", prefix, maxNum+1)
 }
 
-func createAndStartContainer(ccm libcfcontainer.CuttlefishContainerManager, commonArgs *CvdCommonArgs) (string, error) {
+func createAndStartContainer(ccm libcfcontainer.CuttlefishContainerManager, cvdArgs *CvdArgs) (string, error) {
+	commonArgs := cvdArgs.CommonArgs
 	attemptID := uuid.New().String()
 	containerCfg := &container.Config{
 		Env: []string{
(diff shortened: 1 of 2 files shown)


container/src/podcvd/internal/host.go
```diff
@@ -304,6 +305,72 @@ func createAndStartContainer(ccm libcfcontainer.CuttlefishContainerManager, comm
...
+		if strings.Contains(arg, "=") {
+			path = strings.SplitN(arg, "=", 2)[1]
+		}
+		
+		// 1. Filter out empty strings which would incorrectly mount the CWD
+		if strings.TrimSpace(path) == "" {
+			continue
+		}
+
+		if absPath, err := filepath.Abs(path); err == nil {
```
0405ysj: `continue`?

container/src/podcvd/internal/host.go
```diff
@@ -304,6 +305,72 @@ func createAndStartContainer(ccm libcfcontainer.CuttlefishContainerManager, comm
...
+					if !bindSet[key] {
+						pInfo, err := os.Stat(p)
+						if err != nil {
+							fmt.Fprintf(os.Stderr, "warning: failed to stat path %q to mount: %v\n", p, err)
+							continue
+						}
+						opt := "ro"
+						if pInfo.IsDir() {
+							opt = "O"
+						}
```
0405ysj: Do we have any reason for differentiating `ro`(readonly) and `O`(overlayfs), by checking if it's dir or not?

container/src/podcvd/internal/host.go
```diff
@@ -304,6 +305,72 @@ func createAndStartContainer(ccm libcfcontainer.CuttlefishContainerManager, comm
...
+	}
+
+	bindSet := make(map[string]bool)
+	for _, bind := range containerHostCfg.Binds {
+		// Existing bind format is "hostPath:containerPath:options"
+		host, container := parseBind(bind)
+		if host != "" {
+			bindSet[host+":"+container] = true
+		} else {
+			fmt.Fprintf(os.Stderr, "warning: ignoring malformed bind: %q\n", bind)
```
0405ysj: I think we can use `log.Printf` instead, which is used a lot of code in `podcvd`.

container/src/podcvd/internal/host.go
```diff
@@ -304,6 +305,72 @@ func createAndStartContainer(ccm libcfcontainer.CuttlefishContainerManager, comm
...
+			path = strings.SplitN(arg, "=", 2)[1]
+		}
+		
+		// 1. Filter out empty strings which would incorrectly mount the CWD
+		if strings.TrimSpace(path) == "" {
+			continue
+		}
+
+		if absPath, err := filepath.Abs(path); err == nil {
+			if _, statErr := os.Stat(absPath); statErr == nil {
```
0405ysj: `continue`?

container/src/podcvd/internal/host.go
```diff
@@ -304,6 +305,72 @@ func createAndStartContainer(ccm libcfcontainer.CuttlefishContainerManager, comm
...
+					pathsToMount = append(pathsToMount, realPath)
+				}
+
+				// 3. Add to mount configuration while preventing duplicates
+				for _, p := range pathsToMount {
+					key := fmt.Sprintf("%s:%s", p, p)
+					if !bindSet[key] {
+						pInfo, err := os.Stat(p)
+						if err != nil {
+							fmt.Fprintf(os.Stderr, "warning: failed to stat path %q to mount: %v\n", p, err)
```
0405ysj: ditto, use `log.Printf`?

container/src/podcvd/internal/host.go
```diff
@@ -304,6 +305,72 @@ func createAndStartContainer(ccm libcfcontainer.CuttlefishContainerManager, comm
...
+				if strings.HasPrefix(absPath, "/dev/") || strings.HasPrefix(absPath, "/sys/") || strings.HasPrefix(absPath, "/proc/") {
+					continue
+				}
+
+				pathsToMount := []string{absPath}
+				
+				// 2. Resolve symlinks and track the target file to ensure it's accessible inside
+				if realPath, err := filepath.EvalSymlinks(absPath); err == nil && realPath != absPath {
+					pathsToMount = append(pathsToMount, realPath)
+				}
```
0405ysj: Why do we append `realPath` rather than replacing into `realPath`?

container/src/podcvd/internal/host.go
ikicha: I don't understand the comment, did you mean by negating the condition and continue early?

container/src/podcvd/internal/host.go
ikicha: Ideally, I think every input path(both dir/file) of podcvdshould be `ro`, but some dir(like product_out if a user specifies explicitly) still requires to be writeable. To avoid conflicts, just made them be mounted as O. For files, we cannot mount it as O, but I think it is okay for file params.

container/src/podcvd/internal/host.go
ikicha: In this case, we need to modify args in cvd, right? But I think we don't want to modify params directly for the most case. If so, we need to mount both symlink and realpath.

0405ysj: That makes sense that accessing via symlink in the container too.

container/src/podcvd/internal/host.go
0405ysj: Understood, it makes sense as we cannot prepare with `:O` option for files.

0405ysj (approved):

PR #2572, opened 2026-05-15
Add systemd unit file for cuttlefish-podcvd.service

0405ysj:
Context: b/473836983, b/510167998

### DIFF
--- a/container/debian/cuttlefish-podcvd.cuttlefish-podcvd.service
+++ b/container/debian/cuttlefish-podcvd.cuttlefish-podcvd.service
@@ -1,0 +1,21 @@
+[Unit]
+SourcePath=/etc/init.d/cuttlefish-podcvd
+Description=Setup initial environment to use podcvd
+After=network-online.target
+After=remote-fs.target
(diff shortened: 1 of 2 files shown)


rmuthiah (approved):

0405ysj:
Thanks for the review!

container/debian/cuttlefish-podcvd.cuttlefish-podcvd.service
```diff
@@ -0,0 +1,21 @@
+[Unit]
+SourcePath=/etc/init.d/cuttlefish-podcvd
+Description=Setup initial environment to use podcvd
```
ikicha, line 3: QQ: why do we need systemd service for 'initial' env setup? can't it be done in postinstall script in deb pkg?

0405ysj: Because of network interface(named `podcvd`) configuration for linking and routing. It wasn't even triggered during post-installation, when SysV init is not compatible on the machine.

PR #2573, opened 2026-05-15
Set up shared library targets.

3405691582:
Here we create Bazel cc_shared_library targets for the result, libbase, files, and fs libraries. These were determined to have the most usages across all targets.

Since these libraries have external dependencies, they must be resolved by dynamic linkage as well. This necessitates creating some targets to facilitate this. We put this at top level rather than creating new directory structure, since we might want to adjust how this looks.

We also do the needful to ensure these libraries are packaged. This will incur a slight bit of bloat, but we'll perform the migration soon enough that this should be temporary and possibly not even noticed.

### DIFF
--- a/base/cvd/BUILD.bazel
+++ b/base/cvd/BUILD.bazel
@@ -1,5 +1,5 @@
 load("@rules_cc//cc:cc_import.bzl", "cc_import")
-load("@rules_cc//cc:cc_library.bzl", "cc_library")
+load("@rules_cc//cc:cc_shared_library.bzl", "cc_shared_library")
 
 package_group(
     name = "android_cuttlefish",
@@ -77,4 +77,47 @@ cc_import(
     name = "gfxstream_backend_import",
     shared_library = ":gfxstream_backend",
     visibility = ["//visibility:public"],
+)
+
+cc_shared_library(
+    name = "absl_shared",
+    exports_filter = [
+        "@abseil-cpp//:__subpackages__",
+    ],
+    shared_lib_name = "libcuttlefish_absl.so",
+    deps = [
+        "@abseil-cpp//absl/base",
+        "@abseil-cpp//absl/cleanup",
+        "@abseil-cpp//absl/container:flat_hash_map",
+        "@abseil-cpp//absl/container:flat_hash_set",
+        "@abseil-cpp//absl/debugging:stacktrace",
+        "@abseil-cpp//absl/debugging:symbolize",
+        "@abseil-cpp//absl/flags:flag",
+        "@abseil-cpp//absl/flags:parse",
+        "@abseil-cpp//absl/hash",
+        "@abseil-cpp//absl/log",
+        "@abseil-cpp//absl/log:check",
+        "@abseil-cpp//absl/log:globals",
+        "@abseil-cpp//absl/log:initialize",
+        "@abseil-cpp//absl/numeric:int128",
+        "@abseil-cpp//absl/status",
+        "@abseil-cpp//absl/status:statusor",
+        "@abseil-cpp//absl/strings",
+        "@abseil-cpp//absl/strings:cord",
+        "@abseil-cpp//absl/strings:internal",
+        "@abseil-cpp//absl/strings:str_format",
+        "@abseil-cpp//absl/synchronization",
+        "@abseil-cpp//absl/time",
+    ],
+)
+
+cc_shared_library(
+    name = "fmt_shared",
+    exports_filter = [
+        "@fmt//:__subpackages__",
+    ],
+    shared_lib_name = "libcuttlefish_fmt.so",
+    deps = [
+        "@fmt",
+    ],
 )
--- a/base/cvd/cuttlefish/result/BUILD.bazel
+++ b/base/cvd/cuttlefish/result/BUILD.bazel
@@ -1,3 +1,4 @@
+load("@rules_cc//cc:cc_shared_library.bzl", "cc_shared_library")
 load("//cuttlefish/bazel:rules.bzl", "cf_cc_library", "cf_cc_test")
 
 package(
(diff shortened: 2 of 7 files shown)


base/cvd/cuttlefish/result/BUILD.bazel
```diff
@@ -27,6 +27,19 @@ cf_cc_library(
     ],
 )
 
+cc_shared_library(
+    name = "result_shared",
```
Databean, line 32: Is the intention to eventually replace the non-shared versions of these targets?

base/cvd/BUILD.bazel
```diff
@@ -78,3 +78,46 @@ cc_import(
     shared_library = ":gfxstream_backend",
     visibility = ["//visibility:public"],
 )
+
+cc_shared_library(
+    name = "absl_shared",
```
Databean, line 83: Another approach is to start with the `cc_binary` executable targets, and set [`linkstatic = False`](https://bazel.build/reference/be/c-cpp#cc_binary.linkstatic).

The approach of creating a distinct `*_shared` target risks including the same library both dynamically and statically through different include paths, which can lead to issues like duplicates of global constants between the dynamic and statically included version.

base/cvd/cuttlefish/result/BUILD.bazel
3405691582: They will eventually be used as the dependency target, but AIUI both targets are necessary.

base/cvd/BUILD.bazel
3405691582: If I remember correctly, doing that means all library dependencies will be dynamic, and that makes things difficult on the packaging side. This affords more control: letting us specify which libraries we want to be dynamic and which dependencies are still linked statically.

Databean: The control doesn't extend into external targets. Without `linkstatic = False`, a library like abseil can still choose to statically link internally. As a specific example, consider the [`logging_initialized`](https://github.com/abseil/abseil-cpp/blob/b0e44a92f83267ecc17c7423e9aa8e4e5d7485b3/absl/log/internal/globals.cc#L39) global variable. If one intermediate library in Cuttlefish depends on `absl_shared` while another depends on `@abseil-cpp//absl/log`, then the two paths will lead to different locations in memory for this global.

Even if every Cuttlefish target is modified to depend on the new `absl_shared` target, it doesn't help for other libraries that we depend on. Targets in `@protobuf` will always [link directly to](https://github.com/protocolbuffers/protobuf/blob/a37dbd643372b05de834190941671a1563d7e499/src/google/protobuf/util/BUILD.bazel#L63) `@abseil-cpp//absl/log:absl_log`.

3405691582: > The control doesn't extend into external targets. 

If I take cvdalloc and add linkstatic = False:

```
$ llvm-objdump -p bazel-bin/cuttlefish/host/commands/cvdalloc/cvdalloc | egrep 'RUN|NEED'
...
  NEEDED       libcuttlefish_Sresult_Sliberror_Utype.so
  NEEDED       liblibbase_Sliblibbase.so
  NEEDED       liblibbase_Sliblibbase_Uposix_Ustrerror.so
  NEEDED       libexternal_Sfmt+_Slibfmt.so
  NEEDED       libexternal_Sabseil-cpp+_Sabsl_Sflags_Slibparse.so
  NEEDED       libexternal_Sabseil-cpp+_Sabsl_Sflags_Slibusage.so
...
$ llvm-objdump -p bazel-bin/cuttlefish/host/commands/cvdalloc/cvdalloc | egrep 'NEED' | wc -l
155
```

On the other hand, following the approach in this pr, when migrated:

```
$ llvm-objdump -p bazel-bin/cuttlefish/host/commands/cvdalloc/cvdalloc | egrep 'NEED'
  NEEDED       libm.so.6
  NEEDED       libcuttlefish_absl.so
  NEEDED       libcuttlefish_fmt.so
  NEEDED       libcuttlefish_result.so
  NEEDED       libcuttlefish_libbase.so
  NEEDED       libc.so.6
  NEEDED       ld-linux-x86-64.so.2
  VERNEED      0x0000000000008a44
  VERNEEDNUM   0x0000000000000003
```

> If one intermediate library in Cuttlefish depends on absl_shared while another depends on @abseil-cpp//absl/log, then the two paths will lead to different locations in memory for this global.

My vague understanding of how bazel handles this is that bazel will actually ensure that the shared version of the library will be used when specified correctly. I'll test that empirically next week.

3405691582: >> If one intermediate library in Cuttlefish depends on absl_shared while another depends on @abseil-cpp//absl/log, then the two paths will lead to different locations in memory for this global.
>
> My vague understanding of how bazel handles this is that bazel will actually ensure that the shared version of the library will be used when specified correctly. I'll test that empirically next week.

I've empirically tested this briefly. `//cuttlefish/host/commands/defaults:cf_defaults` depends on `//cuttlefish/common/libs/key_equals_value` which depends on `@abseil-cpp//absl/strings`. If I do not use `dynamic_deps` and build:

```
$ nm bazel-bin/cuttlefish/host/commands/defaults/cf_defaults | grep ' T ' | grep -i absl | wc -l
1036
```

as expected: absl text symbols are statically linked into the binary.

If I add `dynamic_deps = ["//cuttlefish/result:result_shared", "//libbase:libbase_shared"]` to the `cf_defaults` target and rebuild:

```
$ nm bazel-bin/cuttlefish/host/commands/defaults/cf_defaults | grep ' T ' | grep -i absl | wc -l
0
```

as expected.

Does that alleviate your concern?

Databean: Yes, thank you for doing this validation.

Databean (approved):

3405691582:
Reverting again in #2625 since this may have unintended side effects.

PR #2580, opened 2026-05-19
Remove acl-restore.sh, and add do setfacl via udev instead

0405ysj:
Context: b/514558958


container/debian/podcvd-setup
```diff
@@ -131,4 +125,31 @@ else
...
+    if ! id -nG "$username" | grep -qw "$group"; then
+        groups_to_add+=("$group")
+    fi
+done
+if [ ${#groups_to_add[@]} -gt 0 ]; then
+    unique_groups=($(echo "${groups_to_add[@]}" | tr ' ' '\n' | sort -u | tr '\n' ' '))
+    for unique_group in "${unique_groups[@]}"; do
+        sudo usermod -aG "$unique_group" "$username"
+    done
+    echo "Warning: Please reboot the machine (or log out and log back in) before using podcvd."
```
jemoreira: The whole point of having the restore acls service is that logging out and back in is not necessary. Also, the groups for the device nodes can change, in fact cuttlefish-base changes them, so depending on those can be risky.

0405ysj: I believe the change around `cuttlefish-base` is https://github.com/google/android-cuttlefish/pull/2589, but it's fine as it doesn't introduce any new group in overall.(Needed `kvm` and `cvdnetwork` before, but will need `kvm` only) Though I agree that groups for device nodes could be changed, I don't think doing `chown` on device nodes is a common behavior.

The most critical problem I faced at was using `setfacl` could be problematic for some machines. Machine details are described at b/514558958. While creating Cuttlefish instances, udev change action on `/dev/kvm` is triggered, and it flushes existing ACL but not recover it.

I think we can make a conversation around `acl-restore.sh` later but resolving the critical problem is urgent, because it's highly coupled with deployment of its skill. So I'd separate this PR into two, one for doing `sudo usermod -aG "$unique_group" "$username"` (https://github.com/google/android-cuttlefish/pull/2590) and the other for removing `acl-restore.sh`. I'm targeting to merge former one within today, but please feel free to leave comments afterwards.

.github/workflows/presubmit.yaml
```diff
@@ -522,8 +522,9 @@ jobs:
       run: |
         # Grant permission on devices as it's hard to grant on test env
         sudo chmod 666 /dev/kvm /dev/vhost-net /dev/vhost-vsock
+        sudo setfacl -m u:$USER:rw /dev/vhost-vsock
         cd e2etests
-        bazel test \
+        sudo -E -u $USER bazel test \
```
jemoreira: why is this `sudo` call needed here? The `setfacl` call above should take effect immediately, maybe you need to run a similar one for kvm and vhost-net?

container/debian/podcvd-setup
```diff
@@ -99,18 +99,12 @@ echo "${username}:${assigned_range}" | sudo tee -a "$USER_CONFIG" > /dev/null
 
 echo "Successfully allocated range $assigned_range to user '$username'."
 
-# 5. Apply ACLs immediately (requires sudo)
-echo "Configuring device permissions..."
-sudo setfacl -m "u:$username:rw" /dev/kvm
-sudo setfacl -m "u:$username:rw" /dev/vhost-net
-sudo setfacl -m "u:$username:rw" /dev/vhost-vsock
```
jemoreira, line 106: I understand adding the user to the `kvm` group eliminates the need for restore-acls. But why detele these? without them the user will have to create a new session.

0405ysj: Because creating a new session may be necessary though setting ACL via `setfacl`. Those settings could be flushed by triggering udev change action while creating Cuttlefish instance.

jemoreira: > triggering udev change action while creating Cuttlefish instance

I am missing something here. Creating a cuttlefish instance is done without root privileges, how can that trigger changes in device nodes?

.github/workflows/presubmit.yaml
0405ysj: I'd try to simplify this part. There's a complicated problem here for handling device node permissions, as both doing `chmod 666` or `setfacl` can be flushed depending on its udev configuration.

container/debian/podcvd-setup
0405ysj: 1. While creating cuttlefish instance, it triggers udev change action. I don't know the exact mechanism how udev change action is triggered, but I'm speculating that an event is triggered while interacting with those devices.
1. From udev rule for `/dev/kvm`, `uaccess` is set. When it's set, it flushes current ACL setting and sets ACL dynamically depending on 'seat'. Physical workstations were okay, but it mattered on virtual environments.

.github/workflows/presubmit.yaml
0405ysj: Removed `sudo -E -u $USER` with some further checks.

container/debian/podcvd-setup
jemoreira: ~~`uaccess` shouldn't trigger for a user connected remotely, were you using CRD?~~

Instead of creating ephemeral ACLs, can `podcvd setup` add udev rules to grant the current user access to the devices? And then make udev reload devices.

Gemini suggested rules like this:
```
# Apply custom ACLs every time the device is created/triggered
KERNEL=="vhost-net", RUN+="/usr/bin/setfacl -m u:theuser:rw /dev/%k"
KERNEL=="vhost-vsock", RUN+="/usr/bin/setfacl -m u:theuser:rw /dev/%k"
KERNEL=="kvm", RUN+="/usr/bin/setfacl -m u:theuser:rw /dev/%k"
```

0405ysj: Oh, I guess that would be working! Let me try.

I was using CRD, and I was speculating that was the underlying cause why `uaccess` mattered.

0405ysj: Huge thanks, it's working! I've checked from CRD environment with and without rebooting.

jemoreira (approved):

PR #2582, opened 2026-05-19
Migrate to using shared libraries.

3405691582:
In #2573 we set up some new targets to build shared libraries. This commit migrates all binary targets with dependencies on //cuttlefish/result, //libbase, //cuttlefish/common/libs/fs:fs, and //cuttlefish/common/utils:files to use the shared library targets instead.

This results in the generated deb decreasing from 110M to 103M and the reported Installed-Size decreasing from 450179K to 414364K, a reduction of 35M. Actual on-disk build size has reduced from 2.4G to 2.2G. More aggressive reductions could likely be achieved by migrating other common libraries to shared libraries.

While we are here: we also ensure //cuttlefish/result has a dependency on fmt, to ensure that inline usages (like in casimir_control_server) correctly depend on fmt.

### DIFF
--- a/base/cvd/cuttlefish/common/frontend/socket_vsock_proxy/BUILD.bazel
+++ b/base/cvd/cuttlefish/common/frontend/socket_vsock_proxy/BUILD.bazel
@@ -13,6 +13,11 @@ cf_cc_binary(
         "server.h",
         "socket_vsock_proxy.cpp",
     ],
+    dynamic_deps = [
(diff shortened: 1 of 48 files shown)


3405691582:
(In fact, pulling out //cuttlefish/host/libs/config:cuttlefish_config will result in more storage reduction, but we'll handle this in a separate pr.)

Databean (approved):

PR #2584, opened 2026-05-19
Replace `cvd load` with `cvd create --config_file` in e2etests

chihchiachen:
`cvd load` is deprecated

Bug: 512983143
Test: `bazel build //cvd/cvd_load_tests:cvd_load_tests`

### DIFF
--- a/e2etests/cvd/common/common.go
+++ b/e2etests/cvd/common/common.go
@@ -285,13 +285,13 @@ func (tc *TestContext) CVDPowerwash() error {
 	return nil
 }
 
-// Common parameters for `cvd load`.
+// Common parameters for `cvd create --config_file`.
 type LoadArgs struct {
 	LoadConfig string
 }
 
-// Performs `cvd load`.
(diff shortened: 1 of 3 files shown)


e2etests/cvd/common/common.go
```diff
@@ -285,12 +285,12 @@ func (tc *TestContext) CVDPowerwash() error {
...
 }
 
-// Common parameters for `cvd load`.
+// Common parameters for `cvd create --config_file`.
 type LoadArgs struct {
 	LoadConfig string
 }
 
-// Performs `cvd load`.
+// Performs `cvd create --config_file`.
```
jemoreira, line 293: You should rename the function too. `CVDCreateWithConfigFile` would be my choice.

jemoreira (approved):

Databean (approved):

e2etests/cvd/common/common.go
chihchiachen: Sounds good - let me rename the function as you suggested.

PR #2588, opened 2026-05-19
Add selector flags to `cvd help start`

jemoreira:
New `GflagsCompatFlag` overloads allow creating flags bound to `std::optional<std::string>` and `std::optional<std::vector<std::string>>`.

Bug: b/515472673


base/cvd/cuttlefish/common/libs/utils/flag_parser.cpp
```diff
@@ -60,6 +61,123 @@ constexpr std::array help_str_opts{
...
+  }
+}
+
+template <typename T>
+std::string FlagValueToString(const T& value) {
+  return fmt::format("{}", value);
+}
+
+template <typename T>
+Result<void> ParseFlagValue(std::string_view str, T& value);
```
Databean: Rather than accept a `T& value`, can this be written as
```c++
template<typename T>
Result<T> ParseFlagValue(std::string_view);
```

That satisfies the [style guide](https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs) rule,
>
> Prefer using return values over output parameters: they improve readability, and often provide the same or better performance. See [TotW #176](https://abseil.io/tips/176).

and the current return type has some room to pass more information.

base/cvd/cuttlefish/common/libs/utils/flag_parser.cpp
Databean, line 810: This file is starting to get kind of long. What do you think about splitting out all the `GflagsCompatFlag` implementations into a separate file?

In order to avoid having to fix the include paths in the rest of the codebase at the same time while keeping [include cleaner](https://clangd.llvm.org/design/include-cleaner) happy, `flag_parser.h` can be replaced with a header that includes the other relevant headers with [`// IWYU pragma: export`](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md#iwyu-pragma-export). An example of this is [`result.h`](https://github.com/google/android-cuttlefish/blob/d2257473dfef9750deff5f165695769d8cd379d6/base/cvd/cuttlefish/result/result.h#L18). Alternatively, there's also the strategy used by [`cuttlefish_config.h`](https://github.com/google/android-cuttlefish/blob/d2257473dfef9750deff5f165695769d8cd379d6/base/cvd/cuttlefish/host/libs/config/BUILD.bazel#L109) which has multiple source files implement one header.

base/cvd/cuttlefish/host/commands/cvd/cli/selector/BUILD.bazel
```diff
@@ -53,6 +42,7 @@ cf_cc_library(
     srcs = ["parser_ids_helper.cpp"],
     hdrs = ["parser_ids_helper.h"],
     deps = [
+        "selector_common_parser",
```
Databean: nit: `:selector_common_parser`?

base/cvd/cuttlefish/common/libs/utils/flag_parser_test.cpp
```diff
@@ -424,4 +424,108 @@ TEST(FlagParser, ConsumesConstrainedSeparated) {
...
+TEST(FlagParser, OptionalStringFlag) {
+  std::optional<std::string> value;
+
+  // Default options: FlagNotPresent
+  auto flag_default = GflagsCompatFlag("myflag", value);
+
+  // Not present, should remain default (nullopt)
+  value = std::nullopt;
+  ASSERT_THAT(flag_default.Parse({}), IsOk());
+  ASSERT_FALSE(value.has_value());
```
Databean, line 436: `ASSERT_*` statements are fatal when they fail, in contrast to `EXPECT_*` statements. The value in non-fatal assertions is that the test can continue to validate more properties or use cases and produce more information without stopping at the first failure.

In this function, both the `Parse` calls and the the terminating assertion for each block could likely be an `EXPECT_*` statement rather than an `ASSERT_*` statement. For this block, even if this line fails, the test could still continue safely. For the block below, `ASSERT_TRUE(value.has_value());` has to be an `ASSERT` so that `*value` can be evaluated without crashing, but nothing after `ASSERT_EQ(*value, "foo");` relies on that property being true, so it can be an `EXPECT`.

Same concern for `OptionalStringVectorFlag`.

base/cvd/cuttlefish/host/commands/cvd/cli/selector/selector_common_parser.cpp
```diff
@@ -55,46 +50,35 @@ Result<std::vector<std::string>> HandleInstanceNames(
...
   }
 
-  if (instance_names) {
-    ret.instance_names = CF_EXPECT(HandleInstanceNames(*instance_names));
+  if (opts.instance_names) {
+    CF_EXPECT(ValidateInstanceNames(*opts.instance_names));
   }
-  return ret;
+  return {};
 }
```
Databean: Part of the benefit of supporting arbitrary callbacks in `Flag::Setter` is that it can build validation into the the parsing logic. Rather than relying only on `GflagsCompatFlag`, can `BuildCommonSelectorFlags` bake this validation into the `Flag`s produced? Then the callers of `BuildCommonSelectorFlags` don't have to follow the (currently implicit) expectation that they will also call `ValidateSelectorOptions`.

The property was upheld for `ParseCommonSelectorArguments` before as the caller would never see an invalid `SelectorOptions` but with this implementation the caller has to call the two functions together.

base/cvd/cuttlefish/common/libs/utils/flag_parser.cpp
```diff
@@ -60,6 +61,123 @@ constexpr std::array help_str_opts{
...
+      return "unset";
+    case EmptyOptionalOptions::EmptyString:
+      return "";
+  }
+}
+
+template <typename T>
+std::string FlagValueToString(const T& value) {
+  return fmt::format("{}", value);
+}
```
Databean, line 89: I think this brings us up to 5 mechanisms to convert arbitrary values to strings:

- `operator<<(std::ostream&, const T&)`
- `fmt::format`
- `AbslStringify`
- `cuttlefish::Pretty`
- `FlagValueToString`

Maybe worth consolidating at some point.

jemoreira: True, but this PR didn't introduce this one, it's copied from one of the `Getter`'s in this file. Having it here just makes it usable from the templates for optional and vector.

base/cvd/cuttlefish/common/libs/utils/flag_parser.cpp
jemoreira: Done

base/cvd/cuttlefish/common/libs/utils/flag_parser.cpp
jemoreira: If we split it into different files we should move it to its own directory (common/libs/flag_parser), otherwise we'd be simplifying one thing to complicate another (the utils directory). So updating headers everywhere is unavoidable. I'd prefer to do this in a different PR, filed b/515490829.

base/cvd/cuttlefish/common/libs/utils/flag_parser_test.cpp
jemoreira: Split these two into multiple smaller tests instead. Once one check fails in one of the smaller tests, the remaining checks are just noise, so kept the ASSERT_ statements.

base/cvd/cuttlefish/host/commands/cvd/cli/selector/BUILD.bazel
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/selector/selector_common_parser.cpp
jemoreira: Kept the GflagsCompatFlag calls and added validators to Flag.

base/cvd/cuttlefish/common/libs/utils/flag_parser.h
```diff
@@ -194,6 +203,18 @@ Flag GflagsCompatFlag(const std::string& name, bool& value);
 Flag GflagsCompatFlag(const std::string& name, std::vector<std::string>& value);
 Flag GflagsCompatFlag(const std::string& name, std::vector<bool>& value,
                       bool default_value);
+// Indicates when to assign std::nullopt to the std::optional backing the flag.
+enum class EmptyOptionalOptions {
+  FlagNotPresent, // When the flag is not present in the arguments
```
Databean: The naming is a bit odd here: the `std::optional` will still be `std::nullopt` if the flag is not present even in the other two cases, this is distinguished by it being the _only_ way that the flag remains `std::nullopt`.

What do you think of renaming `EmptyOptionalOptions` to `CoerceToNullopt` and renaming the first option to `None`?

base/cvd/cuttlefish/host/commands/cvd/cli/selector/selector_common_parser.cpp
```diff
@@ -35,66 +32,57 @@
...
-    cvd_common::Args& args) {
-  // Handling name-related options
-  std::optional<std::string> group_name_opt;
+std::vector<Flag> BuildCommonSelectorFlags(SelectorOptions& opts) {
   Flag group_name_flag =
-      GflagsCompatFlag(SelectorFlags::kGroupName)
-          .Setter([&group_name_opt](const FlagMatch& match) -> Result<void> {
-            group_name_opt = match.value;
+      GflagsCompatFlag(SelectorFlags::kGroupName, opts.group_name)
+          .AddValidator([opts]() -> Result<void> {
```
Databean: This and the following validator anonymous function capture `opts` by value, which means it won't see the changes applied to `opts` by the setter, and instead will operate on the default instance, which returns early and always approves `std::nullopt` values.

To work as expected, it would need to accept [`&opts`].

This seems like an easy pitfall for validator users as well...

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/selector/selector_common_parser.cpp
jemoreira: Done

base/cvd/cuttlefish/common/libs/utils/flag_parser.h
jemoreira: Done

PR #2589, opened 2026-05-19
Remove udev rules for /dev/vhost-vsock and /dev/vhost-net

Databean:
In [PR #10 (2019)](https://www.github.com/google/android-cuttlefish/pull/10) we added a udev rule to put `/dev/vhost-vsock` in the `cvdnetwork` group. IIRC before that the `/dev/vhost-vsock` device was owned by both the `root` user and `root` group. systemd's
[PR #18214](https://github.com/systemd/systemd/pull/18214) (2021) added the `/dev/vhost-vsock` and `/dev/vhost-net` devices to the `kvm` group so we can probably drop our own udev rules now as we already require users to be in the `kvm` group.

Issue brought up by @0405ysj in the team notes.

Bug: b/514792751

### DIFF
--- a/base/debian/cuttlefish-base.udev
+++ b/base/debian/cuttlefish-base.udev
@@ -1,5 +1,2 @@
-ACTION=="add", KERNEL=="vhost-net", SUBSYSTEM=="misc", MODE="0660", GROUP="cvdnetwork"
-ACTION=="add", KERNEL=="vhost-vsock", SUBSYSTEM=="misc", MODE="0660", GROUP="cvdnetwork"
-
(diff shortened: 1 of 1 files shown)


0405ysj (approved):
LGTM. However, I think chown logic on cuttlefish-host-resources.init should be removed together. I disabled the merge just for that part.

https://github.com/google/android-cuttlefish/blob/52a4059ce4b47d01cb3b5f43d957274e49978064/base/debian/cuttlefish-base.cuttlefish-host-resources.init#L297-L298

Databean:
I'm not against changing the behavior in the docker case, but I don't think the udev rules were ever applied inside a container anyway. If the docker-specific `chown` commands can be removed, it'll be for a different reason than the reason to remove the udev rules, so I'll leave that for another change.

PR #2593, opened 2026-05-20
Add a boot timeout thread.

3405691582:
Add a timeout thread to the state machine. Once both threads start, the timeout thread will wait until the boot completion event is received, the boot affirmatively fails, or the timeout gets triggered. Then we attempt to inform the launcher, otherwise we abend. This way, we should have more deterministic stopping conditions.

We should perhaps expose the timeout flag more to the user, but this will suffice for now.

Bug: 508320217

### DIFF
--- a/base/cvd/cuttlefish/host/commands/run_cvd/boot_state_machine.cc
+++ b/base/cvd/cuttlefish/host/commands/run_cvd/boot_state_machine.cc
@@ -22,8 +22,10 @@
 #include <sys/poll.h>
 #include <unistd.h>
 
+#include <chrono>
+#include <condition_variable>
 #include <cstdlib>
-#include <memory>
+#include <mutex>
 #include <optional>
 #include <sstream>
 #include <string>
@@ -33,6 +35,7 @@
 
 #include "absl/log/check.h"
 #include "absl/log/log.h"
+#include "absl/time/time.h"
 #include "android-base/file.h"
 #include "fruit/component.h"
 #include "fruit/fruit_forward_decls.h"
@@ -74,6 +77,9 @@ using openwrtcontrolserver::OpenwrtIpaddrReply;
 
 DEFINE_int32(reboot_notification_fd, CF_DEFAULTS_REBOOT_NOTIFICATION_FD,
              "A file descriptor to notify when boot completes.");
+
+DEFINE_int32(boot_timeout_secs, 600,
+             "Wait for completed boot before failing. On qemu and gem5, this timeout is disabled unless flag is specified");
 
 namespace cuttlefish {
 namespace {
@@ -286,6 +292,8 @@ class CvdBootStateMachine : public SetupFeature, public KernelLogPipeConsumer {
         state_(kBootStarted) {}
 
   ~CvdBootStateMachine() {
+    CancelTimeout();
+
     if (interrupt_fd_write_->IsOpen()) {
       char c = 1;
       CHECK_EQ(interrupt_fd_write_->Write(&c, 1), 1)
@@ -293,6 +301,9 @@ class CvdBootStateMachine : public SetupFeature, public KernelLogPipeConsumer {
     }
     if (boot_event_handler_.joinable()) {
       boot_event_handler_.join();
+    }
+    if (timeout_thread_.joinable()) {
+      timeout_thread_.join();
     }
     if (restore_complete_stop_write_->IsOpen()) {
       char c = 1;
@@ -302,6 +313,14 @@ class CvdBootStateMachine : public SetupFeature, public KernelLogPipeConsumer {
     if (restore_complete_handler_.joinable()) {
       restore_complete_handler_.join();
     }
+  }
+
+  void CancelTimeout() {
+    {
+      std::unique_lock<std::mutex> lock(timeout_mutex_);
+      timeout_done_ = true;
+    }
+    timeout_cv_.notify_all();
   }
 
   // SetupFeature
@@ -420,6 +439,8 @@ class CvdBootStateMachine : public SetupFeature, public KernelLogPipeConsumer {
           ThreadLoop(boot_events_pipe, restore_complete_pipe);
         });
 
+    timeout_thread_ = std::thread([this]() { TimeoutThreadLoop(); });
+
     return {};
   }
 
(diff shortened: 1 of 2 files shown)


base/cvd/cuttlefish/host/commands/run_cvd/boot_state_machine.cc
```diff
@@ -644,6 +712,10 @@ class CvdBootStateMachine : public SetupFeature, public KernelLogPipeConsumer {
 
   std::thread boot_event_handler_;
   std::thread restore_complete_handler_;
+  std::thread timeout_thread_;
```
Databean: Note that the [`std::thread` destructor](https://en.cppreference.com/cpp/thread/thread/~thread) will call [`std::terminate`](https://en.cppreference.com/cpp/error/terminate) if the instance is `joinable()`. The `CvdBootStateMachine` destructor already calls `CancelTimeout()` which should signal the thread to stop, but it should also `join()` the thread instance to make sure the thread has actually exited before returning and allowing the member destructors to run.

base/cvd/cuttlefish/host/commands/run_cvd/boot_state_machine.cc
```diff
@@ -524,6 +549,45 @@ class CvdBootStateMachine : public SetupFeature, public KernelLogPipeConsumer {
     }
   }
 
+  void TimeoutThreadLoop() {
+    auto timeout = std::chrono::seconds(FLAGS_boot_timeout_secs);
+    bool done = false;
+    std::string formatted_timeout =
+        absl::FormatDuration(absl::Seconds(FLAGS_boot_timeout_secs));
```
Databean, line 552: `absl::Seconds(FLAGS_boot_timeout_secs)` could also be [`absl::FromChrono(timeout)`](https://github.com/abseil/abseil-cpp/blob/9346a86073dc6c3620c8277fc7c89e70fd4e9054/absl/time/time.h#L667).

base/cvd/cuttlefish/host/commands/run_cvd/boot_state_machine.cc
```diff
@@ -286,6 +292,8 @@ class CvdBootStateMachine : public SetupFeature, public KernelLogPipeConsumer {
         state_(kBootStarted) {}
 
   ~CvdBootStateMachine() {
+    CancelTimeout();
```
Databean, line 295: This is the only call to `CancelTimeout` that isn't paired with a call to `MaybeWriteNotification`.

Should `CancelTimeout` be combined with `MaybeWriteNotification`? It also seems reasonable to run the `MaybeWriteNotification` in this case as well.

base/cvd/cuttlefish/host/commands/run_cvd/boot_state_machine.cc
```diff
@@ -75,6 +78,9 @@ using openwrtcontrolserver::OpenwrtIpaddrReply;
 DEFINE_int32(reboot_notification_fd, CF_DEFAULTS_REBOOT_NOTIFICATION_FD,
              "A file descriptor to notify when boot completes.");
 
+DEFINE_int32(boot_timeout_secs, 300,
+             "How long to wait for the boot completed event in seconds before failing.");
```
Databean: FYI: adding a flag here taps into some unintuitive code paths, but it's _supposed_ to work.

- `launch_cvd` aka `cvd_internal_start` presents a [composite](https://github.com/google/android-cuttlefish/blob/d2257473dfef9750deff5f165695769d8cd379d6/base/cvd/cuttlefish/host/commands/start/main.cc#L228) of its own flags, `assemble_cvd` flags, and `run_cvd` flags. It collects `assemble_cvd` and `run_cvd` flags at runtime by invoking the executables with `--helpxml`. In practice this has amounted to the `assemble_cvd` flags plus a few flags from `cvd_internal_start` itself.
- `run_cvd` has a code path where it [`exec`s itself](https://github.com/google/android-cuttlefish/blob/d2257473dfef9750deff5f165695769d8cd379d6/base/cvd/cuttlefish/host/commands/run_cvd/server_loop_impl.cpp#L458) in response to a request from `restart_cvd` or `powerwash_cvd`. It tries to reconstruct the same `argv` it was initially launched with.

The alternative is to put the flag in `assemble_cvd` and store state in `CuttlefishConfig`, which is how most other flags work.

Databean: FYI: I was under the impression that there were some targets in CI that weren't expected to get to `VIRTUAL_DEVICE_BOOT_COMPLETED`. CI validated those targets had gotten "far enough" in some other way, and expected this process to hang forever.

However, I'm not sure this is still true. The two targets I remember doing this were the minidroid and riscv64 targets. However, looking at http://ab/git_main-throttled , the minidroid target hasn't built successfully in a year and a half, and the riscv64 target is not running tests at all.

base/cvd/cuttlefish/host/commands/run_cvd/boot_state_machine.cc
```diff
@@ -75,6 +78,9 @@ using openwrtcontrolserver::OpenwrtIpaddrReply;
 DEFINE_int32(reboot_notification_fd, CF_DEFAULTS_REBOOT_NOTIFICATION_FD,
              "A file descriptor to notify when boot completes.");
 
+DEFINE_int32(boot_timeout_secs, 300,
```
Databean: Note: there is some prior art for timeouts, which CI platforms have imposed over invoking the Cuttlefish tools. They've exposed configuration flags, and started with different default values:

- `acloud` default: [450 seconds](https://cs.android.com/android/platform/superproject/+/android-latest-release:tools/acloud/internal/constants.py;l=175;drc=86219821d73808761a317e1b48a0c0cebf3c1fd5)
- `acloud` configuration with `--boot-timeout`: [various](http://cs/file:.gcl%20content:%22--boot-timeout%22), 270 to 900 seconds
- Oxygen default: [8 minutes](http://cs/file:LeaseManager.java%20content:MAX_DEVICE_STARTUP_TIME)
- Oxygenation default: [10 minutes](http://cs/file:BootCompletePoller.java%20content:OXY_ON_DEMAND_DEVICE_CREATION_BOOT_TIMING)
- tradefed's `gce-boot-timeout` default: [30 minutes](http://cs/file:TestDeviceOptions.java%20content:gce-boot-timeout)
- tradefed configuration of `gce-boot-timeout`: [various](http://cs/file:/.gcl%20content:gce-boot-timeout), 5 minutes to 30 minutes

To make things more confusing, these timeouts are sometimes applied to device launching, and sometimes to fetch-and-launch. This is partially our team's own fault, as we've been pushing CI platforms to use `cvd load` / `cvd create --config_file`, which is fetch-and-launch as a single operation.

300 seconds is generous for AOSP targets running locally, but not as generous for targets with google play service and google apps running under nested virtualization in GCE. Another case to consider is the QEMU and gem5 architecture emulation flows, which can be very slow. My understanding is that gem5 can take more than 8 hours.

Databean (approved):

base/cvd/cuttlefish/host/commands/run_cvd/boot_state_machine.cc
3405691582: I thought of this, but https://github.com/google/android-cuttlefish/pull/2593/changes/BASE..7fc6ed3d8353adf0a039d1879976cb4853434565#diff-47382b1121a85f9053270f03c189ea363f305c4c6d9dd7556bd93103e132557fL609 is another.

base/cvd/cuttlefish/host/commands/run_cvd/boot_state_machine.cc
3405691582: Done.

base/cvd/cuttlefish/host/commands/run_cvd/boot_state_machine.cc
3405691582: Yes, I checked and it seems to work fine. Let me know if you'd want this to exist elsewhere, though.

3405691582: I have added a small amount of logic that using a timeout less than or equal to 0 will disable the timeout thread. This should provide an alternative for this and the case you mention below.

base/cvd/cuttlefish/host/commands/run_cvd/boot_state_machine.cc
3405691582: I'll up the default to 10 minutes.

base/cvd/cuttlefish/host/commands/run_cvd/boot_state_machine.cc
Databean: Can it default to no-timeout in the cases where QEMU or gem5 are the vm manager? I think the QEMU/gem5 usage is pretty low anyway, but those users much more likely to get surprised by a timeout.

base/cvd/cuttlefish/host/commands/run_cvd/boot_state_machine.cc
Databean: Sorry, link is dead since the force-push (GitHub drops commits not referenced by any branch?) I'm assuming you're referring to this piece, based on `L609` at the end of the link:

```c++
    if ((*read_result)->event == monitor::Event::BootCompleted) {
      state_ |= kGuestBootCompleted;
      CancelTimeout();
      if (!instance_.vcpu_config_path().empty()) {
        auto res = WattsonRebalanceThreads(instance_.id());
        if (!res.ok()) {
          LOG(ERROR) << res.error();
        }
      }
    } else if ((*read_result)->event == monitor::Event::BootFailed) {
      LOG(ERROR) << "Virtual device failed to boot";
      state_ |= kGuestBootFailed;
      CancelTimeout();
    }  // Ignore the other signals

    return MaybeWriteNotification();
```

The `MaybeWriteNotification()` part is still there, it's after the conditional.

Databean (approved):

base/cvd/cuttlefish/host/commands/run_cvd/boot_state_machine.cc
3405691582: Done. I've also made sure that MaybeWriteNotification uses BootFailed(), which should be safe and looks a bit more readable.

base/cvd/cuttlefish/host/commands/run_cvd/boot_state_machine.cc
3405691582: Done.

Databean: The implementation now is more like "the flag is defunct if QEMU or gem5 is the VM manager" which might be unintuitive. This is fine, but it would also be possible to check `gflags::GetCommandLineFlagInfoOrDie("boot_timeout_secs").is_default` to see whether the user has passed a value or not.

3405691582: I've updated the helptext and added the necessary logic.

PR #2601, opened 2026-05-22
run_cvd: Improve tap device validation error messages

bridadan:
Refactor ValidateTapInterfaceUnused to check if the interface exists first, avoiding misleading "Operation not permitted" errors when the device simply does not exist.

Update ValidateTapDevices to remove the assumption that another device is running, while keeping the actionable suggestions (cvd reset, pkill).

Also remove redundant "Device in use" messages in TestTapDevices to keep logs terse.

Bug: 512536991
Test: Built it with a intentionally incorrect interface name, ensured
      output stated interface did not exist instead of a misleading error

### DIFF
--- a/base/cvd/allocd/alloc_utils.cpp
+++ b/base/cvd/allocd/alloc_utils.cpp
@@ -28,6 +28,7 @@
 #include "absl/strings/str_format.h"
 
 #include "allocd/alloc_driver.h"
+#include "cuttlefish/common/libs/utils/network.h"
 #include "cuttlefish/common/libs/utils/subprocess.h"
 #include "cuttlefish/common/libs/utils/files.h"
 #include "cuttlefish/host/commands/cvd/utils/common.h"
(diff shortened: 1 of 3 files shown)


base/cvd/allocd/alloc_utils.cpp
```diff
@@ -158,6 +159,9 @@ bool CreateTap(std::string_view name) {
 
 #ifdef __linux__
 Result<void> ValidateTapInterfaceIsUsable(const std::string& interface_name) {
+  CF_EXPECTF(NetworkInterfaceExists(interface_name),
+             "Tap interface {} does not exist", interface_name);
```
ser-io: nit: wondering whether we should reference "cuttlefish-host-resources.service" in this error message for a more actionable message. 

For example: "Tap interface {} does not exist. Cuttlefish tap devices should have been created when running cuttlefish-host-resources.service"

I'll leave it up to you for the final message.

ser-io (approved):
Thanks for updating this!

base/cvd/allocd/alloc_utils.cpp
bridadan: Yeah agreed! I updated the message, let me know what you think

Databean (approved):

PR #2603, opened 2026-05-22
Revert "Revert "Migrate iptables and ebtables usages in host resources script to nftables""

dxapd:
Reverts google/android-cuttlefish#2598

Adding the packages to base/BUILD resolves the internal test failure, so this can be cleanly unreverted as-is.


dxapd:
Unrevert needs to be done manually due to the repo's restrictions.

PR #2604, opened 2026-05-22
Reapply the nftables migration changes.

dxapd:
These changes can be re-applied as-is, they just need the dependencies pre-emptively added internally, so this can be merged once the internal CL to do that is submitted.


Databean (approved):

dxapd:
Seems like the presubmit passes now: cl/921490033

PR #2605, opened 2026-05-23
cvd: Better help for load and start subcommands

jemoreira:
Added a default implementation of DetailedHelp that prints in the same format for all commands (unless they override it, which most still do):
```
cvd `CmdList()[0]` - `SummaryHelp()`

`Description()`

`Flags()`
```
Both `load` and `start` make use of that default implementation.

Added more information and missing flags to both command's help output.

Some QoL improvements to the load command.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/BUILD.bazel
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/BUILD.bazel
@@ -41,6 +41,18 @@ cf_cc_test(
 )
 
 cf_cc_library(
+    name = "help_format",
+    srcs = ["help_format.cpp"],
+    hdrs = ["help_format.h"],
+    deps = [
+        "//cuttlefish/common/libs/utils:flag_parser",
+        "@abseil-cpp//absl/log",
+        "@abseil-cpp//absl/log:check",
+        "@abseil-cpp//absl/strings",
+    ],
+)
+
+cf_cc_library(
     name = "interruptible_terminal",
     srcs = ["interruptible_terminal.cpp"],
     hdrs = ["interruptible_terminal.h"],
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/BUILD.bazel
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/BUILD.bazel
@@ -116,6 +116,7 @@ cf_cc_library(
         "//cuttlefish/common/libs/utils:subprocess",
         "//cuttlefish/common/libs/utils:users",
         "//cuttlefish/host/commands/cvd/cli:command_request",
+        "//cuttlefish/host/commands/cvd/cli:help_format",
         "//cuttlefish/host/commands/cvd/cli:types",
         "//cuttlefish/host/commands/cvd/cli/commands:bugreport",
         "//cuttlefish/host/commands/cvd/cli/commands:cache",
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/command_handler.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/command_handler.cpp
@@ -16,8 +16,127 @@
 
 #include "cuttlefish/host/commands/cvd/cli/commands/command_handler.h"
 
+#include <stddef.h>
+
+#include <algorithm>
+#include <sstream>
+#include <string>
+#include <string_view>
+#include <vector>
+
+#include "absl/strings/str_join.h"
+#include "absl/strings/str_split.h"
+
+#include "cuttlefish/common/libs/utils/flag_parser.h"
+#include "cuttlefish/host/commands/cvd/cli/command_request.h"
+#include "cuttlefish/host/commands/cvd/cli/selector/selector_common_parser.h"
+#include "cuttlefish/host/commands/cvd/cli/types.h"
+#include "cuttlefish/result/result.h"
+
 namespace cuttlefish {
+namespace {
+
+constexpr size_t kMaxLineLength = 80;
+
+std::vector<std::string> WrapAroundLine(
+    std::string_view str, size_t max_line_length = kMaxLineLength) {
+  std::vector<std::string> ret;
+  std::vector<std::string_view> words = absl::StrSplit(str, ' ');
+  size_t total_word_sizes = 0;
+  std::vector<std::string_view> line;
+  for (std::string_view word : words) {
+    // line.size() accounts for the spaces added when joining the words.
+    if (total_word_sizes + word.size() + line.size() >= max_line_length) {
+      // If the line is empty at this point it means the current word is too
+      // long, it will be added to the line anyways and printed on its own in
+      // the next iteration.
+      if (!line.empty()) {
+        ret.push_back(absl::StrJoin(line, " "));
+      }
+      total_word_sizes = 0;
+      line.clear();
+    }
+    total_word_sizes += word.size();
+    line.push_back(word);
+  }
+  if (!line.empty()) {
+    // Print the last line
+    ret.push_back(absl::StrJoin(line, " "));
+  }
+  return ret;
+}
+
+}  // namespace
 
 bool CvdCommandHandler::RequiresDeviceExists() const { return false; }
+
+std::vector<std::string> CvdCommandHandler::Description() const { return {}; }
+
+Result<std::vector<Flag>> CvdCommandHandler::Flags(const CommandRequest&) {
+  return {};
+}
+
+Result<std::string> CvdCommandHandler::DetailedHelp(
+    const CommandRequest& request) {
+  std::stringstream ss;
+  std::vector<std::string> cmd_list = CmdList();
+  CF_EXPECT(!cmd_list.empty(), "Command aliases list is empty");
+
+  ss << "cvd " << cmd_list[0] << " - " << SummaryHelp() << "\n";
+  ss << "\n";
+  for (const std::string& paragraph : Description()) {
+    for (std::string_view line : WrapAroundLine(paragraph)) {
+      ss << line << "\n";
+    }
+    // Empty line after paragraphs
+    ss << "\n";
+  }
+
+  std::vector<Flag> flags = CF_EXPECT(Flags(request));
+  // Consume flags to ensure "current value" is accurate
+  cvd_common::Args args = request.SubcommandArguments();
+
+  CF_EXPECT(ConsumeFlags(flags, args));
+  selector::SelectorOptions selector_options = request.Selectors();
+  if (RequiresDeviceExists()) {
+    // Add the common selector flags if the command supports them. This doesn't
+    // need to hapen before consuming as the selector flags were consumed
+    // already. Using the selectors from the request ensures the flags's
+    // "current value" is correct.
+    std::vector<Flag> selector_flags =
+        selector::BuildCommonSelectorFlags(selector_options);
+    flags.insert(flags.end(), selector_flags.begin(), selector_flags.end());
+  }
+
+  // Make sure the flags are in alphabetical order
+  std::sort(flags.begin(), flags.end());
+
+  for (const Flag& flag : flags) {
+    std::stringstream ss2;
+    ss2 << flag;
+    std::string flag_help = ss2.str();
+    std::vector<std::string_view> flag_help_lines =
+        absl::StrSplit(flag_help, '\n');
+    bool first_line = true;
+    for (std::string_view& line : flag_help_lines) {
+      if (line.empty()) {
+        continue;
+      }
+      if (first_line) {
+        // The first line contains the --flag=value examples, don't indent or
+        // wrap those.
+        ss << line << "\n";
+        first_line = false;
+      } else {
+        for (std::string_view l : WrapAroundLine(line, kMaxLineLength - 4)) {
+          ss << "    " << l << "\n";
+        }
+      }
+    }
+    ss << "\n";
+  }
+
+  return ss.str();
+}
 
 }  // namespace cuttlefish
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/command_handler.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/command_handler.cpp
@@ -21,52 +21,16 @@
 #include <algorithm>
 #include <sstream>
 #include <string>
-#include <string_view>
 #include <vector>
-
-#include "absl/strings/str_join.h"
-#include "absl/strings/str_split.h"
 
 #include "cuttlefish/common/libs/utils/flag_parser.h"
 #include "cuttlefish/host/commands/cvd/cli/command_request.h"
+#include "cuttlefish/host/commands/cvd/cli/help_format.h"
 #include "cuttlefish/host/commands/cvd/cli/selector/selector_common_parser.h"
 #include "cuttlefish/host/commands/cvd/cli/types.h"
 #include "cuttlefish/result/result.h"
 
 namespace cuttlefish {
-namespace {
-
-constexpr size_t kMaxLineLength = 80;
-
-std::vector<std::string> WrapAroundLine(
-    std::string_view str, size_t max_line_length = kMaxLineLength) {
-  std::vector<std::string> ret;
-  std::vector<std::string_view> words = absl::StrSplit(str, ' ');
-  size_t total_word_sizes = 0;
-  std::vector<std::string_view> line;
-  for (std::string_view word : words) {
-    // line.size() accounts for the spaces added when joining the words.
-    if (total_word_sizes + word.size() + line.size() >= max_line_length) {
-      // If the line is empty at this point it means the current word is too
-      // long, it will be added to the line anyways and printed on its own in
-      // the next iteration.
-      if (!line.empty()) {
-        ret.push_back(absl::StrJoin(line, " "));
-      }
-      total_word_sizes = 0;
-      line.clear();
-    }
-    total_word_sizes += word.size();
-    line.push_back(word);
-  }
-  if (!line.empty()) {
-    // Print the last line
-    ret.push_back(absl::StrJoin(line, " "));
-  }
-  return ret;
-}
-
-}  // namespace
 
 bool CvdCommandHandler::RequiresDeviceExists() const { return false; }
 
@@ -84,19 +48,13 @@ Result<std::string> CvdCommandHandler::DetailedHelp(
 
   ss << "cvd " << cmd_list[0] << " - " << SummaryHelp() << "\n";
   ss << "\n";
-  for (const std::string& paragraph : Description()) {
-    for (std::string_view line : WrapAroundLine(paragraph)) {
-      ss << line << "\n";
-    }
-    // Empty line after paragraphs
-    ss << "\n";
-  }
+  ss << FormatHelpText(Description());
 
   std::vector<Flag> flags = CF_EXPECT(Flags(request));
   // Consume flags to ensure "current value" is accurate
   cvd_common::Args args = request.SubcommandArguments();
-
   CF_EXPECT(ConsumeFlags(flags, args));
+
   selector::SelectorOptions selector_options = request.Selectors();
   if (RequiresDeviceExists()) {
     // Add the common selector flags if the command supports them. This doesn't
@@ -111,30 +69,7 @@ Result<std::string> CvdCommandHandler::DetailedHelp(
   // Make sure the flags are in alphabetical order
   std::sort(flags.begin(), flags.end());
 
-  for (const Flag& flag : flags) {
-    std::stringstream ss2;
-    ss2 << flag;
-    std::string flag_help = ss2.str();
-    std::vector<std::string_view> flag_help_lines =
-        absl::StrSplit(flag_help, '\n');
-    bool first_line = true;
-    for (std::string_view& line : flag_help_lines) {
-      if (line.empty()) {
-        continue;
-      }
-      if (first_line) {
-        // The first line contains the --flag=value examples, don't indent or
-        // wrap those.
-        ss << line << "\n";
-        first_line = false;
-      } else {
-        for (std::string_view l : WrapAroundLine(line, kMaxLineLength - 4)) {
-          ss << "    " << l << "\n";
-        }
-      }
-    }
-    ss << "\n";
-  }
+  ss << FormatFlagsHelp(flags);
 
   return ss.str();
 }
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/command_handler.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/command_handler.cpp
@@ -34,7 +34,7 @@ namespace cuttlefish {
 
 bool CvdCommandHandler::RequiresDeviceExists() const { return false; }
 
-std::vector<std::string> CvdCommandHandler::Description() const { return {}; }
+std::vector<HelpParagraph> CvdCommandHandler::Description() const { return {}; }
 
 Result<std::vector<Flag>> CvdCommandHandler::Flags(const CommandRequest&) {
   return {};
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/command_handler.h
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/command_handler.h
@@ -20,6 +20,7 @@
 
 #include "cuttlefish/common/libs/utils/flag_parser.h"
 #include "cuttlefish/host/commands/cvd/cli/command_request.h"
+#include "cuttlefish/host/commands/cvd/cli/help_format.h"
 #include "cuttlefish/host/commands/cvd/cli/types.h"
 #include "cuttlefish/result/result.h"
 
@@ -36,12 +37,7 @@ class CvdCommandHandler {
   virtual std::string SummaryHelp() const = 0;
   virtual bool RequiresDeviceExists() const;
   virtual Result<std::string> DetailedHelp(const CommandRequest&);
-  // Each string returned by Description() is treated as a paragraph. The
-  // default implementation of DetailedHelp will visually separate each
-  // paragraph. That implementation will also split each paragraph in lines not
-  // longer than 80 characters, so implementations of Description() don't need
-  // to (and probably shouldn't) use line breaks for formatting.
-  virtual std::vector<std::string> Description() const;
+  virtual std::vector<HelpParagraph> Description() const;
   virtual Result<std::vector<Flag>> Flags(const CommandRequest&);
 
   virtual bool RequiresHostConfiguration() const { return true; }
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/command_handler.h
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/command_handler.h
@@ -18,6 +18,7 @@
 
 #include <string>
 
+#include "cuttlefish/common/libs/utils/flag_parser.h"
 #include "cuttlefish/host/commands/cvd/cli/command_request.h"
 #include "cuttlefish/host/commands/cvd/cli/types.h"
 #include "cuttlefish/result/result.h"
@@ -34,8 +35,16 @@ class CvdCommandHandler {
   // used for command help text
   virtual std::string SummaryHelp() const = 0;
   virtual bool RequiresDeviceExists() const;
+  virtual Result<std::string> DetailedHelp(const CommandRequest&);
+  // Each string returned by Description() is treated as a paragraph. The
+  // default implementation of DetailedHelp will visually separate each
+  // paragraph. That implementation will also split each paragraph in lines not
+  // longer than 80 characters, so implementations of Description() don't need
+  // to (and probably shouldn't) use line breaks for formatting.
+  virtual std::vector<std::string> Description() const;
+  virtual Result<std::vector<Flag>> Flags(const CommandRequest&);
+
   virtual bool RequiresHostConfiguration() const { return true; }
-  virtual Result<std::string> DetailedHelp(const CommandRequest&) = 0;
 };
 
 }  // namespace cuttlefish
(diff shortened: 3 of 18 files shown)


base/cvd/cuttlefish/host/commands/cvd/cli/commands/command_handler.h
```diff
@@ -34,8 +35,15 @@ class CvdCommandHandler {
   // used for command help text
   virtual std::string SummaryHelp() const = 0;
   virtual bool RequiresDeviceExists() const;
+  virtual Result<std::string> DetailedHelp(const CommandRequest&);
+  // Each element in the description will be printed as a separate paragraph.
+  // This implementation will break each paragraph in lines not longer than 80
+  // characters.
```
Databean: "This implementation" sounds like it refers to the default implementation of `Description()`, but that doesn't make sense: `Description()` doesn't have anything to do with formatting.

The comment is really referring to the default implementation of `DetailedHelp`, so the comment should say that instead.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/command_handler.cpp
```diff
@@ -16,8 +16,63 @@
...
+    // already. Using the selectors from the request ensures the flags's
+    // "current value" is correct.
+    std::vector<Flag> selector_flags =
+        selector::BuildCommonSelectorFlags(selector_options);
+    flags.insert(flags.end(), selector_flags.begin(), selector_flags.end());
+  }
+
+  // Make sure the flags are in alphabetical order
+  std::sort(flags.begin(), flags.end(),
+            [](auto f1, auto f2) { return f1.Name() < f2.Name(); });
```
Databean: Maybe worth defining `operator<` on flags, or at least a public function `CompareName(const Flag&, const Flag&)`.

base/cvd/cuttlefish/host/commands/cvd/cli/help_helpers.cpp
```diff
@@ -0,0 +1,114 @@
...
+        ss << "    " << l << "\n";
+      }
+    }
+    ss << "\n";
+  }
+  return ss.str();
+}
+
+std::string MarkAsRawText(std::string_view str) {
+  return std::string(kRawTextMark).append(str);
```
Databean: nit: `absl::StrCat`

base/cvd/cuttlefish/host/commands/cvd/cli/BUILD.bazel
```diff
@@ -40,6 +40,18 @@ cf_cc_test(
     ],
 )
 
+cf_cc_library(
+    name = "help_helpers",
```
Databean: `format_help` or `help_format`?

base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_configs_parser.h
```diff
@@ -35,26 +35,16 @@ struct CvdFlags {
...
-};
-
-std::ostream& operator<<(std::ostream& out, const Override& override);
-
 struct LoadFlags {
-  std::vector<Override> overrides;
-  std::string config_path;
-  std::string credential_source;
-  std::string project_id;
+  std::map<std::string, std::string> overrides;
```
Databean: If this was `std::map<std::string, std::string, std::less<void>>`, it would be possible to use methods like `operator[]` or `find` with `std::string_view` arguments.

base/cvd/cuttlefish/host/commands/cvd/cli/help_helpers.cpp
```diff
@@ -0,0 +1,114 @@
...
+    // Print the last line
+    ret.push_back(absl::StrJoin(line, " "));
+  }
+  return ret;
+}
+
+std::vector<std::string> GetFlagHelpMessage(const Flag& flag) {
+  std::stringstream ss;
+  ss << flag;
+  std::vector<std::string> flag_help_lines = absl::StrSplit(ss.str(), '\n');
```
Databean: Would it work to write `absl::StrSplit(absl::StrCat(flag), '\n')`?

base/cvd/cuttlefish/host/commands/cvd/cli/help_helpers.cpp
```diff
@@ -0,0 +1,114 @@
...
+  std::erase_if(flag_help_lines,
+                [](const std::string& str) { return str.empty(); });
+  return flag_help_lines;
```
Databean: Two better ways of doing this ([compiler explorer](https://godbolt.org/z/77EWPP6Gj)):

`std::erase_if(flag_help_lines, &std::string::empty);`

or use [`absl::SkipWhitespace`](https://github.com/abseil/abseil-cpp/blob/147d631c216ea39af76eef5640b33cc63033d6ae/absl/strings/str_split.h#L377):

`return absl::StrSplit(ss.str(), '\n', absl::SkipWhitespace());`

base/cvd/cuttlefish/host/commands/cvd/cli/help_helpers.cpp
```diff
@@ -0,0 +1,114 @@
...
+namespace cuttlefish {
+namespace {
+
+constexpr char kRawTextMark[] = "_RAW_TEXT:";
+
+std::vector<std::string> WrapAroundLine(std::string_view str,
+                                        size_t max_line_length) {
+  if (absl::ConsumePrefix(&str, kRawTextMark)) {
+    return {std::string(str)};
+  }
```
Databean: I'm not a fan of how `_RAW_TEXT:` ends up in the string contents. If this were Rust I'd ask for an `enum` with variants for raw text and to-be-formatted text, but `std::variant` is also pretty clumsy for this.

A somewhat ergonomic alternative is to define a type like

```c++
class HelpParagraph {
 public:
  static HelpParagraph LineWrapped(std::string);
  static HelpParagraph Raw(std::string);

  friend std::string FormatHelpText(const std::vector<HelpParagraph>& text, size_t max_line_width);
 private:
  std::string text_;
  bool raw_;
};

File: base/cvd/cuttlefish/host/commands/cvd/cli/help_helpers.cpp [hunk 9]
```diff
@@ -0,0 +1,114 @@
...
+  std::vector<std::string> ret;
+  std::vector<std::string_view> words = absl::StrSplit(str, ' ');
```
Databean: Maybe use `absl::ByAnyChar(" \t\r\n")` to consume other kinds of whitespace that are added unintentionally?

base/cvd/cuttlefish/host/commands/cvd/cli/help_format.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+#include "cuttlefish/common/libs/utils/flag_parser.h"
+
+namespace cuttlefish {
+namespace {
+
+constexpr char kRawTextMark[] = "_RAW_TEXT:";
+
+std::vector<std::string> WrapAroundLine(std::string_view str,
+                                        size_t max_line_length) {
```
Databean, line 33: FYI this assumes ASCII text. If this ever needs to handle more unicode (in practice, maybe emoji), it should use something like ICU for [boundary analysis](https://unicode-org.github.io/icu/userguide/boundaryanalysis/). This can be used for counting characters and for [break rules](https://unicode-org.github.io/icu/userguide/boundaryanalysis/break-rules.html). This can use [`icu::BreakIterator`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/classicu_1_1BreakIterator.html)'s `createWordInstance` to determine word boundaries for inserting line breaks and `createCharacterInstance` to determine character boundaries for counting characters.

> Line boundary analysis determines where a text string can be broken when line-wrapping.

That also doesn't take into account escape codes for coloring text, which can take up multiple bytes but no visible space on the screen.

base/cvd/cuttlefish/host/commands/cvd/cli/help_format.h
```diff
@@ -0,0 +1,48 @@
...
+
+#include <string>
+#include <string_view>
+#include <vector>
+
+#include "cuttlefish/common/libs/utils/flag_parser.h"
+
+namespace cuttlefish {
+
+constexpr size_t kDefaultMaxLineWidth = 80;
```
Databean, line 28: Rather than using a fixed value, what do you think of dynamically loading the actual terminal width?

base/cvd/cuttlefish/host/commands/cvd/cli/help_format.cpp
```diff
@@ -0,0 +1,114 @@
...
+      // If the line is empty at this point it means the current word is too
+      // long, it will be added to the line anyways and printed on its own in
+      // the next iteration.
+      if (!line.empty()) {
+        ret.push_back(absl::StrJoin(line, " "));
+      }
+      total_word_sizes = 0;
+      line.clear();
+    }
+    total_word_sizes += word.size();
```
Databean, line 49: `+ line.size()` can be left out above if this is changed to
```c++
total_word_sizes += word.size() + 1;

File: base/cvd/cuttlefish/host/commands/cvd/cli/help_helpers.cpp [hunk 13]
```diff
@@ -0,0 +1,114 @@
...
+
+std::string FormatFlagsHelp(const std::vector<Flag>& flags,
+                            size_t max_line_width) {
+  std::stringstream ss;
+  for (const Flag& flag : flags) {
+    std::vector<std::string> help_lines = GetFlagHelpMessage(flag);
+    CHECK(!help_lines.empty())
+        << "Flag produced empty help message: " << flag.Name();
+    // The first line contains the --flag=value examples, don't indent or wrap
+    // those.
```
Databean: This relies on an implementation detail of `std::ostream& operator<<std::ostream&, const Flag&)`.

Should `Flag` expose some more fine-grained operations so that this code can have separate logic for printing flag help text and flag aliases?

base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_configs_parser.cpp
```diff
@@ -266,6 +244,61 @@ EnvironmentSpecification FillEmptyInstanceNames(
...
+  flags.emplace_back(
+      GflagsCompatFlag("credential_source")
+          .Help("Source of credentials to access the Android Build Server API. "
+                "Can be left empty in most cases, see the help for `login` and "
+                "`fetch` for details.")
+          .Setter([&load_flags](const FlagMatch& match) -> Result<void> {
+            CF_EXPECTF(load_flags.overrides.count(std::string(kCredentialSourceOverride)) == 0,
+                       "Specifying both --override={} and the --credential_source flag is not allowed.",
+                       kCredentialSourceOverride);
+            load_flags.overrides[std::string(kCredentialSourceOverride)] = match.value;
```
Databean: For this and the next `Flag`:

`ConsumeFlags` handles flags in order present in the `std::vector<Flag>`, not in the order they are given in the command line vector. This means that validating against `load_flags.overrides` here will error out on multiple `--credential_source` arguments, but will not detect a conflicting `--override` argument.

To handle this ordering conflict correctly, the `--override` flag should be added to `std::vector<Flag> flags` first.

base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_configs_parser.cpp
```diff
@@ -266,6 +244,61 @@ EnvironmentSpecification FillEmptyInstanceNames(
...
+              "Parent directory for artifacts and runtime files. When not "
+              "provided a new directory under {}/{} will be created.",
+              CvdDir(), getuid())));
+  flags.emplace_back(
+      GflagsCompatFlagOverride("override", load_flags.overrides)
+          .Help(
+              "Override or add new properties to the group specification. This "
+              "flag may appear multiple times to affect different properties. "
+              "The format is `--override=<path.to.property>:<value>`. "
+              "Forexample: `--override=instance.0.vm.cpus=16`."));
```
Databean: nit: "For example"

base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_configs_parser.cpp
```diff
@@ -71,35 +71,39 @@ bool IsLocalBuild(std::string path) {
...
                    "Unable to find separator \"{}\" in input \"{}\"",
                    kOverrideSeparator, match.value);
-        auto result =
-            Override{.config_path = match.value.substr(0, separator_index),
-                     .new_value = match.value.substr(separator_index + 1)};
-        CF_EXPECTF(!result.config_path.empty(),
+        auto path = match.value.substr(0, separator_index);
+        auto val = match.value.substr(separator_index + 1);
+        CF_EXPECTF(values.count(path) == 0,
+                   "Property \"{}\" is already overridden", path);
```
Databean: It's a bit confusing that `val` refers to a config property while `values` refers to a map from config paths to config values.

Maybe `overrides.count(path)` or `path_to_value(path)` would make this easier to understand.

base/cvd/cuttlefish/host/commands/cvd/cli/parser/flags_parser_test.cc
```diff
@@ -170,4 +174,126 @@ TEST(BootFlagsParserTest, ParseNetSimFlagEnabled) {
       << "netsim_uwb flag is missing or wrongly formatted";
 }
 
+TEST(CvdLoadFlagsTest, CredentialSourceGetterSetter) {
+  LoadFlags load_flags;
+
+  // Test Setter
+  {
```
Databean: The tests added in this file look like they could be easily split up into multiple tests around these blocks. The only shared state is in the `load_flags` variable, and IMO the getter test would be easier to read if it didn't rely on the setter test running first.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/load_configs.cpp
```diff
@@ -250,9 +241,73 @@ cvd_common::Args LoadConfigsCommand::CmdList() const { return {kLoadSubCmd}; }
...
+      "properties (default_build, kernel.build, bootloader.build, etc) require "
+
+      "more explanation. These properties support two types of values:",
+
+      MarkAsRawText(
+          R"( - "@ab/<branch_or_build_id>[/<target>[{<filepath>}]]"
+ - "<absolute_path>")"),
+
+      "If the build value starts with \"@ab\", cvd will fetch the specified "
+      "Android build target from the Android  build servers. By default it "
```
Databean: nit: two spaces between "Android" and "build"

base/cvd/cuttlefish/host/commands/cvd/cli/commands/load_configs.cpp
```diff
@@ -250,9 +241,73 @@ cvd_common::Args LoadConfigsCommand::CmdList() const { return {kLoadSubCmd}; }
...
+
+      "A complete reference of the specification file format can be found in "
+      "https://github.com/google/android-cuttlefish/blob/main/base/cvd/"
+      "cuttlefish/host/commands/cvd/cli/parser/load_config.proto.",
+
+      "While most config file properties are self explanatory, the build "
+
+      "properties (default_build, kernel.build, bootloader.build, etc) require "
+
+      "more explanation. These properties support two types of values:",
```
Databean: nit: odd blank lines here?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
```diff
@@ -521,11 +507,24 @@ Result<void> CvdStartCommandHandler::LaunchDeviceInterruptible(
...
+      "instances because Cuttlefish instances in the same group must all be "
+      "started (and stopped) in unisom. The group to be started is chosen "
+      "using the standard selector flags.",
+      "Flags that modify individual instances accept a comma separated list of "
+      "values. If the number of values in one of these flags is less than the "
+      "number of instances, then the last instances in the group will default "
+      "to the first value in the list (as a result a single value provided "
+      "applies to all instances). The empty string is interpreted as a valid "
+      "value, to force a particular instance to use its intended default value "
+      "the special value \"unset\" must be given."};
```
Databean: nit: maybe use a `R"(...)"` string here?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/load_configs.cpp
```diff
@@ -250,9 +241,73 @@ cvd_common::Args LoadConfigsCommand::CmdList() const { return {kLoadSubCmd}; }
 
 std::string LoadConfigsCommand::SummaryHelp() const { return kSummaryHelpText; }
 
-Result<std::string> LoadConfigsCommand::DetailedHelp(
+std::vector<std::string> LoadConfigsCommand::Description() const {
+  return {
+      "This command is an alias of `cvd create --config_file=<config_filepath> "
```
Databean, line 247: nit: maybe use `R"(...)"` strings even for the non-`MarkAsRawText` help strings?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/command_handler.cpp
jemoreira: Added operator<

base/cvd/cuttlefish/host/commands/cvd/cli/commands/command_handler.h
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/load_configs.cpp
jemoreira: These strings are meant to be long, the `DetailedHelp` implementation in `CvdCommandHandler` will break them in the correct length. If I use raw string literals, clang-format won't cut them to size and will be hard to read.

I wanted to use a runtime approach to wrapping lines to make it consisted across all commands and to maybe make it able to figure out the terminal's width and adjust the line length dynamically.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/load_configs.cpp
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/load_configs.cpp
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
jemoreira: Same as before: I don't want line breaks in these strings, raw strings literals will be too wide.

base/cvd/cuttlefish/host/commands/cvd/cli/parser/flags_parser_test.cc
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_configs_parser.cpp
jemoreira: Renamed:

values -> overrides
path -> property_path
val -> new_value

base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_configs_parser.cpp
jemoreira: Done.

It would still be caught by the check that overrides are not duplicated, but the message wouldn't be as clear.

base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_configs_parser.cpp
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_configs_parser.h
jemoreira: Didn't work when I tried it:
```
cuttlefish/host/commands/cvd/cli/parser/load_configs_parser.cpp:267:33: error: no viable overloaded operator[] for type 'EnvironmentOverrides' (aka 'map<basic_string<char>, basic_string<char>, less<void>>')
  267 |             load_flags.overrides[kCredentialSourceOverride] = match.value;
      |             ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
external/toolchains_llvm++llvm+llvm_toolchain_llvm/bin/../include/c++/v1/map:1151:38: note: candidate function not viable: no known conversion from 'const std::string_view' (aka 'const basic_string_view<char>') to 'const key_type' (aka 'const std::string') for 1st argument
 1151 |   _LIBCPP_HIDE_FROM_ABI mapped_type& operator[](const key_type& __k);
```

base/cvd/cuttlefish/host/commands/cvd/cli/BUILD.bazel
jemoreira: done

base/cvd/cuttlefish/host/commands/cvd/cli/help_format.cpp
jemoreira: ack

base/cvd/cuttlefish/host/commands/cvd/cli/help_helpers.cpp
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/help_format.cpp
jemoreira: It would be less readable IMO because the variable would no longer represent the sum of word sizes, but the sum or word sizes plus the number of words. That number doesn't represent the length of the current line either (the line will have 1 less character).

base/cvd/cuttlefish/host/commands/cvd/cli/help_helpers.cpp
jemoreira: No:
```
cuttlefish/host/commands/cvd/cli/help_format.cpp:68:61: error: no matching function for call to 'StrCat'
   68 |   std::vector<std::string> flag_help_lines = absl::StrSplit(absl::StrCat(flag), '\n');
      |                                                             ^~~~~~~~~~~~
external/abseil-cpp+/absl/strings/str_cat.h:553:34: note: candidate function not viable: no known conversion from 'const Flag' to 'const AlphaNum' for 1st argument
```

base/cvd/cuttlefish/host/commands/cvd/cli/help_helpers.cpp
jemoreira: done with absl::SkipWhitespace.

base/cvd/cuttlefish/host/commands/cvd/cli/help_helpers.cpp
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/help_format.h
jemoreira: That's the plan, but will require some more refactoring. I want to get all commands using the same utilities for help before I start making it even prettier.

base/cvd/cuttlefish/host/commands/cvd/cli/help_helpers.cpp
jemoreira: I'm planning a change to the Flag texts that will make this hack unnecessary. In the meantime I've removed it since it's only a handful of flags in `cvd start` that need it.

base/cvd/cuttlefish/host/commands/cvd/cli/help_helpers.cpp
jemoreira: done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/load_configs.cpp
Databean: With the change to use `absl::StrSplit(str, absl::ByAnyChar(" \t\r\n"))` in `WrapAroundLine` all internal whitespace is collapsed anyway, so it doesn't really matter how it is formatted in the source file.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
Databean: With the change to use `absl::StrSplit(str, absl::ByAnyChar(" \t\r\n"))` in `WrapAroundLine` all internal whitespace is collapsed anyway, so it doesn't really matter how it is formatted in the source file.

base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_configs_parser.h
Databean: Well, at least `.count`, `.emplace`, and `.find`  would work. See [compiler explorer](https://godbolt.org/z/ffM1chr39)

jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/help_format.cpp
```diff
@@ -0,0 +1,117 @@
...
+namespace {
+
+std::vector<std::string> WrapAroundLine(std::string_view str,
+                                        size_t max_line_length) {
+  std::vector<std::string> ret;
+  std::vector<std::string_view> words =
+      absl::StrSplit(str, absl::ByAnyChar(" \t\r\n"));
+  size_t total_word_sizes = 0;
+  std::vector<std::string_view> line;
+  for (std::string_view word : words) {
```
Databean: nit: can avoid the `words` intermediate by writing something like

```c++
const absl::ByAnyChar whitespace(" \t\r\n");
for (std::string_view word : absl::StrSplit(str, whitespace)) {
  ...
}
```

base/cvd/cuttlefish/host/commands/cvd/cli/commands/load_configs.cpp
jemoreira: Tried many options, none was much better than this.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
jemoreira: Tried many options, none was much better than this.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/command_handler.h
```diff
@@ -34,8 +36,16 @@ class CvdCommandHandler {
   // used for command help text
   virtual std::string SummaryHelp() const = 0;
   virtual bool RequiresDeviceExists() const;
+  virtual Result<std::string> DetailedHelp(const CommandRequest&);
+  // Each string returned by Description() is treated as a paragraph. The
+  // default implementation of DetailedHelp will visually separate each
+  // paragraph. That implementation will also split each paragraph in lines not
+  // longer than 80 characters, so implementations of Description() don't need
+  // to (and probably shouldn't) use line breaks for formatting.
```
Databean: nit: maybe some parts of this comment are redundant now that the type is named `HelpParagraph`?

base/cvd/cuttlefish/host/commands/cvd/cli/parser/flags_parser_test.cc
```diff
@@ -170,4 +174,131 @@ TEST(BootFlagsParserTest, ParseNetSimFlagEnabled) {
...
+  LoadFlags load_flags;
+
+  auto flags = BuildCvdLoadFlags(load_flags);
+  std::vector<std::string> args = {"--credential_source=foo"};
+  auto result = ConsumeFlags(flags, args);
+  ASSERT_TRUE(result.ok()) << result.error().Trace();
+
+  ASSERT_EQ(load_flags.overrides.size(), 1);
+  EXPECT_EQ(load_flags.overrides.count("fetch.credential_source"), 1);
+  EXPECT_EQ(load_flags.overrides.at("fetch.credential_source"), "foo");
```
Databean: `.at` is not recommended: [go/totw/202#access-with-presence-checking-at](http://go/totw/202#access-with-presence-checking-at)

base/cvd/cuttlefish/host/commands/cvd/cli/parser/flags_parser_test.cc
```diff
@@ -170,4 +174,131 @@ TEST(BootFlagsParserTest, ParseNetSimFlagEnabled) {
...
+}
+
+TEST(CvdLoadFlagsTest, CredentialSourceDuplicated) {
+  LoadFlags load_flags;
+
+  auto flags = BuildCvdLoadFlags(load_flags);
+  std::vector<std::string> args = {"--credential_source=first_val",
+                                   "--credential_source=second_val"};
+  auto result = ConsumeFlags(flags, args);
+  EXPECT_FALSE(result.ok()) << "Expected duplicate flag to fail";
```
Databean: This can be `EXPECT_THAT(result, IsError())`

https://github.com/google/android-cuttlefish/blob/deffb1072c07e29c8fcb50e18bc37884cfcf1aa2/base/cvd/cuttlefish/result/result_matchers.h#L37-L44

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/command_handler.h
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/parser/flags_parser_test.cc
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/parser/flags_parser_test.cc
jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/help_format.cpp
jemoreira: Done

PR #2606, opened 2026-05-23
cvd: Begin simplifying creation analyzer

jemoreira:
Most of the simplification is done in StartSelectorParser.

The most significant change is that it no longer considers the `$CUTTLEFISH_INSTANCE` environment variable when trying to figure the number of instances and its ids. `cvd_internal_start` never expected the user to set that variable, expecting the user to use flags instead. That variable is unconditionally set by `cvd_internal_start` so `run_cvd` and its children know which instance they are running.

NOTE TO REVIEWER: reviewing each commit separately is going to be a lot easier than the entire PR at once.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/selector/start_selector_parser.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/selector/start_selector_parser.cpp
@@ -226,10 +226,6 @@ StartSelectorParser::HandleInstanceIds(
 }
 
 Result<void> StartSelectorParser::ParseOptions() {
-  std::optional<bool> use_cvdalloc;
-  CF_EXPECT(FilterSelectorFlag(cmd_args_, "use_cvdalloc", use_cvdalloc));
-  use_cvdalloc_ = use_cvdalloc.value_or(false);
-
   std::optional<std::string> num_instances;
   std::optional<std::string> instance_nums;
   std::optional<std::string> base_instance_num;
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/selector/start_selector_parser.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/selector/start_selector_parser.cpp
@@ -226,14 +226,13 @@ StartSelectorParser::HandleInstanceIds(
 }
 
 Result<void> StartSelectorParser::ParseOptions() {
-  std::optional<std::string> num_instances;
-  std::optional<std::string> instance_nums;
-  std::optional<std::string> base_instance_num;
-  // set num_instances as std::nullptr or the value of --num_instances
-  CF_EXPECT(FilterSelectorFlag(cmd_args_, "num_instances", num_instances));
-  CF_EXPECT(FilterSelectorFlag(cmd_args_, "instance_nums", instance_nums));
-  CF_EXPECT(
-      FilterSelectorFlag(cmd_args_, "base_instance_num", base_instance_num));
+  // set num_instances as std::nullopt or the value of --num_instances
+  std::optional<std::string> num_instances =
+      CF_EXPECT(FilterSelectorFlag<std::string>(cmd_args_, "num_instances"));
+  std::optional<std::string> instance_nums =
+      CF_EXPECT(FilterSelectorFlag<std::string>(cmd_args_, "instance_nums"));
+  std::optional<std::string> base_instance_num = CF_EXPECT(
+      FilterSelectorFlag<std::string>(cmd_args_, "base_instance_num"));
 
   InstanceIdsParams instance_nums_param{
       .num_instances = std::move(num_instances),
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/selector/start_selector_parser.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/selector/start_selector_parser.cpp
@@ -31,7 +31,6 @@
 #include "absl/strings/numbers.h"
 
 #include "cuttlefish/common/libs/utils/contains.h"
-#include "cuttlefish/host/commands/cvd/cli/selector/selector_common_parser.h"
 #include "cuttlefish/host/commands/cvd/cli/selector/selector_option_parser_utils.h"
 #include "cuttlefish/host/commands/cvd/cli/types.h"
 #include "cuttlefish/host/libs/config/config_constants.h"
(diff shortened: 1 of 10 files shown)


base/cvd/cuttlefish/host/commands/cvd/cli/selector/start_selector_parser.cpp
```diff
@@ -226,14 +226,13 @@ StartSelectorParser::HandleInstanceIds(
...
-  CF_EXPECT(FilterSelectorFlag(cmd_args_, "instance_nums", instance_nums));
-  CF_EXPECT(
-      FilterSelectorFlag(cmd_args_, "base_instance_num", base_instance_num));
+  // set num_instances as std::nullopt or the value of --num_instances
+  std::optional<std::string> num_instances =
+      CF_EXPECT(FilterSelectorFlag<std::string>(cmd_args_, "num_instances"));
+  std::optional<std::string> instance_nums =
+      CF_EXPECT(FilterSelectorFlag<std::string>(cmd_args_, "instance_nums"));
+  std::optional<std::string> base_instance_num = CF_EXPECT(
+      FilterSelectorFlag<std::string>(cmd_args_, "base_instance_num"));
```
Databean, line 170: Can `num_instances` and `base_instance_num` be `std::optional<uintN_t>` instead of `std::optional<std::string>`? Can `instance_nums` be a `std::vector<uintN_t>`?

base/cvd/cuttlefish/host/commands/cvd/cli/selector/start_selector_parser.cpp
```diff
@@ -123,15 +117,15 @@ Result<unsigned> StartSelectorParser::VerifyNumOfInstances(
...
+  const auto& num_instance_names = params.num_instance_names;
   const auto& instance_nums_flag = params.instance_nums_flag;
 
   std::optional<unsigned> num_instances;
   if (num_instances_flag) {
     num_instances = CF_EXPECT(ParsePositiveNumber(*num_instances_flag));
   }
-  if (instance_names && !instance_names->empty()) {
-    auto implied_n_instances = instance_names->size();
+  if (num_instance_names && *num_instance_names > 0) {
```
Databean: maybe `num_instance_names.value_or(0) > 0`?

base/cvd/cuttlefish/host/libs/config/instance_nums.cpp
```diff
@@ -223,10 +223,10 @@ Result<std::vector<int32_t>> InstanceNumsCalculator::CalculateFromFlags() {
   }
   // exactly one of these two should be given
   CF_EXPECT(!instance_nums_opt || !base_instance_num_,
+            "InstanceNums and BaseInstanceNum are mutually exclusive");
```
Databean: nit: IMO this would be easier to read as `!(instance_nums_opt && base_instance_num_)`

base/cvd/cuttlefish/host/commands/cvd/cli/selector/start_selector_parser.h
```diff
@@ -45,19 +45,15 @@ namespace selector {
...
  public:
   static Result<StartSelectorParser> ConductSelectFlagsParser(
-      const SelectorOptions& selector_options, const cvd_common::Args& cmd_args,
-      const cvd_common::Envs& envs);
-  std::optional<std::string> GroupName() const;
-  std::optional<std::vector<std::string>> PerInstanceNames() const;
-  const std::vector<unsigned>& InstanceIds() const {
-    return instance_ids_;
-  }
+      const std::optional<std::vector<std::string>>& instance_names_opt,
```
Databean, line 48: Rather than `optional<vector<string>>` can it be `vector<string>`?

Is there a significant semantic difference between an empty vector and an empty optional? Based on usage elsewhere it seems like a provided empty `--instance_names=` is an error, but maybe it could be legal instead and handled the same as a missing flag.

base/cvd/cuttlefish/host/commands/cvd/cli/selector/start_selector_parser.cpp
```diff
@@ -135,26 +134,20 @@ StartSelectorParser::HandleInstanceIds(
...
-    unsigned base = CF_EXPECT(ParsePositiveNumber(*base_instance_num));
-    calculator.BaseInstanceNum(static_cast<int32_t>(base));
+  // base_instance_num must be set since instance_nums is not set and we checked
+  // that at least one of them is set above
+  unsigned base = CF_EXPECT(ParsePositiveNumber(*base_instance_num));
+  std::vector<unsigned> instance_ids_vector;
+  for (unsigned offset = 0; offset < num_instances; ++offset) {
+    instance_ids_vector.emplace_back(base + offset);
   }
-  auto instance_ids = CF_EXPECT(calculator.CalculateFromFlags());
```
Databean, line 225: > StartSelectorParser already duplicates a lot of its logic

so maybe it shouldn't duplicate its logic?

base/cvd/cuttlefish/host/commands/cvd/cli/selector/start_selector_parser.cpp
jemoreira: Yes, I'm working on that for my next PR. This one just copied the flags without modification

base/cvd/cuttlefish/host/commands/cvd/cli/selector/start_selector_parser.cpp
jemoreira: I didn't verify that the code in start_selector_parser does exactly the same as the calculator, so I didn't want to replace it. I'll drop this commit from the PR and keep it using the calculator for the final part only.

base/cvd/cuttlefish/host/commands/cvd/cli/selector/start_selector_parser.h
jemoreira: Probably, but that change would have to be in `SelectorOptions` which would affect many more commands than just `create`. I'd rather not do that in this PR.

base/cvd/cuttlefish/host/libs/config/instance_nums.cpp
Databean: Or maybe this code can actually use the exclusive or operator `^`?

jemoreira: Done

base/cvd/cuttlefish/host/commands/cvd/cli/selector/start_selector_parser.cpp
jemoreira: Done

Databean (approved):

PR #2607, opened 2026-05-26
Implement GET /cvds/{group}/{name} endpoint

SuperStrongDinosaur:
Implement the `GET /cvds/{group}/{name}` endpoint in the host orchestrator to allow retrieving a specific CVD instance by its group and name.

This is achieved by:
- Extending `ListCVDsActionOpts` and `ListCVDsAction` to support filtering by CVD name.
- Updating `listCVDsHandler.Handle` to extract the `name` variable from the request and pass it to `ListCVDsActionOpts`.
- Registering the `GET /cvds/{group}/{name}` route in `Controller.AddRoutes` using `listCVDsHandler`.
- Adding unit tests in `listcvdsaction_test.go` to verify the filtering logic and error handling (404 when not found).
- Adding a routing test in `controller_test.go` to ensure the new endpoint is correctly registered.
  
This change preserves the original behavior of `GET /cvds` and `GET /cvds/{group}`.

### DIFF
--- a/frontend/src/host_orchestrator/orchestrator/listcvdsaction.go
+++ b/frontend/src/host_orchestrator/orchestrator/listcvdsaction.go
@@ -25,12 +25,14 @@ import (
 
 type ListCVDsActionOpts struct {
 	Group       string
+	Name        string
 	Paths       IMPaths
 	ExecContext exec.ExecContext
 }
 
 type ListCVDsAction struct {
 	group  string
+	name   string
 	paths  IMPaths
 	cvdCLI *cvd.CLI
 }
@@ -38,6 +40,7 @@ type ListCVDsAction struct {
 func NewListCVDsAction(opts ListCVDsActionOpts) *ListCVDsAction {
 	return &ListCVDsAction{
 		group:  opts.Group,
+		name:   opts.Name,
 		paths:  opts.Paths,
 		cvdCLI: cvd.NewCLI(opts.ExecContext),
 	}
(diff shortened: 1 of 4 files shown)


frontend/src/host_orchestrator/orchestrator/listcvdsaction.go
```diff
@@ -59,5 +62,17 @@ func (a *ListCVDsAction) Run() (*apiv1.ListCVDsResponse, error) {
 	for _, g := range groups {
 		cvds = append(cvds, CvdGroupToAPIObject(g)...)
 	}
+	if a.name != "" {
+		filtered := []*apiv1.CVD{}
```
ser-io: Reuse existing `findInstance` and `CvdInstanceToAPIObject` functions. Code should look like the following.


```
cvds := []*apiv1.CVD{}
if a.name != "" {
  // assert len(groups) == 1
  found, ins := findInstance(groups[0], a.name)
  if !found {
    return err
  }
  cvds = []*apiv1.CVD{CvdInstanceToAPIObject(ins, a.group)}
} else {
  for _, g := range groups {
    cvds = append(cvds, CvdGroupToAPIObject(g)...)
  }
}
```

ser-io (changes requested):

frontend/src/host_orchestrator/orchestrator/listcvdsaction.go
SuperStrongDinosaur: done

ser-io (approved):

PR #2611, opened 2026-05-27
Move the new gpu mode opt-in behind a stubbed googler guard

jmacnak:
Bug: b/517195732

### DIFF
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/graphics_flags.cc
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/graphics_flags.cc
@@ -431,6 +431,10 @@ GetNeededVhostUserGpuHostRendererFeatures(
 }
 
 #ifndef __APPLE__
+
+// TODO(b/503397840): remove after default updated.
+bool EnableHostRenderingByDefault() { return false; }
+
 std::vector<GpuMode> GetGpuModeCandidates(const GuestConfig& guest_config) {
   if (!guest_config.gpu_mode_candidates.empty()) {
     VLOG(0) << "GPU mode candidates provided by guest.";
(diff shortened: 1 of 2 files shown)


base/cvd/cuttlefish/host/commands/assemble_cvd/graphics_flags.cc
```diff
@@ -445,13 +446,25 @@ std::vector<GpuMode> GetGpuModeCandidates(const GuestConfig& guest_config) {
...
 
-  gpu_mode_candidates.push_back(GpuMode::Gfxstream);
-  gpu_mode_candidates.push_back(GpuMode::GfxstreamGuestAngle);
-  gpu_mode_candidates.push_back(GpuMode::GfxstreamGuestAngleHostSwiftshader);
-  gpu_mode_candidates.push_back(GpuMode::GfxstreamGuestAngleHostLavapipe);
-  gpu_mode_candidates.push_back(GpuMode::GuestSwiftshader);
+  // Opt Googlers into the more experimental modes early.
+  //
+  // TODO(b/503397840): remove after default updated.
+  if (IsGoogler()) {
```
3405691582: I think you should make this more feature flag-gy, because AIUI there is nothing intrinsic about gfxstream that should be used with employees. This should probably be called UseGfxstream() or ShouldUseGfxstream() or something similar.

3405691582 (approved):
Just naming.

base/cvd/cuttlefish/host/commands/assemble_cvd/graphics_flags.cc
jmacnak: Updated.

PR #2612, opened 2026-05-27
Resolve filesystem isolation issue on podcvd bugreport

0405ysj:
Context:b/491258372

### DIFF
--- a/container/src/podcvd/internal/parser.go
+++ b/container/src/podcvd/internal/parser.go
@@ -111,6 +111,32 @@ func (a *CvdArgs) GetStringFlagValueOnSubCommandArgs(flagName string) string {
 	return ""
 }
 
+func (a *CvdArgs) ReplaceFlagValueOnSubCommandArgs(flagName, newValue string) {
+	flags := make(map[string]struct{})
+	flags["-"+flagName] = struct{}{}
+	flags["--"+flagName] = struct{}{}
+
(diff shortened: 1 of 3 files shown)


container/src/podcvd/internal/parser.go
```diff
@@ -126,3 +126,29 @@ func mapSubcommand(subcmd string) string {
...
+		splitArg := strings.SplitN(arg, "=", 2)
+		if len(splitArg) == 2 {
+			if _, exists := flags[splitArg[0]]; exists {
+				a.SubCommandArgs[idx] = fmt.Sprintf("%s=%s", splitArg[0], newValue)
+				return
+			}
+		}
+	}
+
+	a.SubCommandArgs = append(a.SubCommandArgs, fmt.Sprintf("-%s=%s", flagName, newValue))
```
k311093, line 137: Not sure the flag like `-output=...` is valid, can we pass the full name at the parameter like `-o` or `--output` ?

k311093 (approved):

container/src/podcvd/internal/parser.go
0405ysj: `fmt.Sprintf("-%s=%s", flagName, newValue)` should be working, I verified it. Also I believe this is being checked by podcvd E2E test.

PR #2614, opened 2026-05-28
Support per-instance stop in stop_cvd

SuperStrongDinosaur:
### **Summary**

Currently, stop_cvd indiscriminately stops all Cuttlefish instances. This change introduces granular control, allowing users to stop a specific instance (or list of instances) in a running group without affecting others.

### **Changes**

stop_cvd update: Introduced the --instance_nums flag to stop_cvd. It accepts a comma-separated list of instance IDs to stop.
Instance Filtering: Updated the main loop in stop_cvd to check the --instance_nums set. If populated, it only sends the termination signal to instances matching those IDs.
cvd stop Integration: Modified the cvd stop command handler to parse selectors. It resolves targeted instance names to their IDs and propagates them as --instance_nums to the stop_cvd backend.
API Propagation: Updated InstanceManager::StopInstanceGroup and RunStopCvd to accept and pass the instance_nums string.
Database Status Update Fix: Updated StopInstanceGroup in instance_manager.cpp to only set the state to STOPPED in the database for the instances that were actually targeted and stopped. 

### **Bug**
b/459780275

### **Document**

### **Frontend pr**
https://github.com/google/android-cuttlefish/pull/2615

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/stop.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/stop.cpp
@@ -34,6 +34,7 @@
 #include "cuttlefish/host/commands/cvd/cli/types.h"
 #include "cuttlefish/host/commands/cvd/cli/utils.h"
 #include "cuttlefish/host/commands/cvd/instances/instance_manager.h"
+#include "cuttlefish/host/commands/cvd/instances/local_instance.h"
 #include "cuttlefish/host/libs/metrics/device_metrics_orchestration.h"
 #include "cuttlefish/result/result.h"
 
@@ -94,10 +95,24 @@ Result<void> CvdStopCommandHandler::Handle(const CommandRequest& request) {
   if (flags.wait_for_launcher_secs > 0) {
     launcher_timeout.emplace(flags.wait_for_launcher_secs);
   }
+
+  std::vector<unsigned> instance_nums;
+  if (request.Selectors().instance_names) {
+    for (const auto& name : *request.Selectors().instance_names) {
+      std::vector<LocalInstance> instances = group.FindByInstanceName(name);
+      CF_EXPECTF(!instances.empty(), "Instance '{}' not found in group '{}'",
+                 name, group.GroupName());
+      for (const auto& inst : instances) {
+        instance_nums.push_back(inst.Id());
+      }
+    }
+  }
+
   Result<void> stop_outcome = instance_manager_.StopInstanceGroup(
       group, launcher_timeout,
       flags.clear_instance_dirs ? InstanceDirActionOnStop::Clear
-                                : InstanceDirActionOnStop::Keep);
+                                : InstanceDirActionOnStop::Keep,
+      instance_nums);
 
   GatherVmStopMetrics(group);
 
--- a/base/cvd/cuttlefish/host/commands/cvd/instances/instance_manager.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/instances/instance_manager.cpp
@@ -209,7 +209,21 @@ Result<void> InstanceManager::UpdateInstanceGroup(
 Result<void> InstanceManager::StopInstanceGroup(
     LocalInstanceGroup& group,
     std::optional<std::chrono::seconds> launcher_timeout,
-    InstanceDirActionOnStop instance_dir_action) {
+    InstanceDirActionOnStop instance_dir_action,
+    const std::vector<unsigned>& instance_nums) {
+  // Validate that the requested instances actually belong to this group
+  std::set<unsigned> valid_ids;
+  for (const auto& instance : group.Instances()) {
+    valid_ids.insert(instance.Id());
+  }
+  std::set<unsigned> ids_to_stop;
+  for (unsigned num : instance_nums) {
+    CF_EXPECTF(valid_ids.count(num) != 0,
+               "Instance ID {} does not belong to group '{}'", num,
+               group.GroupName());
+    ids_to_stop.insert(num);
+  }
+
   const auto stop_bin = CF_EXPECT(StopBin(group.HostArtifactsPath()));
   const auto stop_bin_path = group.HostArtifactsPath() + "/bin/" + stop_bin;
   int wait_for_launcher_secs = 0;
@@ -222,13 +236,23 @@ Result<void> InstanceManager::StopInstanceGroup(
       .wait_for_launcher_secs = wait_for_launcher_secs,
       .clear_runtime_dirs =
           instance_dir_action == InstanceDirActionOnStop::Clear,
+      .instance_nums = instance_nums,
   });
   if (!cmd_result.ok()) {
     LOG(WARNING)
         << "Warning: error stopping instances for dir \"" + group.HomeDir() +
                "\".\nThis can happen if instances are already stopped.\n";
+    return cmd_result;
   }
-  group.SetAllStates(cvd::INSTANCE_STATE_STOPPED);
+  if (instance_nums.empty()) {
+    group.SetAllStates(cvd::INSTANCE_STATE_STOPPED);
+  } else {
+    for (auto& instance : group.Instances()) {
+      if (ids_to_stop.count(instance.Id()) > 0) {
+        instance.SetState(cvd::INSTANCE_STATE_STOPPED);
+      }
+    }
+  }
   // TODO: b/471069557 - diagnose unused
   Result<void> unused = instance_db_.UpdateInstanceGroup(group);
   return {};
--- a/base/cvd/cuttlefish/host/commands/cvd/instances/instance_manager.h
+++ b/base/cvd/cuttlefish/host/commands/cvd/instances/instance_manager.h
@@ -83,7 +83,8 @@ class InstanceManager {
   Result<void> StopInstanceGroup(
       LocalInstanceGroup& group,
       std::optional<std::chrono::seconds> launcher_timeout,
-      InstanceDirActionOnStop instance_dir_action);
+      InstanceDirActionOnStop instance_dir_action,
+      const std::vector<unsigned>& instance_nums = {});
 
  private:
   struct InternalInstanceDesc {
--- a/base/cvd/cuttlefish/host/commands/cvd/instances/stop.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/instances/stop.cpp
@@ -32,6 +32,7 @@
 #include <fmt/core.h>
 #include <fmt/ranges.h>  // NOLINT(misc-include-cleaner): version difference
 #include "absl/log/log.h"
+#include "absl/strings/str_join.h"
 
 #include "cuttlefish/common/libs/utils/contains.h"
 #include "cuttlefish/common/libs/utils/files.h"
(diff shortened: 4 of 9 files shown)


base/cvd/cuttlefish/host/commands/stop/main.cc
```diff
@@ -173,6 +176,9 @@ FlagVaules GetFlagValues(int argc, char** argv) {
       GflagsCompatFlag("clear_instance_dirs", clear_instance_dirs)
           .Help("If provided, deletes the instance dir after attempting to "
                 "stop each instance."));
+  flags.emplace_back(
+      GflagsCompatFlag("instance_nums", instance_nums)
```
jemoreira, line 180: As of today `GflagsCompatFlag` supports passing a reference to a `vector<unsigned>` to parse a comma separated list of non negative integers. This should use that or add a new overload with the type you need, but it shouldn't do the parsing here.

base/cvd/cuttlefish/host/commands/cvd/instances/instance_manager.h
```diff
@@ -83,7 +83,8 @@ class InstanceManager {
   Result<void> StopInstanceGroup(
       LocalInstanceGroup& group,
       std::optional<std::chrono::seconds> launcher_timeout,
-      InstanceDirActionOnStop instance_dir_action);
+      InstanceDirActionOnStop instance_dir_action,
+      const std::string& instance_nums = "");
```
jemoreira: This should be `set<unsigned>` or `vector<unsigned>`. The instance manager should not be parsing strings and must validate that instances with those ids are in fact in the group.

base/cvd/cuttlefish/host/commands/cvd/instances/instance_manager.cpp
```diff
@@ -222,13 +225,30 @@ Result<void> InstanceManager::StopInstanceGroup(
...
   }
-  group.SetAllStates(cvd::INSTANCE_STATE_STOPPED);
+  if (instance_nums.empty()) {
+    group.SetAllStates(cvd::INSTANCE_STATE_STOPPED);
+  } else {
+    std::vector<std::string_view> tokens = absl::StrSplit(instance_nums, ",");
+    std::set<unsigned> ids_to_stop;
+    for (const auto& token : tokens) {
+      unsigned id;
+      if (android::base::ParseUint(std::string(token), &id)) {
```
jemoreira: use `absl::SimpleAtoi` to parse integers.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/stop.cpp
```diff
@@ -93,10 +95,26 @@ Result<void> CvdStopCommandHandler::Handle(const CommandRequest& request) {
...
+    std::vector<std::string> ids;
+    for (const auto& name : *request.Selectors().instance_names) {
+      auto instances = group.FindByInstanceName(name);
+      if (!instances.empty()) {
+        ids.push_back(std::to_string(instances.front().id()));
+      } else {
+        return CF_ERRF("Instance '{}' not found in group '{}'", name, group.GroupName());
+      }
+    }
+    instance_nums = android::base::Join(ids, ",");
```
jemoreira: Don't use android::base when there is an absl equivalent (absl::StrJoin in this case). We have a copy of android::base instead of a real dependency, so we want to get rid of it.

jemoreira (commented):
This is not very useful unless `cvd start` also supports specific instances. Are you planning to implement that too?

SuperStrongDinosaur:
> This is not very useful unless `cvd start` also supports specific instances. Are you planning to implement that too?

Yes, I am currently finishing up `cvd start` changes also

base/cvd/cuttlefish/host/commands/stop/main.cc
SuperStrongDinosaur: Done. Added template specializations

base/cvd/cuttlefish/host/commands/cvd/instances/instance_manager.h
SuperStrongDinosaur: Done. Updated StopInstanceGroup to accept new type

base/cvd/cuttlefish/host/commands/cvd/instances/instance_manager.cpp
SuperStrongDinosaur: Done. Since std::vector<unsigned> changes, no need to do this

base/cvd/cuttlefish/host/commands/cvd/cli/commands/stop.cpp
SuperStrongDinosaur: Done. again I am using now std::vector<unsigned> instance_nums

base/cvd/cuttlefish/host/commands/stop/main.cc
```diff
@@ -182,22 +187,34 @@ FlagVaules GetFlagValues(int argc, char** argv) {
...
   }
 
+  std::set<unsigned> instance_ids(instance_nums.begin(), instance_nums.end());
+
   int exit_code = 0;
   auto instances = config->Instances();
   std::vector<std::future<int>> exit_state_futures;
   exit_state_futures.reserve(instances.size());
   for (const auto& instance : instances) {
+    unsigned id;
```
jemoreira, line 223: IIRC stopping the first instance causes all other instances to stop too (or crash). Did you test that case?

jemoreira: I tested this myself and can confirm that stopping the first one also kills the second one (becomes unreachable). I think you're going to have to land the separate starts before being able to land this one.

You probably already know this, but just in case: The first instance runs some services that are used by all instances. To support this feature you're likely going to have to extract those to run as a standalone thing as part of the group.

SuperStrongDinosaur: I'm aware of this issue. I plan to introduce validation in a separate PR to prevent cvd stop and cvd restart from executing on the main instance, as doing so currently crashes all other instances.

Long-term, the ideal solution is to extract those daemons from the main instance. However, that is highly complex :), and we need the ability to manipulate instances separately within a reasonable timeframe. Given that this specific limitation isn't a blocker for our use case, the validation check is the best step for now

Databean: > I plan to introduce validation in a separate PR

Would rather see it in this PR. `stop/main.cc` can already load the `CuttlefishConfig` structure, which reports what the first instance id is. Something like
```c++
CuttlefishConfig::Get()->Instances()[0].id()
```
should produce the first instance id.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/stop.cpp
```diff
@@ -93,10 +93,24 @@ Result<void> CvdStopCommandHandler::Handle(const CommandRequest& request) {
...
+
+  std::vector<unsigned> instance_nums;
+  if (request.Selectors().instance_names) {
+    for (const auto& name : *request.Selectors().instance_names) {
+      auto instances = group.FindByInstanceName(name);
+      if (!instances.empty()) {
+        instance_nums.push_back(instances.front().id());
+      } else {
+        return CF_ERRF("Instance '{}' not found in group '{}'", name, group.GroupName());
+      }
```
Databean: nit: can be written as

```c++
CF_EXPECTF(!instances.empty(), "Instance '{}' not found in group '{}'", name, group.GroupName());
instance_nums.push_back(instances.front().id());
```

base/cvd/cuttlefish/host/commands/cvd/cli/commands/stop.cpp
```diff
@@ -93,10 +93,24 @@ Result<void> CvdStopCommandHandler::Handle(const CommandRequest& request) {
   if (flags.wait_for_launcher_secs > 0) {
     launcher_timeout.emplace(flags.wait_for_launcher_secs);
   }
+
+  std::vector<unsigned> instance_nums;
+  if (request.Selectors().instance_names) {
+    for (const auto& name : *request.Selectors().instance_names) {
+      auto instances = group.FindByInstanceName(name);
```
Databean: nit: [abseil.io/tips/232](https://abseil.io/tips/232) writes

> _Use type deduction only if it makes the code clearer to readers who aren’t familiar with the project, or if it makes the code safer. Do not use it merely to avoid the inconvenience of writing an explicit type._

base/cvd/cuttlefish/host/commands/cvd/instances/instance_manager.cpp
```diff
@@ -209,7 +209,21 @@ Result<void> InstanceManager::UpdateInstanceGroup(
...
+  // Validate that the requested instances actually belong to this group
+  std::set<unsigned> valid_ids;
+  for (const auto& instance : group.Instances()) {
+    valid_ids.insert(instance.id());
+  }
+  std::set<unsigned> ids_to_stop;
+  for (unsigned num : instance_nums) {
+    if (valid_ids.count(num) == 0) {
+      return CF_ERRF("Instance ID {} does not belong to group '{}'", num, group.GroupName());
+    }
```
Databean: nit: can be written as
```c++
CF_EXPECTF(valid_ids.count(num) == 0, "Instance ID {} does not belong to group '{}'", num, group.GroupName());
```

base/cvd/cuttlefish/host/commands/cvd/instances/stop.cpp
```diff
@@ -17,6 +17,7 @@
 #include "cuttlefish/host/commands/cvd/instances/stop.h"
 
 #include <errno.h>
+#include "absl/strings/str_join.h"
```
Databean: nit: include should be with the other abseil includes

SuperStrongDinosaur: done

base/cvd/cuttlefish/host/commands/cvd/instances/instance_manager.cpp
SuperStrongDinosaur: done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/stop.cpp
SuperStrongDinosaur: done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/stop.cpp
SuperStrongDinosaur: done

base/cvd/cuttlefish/host/commands/stop/main.cc
SuperStrongDinosaur: I added the check that prevents main instance to stop. Also I created a separate pr with this change for cvd restart 
https://github.com/google/android-cuttlefish/pull/2649

base/cvd/cuttlefish/host/commands/cvd/cli/commands/stop.cpp
```diff
@@ -93,10 +94,22 @@ Result<void> CvdStopCommandHandler::Handle(const CommandRequest& request) {
...
     launcher_timeout.emplace(flags.wait_for_launcher_secs);
   }
+
+  std::vector<unsigned> instance_nums;
+  if (request.Selectors().instance_names) {
+    for (const auto& name : *request.Selectors().instance_names) {
+      std::vector<LocalInstance> instances = group.FindByInstanceName(name);
+      CF_EXPECTF(!instances.empty(), "Instance '{}' not found in group '{}'",
+                 name, group.GroupName());
+      instance_nums.push_back(instances.front().id());
```
Databean: Rather than looking only at `instances.front()`, can it either validate that `instances.size() == 1` or apply to every member of `instances`?

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/stop.cpp
SuperStrongDinosaur: Yes, that works. Done.

SuperStrongDinosaur:
Rebased + squashed commits

PR #2617, opened 2026-05-28
Support per-instance start bypassing assembly

SuperStrongDinosaur:
### **Summary**

Running cvd start typically triggers assemble_cvd, which regenerates the cuttlefish_config.json for all configured instances. Running this in a shared environment for a single instance can overwrite and corrupt the configuration of already running instances. This change allows starting a specific stopped instance using its existing assembled configuration.

### **Key Changes**

Assembly Bypass: Modified CvdStartCommandHandler::Handle to detect if a specific instance is targeted via selectors. Added a disk check for cuttlefish_config.json before bypassing assembly. If the configuration is missing, it correctly falls back to the normal group start path, runs assemble_cvd, and boots all instances.
LaunchSingleInstance Implementation: Added LaunchSingleInstance which directly constructs and invokes the run_cvd command for the target instance, bypassing assemble_cvd.
Environment Setup: Ensures LaunchSingleInstance sets up the exact environment variables required for run_cvd to boot the specific instance.
Database Status Update Fix: Updated LaunchSingleInstance to only update the database state for the targeted instance being started, preventing other stopped instances in the group from being incorrectly marked as active.

### Bug
b/459780275

### Document

### Frontend pr
https://github.com/google/android-cuttlefish/pull/2618

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
@@ -16,6 +16,7 @@
 
 #include "cuttlefish/host/commands/cvd/cli/commands/start.h"
 
+#include <fcntl.h>
 #include <signal.h>  // IWYU pragma: keep
 #include <stddef.h>
 #include <stdlib.h>
@@ -40,6 +41,7 @@
 #include "absl/strings/str_join.h"
 #include "absl/strings/str_split.h"
 
+#include "cuttlefish/common/libs/fs/shared_fd.h"
 #include "cuttlefish/common/libs/utils/contains.h"
 #include "cuttlefish/common/libs/utils/files.h"
 #include "cuttlefish/flag_parser/flag.h"
@@ -327,6 +329,21 @@ static Result<void> ConsumeDaemonModeFlag(cvd_common::Args& args) {
   return {};
 }
 
+static bool HasUnsafeFlagsForBypass(const std::vector<std::string>& args) {
+  std::vector<std::string> args_copy = args;
+  bool daemon = true;
+  std::string report_anonymous = "";
+  std::vector<Flag> safe_flags = {
+      GflagsCompatFlag("daemon", daemon),
+      GflagsCompatFlag("report_anonymous_usage_stats", report_anonymous),
+  };
+  auto res = ConsumeFlags(safe_flags, args_copy);
+  if (!res.ok()) {
+    return true;
+  }
+  return !args_copy.empty();
+}
+
 Result<void> CvdStartCommandHandler::Handle(const CommandRequest& request) {
   std::vector<std::string> subcmd_args = request.SubcommandArguments();
   CF_EXPECT(!GetConfigPath(subcmd_args).has_value(),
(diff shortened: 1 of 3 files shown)


base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
```diff
@@ -337,6 +339,19 @@ Result<void> CvdStartCommandHandler::Handle(const CommandRequest& request) {
...
   }
 
+  if (request.Selectors().instance_names) {
+    auto [instance, group] =
+        CF_EXPECT(selector::SelectInstance(instance_manager_, request));
+
+    auto config_path =
+        group.AssemblyDir() + "/cuttlefish_config.json";
+    if (FileExists(config_path)) {
+      return LaunchSingleInstance(instance, group, request);
```
Databean: nit: `return CF_EXPECT(LaunchSingleInstance(...))`

The macro is necessary to add an additional member to the stack trace held inside the `Result`.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
```diff
@@ -337,6 +339,19 @@ Result<void> CvdStartCommandHandler::Handle(const CommandRequest& request) {
     return CF_ERR(NoGroupMessage(request));
   }
 
+  if (request.Selectors().instance_names) {
+    auto [instance, group] =
+        CF_EXPECT(selector::SelectInstance(instance_manager_, request));
+
+    auto config_path =
+        group.AssemblyDir() + "/cuttlefish_config.json";
```
Databean: nit: the [style guide](https://google.github.io/styleguide/cppguide.html#Type_deduction) writes

> Use type deduction only if it makes the code clearer to readers who aren't familiar with the project, or if it makes the code safer. Do not use it merely to avoid the inconvenience of writing an explicit type.

The destructuring `auto` above is fair since it's the only way to write something like that, but in this case it's helpful while reading to know what the types are when possible.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
```diff
@@ -337,6 +339,19 @@ Result<void> CvdStartCommandHandler::Handle(const CommandRequest& request) {
...
+    if (FileExists(config_path)) {
+      return LaunchSingleInstance(instance, group, request);
+    } else {
+      LOG(INFO) << "Group configuration does not exist on disk. Proceeding with normal group start.";
```
Databean: `INFO` is kind of high. Any launch from `cvd load` or `cvd create --config_file` will trigger this:

https://github.com/google/android-cuttlefish/blob/caa0fd4f25ad37de966a412537a725e47bb72a71/base/cvd/cuttlefish/host/commands/cvd/cli/parser/selector_parser.cpp#L32-L35

Maybe `DEBUG`?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
Databean: From the perspective of `cvd`, whether `cuttlefish_config.json` exists is an implementation detail of `assemble_cvd` and it would be better to avoid more dependency on this fact than we already have.

Is it possible to determine the conditions for re-launching an instance through whether `cvd start` is being invoked on an instance state matching `INSTANCE_STATE_STOPPED`?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
SuperStrongDinosaur: I've updated the code to use  CF_EXPECT  to propagate the stack trace.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
SuperStrongDinosaur: refactored and removed

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
SuperStrongDinosaur: Done.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
SuperStrongDinosaur: Done. Replaced the  FileExists  check on the config file with  instance.State() == cvd::INSTANCE_STATE_STOPPED

Databean (approved):

PR #2618, opened 2026-05-28
Route per-instance start API

SuperStrongDinosaur:
### **Summary**
Exposes per-instance start functionality through the HO API. This change ensures that users can start a specific stopped instance, with support for passing a SnapshotID in the request body to boot the instance from a snapshot.

### **Key Changes**
Dedicated Handler: Implemented newStartCVDInstanceHandler and startCVDInstanceHandler in controller.go. This handler correctly decodes the apiv1.StartCVDRequest from the POST request body.
StartCVDInstanceAction: Added StartCVDInstanceAction in startcvdaction.go to handle the logic. It validates the request, locates the snapshot if provided, and invokes cvdCLI.LazySelectInstance().Start(opts).
Route Redirection: Updated controller.go to map /cvds/{group}/{name}/:start to newStartCVDInstanceHandler instead of the group instance command handler.

### **Bug**
b/459780275

### **Document**

### **Backend pr**
https://github.com/google/android-cuttlefish/pull/2617


jemoreira (commented):
LG but please wait until support for the operation is added in `cvd` before merging this one

jemoreira (approved):
I don't see anything related to snapshots in this change, should the title be updated?

PR #2619, opened 2026-05-28
Make execution of `podcvd-setup` with sudo permision being available.

0405ysj:
Current `podcvd-setup` only allows regular users, but not the root user. However if the root user can execute `podcvd-setup`, we may not even need to execute `podcvd-setup` manually by adding it into post-installation step of `cuttlefish-podcvd`. This PR aims for supporting `sudo podcvd-setup [username]`, and also refactored some.

For more contexts including motivation and how `sudo podcvd-setup [username]` will be used, please refer to the context.

Context: b/514860208

### DIFF
--- a/container/debian/podcvd-setup
+++ b/container/debian/podcvd-setup
@@ -16,6 +16,17 @@
 
 set -e
 
+# Execute this script with sudo again, with passing username as an argument.
+if [ "$(id -u)" -ne 0 ]; then
+    username="${USER:-$(id -un)}"
+    if [ -n "$1" ] && [ "$1" != "$username" ]; then
+        echo "Error: Cannot setup for another user ('$1') when running as regular user."
+        exit 1
+    fi
+    echo "Requesting sudo privileges to setup podcvd for user '$username'..."
+    exec sudo "$0" "$username"
+fi
+
 setup_cidr() {
     # Read default CIDR from /etc/default/cuttlefish-podcvd or fallback
     local podcvd_cidr="192.168.112.0/20"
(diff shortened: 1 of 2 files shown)


container/debian/podcvd-setup
```diff
@@ -16,6 +16,18 @@
 
 set -e
 
+# Execute this script with sudo again, with passing username as an argument.
```
jemoreira, line 19: The username argument isn't needed. When a command is run with `sudo` it normally defines the environment variables `SUDO_USER`, `SUDO_TTY`, `SUDO_GID` and `SUDO_UID`.

jemoreira: `sudo podcvd-setup <username>` would be useful for an admin to give another user access, but the username parameter shouldn't be required to give yourself access.

0405ysj: I'm currently thinking of executing `podcvd-setup` from the installation step of the debian package. As I know, environment variables such as `SUDO_USER` can't be working for this case.

jemoreira: Then please change the title and description of the PR to better define the problem you're trying to solve

jemoreira: As I hinted at in my second comment: You can still land this change supporting `sudo podcvd {user}` as long as `sudo podcvd` defaults to the user executing sudo.

0405ysj: I made the username parameter as optional on `podcvd-setup`. Please reopen this comment if I misunderstood your intention. Thanks!

container/debian/podcvd-setup
```diff
@@ -16,125 +16,147 @@
...
-    exit 1
+# Execute this script with sudo again, with passing username as an argument.
+if [ "$(id -u)" -ne 0 ]; then
+    username="${USER:-$(id -un)}"
+    if [ -n "$1" ] && [ "$1" != "$username" ]; then
+        echo "Error: Cannot setup for another user ('$1') when running as regular user."
+        exit 1
+    fi
+    echo "Requesting sudo privileges to setup podcvd for user '$username'..."
+    SCRIPT_PATH=$(realpath "$0")
```
jemoreira: nit: You probably don't need `realpath` here. A relative path will work too because you haven't changed directory. As a user, if I type `thecommand dosomething` I'd expect it to execute `sudo thecommand dosomething` not `sudo /path/to/thecommand dosomething`.

container/debian/podcvd-setup
```diff
@@ -16,125 +16,147 @@
...
+        usermod --add-subuids "$id_range" --add-subgids "$id_range" "$username"
+    fi
+
+    # Migrate subuid/subgid setting into rootless podman
+    echo "Migrating subuid/subgid setting into rootless podman..."
+    sudo -H -u "$username" podman system migrate
+}
+
+USER_CONFIG="/etc/podcvd.users"
+username="${1:-$SUDO_USER}"
```
jemoreira: `username="${1:-${SUDO_USER:-$(id -un)}}"`

Without it, if the user logs in as root (not using sudo) and runs `podcvd-setup` they would get an error message like "Error: User '' does not exist.". With that line they'll get "Error: Target user 'root' cannot be root."

container/debian/podcvd-setup
```diff
@@ -16,125 +16,147 @@
...
+if [ -z "$username" ] || ! id "$username" >/dev/null 2>&1; then
+    echo "Error: User '$username' does not exist."
+    exit 1
+fi
+if [ "$(id -u "$username")" -eq 0 ]; then
+    echo "Error: Target user '$username' cannot be root."
```
jemoreira: nit: This will print "Error: Target user 'root' cannot be root.", drop the '$username' so it prints "Error: Target user cannot be root." instead.

jemoreira (approved):

container/debian/podcvd-setup
0405ysj: Thanks for catching it, done.

container/debian/podcvd-setup
0405ysj: Agree with your opinion. Removed the use of `realpath`. Done by just executing $0.

container/debian/podcvd-setup
0405ysj: Thanks, it sounds good to me, as the username won't be empty with that approach.

0405ysj:
Thanks for the review!

PR #2627, opened 2026-05-28
Resolve race condition for unlinking file on base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc

0405ysj:
While launching multiple Cuttlefish instances simultaneously by `cvdr create`, I observed ENOENT error could be observed from the given line though it's guarded by `FileExists`. So it makes paralleled execution of `cvdr create` being flaky. This commit fixes the issue by just trying to do unlink and throw an error only when the error is not ENOENT.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
+++ b/base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
@@ -64,9 +64,8 @@ Result<void> Substitute(const std::string& target,
 
   CF_EXPECT(EnsureDirectoryExists(android::base::Dirname(full_link_name)));
 
-  if (FileExists(full_link_name)) {
-    CF_EXPECTF(unlink(full_link_name.c_str()) == 0, "{}", StrError(errno));
-  }
+  int unlink_res = unlink(full_link_name.c_str());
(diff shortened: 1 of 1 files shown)


base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
```diff
@@ -64,8 +64,8 @@ Result<void> Substitute(const std::string& target,
 
   CF_EXPECT(EnsureDirectoryExists(android::base::Dirname(full_link_name)));
 
-  if (FileExists(full_link_name)) {
-    CF_EXPECTF(unlink(full_link_name.c_str()) == 0, "{}", StrError(errno));
+  if (unlink(full_link_name.c_str()) != 0) {
+    CF_EXPECTF(errno == ENOENT, "{}", StrError(errno));
   }
```
jemoreira: nit: Rewrite as
```
int unlink_res = unlink(full_link_name.c_str());
CF_EXPECTF(unlink_res == 0 || errno == ENOENT, StrError(errno));
```

jemoreira (approved):

base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
0405ysj: Fixed, thanks!

PR #2629, opened 2026-05-29
Overhaul `cvd help create` output

jemoreira:
More detailed description, with clear separation between having the config file and not.

References the flags from load and create prevent drift going forward.

Refactors creation analyzer and start selector parser.

Bug: b/517657871


base/cvd/cuttlefish/common/libs/utils/flag_parser.cpp
```diff
@@ -790,6 +824,18 @@ Flag GflagsCompatFlag(const std::string& name,
...
+                      std::optional<std::vector<unsigned>>& value,
+                      CoerceToNullopt opt) {
+  return GflagsCompatFlag(name)
+      .Getter([&value, opt]() { return FlagValueToString(value, opt); })
+      .Setter([&value, opt](const FlagMatch& match) -> Result<void> {
+        value = CF_EXPECT(
+            ParseFlagOptionalValue<std::vector<unsigned>>(match.value, opt));
+        return {};
+      });
+}
```
Databean, line 801: These 5 implementations of `GflagsCompatFlag` all look mostly the same, is it possible to deduplicate them with templates?

base/cvd/cuttlefish/common/libs/utils/flag_parser.cpp
```diff
@@ -61,6 +61,24 @@ constexpr std::array help_str_opts{
...
+  errno = 0;
+  auto r = strtoll(base, &end, /* auto-detect */ 0);
+  if (errno != 0 || end != base + value.size()) {
+    return {};
+  }
+  if (static_cast<T>(r) != r) {
+    return {};
+  }
+  return r;
+}
```
Databean: Maybe this function or its callers could use `absl::SimpleAtoi`?

https://github.com/abseil/abseil-cpp/blob/fe126c727b925b7c3b2742f6f54690973216952b/absl/strings/numbers.h#L58-L66

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
```diff
@@ -265,26 +203,67 @@ Result<void> CreateSymlinks(const LocalInstanceGroup& group) {
...
+// are parsed before calling the handler's functions so this is not possible.
+// The other option would be to use differently named flags altogether, but that
+// would break backwards compatibility at this point.
+// This function produces the same flags with different help message.
+std::vector<Flag> BuildSelectorFlagsForCreateHelp(
+    const selector::SelectorOptions& selectors) {
+  return {
+      GflagsCompatFlag(selector::SelectorFlags::kGroupName)
+          .Getter([selectors]() { return selectors.group_name.value_or(""); })
+          .Setter([](const FlagMatch&) -> Result<void> { return {}; })
```
Databean: Do the `Setter`s of these `Flag`s need to be defined if they are never going to be called?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
```diff
@@ -323,13 +302,138 @@ Result<void> CvdCreateCommandHandler::Handle(const CommandRequest& request) {
...
+          "depending on whether the `--config_file` flag is provided."),
+
+      HelpParagraph("ENVIRONMENT SPECIFICATION FILE MODE"),
+
+      HelpParagraph(
+          "When `--config_file` is provided, `cvd create` loads a JSON "
+          "configuration file to define and the instance group. This allows "
+          "launching complex environments with multiple instances, custom VM "
+          "configurations (CPUs, memory), and mixed build sources. In this "
+          "mode the instances in the group are always started immediately."),
```
Databean: Can this reuse the same text as `load_configs.cpp` without copying it? The two places could get out of sync.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
```diff
@@ -323,13 +302,138 @@ Result<void> CvdCreateCommandHandler::Handle(const CommandRequest& request) {
...
+      "--nostart is provided):")});
+
+  ss << FormatFlagsHelp(
+      (CF_EXPECT(CvdStartCommandHandler(instance_manager_).Flags(request))));
+
+  return ss.str();
+}
+
+std::vector<Flag> CvdCreateCommandHandler::ConfigFileModeFlags() {
+  own_flags_.config_file = "";
```
Databean, line 395: For this and FlagModeFlags, it's unintuitive that calling it already modifies state. This extends to DetailedHelp which calls the two *Flags functions.

If I understand correctly `Handle` and `DetailedHelp` are both top-level entry points now. Can they both call a new function that sets all the defaults based on the selector flags?

base/cvd/cuttlefish/host/commands/cvd/cli/selector/num_instances_parser.cpp
```diff
@@ -0,0 +1,121 @@
...
+    const selector::SelectorOptions& selector_options) {
+  if (selector_options.instance_names) {
+    num_instance_names_ = selector_options.instance_names->size();
+  }
+  return {
+      GflagsCompatFlag("num_instances", num_instances_)
+          .AddValidator([this]() -> Result<void> {
+            CF_EXPECT(Validate());
+            return {};
+          })
```
Databean: It loses the stack trace element of this anonymous function, but it would work to write
```c++
.AddValidator(std::bind_front(&NumInstancesParser::Validate, this))
```
[Compiler explorer](https://godbolt.org/z/Wfr9K5sfK)

base/cvd/cuttlefish/host/commands/cvd/cli/selector/num_instances_parser.cpp
```diff
@@ -0,0 +1,121 @@
...
+  // The provided instance numbers must be distinct
+  if (instance_nums_) {
+    std::set<int32_t> instance_nums;
+    for (int32_t num : *instance_nums_) {
+      CF_EXPECTF(instance_nums.find(num) == instance_nums.end(),
+                 "Instance ids must be unique, but {} is repeated.", num);
+    }
+  }
+
+  // --base_instance_nums and --instance_nums are mutually exclusive
```
Databean: nit: comment is redundant with the condition

base/cvd/cuttlefish/host/commands/cvd/cli/selector/num_instances_parser_test.cpp
```diff
@@ -0,0 +1,140 @@
...
+#include "cuttlefish/host/commands/cvd/cli/types.h"
+#include "cuttlefish/result/result_matchers.h"
+
+namespace cuttlefish {
+namespace selector {
+
+struct NumInstancesParserTestParam {
+  SelectorOptions selector_options;
+  cvd_common::Args args;
+  bool should_fail;
```
Databean: nit: this should have a default value of `false`, I'm not sure that `bool` has a reasonable default initializer if unspecified.

base/cvd/cuttlefish/host/commands/cvd/cli/selector/num_instances_parser_test.cpp
```diff
@@ -0,0 +1,140 @@
...
+TEST_P(NumInstancesParserTest, ParseFlags) {
+  NumInstancesParser parser;
+  cvd_common::Args args = GetParam().args;
+  Result<void> parse_result =
+      ConsumeFlags(parser.Flags(GetParam().selector_options), args);
+  if (GetParam().should_fail) {
+    ASSERT_THAT(parse_result, IsError());
+  } else {
+    ASSERT_THAT(parse_result, IsOk());
+  }
```
Databean, line 71: `GetParam().should_fail` is the only parameter actually used for validation. `expected_num_instances` and `expected_instance_nums` aren't compared against anything.

Can this test validate that `expected_num_instances` and `expected_instance_nums` are correct following the struct?

base/cvd/cuttlefish/host/commands/cvd/cli/selector/num_instances_parser_test.cpp
```diff
@@ -0,0 +1,140 @@
...
+  NumInstancesParserTestParam& ExpectNum(size_t num) {
+    expected_num_instances = num;
+    return *this;
+  }
+  NumInstancesParserTestParam& ExpectIds(std::vector<unsigned> ids) {
+    expected_num_instances = ids.size();
+    expected_instance_nums = ids;
+    return *this;
+  }
```
Databean, line 55: Not sure it'll be an improvement, but an alternative is having the test instantiation use [aggregate initialization](https://en.cppreference.com/cpp/language/aggregate_initialization)

Databean (approved):

base/cvd/cuttlefish/common/libs/utils/flag_parser.cpp
jemoreira: done

base/cvd/cuttlefish/common/libs/utils/flag_parser.cpp
jemoreira: done

also did it for vector and integer types

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
jemoreira: no

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
jemoreira: done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
jemoreira: The Flags() function receives the request as parameter precisely so the default values can be set accordingly. This class needs to do something else and doesn't implement that function, but these are very similar in spirit. The flags are used in two different places, the common function that gets called is the one that creates the flags, I think that's the best place to set the default.
IMO it's unfortunate that the defaults depend on the request itself, but even if that's wrong I don't want to change it in this PR.

base/cvd/cuttlefish/host/commands/cvd/cli/selector/num_instances_parser.cpp
jemoreira: done

base/cvd/cuttlefish/host/commands/cvd/cli/selector/num_instances_parser.cpp
jemoreira: done

base/cvd/cuttlefish/host/commands/cvd/cli/selector/num_instances_parser_test.cpp
jemoreira: done

base/cvd/cuttlefish/host/commands/cvd/cli/selector/num_instances_parser_test.cpp
jemoreira: That's how parser_ids_test.cpp was implemented and how I initially implemented it, but the tests themselves were hard to read. Having these allows the test to be written as `Param().WithX().WithY().ExpectZ()` which is easier to understand IMHO.

base/cvd/cuttlefish/host/commands/cvd/cli/selector/num_instances_parser_test.cpp
jemoreira: yeah, I apparently forgot to finish implementing the test.
Removed the optionals and made it check both every time should_fail is false to ensure InstaceIds() returns empty when  no ids are given.

PR #2635, opened 2026-06-01
Fix unstable channel workflow trigger

3405691582:
Update the branch pattern to 'version-*-dev' to support 3-digit version branches like 'version-1.54.1-dev'. Also add 'workflow_dispatch' to support manual triggering.

### DIFF
--- a/.github/workflows/unstable-channel.yaml
+++ b/.github/workflows/unstable-channel.yaml
@@ -3,7 +3,8 @@ name: Unstable channel
 on:
   push:
     branches:
-      - 'version-[0-9]+.[0-9]+-dev'
+      - 'version-*-dev'
+  workflow_dispatch:
 
 permissions:
   contents: read


ser-io (approved):

0405ysj:
My understanding was having 2-digit versioned branches(`version-{major}.{minor}-dev`), and have 3-digit versioned tag(`v{major}.{minor}.{release}`) there. So I've thought there are no needs for 3-digit versioned branches until now. Would there be any policy changes for the release?

I wonder about it as I accordingly need to modify [update-cache-and-deployment.yaml](https://github.com/google/android-cuttlefish/blob/main/.github/workflows/update-cache-and-deployment.yaml) to support 3-digit versioned branches (and also manual triggering).

3405691582:
This was intended for a .1 dot release with cherrypicks. Without this, we won't be running the "Unstable Channel" workflow for dot releases, which is likely not what was intended.

0405ysj:
I may need to change the question, would 3-digit versioned branches(version-1.54.1-dev) be general for the future releases, or just for triggering "Unstable Channel" for this time? I'm understanding that all of .0, .1, or .X dot releases were dealt on top of 2-digit versioned branches. For [example](https://github.com/google/android-cuttlefish/commits/version-1.52-dev/), both 1.52.0 and 1.52.1 were at version-1.52-dev.

PR #2643, opened 2026-06-02
Migrate hinge angle injection to sensors simulator

dimorinny:
The cuttlefish_sensor_injection tool does not work with goldfish multihal sensors, meaning hinge angle changes currently have no effect

To fix this, redirect WebRTC hinge angle updates to the sensors simulator instead.

Test: hinge angle is updated from web ui
Bug: 517370262

### DIFF
--- a/base/cvd/cuttlefish/host/commands/sensors_simulator/sensors_hal_proxy.cpp
+++ b/base/cvd/cuttlefish/host/commands/sensors_simulator/sensors_hal_proxy.cpp
@@ -16,6 +16,9 @@
 
 #include "cuttlefish/host/commands/sensors_simulator/sensors_hal_proxy.h"
 
+#include <mutex>
+#include <string>
+
 #include "absl/log/check.h"
 #include "absl/log/log.h"
 
@@ -34,6 +37,13 @@ static constexpr SensorsMask kContinuousModeSensors =
     (1 << kAccelerationId) | (1 << kGyroscopeId) | (1 << kMagneticId) |
     (1 << kPressureId) | (1 << kUncalibGyroscopeId) |
     (1 << kUncalibAccelerationId) | (1 << kLightId);
+
+/*
+  On-change sensors are only reported to the guest when their value changes
+  (and once after the guest HAL (re)starts), instead of being streamed every
+  cycle like the continuous mode sensors above.
+*/
+static constexpr SensorsMask kOnChangeSensors = (1 << kHingeAngle0Id);
 
 Result<std::string> SensorIdToName(int id) {
   switch (id) {
@@ -72,10 +82,10 @@ Result<std::string> SensorIdToName(int id) {
 
 Result<void> SendResponseHelper(transport::SharedFdChannel& channel,
                                 const std::string& msg) {
-  auto size = msg.size();
-  auto cmd = sensors::kUpdateHal;
-  auto response = CF_EXPECT(transport::CreateMessage(cmd, msg.size()),
-                            "Failed to allocate message.");
+  size_t size = msg.size();
+  int cmd = sensors::kUpdateHal;
+  transport::ManagedMessage response = CF_EXPECT(
+      transport::CreateMessage(cmd, msg.size()), "Failed to allocate message.");
   std::memcpy(response->payload, msg.data(), size);
   CF_EXPECT(channel.SendResponse(*response), "Can't update sensor HAL.");
   return {};
@@ -84,12 +94,12 @@ Result<void> SendResponseHelper(transport::SharedFdChannel& channel,
 Result<void> ProcessHalRequest(transport::SharedFdChannel& channel,
                                std::atomic<bool>& hal_activated,
                                uint32_t mask) {
-  auto request =
+  transport::ManagedMessage request =
       CF_EXPECT(channel.ReceiveMessage(), "Couldn't receive message.");
   std::string payload(reinterpret_cast<const char*>(request->payload),
                       request->payload_size);
   if (payload.rfind("list-sensors", 0) == 0) {
-    auto msg = std::to_string(mask) + END_OF_MSG;
+    std::string msg = std::to_string(mask) + END_OF_MSG;
     CF_EXPECT(SendResponseHelper(channel, msg));
     hal_activated = true;
   }
@@ -107,7 +117,7 @@ Result<void> UpdateSensorsHal(const std::string& sensors_data,
   while (mask) {
     if (mask & 1) {
       CF_EXPECT(static_cast<bool>(sensors_data_stream >> report));
-      auto result = SensorIdToName(id);
+      Result<std::string> result = SensorIdToName(id);
       if (result.ok()) {
         reports.push_back(result.value() + INNER_DELIM + report + END_OF_MSG);
       }
@@ -162,6 +172,10 @@ SensorsHalProxy::SensorsHalProxy(SharedFD control_from_guest_fd,
       sensors_simulator_(sensors_simulator) {
   const SensorsMask host_enabled_sensors = HostEnabledSensors(device_type);
 
+  sensors_simulator_.SetSensorsChangedCallback(
+      host_enabled_sensors & kOnChangeSensors,
+      [this](SensorsMask changed) { ReportToGuest(changed); });
+
   req_responder_thread_ = std::thread([this, host_enabled_sensors] {
     while (running_) {
       auto result = ProcessHalRequest(control_channel_, hal_activated_,
@@ -174,24 +188,14 @@ SensorsHalProxy::SensorsHalProxy(SharedFD control_from_guest_fd,
   });
   data_reporter_thread_ = std::thread([this, host_enabled_sensors] {
     while (running_) {
-      if (hal_activated_) {
-        SensorsMask host_update_sensors =
-            host_enabled_sensors & kContinuousModeSensors;
-        auto sensors_data =
-            sensors_simulator_.GetSensorsData(host_update_sensors);
-        auto result =
-            UpdateSensorsHal(sensors_data, data_channel_, host_update_sensors);
-        if (!result.ok()) {
-          running_ = false;
-          LOG(ERROR) << result.error();
-        }
-      }
+      ReportToGuest(host_enabled_sensors & kContinuousModeSensors);
       std::this_thread::sleep_for(std::chrono::milliseconds(kIntervalMs));
     }
   });
   reboot_monitor_thread_ = std::thread([this] {
     while (kernel_events_fd_->IsOpen()) {
-      auto read_result = monitor::ReadEvent(kernel_events_fd_);
+      Result<std::optional<monitor::ReadEventResult>> read_result =
+          monitor::ReadEvent(kernel_events_fd_);
       CHECK(read_result.ok()) << read_result.error();
       CHECK(read_result->has_value()) << "EOF in kernel log monitor";
       if ((*read_result)->event == monitor::Event::BootloaderLoaded) {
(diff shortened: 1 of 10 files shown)


base/cvd/cuttlefish/host/commands/sensors_simulator/sensors_hal_proxy.cpp
```diff
@@ -201,5 +209,18 @@ SensorsHalProxy::SensorsHalProxy(SharedFD control_from_guest_fd,
   });
 }
 
+void SensorsHalProxy::ReportToGuest(SensorsMask mask) {
+  if (!mask) {
+    return;
+  }
+  auto sensors_data = sensors_simulator_.GetSensorsData(mask);
+  std::lock_guard<std::mutex> lock(report_mtx_);
+  auto result = UpdateSensorsHal(sensors_data, data_channel_, mask);
```
Databean: I see that this was moved from `data_reporter_thread_`, but note [abseil.io/tips/232](https://abseil.io/tips/232)

> Use type deduction only if it makes the code clearer to readers who aren’t familiar with the project, or if it makes the code safer. Do not use it merely to avoid the inconvenience of writing an explicit type.

base/cvd/cuttlefish/host/commands/sensors_simulator/sensors_hal_proxy.cpp
```diff
@@ -162,6 +172,13 @@ SensorsHalProxy::SensorsHalProxy(SharedFD control_from_guest_fd,
       sensors_simulator_(sensors_simulator) {
   const SensorsMask host_enabled_sensors = HostEnabledSensors(device_type);
 
+  sensors_simulator_.SetOnChangeCallback(
+      [this, host_enabled_sensors](SensorsMask changed) {
+        if (hal_activated_) {
+          ReportToGuest(changed & host_enabled_sensors & kOnChangeSensors);
+        }
```
Databean: Both callers of `ReportToGuest` check `hal_activated_`, should this happen inside the `ReportToGuest method instead of outside?

base/cvd/cuttlefish/host/commands/sensors_simulator/sensors_simulator.cpp
```diff
@@ -132,6 +137,19 @@ void SensorsSimulator::RefreshSensors(double x, double y, double z) {
...
+  {
+    std::lock_guard<std::mutex> lock(sensors_data_mtx_);
+    if (sensors_data_[kHingeAngle0Id].f == angle) {
+      return;
+    }
+    sensors_data_[kHingeAngle0Id].f = angle;
+  }
+  if (on_change_callback_) {
+    on_change_callback_(1 << kHingeAngle0Id);
+  }
```
Databean: Is there anything else that should also invoke this new callback? `SetMotion` also affects state and seems relevant to something called "on change callback".

base/cvd/cuttlefish/host/frontend/webrtc/sensors_handler.cpp
```diff
@@ -36,61 +36,49 @@ static constexpr sensors::SensorsMask kUiSupportedSensors =
...
 
-Result<void> SensorsHandler::RefreshSensors(const double x, const double y,
-                                            const double z) {
-  std::stringstream ss;
-  ss << x << sensors::INNER_DELIM << y << sensors::INNER_DELIM << z;
-  auto msg = ss.str();
-  auto size = msg.size();
-  auto cmd = sensors::kUpdateRotationVec;
+Result<void> SensorsHandler::SendCommand(uint32_t cmd,
+                                         const std::string& payload) {
```
Databean: nit: this could accept a `std::string_view payload`, since it makes a copy of the data stored in it anyway.

base/cvd/cuttlefish/host/frontend/webrtc/html_client/js/app.js
```diff
@@ -352,18 +352,11 @@ class DeviceControlApp {
...
               this.#getCustomDeviceStateButtonCb(button.device_states),
               'control-panel-custom-buttons');
-          for (const device_state of button.device_states) {
-            // hinge_angle is currently injected via an adb shell command that
-            // triggers a guest binary.
-            if ('hinge_angle_value' in device_state) {
-              element.dataset.adb = true;
-            }
-          }
+          element.dataset.adb = true;
```
Databean: Should this still be here? If I understand correctly, the code before amounted to something like "if the device state includes `hinge_angle`, make sure `adb` is connected."

As written this commit changes it to "always connect to adb," but I would expect "the hinge angle is not sent via adb" to imply "adb no longer matters for sensors, don't set or unset the adb connection requirement."

I may be misunderstanding what the `.adb` variable is for. FYI @jemoreira

dimorinny: From the change perspective this makes sense, I'm agree.

However, after checking the `getCustomDeviceStateButtonCb` [implementation](https://github.com/google/android-cuttlefish/blob/main/base/cvd/cuttlefish/host/frontend/webrtc/html_client/js/app.js#L736), it looks to me like any action on that panel [triggers](https://github.com/google/android-cuttlefish/blob/main/base/cvd/cuttlefish/host/frontend/webrtc/html_client/js/app.js#L742) `adbShell('cmd device_state state reset')`, meaning ADB is executed regardless of whether we change the hinge angle, the lid, or anything else, so I assume the entire panel should be disabled if ADB isn't available.

Do you think I'm missing something?

base/cvd/cuttlefish/host/commands/sensors_simulator/sensors_hal_proxy.cpp
dimorinny: Good call. Moved the hal_activated_ check into ReportToGuest and removed it from both callers.

base/cvd/cuttlefish/host/commands/sensors_simulator/sensors_hal_proxy.cpp
dimorinny: Agreed with making `sensors_data` explicit (`std::string`) above, the type isn't obvious from `GetSensorsData`, but I kept result as `auto` since `auto result = …; if (!result.ok())` for `Result<void>` is the idiom used throughout this file and the wider codebase. Happy to spell it out too if you'd rather go the other way.

base/cvd/cuttlefish/host/commands/sensors_simulator/sensors_simulator.cpp
dimorinny: Good catch, you're right it's odd for a "changed" callback to fire only from `SetHingeAngle`.

Context: the HAL proxy splits sensors into continuous (`kContinuousModeSensors`, streamed to the guest every cycle) and on-change (`kOnChangeSensors`, currently just the hinge angle). Motion sensors are continuous, so the polling loop already reports them regularly; there's nothing to push when `SetMotion` runs.

The catch is that answering "no, SetMotion shouldn't fire it" means teaching the simulator which sensors are continuous vs on-change. But that's the proxy's concept, and leaking it into the simulator is exactly what made this confusing in the first place.

So I made the callback generic instead: both `SetMotion` and `SetHingeAngle` just report whatever they changed, and the subscriber passes a mask of what it cares about. The proxy subscribes with `kOnChangeSensors`, so motion updates get filtered out there and the continuous/on-change policy stays entirely in the proxy where it belongs.

Do you think it looks cleaner now?

base/cvd/cuttlefish/host/frontend/webrtc/html_client/js/app.js
jemoreira: Looks like adb should have been set regardless of device state including hinge angle or not.

The question that remains is: Is that device reset over adb still necessary if the hinge is handled via sensors?

dimorinny: Looking at the change where this reset was [introduced](https://googleplex-android-review.git.corp.google.com/c/device/google/cuttlefish/+/13922163/15/host/frontend/webrtc_operator/assets/js/app.js) and another very old [change](https://googleplex-android-review.git.corp.google.com/c/platform/vendor/google/products/+/13841865/7/cuttlefish/foldable/config_foldable_pixel.json#26). We used to have buttons triggering the following shell commands:

```
"shell_command":"dumpsys device_state | grep mCommittedState | grep OPENED && cmd device_state state 0 || cmd device_state state 2",
```

So `device_state` can be override via `shell_command` buttons and to ensure old "override" value isn't affecting the new-coming LID injection , this reset was introduced.

Checking the modern custom actions jsons across the internal tree, I cannot see anything similar and I think we should discourage using `device_state` this way and always go through the proper sensors. 

Taken all of that I would prefer getting rid of this adb reset command (so adb dependency), but want to hear @jemoreira opinion after this context?

base/cvd/cuttlefish/host/frontend/webrtc/sensors_handler.cpp
```diff
@@ -36,61 +37,49 @@ static constexpr sensors::SensorsMask kUiSupportedSensors =
...
             "Can't send request for cmd: " << cmd);
   return {};
 }
 
+Result<void> SensorsHandler::SetMotion(const double x, const double y,
+                                       const double z) {
+  std::stringstream ss;
+  ss << x << sensors::INNER_DELIM << y << sensors::INNER_DELIM << z;
+  CF_EXPECT(SendCommand(sensors::kUpdateRotationVec, ss.str()));
+  UpdateSensorsUi();
```
Databean: Similar to [the discussion on `on_change_callback_`](https://github.com/google/android-cuttlefish/pull/2643#discussion_r3344667420), what's the distinction that requires `SetMotion` calls `UpdateSensorsUi` but not `SetHingeAngle`?

base/cvd/cuttlefish/host/commands/sensors_simulator/sensors_hal_proxy.cpp
Databean: I think "idiom used throughout the wider codebase" still doesn't really satisfy the condition of "makes the code clearer to readers who aren't familiar with the project."

base/cvd/cuttlefish/host/commands/sensors_simulator/sensors_simulator.cpp
Databean: Thanks, this is more intuitive now.

Databean (approved):

base/cvd/cuttlefish/host/commands/sensors_simulator/sensors_hal_proxy.cpp
dimorinny: Thanks for the clarification. Get rid of such `auto` usage here and across the change.

base/cvd/cuttlefish/host/frontend/webrtc/sensors_handler.cpp
dimorinny: Unlike motion sensors, the hinge angle doesn't have a UI representation, so the updated value doesn't need to be propagated to all active clients. I'm planning to do some refactoring in b/519984077 to unify how the hinge angle, motion sensors, and other `device_state` variables are delivered between the WebRTC front-end and back-end. For now, however, I have just renamed the subscription and update methods to make it clear that they are motion-specific. Hope this helps!

Databean (approved):

base/cvd/cuttlefish/host/frontend/webrtc/html_client/js/app.js
jemoreira: > Taken all of that I would prefer getting rid of this adb reset command (so adb dependency)

sounds good to me

PR #2645, opened 2026-06-02
Enforce clang-format and include cleaner in cuttlefish/io

Databean:
Assisted-by: Gemini:Next
Bug: b/512215781

### DIFF
--- a/base/cvd/cuttlefish/io/BUILD.bazel
+++ b/base/cvd/cuttlefish/io/BUILD.bazel
@@ -10,6 +10,7 @@ cf_cc_library(
     name = "chroot",
     srcs = ["chroot.cc"],
     hdrs = ["chroot.h"],
+    clang_format_enabled = True,
     deps = [
         "//cuttlefish/io",
         "//cuttlefish/io:filesystem",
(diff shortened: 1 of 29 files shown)


base/cvd/cuttlefish/io/BUILD.bazel
```diff
@@ -108,10 +112,12 @@ cf_cc_test(
     srcs = ["cpio_test.cc"],
     clang_format_enabled = True,
     deps = [
+        ":io",
```
cjreynol: This is inconsistent with the includes in other parts of the file, and I think with how we have generally aimed to use the full build target label.

base/cvd/cuttlefish/io/cpio.cc
```diff
@@ -19,9 +19,16 @@
 #include <string.h>
 
 #include <charconv>
+#include <cstddef>
```
cjreynol: Previously, when I updated from types like `std::size_t` to `size_t` for consistency, we decided based on Google standards to go with `<stddef.h>` and other similar headers.

cjreynol (approved):

base/cvd/cuttlefish/io/cpio.cc
Databean: Replaced uses of the C++ standard versions of C standard headers with the actual C standard headers.

base/cvd/cuttlefish/io/BUILD.bazel
Databean: Changed to use full paths.

Databean:
Thanks for reviewing!

PR #2647, opened 2026-06-02
Add linting for shell scripts using shellcheck.

dxapd:


base/cvd/debian_build/BUILD.bazel
Databean, line 1: It's unintuitive to have this `BUILD.bazel` file outside any bazel repository. Can this be a `BUILD.bazel` file somewhere in the `base/cvd` tree with the `debian` folder symlink in the same directory? Then the `BUILD.bazel` file can still refer to members inside the `debian` directory.

base/cvd/cuttlefish/bazel/rules.bzl
Databean, line 1: For the internal side of this change, consider `fake_lint.bzl` in [cl/865465414](http://cl/865465414). It's worth making the copybara script update in a pending CL internally and making sure it can import this PR before the PR is submitted. Then you can immediately send the copybara script change for review when this PR merges.

Databean (approved):

base/cvd/cuttlefish/bazel/rules.bzl
dxapd: before i do that, how does it work? is there just a different/split version of rules.bzl internally? if I don't make any changes in advance, what would happen during import? would the resulting build just error out encountering the `cf_sh_binary` in base/debian/BUILD.bazel?

Databean: The `rules.bzl` file is mostly shared between the two, but the strategy used for the `cf_cc_*` targets right now is to define no-op versions of the `format_test` and `clang_tidy_test` rules in a different place, and have copybara rewrite the `load` statements in `rules.bzl`.

As this PR is written it would find the `cf_sh_binary` macro, expand it, and then fail to use the `shellcheck_test` rule. If you want to follow the `cf_cc_*` strategy, this PR doesn't need to be changed, but there need to be extra internal changes.

base/cvd/debian_build/BUILD.bazel
dxapd: Done.

base/cvd/cuttlefish/bazel/rules.bzl
dxapd: The test import looks to have worked (cl/933365548). I'll merge this soon.

PR #2648, opened 2026-06-03
Add netsim_args to cvd load config and plumb to launcher

rmuthiah:
- Update load_config.proto to include netsim_args in EnvironmentSpecification and EnvironmentOptions.
- Update launch_cvd_parser.cpp to generate --netsim_args flag from the proto.
- Regenerate Go proto code.
- Add unit test to flags_parser_test.cc.

Generated by Jetski
Bug: b/447440168


cjreynol (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_config.proto
```diff
@@ -45,6 +45,7 @@ message EnvironmentSpecification {
   optional Common common = 4;
   optional bool netsim_bt = 5;
   optional bool netsim_uwb = 6;
+  optional string netsim_args = 7;
```
Databean: This might make more sense as a `repeated string netsim_args`, so that we don't have to deal with users trying to escape whitespace in this string.

Unfortunately the internal `--netsim_args` argument to `assemble_cvd` already is based on [naive string splitting](https://github.com/google/android-cuttlefish/blob/4f1f0a799168930fa379af610c098d9ea57c276e/base/cvd/cuttlefish/host/libs/config/cuttlefish_config.cpp#L309). This is an implementation detail we can fix later, but we should aspire to have a more sensible public API as it will be harder to change later. To make the behavior more predictable, the `launch_cvd_parser.cpp` implementation should introduce the whitespace to create the `netsim_args` flag, and error if the user-provided value has any whitespace. We can lift that restriction later.

Then again, this probably makes life slightly more difficult for all the callers specifying override flags for `netsim_args`...

Same concern applies for `EnvironmentOptions` below.

base/cvd/cuttlefish/host/commands/cvd/cli/parser/flags_parser_test.cc
```diff
@@ -175,6 +175,33 @@ TEST(BootFlagsParserTest, ParseNetSimFlagEnabled) {
...
+}
+  )"""";
+
+  Json::Value json_configs;
+  std::string json_text(test_string);
+
+  EXPECT_TRUE(ParseJsonString(json_text, json_configs))
+      << "Invalid Json string";
+  auto serialized_data = LaunchCvdParserTester(json_configs);
+  EXPECT_TRUE(serialized_data.ok()) << serialized_data.error().Trace();
```
Databean: nit: this can use `EXPECT_THAT(serialized_data, IsOk())`

https://github.com/google/android-cuttlefish/blob/4f1f0a799168930fa379af610c098d9ea57c276e/base/cvd/cuttlefish/result/result_matchers.h#L27-L35

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_config.proto
Databean: Also came across [go/netsim-args-oxygen](http://go/netsim-args-oxygen) which covers a time we talked about this before. I vaguely remember that either @jemoreira or I made the argument that exposing `netsim_args` opened the possibility of mutually contradictory config settings, e.g.
```json
{
  "netsim_bt": false,
  "netsim_args": "-bt=true"
}
```
(not that any such `-bt=true` flag exists)

jemoreira: It was actually Ram's position, we were just the ones in the meeting because of different time zones.

rmuthiah: I still agree with my original position :(. However, given the pressure to get this migration done, I think we'll have to 1. merge this as is now and then 2. allowlist all existing netsim argsm 3. start whittling down the list of netsim args being injected until the list is empty, and then 4. deprecate this parameter.

base/cvd/cuttlefish/host/commands/cvd/cli/parser/flags_parser_test.cc
rmuthiah: Thanks for the suggestion, refactored to use the `EXPECT_THAT` macro.

base/cvd/cuttlefish/host/commands/cvd/cli/parser/load_config.proto
rmuthiah: Good point, I've updated it to a `repeated string` in both structures.

I'm happy to make life more difficult for the callers, we don't want to support this param for long nor do we want it used extensively.

PR #2649, opened 2026-06-03
Validation check for main instance during restart

SuperStrongDinosaur:
Added a check to determine if the target instance is the first instance in the group.                                                                        
If it is the first instance, we check if there are any other active instances in the group. If so, the restart is blocked with the error message

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/restart.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/restart.cpp
@@ -72,9 +72,25 @@ Result<void> CvdDeviceRestartCommandHandler::Handle(
   std::vector<std::string> subcmd_args = request.SubcommandArguments();
   CF_EXPECT(ConsumeFlags(options.Flags(), subcmd_args));
 
-  auto [instance, unused] =
+  auto [instance, group] =
       CF_EXPECT(selector::SelectInstance(instance_manager_, request),
                 "Unable to select an instance");
+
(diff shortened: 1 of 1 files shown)


base/cvd/cuttlefish/host/commands/cvd/cli/commands/restart.cpp
```diff
@@ -71,10 +71,25 @@ Result<void> CvdDeviceRestartCommandHandler::Handle(
...
+    bool other_active = false;
+    for (const auto& inst : instances) {
+      if (inst.id() != instance.id() && inst.IsActive()) {
+        other_active = true;
+        break;
+      }
+    }
+    CF_EXPECTF(!other_active,
+               "Restarting the first instance (ID: {}) is not allowed "
+               "while other instances are running. Restart the entire group instead.",
```
jemoreira: what happens if the entire group is restarted while some instances are stopped? Are those instances [re]started too?

jemoreira (approved):

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/restart.cpp
SuperStrongDinosaur: The cvd restart command is designed to target only a single instance at a time. If you run it on a multi instance group without specifying an instance, it will prompt you to select one. It cannot be used to restart the entire group at once.

PR #2652, opened 2026-06-03
Omnibus for changes to kvm and cvdnetwork checks.

3405691582:
Let's expand on the concept of a sandboxed environment to something more
general. Call the environment in which a platform runs, whether it be a
sandboxed environment or an environment with restricted networking a
_domain_.

We may want to run in domains where the ability to modify the networking
setup or add new tap devices is restricted, so that we want to run with
--enable_tap_devices being false. If we are running with this flag,
then we don't need the cvdnetwork membership check. However, we don't
have easy access to the enable_tap_devices flag where we check for the
group membership.

Instead, we add a new domain for this case. We can't easily infer the
domain like we can for the sandbox or container cases, so for now,
we'll assert we're in the restricted networking domain by checking for
an environment variable.

Also, some execution domains may have the kvm device already
read-writable without the user explicitly needing to be in the kvm
group. Test that the device is read-writable, and if it isn't, then
require membership in the group.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
@@ -38,6 +38,7 @@
 #include "cuttlefish/common/libs/utils/files.h"
 #include "cuttlefish/common/libs/utils/gflags_xml_parser.h"
 #include "cuttlefish/common/libs/utils/subprocess_managed_stdio.h"
+#include "cuttlefish/common/libs/utils/in_sandbox.h"
 #include "cuttlefish/common/libs/utils/users.h"
 #include "cuttlefish/host/commands/cvd/instances/config_path.h"
 #include "cuttlefish/host/commands/cvd/utils/common.h"
@@ -63,7 +64,12 @@ Result<void> CheckProcessExitedNormally(siginfo_t infop,
 
 static Command FixGroupsIfNecessary(const std::string& command_name,
                                     const std::string& bin_path) {
-  if (InGroup("cvdnetwork") && InGroup("kvm")) {
+#ifndef __linux__
+  return Command(command_name).SetExecutable(bin_path);
+#endif
+  const bool has_net = InSandbox() || InGroup("cvdnetwork");
+  const bool has_kvm = IsKvmAccessible();
+  if (has_net && has_kvm) {
     return Command(command_name).SetExecutable(bin_path);
   }
   const std::string refresh_groups =
--- a/base/cvd/cuttlefish/common/libs/utils/users.cpp
+++ b/base/cvd/cuttlefish/common/libs/utils/users.cpp
@@ -126,4 +126,12 @@ Result<std::string> CurrentUserName() {
   return std::string(pwd.pw_name);
 }
 
+bool IsKvmAccessible() {
+#ifdef __linux__
+  return access("/dev/kvm", R_OK | W_OK) == 0 || InGroup("kvm");
+#else
+  return false;
+#endif
+}
+
 } // namespace cuttlefish
--- a/base/cvd/cuttlefish/common/libs/utils/users.h
+++ b/base/cvd/cuttlefish/common/libs/utils/users.h
@@ -25,6 +25,7 @@ namespace cuttlefish {
 
 gid_t GroupIdFromName(const std::string& group_name);
 bool InGroup(const std::string& group);
+bool IsKvmAccessible();
 
 /**
  * returns the user's home defined by the system
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/BUILD.bazel
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/BUILD.bazel
@@ -186,6 +186,7 @@ cf_cc_library(
         "//cuttlefish/common/libs/utils:contains",
         "//cuttlefish/common/libs/utils:files",
         "//cuttlefish/common/libs/utils:gflags_xml_parser",
+        "//cuttlefish/common/libs/utils:in_sandbox",
         "//cuttlefish/common/libs/utils:subprocess",
         "//cuttlefish/common/libs/utils:subprocess_managed_stdio",
         "//cuttlefish/common/libs/utils:users",
--- a/base/cvd/cuttlefish/host/libs/vm_manager/BUILD.bazel
+++ b/base/cvd/cuttlefish/host/libs/vm_manager/BUILD.bazel
@@ -37,6 +37,7 @@ cf_cc_library(
         "//cuttlefish/common/libs/utils:environment",
         "//cuttlefish/common/libs/utils:files",
         "//cuttlefish/common/libs/utils:host_info",
+        "//cuttlefish/common/libs/utils:in_sandbox",
         "//cuttlefish/common/libs/utils:json",
         "//cuttlefish/common/libs/utils:network",
         "//cuttlefish/common/libs/utils:subprocess",
--- a/base/cvd/cuttlefish/host/libs/vm_manager/host_configuration.cpp
+++ b/base/cvd/cuttlefish/host/libs/vm_manager/host_configuration.cpp
@@ -24,6 +24,7 @@
 #include <vector>
 
 #include "cuttlefish/common/libs/utils/container.h"
+#include "cuttlefish/common/libs/utils/in_sandbox.h"
 #include "cuttlefish/common/libs/utils/users.h"
 
 namespace cuttlefish {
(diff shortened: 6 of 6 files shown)


base/cvd/cuttlefish/common/libs/utils/domain.cpp
```diff
@@ -13,12 +13,26 @@
...
+#include "cuttlefish/common/libs/utils/users.h"
+#include <cstdlib>
 #include <unistd.h>
 
 namespace cuttlefish {
 
 bool InSandbox() { return access("/manager.sock", F_OK) == 0; }
 
+bool IsRestrictedNetworkingDomain() {
+  return std::getenv("CUTTLEFISH_DOMAIN_RESTRICTED_NETWORKING") != nullptr;
```
Databean: nit: [`StringFromEnv`](https://github.com/google/android-cuttlefish/blob/a094c8e51669e57b9669b3fbe7a0f981806183ef/base/cvd/cuttlefish/common/libs/utils/environment.cpp#L25)

base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd.cc
```diff
@@ -406,7 +406,8 @@ Result<const CuttlefishConfig*> InitFilesystemAndCreateConfig(
     CF_EXPECT(CleanPriorFiles(preserving, clean_dirs),
               "Failed to clean prior files");
 
-    std::string default_group = "cvdnetwork";
+    std::string default_group =
+        !IsRestrictedNetworkingDomain() ? "cvdnetwork" : "";
```
Databean: nit: please flip the ternary, preferring `X ? Y : Z` over `!X ? Z : Y`.

base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd.cc
Databean, line 415: The `default_group` addition was originally from [aosp/2605752](https://android-review.googlesource.com/c/device/google/cuttlefish/+/2605752). If I remember correctly, the motivation is that some of the members of [`RequiredDirectories`](https://github.com/google/android-cuttlefish/blob/a094c8e51669e57b9669b3fbe7a0f981806183ef/base/cvd/cuttlefish/host/commands/assemble_cvd/required_directories.cc#L54) are in `/tmp` (bunch of steps leads back to [`EnvironmentUdsDir()`](https://github.com/google/android-cuttlefish/blob/a094c8e51669e57b9669b3fbe7a0f981806183ef/base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc#L328)), and if multiple people launch devices on the same host, it shouldn't matter who wins the race to create the `/tmp/...` directories first. The `cvdnetwork` group is used as a shortcut to ensure that all correctly configured user accounts will have access to the created directory.

In practice I don't think there's any overlap between the "multiple user accounts" case and the "restricted networking domain" case, but semantically they are orthogonal concepts. Of course, there's also semantically not much relation between the `cvdnetwork` group and a `/tmp/...` directory either, besides that the directory may hold unix socket files. The `/tmp` directory is used for unix socket files to avoid issues with running into file path length limits for unix sockets that we've experienced keeping them in the user directory.

Is this be a good time to find a way to satisfy the "multiple user accounts" requirement without relying on the `cvdnetwork` shortcut?

base/cvd/cuttlefish/host/commands/cvd/cli/utils.cpp
```diff
@@ -61,7 +62,13 @@ Result<void> CheckProcessExitedNormally(siginfo_t infop,
...
-  if (InGroup("cvdnetwork") && InGroup("kvm")) {
+#ifndef __linux__
+  return Command(command_name).SetExecutable(bin_path);
+#endif
+  if (IsRestrictedNetworkingDomain() && IsKvmAccessible()) {
+    return Command(command_name).SetExecutable(bin_path);
+  }
+  if (InGroup("cvdnetwork") && IsKvmAccessible()) {
     return Command(command_name).SetExecutable(bin_path);
   }
```
Databean, line 74: nit: can the conditions be separated into some boolean variable declarations to avoid duplicating the `Command().SetExecutable()` expression?

base/cvd/cuttlefish/common/libs/utils/domain.cpp
Databean: Is this an environment variable that is going to be part of our public API? Right now it sort of amounts to "`--enable_tap_devices`, but as a custom environment variable."

"restricted networking domain" is also kind of abstract, in comparison to `IsKvmAccessible()`. If I understand correctly the specific issue is that tap devices aren't available, what do you think of starting with `EnvironmentSupportsTapDevices()` which is dependent on things we know are in conflict:

- Non-`__linux__`
- `BORG_TASK_HANDLE`
- `UNITTEST_ON_FORGE`

Databean (approved):

base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd.cc
3405691582: I thought it would read better with the negation, but OK.

base/cvd/cuttlefish/common/libs/utils/domain.cpp
3405691582: Perhaps we could make it specific -- I can see the arguments both ways on this -- but the abstract nature of the naming was the point: to have a generic way to describe environments where networking wasn't fully available, such as in a sandbox, rootless environments, or firewalled.

For specific values of environment variable, I was envisaging tweaking this on the other end of the copybara -- if you are concerned about public API, would you be more comfortable if this was defined `false` here?

3405691582: The other perspective is we go back to the "in_sandbox" nomenclature and treat all of this as just variations on a sandbox. Multi-user concerns also evaporate for sandbox scenarios, too. Let me know what you prefer.

base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd.cc
3405691582: > Is this be a good time to find a way to satisfy the "multiple user accounts" requirement without relying on the cvdnetwork shortcut?

Probably not. We could take one simple approach to change the default_mode, but that is slightly overly permissive, but maybe that doesn't matter so much. 

I think perhaps increasing per-user isolation in the filesystem is better, but with file path length limits, that's a little more fraught and requires further discussion.

base/cvd/cuttlefish/common/libs/utils/domain.cpp
Databean: If you want to cover the space of possible networking capabilities, I see three scenarios right now:

1. Root-in-linux: preallocated or dynamically allocated TAP devices, bridges, packet routing, etc. These are workstations and current Android CI.
2. Userspace networking: the host has a network connection but cannot create privileged networking devices. It might use slirp or an equivalent to still send some networking traffic. These are non-linux hosts or google production.
3. No networking. This is firewalled hosts as you bring up.

If we're speculating into the future (is being future-proof the goal of abstraction here?) the next thing I would worry about is non-linux privileged network devices, but I think we are pretty far off from that.

base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd.cc
Databean: At least worth documenting this interaction here then?

base/cvd/cuttlefish/common/libs/utils/domain.cpp
Databean: The other thing that's bothering me here a little is the methods clash a bit semantically:

- `InSandbox`: domain-focused
- `IsKvmAccessible()`: capability-focused
- `IsRestrictedNetworkingDomain`: domain-focused

The same issue comes up at the `FixGroupsIfNecessary` callsite, which now includes two domain checks:

- `#ifndef __linux__`
- `IsRestrictedNetworkingDomain()`

and two capability checks:

- `InGroup("cvdnetwork")`
- `IsKvmAccessible()`

base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd.cc
3405691582: Good idea. Done.

base/cvd/cuttlefish/common/libs/utils/domain.cpp
3405691582: OK: I'm inclined to just kick the can down the road on public API changes and instead overload IsSandbox() for this purpose. 

(To be fair: the `ifndef __linux__` check in FixGroupsIfNecessary is to avoid hobbling all command executions on other platforms which may not have kvm .)

Databean: Sorry I think I was being too nitpicky here, either way it's still an improvement over what is in the code now.

Databean (approved):

base/cvd/cuttlefish/common/libs/utils/domain.cpp
3405691582: No, it's fine: we can definitely think about adding some further abstractions here, but I don't want to block the other changes. We'll revisit.

Reverted back to InSandbox(), but all the other logic changes have been retained.

PR #2653, opened 2026-06-03
Refactor emulated_camera to use a shared video_capture_device library

bridadan:
Extract shared code from `emulated_camera_mplane` and `emulated_camera_splane` into a new shared library `video_capture_device`. This removes a large amount of copy-pasted code between the two emulated camera implementations.

Key changes:
- Created `video_capture_device` library containing:
  - `CaptureDeviceFormat` trait defining format-specific constants and methods.
  - Generic `EmulatedCamera` and `EmulatedCameraSession` structs implementing `VirtioMediaDevice` and `VirtioMediaIoctlHandler`.
  - Common `Buffer` and `BufferState` implementations.
  - A default implementation of `configure_planes` in the trait, as it was identical for both devices.
- Refactored `emulated_camera_mplane` and `emulated_camera_splane` to:
  - Depend on the new `video_capture_device` library.
  - Implement `CaptureDeviceFormat` for their respective format structs (`MplaneFormat` and `SplaneFormat`).
  - Use type aliases to map their local `EmulatedCamera` to the generic one.
- Updated Cargo.toml and BUILD.bazel files to reflect new dependencies.

TAG=agy
CONV=b4f9ffcf-c393-4dbc-b8fb-5c38416291c2

Bug: 519646531
Test: build and launch with both device types, ensure v4l2-compliance
      passes in the guest

### DIFF
--- a/base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/main.rs
+++ b/base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/main.rs
@@ -12,62 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::path::PathBuf;
-use std::sync::{Arc, RwLock};
-
 use clap::Parser;
-use log::error;
-use thiserror::Error;
+use std::sync::{Arc, RwLock};
 use vhost_user_backend::VhostUserDaemon;
 use vhu_media::VhuMediaBackend;
+use vhu_media::cli::{CmdLineArgs, Config, Error, init_logging, Result};
 use virtio_media::protocol::VirtioMediaDeviceConfig;
 use vm_memory::{GuestMemoryAtomic, GuestMemoryMmap};
 
 mod device;
 
-#[derive(Debug, Error)]
-pub(crate) enum Error {
-    #[error("Could not create daemon: {0}")]
-    CouldNotCreateDaemon(vhost_user_backend::Error),
-    #[error("Fatal error: {0}")]
-    ServeFailed(vhost_user_backend::Error),
-}
-
-type Result<T> = std::result::Result<T, Error>;
-
-#[derive(Parser, Debug)]
-#[clap(author, version, about, long_about = None)]
-struct CmdLineArgs {
-    /// Location of vhost-user Unix domain socket.
-    #[clap(short, long, value_name = "SOCKET")]
-    socket_path: PathBuf,
-    /// Log verbosity, one of Off, Error, Warning, Info, Debug, Trace.
-    #[clap(short, long, default_value_t = log::LevelFilter::Debug)]
-    verbosity: log::LevelFilter,
-}
-
-#[derive(PartialEq, Debug)]
-struct Config {
-    socket_path: PathBuf,
-}
-
-impl TryFrom<CmdLineArgs> for Config {
-    type Error = Error;
-
-    fn try_from(args: CmdLineArgs) -> Result<Self> {
-        Ok(Config {
-            socket_path: args.socket_path,
-        })
-    }
-}
-
-fn init_logging(verbosity: log::LevelFilter) -> Result<()> {
-    env_logger::builder()
-        .format_timestamp_secs()
-        .filter_level(verbosity)
-        .init();
-    Ok(())
-}
 
 const VFL_TYPE_VIDEO: u32 = 0;
 
--- a/base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/main.rs
+++ b/base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/main.rs
@@ -87,7 +87,12 @@ fn start_backend(config: Config) -> Result<()> {
         };
         let backend = Arc::new(RwLock::new(VhuMediaBackend::new(
             config,
-            |event_queue, host_mapper| crate::device::EmulatedCamera::new(event_queue, host_mapper),
+            |event_queue, host_mapper| {
+                vhu_media::devices::EmulatedCamera::<_, _, crate::device::MplaneFormat>::new(
+                    event_queue,
+                    host_mapper,
+                )
+            },
         )));
         let mut daemon = VhostUserDaemon::new(
             String::from("vhost-user-media-backend"),
--- a/base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_splane/src/main.rs
+++ b/base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_splane/src/main.rs
@@ -78,7 +78,12 @@ fn start_backend(config: Config) -> Result<()> {
         };
         let backend = Arc::new(RwLock::new(VhuMediaBackend::new(
             config,
-            |event_queue, host_mapper| crate::device::EmulatedCamera::new(event_queue, host_mapper),
+            |event_queue, host_mapper| {
+                vhu_media::devices::EmulatedCamera::<_, _, crate::device::SplaneFormat>::new(
+                    event_queue,
+                    host_mapper,
+                )
+            },
         )));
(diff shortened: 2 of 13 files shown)


ser-io (changes requested):
Thanks for this needed refactoring.

Can you split the PR into two commits (parts)

1. Refactor code that's clearly duplicated among the two devices implemention without `CaptureDeviceFormat` abstraction.
2. Introduce the new CaptureDeviceFormat abstraction and the refactored code around it.

I'd like to analyze the new abstraction of a commit of its own. Thanks.

bridadan:
@ser-io Here's an attempt at splitting up the changes. The `impl<Q, HM, Reader, Writer> VirtioMediaDevice<Reader, Writer> for EmulatedCamera<Q, HM>` block has some very minor dependencies on the plane type, so if the PR is still hard to read, I could insert a third commit that only deduplicates that implementation based on the plane type. Let me know if that would be useful!

base/cvd/cuttlefish/host/commands/vhost_user_media/video_capture_device/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+pub fn default_fmtdesc(queue: QueueType, pixelformat: u32) -> v4l2_fmtdesc {
+    v4l2_fmtdesc {
+        index: 0,
+        type_: queue as u32,
+        pixelformat,
+        ..Default::default()
+    }
+}
+
+pub fn enum_framesizes(
```
ser-io: This helper function is not extensible. If I were to add a new frame size to my device implementation, I cannot reuse this function.

Let's keep this responsibility on the device implementation side.

base/cvd/cuttlefish/host/commands/vhost_user_media/video_capture_device/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+            }
+        }
+        self.v4l2_buffer.set_flags(flags);
+        self.state = state;
+    }
+}
+
+
+
+pub fn default_fmtdesc(queue: QueueType, pixelformat: u32) -> v4l2_fmtdesc {
```
ser-io: No need to introduce this new helper as it just return a struct. It's more clear to return the struct directly on the device implementation.

base/cvd/cuttlefish/host/commands/vhost_user_media/video_capture_device/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+            discrete: bindings::v4l2_frmsize_discrete {
+                width,
+                height,
+            },
+        },
+        ..Default::default()
+    })
+}
+
+pub fn enum_frameintervals(
```
ser-io: same as enum_framesizes.

base/cvd/cuttlefish/host/commands/vhost_user_media/video_capture_device/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "video_capture_device"
```
ser-io: This should be a new module `pub mod devices` in [lib.rs](https://github.com/google/android-cuttlefish/blob/main/base/cvd/cuttlefish/host/commands/vhost_user_media/vhu_media/src/lib.rs)

The directory "vhu_media" is the library code containing the shared code to be used by the binaries. Eventually we need to split the "lib.rs" file into different files. Feel free to do this if you want in this PR, however it's not a requirement for this PR.

base/cvd/cuttlefish/host/commands/vhost_user_media/video_capture_device/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+            discrete: bindings::v4l2_fract {
+                numerator: 1,
+                denominator: frame_rate,
+            },
+        },
+        ..Default::default()
+    })
+}
+
+pub fn configure_planes(v4l2_buffer: &mut V4l2Buffer, offset: u32, buffer_size: u32) {
```
ser-io: We can be specific with the function name for readability in this case: `set_plane_offset_and_length`.

ser-io (changes requested):

base/cvd/cuttlefish/host/commands/vhost_user_media/video_capture_device/src/lib.rs
bridadan: I've removed this helper function, but I did keep a shared implementation in `CaptureDeviceFormat` since the code only differs on the constant values of each device type at this time.

If you still would prefer to keep it entirely in the device implementation let me know and I can move it.

base/cvd/cuttlefish/host/commands/vhost_user_media/video_capture_device/src/lib.rs
bridadan: Agreed, removed

base/cvd/cuttlefish/host/commands/vhost_user_media/video_capture_device/Cargo.toml
bridadan: Sounds good!

base/cvd/cuttlefish/host/commands/vhost_user_media/vhu_media/src/devices.rs
```diff
@@ -0,0 +1,93 @@
...
+use v4l2r::ioctl::BufferFlags;
+use v4l2r::ioctl::V4l2Buffer;
+use v4l2r::ioctl::V4l2PlanesWithBackingMut;
+use v4l2r::memory::MemoryType;
+
+use virtio_media::ioctl::IoctlResult;
+use virtio_media::memfd::MemFdBuffer;
+
+
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
```
ser-io, line 30: This `Buffer` implementaion is now specific to the `emulated_camera_splane` device only,  see https://github.com/google/android-cuttlefish/pull/2670. Let's keep under `emulated_camera_splane` until it's relevant to move it to a common place, for example,  if there's another single planar format device that needs it.

ser-io (changes requested):

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_splane/src/main.rs
```diff
@@ -12,54 +12,16 @@
...
 
-use std::path::PathBuf;
-use std::sync::{Arc, RwLock};
-
 use clap::Parser;
+use std::sync::{Arc, RwLock};
 use vhost_user_backend::VhostUserDaemon;
 use vhu_media::VhuMediaBackend;
-use vhu_media::cli::Error;
+use vhu_media::cli::{CmdLineArgs, Config, Error, init_logging, Result};
```
ser-io, line 19: Let's use `cli::Result`, `cli::Error` and `cli::init_logging`  down below and don't import these strct/cuntions directly. `Result` and `Error` are very common terms, let's use the `cli` prefix for a better reading.

ser-io (changes requested):

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/main.rs
```diff
@@ -12,62 +12,16 @@
...
-pub(crate) enum Error {
-    #[error("Could not create daemon: {0}")]
-    CouldNotCreateDaemon(vhost_user_backend::Error),
-    #[error("Fatal error: {0}")]
-    ServeFailed(vhost_user_backend::Error),
-}
-
-type Result<T> = std::result::Result<T, Error>;
-
-#[derive(Parser, Debug)]
```
ser-io, line 38: Keep CmdLineArgs and Config per binary.

ser-io (changes requested):

PR #2655, opened 2026-06-03
Invert clang-format default for cuttlefish packages

Databean:
- Change the default of `clang_format_enabled` to `True` in `rules.bzl`.
- Remove explicit `clang_format_enabled = True` from compliant targets.
- Set `clang_format_enabled = False` for non-compliant targets across the workspace.

Also fix a duplicate header in cvdalloc/BUILD.bazel.

Bug: b/512215781
Assisted-by: gemini-next:current

### DIFF
--- a/base/cvd/allocd/net/BUILD.bazel
+++ b/base/cvd/allocd/net/BUILD.bazel
@@ -14,6 +14,7 @@ cf_cc_library(
         "netlink_client.h",
         "netlink_request.h",
     ],
+    clang_format_enabled = False,
     deps = [
         "//cuttlefish/common/libs/fs",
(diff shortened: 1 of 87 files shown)


Databean:
The original PR applied `enable_clang_format = True` where possible, it has been changed to instead flip the default and apply `enable_clang_format = False` to non-compliant files.

The distribution now is:

- Total cf_cc_* targets: 706
- Explicitly set to False (clang_format_enabled = False): 242 targets (non-compliant targets).
- Explicitly set to True: 0 targets (redundant settings removed).
- Implicitly True (default): 464 targets (compliant targets, using the new default).

jemoreira (approved):

PR #2658, opened 2026-06-04
Remove -Icuttlefish from libteeui localization test

Databean:
Assisted-by: Gemini:Next


dxapd:
Seems good, I'm assuming this header is just no longer used there? (or never was)

dxapd (approved):

Databean:
I think this was cargo-culted from a time where it was necessary, before we updated the C++ include paths.

PR #2665, opened 2026-06-04
Use C++20 default comparison operators across Cuttlefish host code

Databean:
Replaces manually defined relational and equality comparison overloads in SharedFD, AndroidBuildKey, DeviceBuildString, DirectoryBuildString, ClientId, and Range with C++20 defaulted comparison operators (= default) and the three-way comparison operator (operator<=>). It also updates the underlying allocd build rules to use `-std=c++20`.

Assisted-by: Gemini:Next
Bug: b/520128494

### DIFF
--- a/base/cvd/allocd/BUILD.bazel
+++ b/base/cvd/allocd/BUILD.bazel
@@ -11,17 +11,17 @@ string_flag(
 )
 
 config_setting(
-  name = "netlink",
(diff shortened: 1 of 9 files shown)


3405691582 (approved):
Fun!

Databean:
Thanks for reviewing!

PR #2666, opened 2026-06-04
Enable clang-format enforcement for targets not touched in 2026

Databean:
Format source files and enable clang-format enforcement across all build targets where none of the files have been modified since the start of 2026.

Using a reduced set of targets makes it less likely to introduce merge conflicts. Some comments are manually preserved with formatting exclusions (see the first commit).

Out of 768 total targets under //cuttlefish/... (243 unformatted), 55 targets were formatted and enabled in this commit, and 188 targets remain unformatted.

Bug: b/512215781
Assisted-by: gemini-next:current


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #2667, opened 2026-06-05
Request higher SafeLevel builds when requesting Android build by branch

Databean:
In cvd fetch, when requesting a build by branch and target, prefer requesting higher SafeLevel builds (PLATINUM, SAFE_2, SAFE_1) when possible before falling back to any complete successful build (SAFE_0 / unspecified).

If users want a specific build id, that is already possible to request through command line arguments. When a user requests a branch by name, they probably do not care as much how recent the build is. In the case of our presubmit, this will hopefully reduce flakes that come from the device side.

The main failure mode here is if a build target / branch combination has previously reported a higher safe level and no longer does, it will get stuck on the older build from the higher safe level indefinitely.

Bug: b/518049553

### DIFF
--- a/base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
+++ b/base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
@@ -275,25 +275,33 @@ Result<std::vector<std::string>> AndroidBuildApi::Headers() {
 
 Result<std::optional<std::string>> AndroidBuildApi::LatestBuildId(
     const std::string& branch, const std::string& target) {
-  const std::string url =
-      android_build_url_.GetLatestBuildIdUrl(branch, target);
-  auto response =
-      CF_EXPECT(HttpGetToJson(http_client_, url, CF_EXPECT(Headers())));
+  for (const SafeLevel safe_level : kAllSafeLevels) {
+    VLOG(0) << "Attempting to download build at safe level '" << safe_level
+            << "' for branch '" << branch << "' and target '" << target << "'";
+    const std::string url =
+        android_build_url_.GetLatestBuildIdUrl(branch, target, safe_level);
+    const std::vector<std::string> headers = CF_EXPECT(Headers());
+    const HttpResponse<Json::Value> response =
+        CF_EXPECT(HttpGetToJson(http_client_, url, headers));
 
-  const Json::Value json = CF_EXPECTF(GetResponseJson(response),
-                                      "Error fetching last known good build "
-                                      "id for:\nbranch \"{}\", target \"{}\"",
-                                      branch, target);
-  if (!json.isMember("builds")) {
-    return std::nullopt;
+    Result<Json::Value> json_res = GetResponseJson(response);
+    if (!json_res.ok()) {
+      continue;
+    }
+    const Json::Value json = *json_res;
+
+    if (!json.isMember("builds")) {
+      continue;
+    }
+
+    CF_EXPECTF(json["builds"].isArray() && json["builds"].size() == 1,
+               "Expected to find a single latest build for branch \"{}\" and "
+               "target \"{}\" in the response array, "
+               "but found {}",
+               branch, target, json["builds"].size());
+    return CF_EXPECT(GetValue<std::string>(json["builds"][0], {"buildId"}));
   }
-
-  CF_EXPECTF(json["builds"].isArray() && json["builds"].size() == 1,
-             "Expected to find a single latest build for branch \"{}\" and "
-             "target \"{}\" in the response array, "
-             "but found {}",
-             branch, target, json["builds"].size());
-  return CF_EXPECT(GetValue<std::string>(json["builds"][0], { "buildId" }));
+  return std::nullopt;
 }
 
 Result<std::unordered_set<std::string>> AndroidBuildApi::Artifacts(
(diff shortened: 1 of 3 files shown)


base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
```diff
@@ -276,25 +276,25 @@ Result<std::vector<std::string>> AndroidBuildApi::Headers() {
...
-      CF_EXPECT(HttpGetToJson(http_client_, url, CF_EXPECT(Headers())));
-
-  const Json::Value json = CF_EXPECTF(GetResponseJson(response),
-                                      "Error fetching last known good build "
-                                      "id for:\nbranch \"{}\", target \"{}\"",
-                                      branch, target);
-  if (!json.isMember("builds")) {
-    return std::nullopt;
+  for (const std::string_view safe_level :
+       {"PLATINUM", "SAFE_2", "SAFE_1", ""}) {
```
cjreynol: Can we store this list somewhere in a constant and leave a comment about where it is sourced from?  That way we have a reference to check against if/when it changes in the future.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
```diff
@@ -276,25 +276,25 @@ Result<std::vector<std::string>> AndroidBuildApi::Headers() {
...
+    const std::string url =
+        android_build_url_->GetLatestBuildIdUrl(branch, target, safe_level);
+    std::vector<std::string> headers = CF_EXPECT(Headers());
```
cjreynol: NIT: the headers will be constant, we could declare this `const` and outside of the loop

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
```diff
@@ -276,25 +276,25 @@ Result<std::vector<std::string>> AndroidBuildApi::Headers() {
...
+    HttpResponse<Json::Value> response =
+        CF_EXPECT(HttpGetToJson(http_client_, url, headers));
+
+    const Json::Value json = CF_EXPECTF(
+        GetResponseJson(response),
+        "Error fetching last known good build id for:\nbranch \"{}\", target "
```
cjreynol: Are there not failure responses when we request a higher safe level than is currently (or ever, for some targets) available?

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
```diff
@@ -276,25 +276,25 @@ Result<std::vector<std::string>> AndroidBuildApi::Headers() {
...
+        "\"{}\", safe_level \"{}\"",
+        branch, target, safe_level);
+    if (json.isMember("builds") && json["builds"].isArray() &&
```
cjreynol: This is not the same as the previous checks.  The separate `builds` call was for the unauthenticated cases, so we could return the `std::nullopt` and get the correct behavior (the special error message) in `GetBuildInfo`.

Also, `builds` not being an array and of size one should still be its own error case that fails due to malformed/unexpected data from the API(unless this safety level changes the return value in a way I would not expect when adding a single query parameter).

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
cjreynol: I wonder if we should add some fetch tests to presubmits to verify the behavior of different targets with different maximum safety levels.  

On the bug it looks like you already did some digging, could you at least add some test cases to the PR description so I can run some tests manually to see if there is anything we are missing when updating this logic?

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
cjreynol: Nevermind, those docs are probably internal.  Which is why we do not have a reference for the terminal statuses, either.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
Databean: They are not constant, [`Headers()`](https://github.com/google/android-cuttlefish/blob/5cb8b388abd4ce57a53eed629d3912daa0696c60/base/cvd/cuttlefish/host/libs/web/android_build_api.cpp#L270) invokes `credential_source_->Credential()`. Some of the `CredentialSource` implementations are caching, but that is an implementation detail.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
Databean: I posted some test cases in [b/518049553#comment3](http://b/518049553#comment3).

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
Databean: I've updated the logic. I see a `{}` response when there are no matching builds, so now it first checks that case, and then asserts that non-empty responses are structured correctly.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
Databean: Created an `enum class SafeLevel` type.

Databean (commented):
Thanks for reviewing!

cjreynol (approved):

Databean:
Thanks for reviewing!

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
Databean: The new checks failed on requesting builds by specific IDs, rather than branches. Reverting to something more like the older checks.

PR #2670, opened 2026-06-05
Support multi-planar properly with YUV420M format

chihchiachen:
To support V4L2 multi-planar API:

- Allocate separate physical `MemFdBuffer` memory descriptors for each color plane of a multi-planar buffer, so that VMM (e.g. crosvm) can map each plane independently.

- Register each plane's physical memory map offset independently.

- Change the advertised pixel format from `YU12` (`YUV420`) to `YM12` (`YUV420M`).

BUG: 520122895
TEST: `v4l2-compliance -d1 -s` passes 100% successfully on emulated camera

### DIFF
--- a/base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
+++ b/base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
@@ -64,27 +64,30 @@ enum BufferState {
     },
 }
 
+/// Information about a single plane of a multi-planar buffer.
+struct Plane {
+    /// Backing memory file descriptor.
+    fd: MemFdBuffer,
+    /// Offset that can be used to map the plane's memory.
+    offset: u32,
+}
+
 /// Information about a single buffer.
 struct Buffer {
     /// Current state of the buffer.
     state: BufferState,
     /// V4L2 representation of this buffer to be sent to the guest when requested.
     v4l2_buffer: V4l2Buffer,
-    /// Backing storage for the buffer.
-    fd: MemFdBuffer,
-    /// Offset that can be used to map the buffer.
-    ///
-    /// Cached from `v4l2_buffer` to avoid doing a match.
-    offset: u32,
+    /// Backing storage and offsets for the planes of the buffer.
+    planes: [Plane; 3],
 }
 
 impl Buffer {
-    fn new(v4l2_buffer: V4l2Buffer, fd: MemFdBuffer, offset: u32) -> Self {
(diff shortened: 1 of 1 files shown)


base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -70,21 +70,19 @@ struct Buffer {
     state: BufferState,
     /// V4L2 representation of this buffer to be sent to the guest when requested.
     v4l2_buffer: V4l2Buffer,
-    /// Backing storage for the buffer.
-    fd: MemFdBuffer,
-    /// Offset that can be used to map the buffer.
-    ///
-    /// Cached from `v4l2_buffer` to avoid doing a match.
-    offset: u32,
+    /// Backing storage for the planes of the buffer.
```
ser-io: What do you think of creating a new `Plane` struct  and have single array `planes: [Plane; 3]`. Rather than keeping two different arrays `fds` and `plane_offsets` in sync?

```
struct Plane {
  fd: MemFdBuffer,
  offset: u32,
}
```

ser-io (changes requested):

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
chihchiachen: I like your idea - it will make things a bit cleaner. Please let me know if you have any other suggestions so that I can address them all at once.

ser-io: That's it, rest looks good to me.

chihchiachen: Updated, thanks!

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -64,27 +64,27 @@ enum BufferState {
     },
 }
 
+struct Plane {
```
ser-io, line 68: nit: add documentation.

ser-io (approved):

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
chihchiachen: Done, thanks.

PR #2673, opened 2026-06-06
Set up `AndroidBuildApi` for testing

Databean:
This PR attempts to move `AndroidBuildApi` in the direction of "can be tested by mocking its dependencies."

- `catchall_api_key` is removed as a constructor argument, since it is only ever given one value.
- `AndroidBuildApi::AndroidBuildApi` accepts `AndroidBuildUrl` by reference, since it is mandatory
- Non-reference `AndroidBuildApi::AndroidBuildApi` arguments are given sensible defaults
- `FakeHttpClient` gets some more utilities
- `AndroidBuildApi::FileReader` gets a test

Outstanding issues that are inconvenient for tests:

- `DownloadFile*` interacts with the filesystem
- `GetBuild` is kind of complicated with multiple API roundtrips


cjreynol (approved):

Databean:
Thanks for reviewing!

PR #2675, opened 2026-06-08
Migrate rotation buttons from injector to sensors HAL

dimorinny:
Since CF migrated to proper sensors HAL, injector doesn't work anymore. Motion-based rotation based functionality:

https://github.com/user-attachments/assets/e1a0fad6-2135-43b5-9cae-9fa33cab30ae

### DIFF
--- a/base/cvd/cuttlefish/host/frontend/webrtc/html_client/js/app.js
+++ b/base/cvd/cuttlefish/host/frontend/webrtc/html_client/js/app.js
@@ -509,7 +509,7 @@ class DeviceControlApp {
     var acc_update = sensor_vals[0].split(":").map((val) => parseFloat(val).toFixed(3));
     var gyro_update = sensor_vals[1].split(":").map((val) => parseFloat(val).toFixed(3));
     var mgn_update = sensor_vals[2].split(":").map((val) => parseFloat(val).toFixed(3));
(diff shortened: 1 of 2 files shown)


base/cvd/cuttlefish/host/frontend/webrtc/html_client/js/app.js
```diff
@@ -541,37 +541,19 @@ class DeviceControlApp {
     deviceConnection.sendSensorsMessage(`${xyz[0]} ${xyz[1]} ${xyz[2]}`);
   }
 
-  // Gradually rotate to a fixed orientation.
```
jemoreira, line 544: why drop the gradual rotation animation?

dimorinny: I don't see much value in this logic. It produces significantly more events on the WebRTC channel, which aren't particularly useful given that the rate from the sensors simulator to the sensors HAL is only 1 Hz. It also introduces a minor delay on the rotation buttons. Everything feels simpler without it, so I've decided to get rid of it.

jemoreira (approved):

base/cvd/cuttlefish/host/frontend/webrtc/html_client/js/app.js
jemoreira: SG

PR #2679, opened 2026-06-08
Use the correct disk type on ARM GCE machine types.

3405691582:
Newer GCE ARM machine types require hyperdisk, but older GCE ARM machine types must use pd. To fix this, we first create the instance, then the disk, as opposed to creating the disk first. Then, we can use and not ignore the returned instance data to inform what the correct disk type we should use when we create it.

### DIFF
--- a/tools/baseimage/pkg/gce/gce.go
+++ b/tools/baseimage/pkg/gce/gce.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"log"
 	"os/exec"
+	"path"
 	"strings"
 	"time"
 
@@ -260,8 +261,24 @@ func (h *GceHelper) BuildImage(project, zone string, opts BuildImageOpts) error 
 	insName := opts.ImageName
 	attachedDiskName := fmt.Sprintf("%s-attached-disk", insName)
 
+	log.Println("creating instance...")
+	inst, err := h.CreateInstance(insName, opts.Arch)
+	if err != nil {
+		return fmt.Errorf("failed to create instance: %w", err)
+	}
+	defer h.cleanupDeleteInstance(insName)
+	log.Printf("instance created: %q", insName)
+
 	if opts.Arch == ArchArm {
-		opts.CreateAttachedDiskOpts.Type = fmt.Sprintf("zones/%s/diskTypes/hyperdisk-balanced", zone)
+		machineType := path.Base(inst.MachineType)
+		var diskType string
+		switch machineType {
+		case "t2a-standard-16":
+			diskType = "pd-balanced"
+		default:
+			diskType = "hyperdisk-balanced"
+		}
+		opts.CreateAttachedDiskOpts.Type = fmt.Sprintf("zones/%s/diskTypes/%s", zone, diskType)
 	}
 
 	log.Println("creating disk...")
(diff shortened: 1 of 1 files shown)


tools/baseimage/pkg/gce/gce.go
```diff
@@ -260,8 +261,24 @@ func (h *GceHelper) BuildImage(project, zone string, opts BuildImageOpts) error
...
+	}
+	defer h.cleanupDeleteInstance(insName)
+	log.Printf("instance created: %q", insName)
+
 	if opts.Arch == ArchArm {
-		opts.CreateAttachedDiskOpts.Type = fmt.Sprintf("zones/%s/diskTypes/hyperdisk-balanced", zone)
+		machineType := path.Base(inst.MachineType)
+		var diskType string
+		switch machineType {
+		case "t2a-standard-16":
```
ser-io, line 276: can we create the instance with machine type other than "t2a-standard-16" and avoid this `switch` logic?

ser-io (changes requested):

tools/baseimage/pkg/gce/gce.go
3405691582: We have the older t2a, and the rest are the newer c4a, n4a, or a4x types. Both the older machine types and newer machine types can stock out, presumably because ARM machines are less common than amd64-based machine types. If we want to support both machine types, then we need this logic.

ser-io: I prefer to keep things simple. There's no need to support both machine types.assuming users won't be blocked if we don't support t2a. I'll leave it up to your judgement.

ser-io (approved):

tools/baseimage/pkg/gce/gce.go
3405691582: If Cloud is stocking out of one machine type, one way of mitigating this error is to try a different machine type. We already made that change in a previous pull request. To properly support the different machine type, we need to support the newer disk type.

The alternative is to try different zones or regions when Cloud is stocked out, but that is also additional logic we'd need to write. Handling Cloud stockout conditions is going to involve specific logic no matter what we do.

ser-io: We don't have automated tests for these scripts, nor ways to recreate the stock out scenarios. The script might as well fail due a different reason when cloud is stocking out cause we don't have ways to verify this behavior on our end when changing this scripts. We only do manual verification for these scripts and we are only testing the rosie path.  I'm in favor of having a more robust solution as long we can back it up with tests.

On that regard, did you you test the script using both type of disks? Wondering if they are mounted in different paths: see https://github.com/google/android-cuttlefish/blob/main/tools/baseimage/pkg/gce/scripts/mount_attached_disk.sh#L23?

3405691582: No, but it's really annoying to have to invoke this script and it failing because of a stockout, which I have experienced trying to deal with this first-hand. Note that the mount point won't change, it's just the disk type at the Cloud layer. 

The testing for this wouldn't be too difficult if we mock out or use DI for the GCE interactions.

ser-io: I understand, I haven't worked with ARM resources, and from your experience stock out is very common, hence there's a chance both paths are tested frequently when running the script manually. Just be aware that if one path starts to become less common when trying the script manually, it would become a liability to the code base.

PR #2682, opened 2026-06-08
Prettier printing of flag description in cvd help commands

jemoreira:
Removes the aliases and alias modes so that the  Flag class knows more about the details of how the command line arguments are matched and can produce a more concise help output.

Bug: b/515490829

### DIFF
--- a/base/cvd/cuttlefish/flag_parser/flag.cpp
+++ b/base/cvd/cuttlefish/flag_parser/flag.cpp
@@ -121,22 +121,6 @@ const std::string& Flag::Description() const { return help_; }
 
 std::string Flag::CurrentValue() const { return getter_(); }
 
-Result<void> Flag::Parse(std::vector<std::string>& arguments) const {
-  for (auto it = arguments.begin(); it != arguments.end();) {
-    size_t consumed = CF_EXPECT(Match(*it, std::span(it + 1, arguments.end())));
-    if (consumed == 0) {
-      ++it;
-    } else {
-      it = arguments.erase(it, it + consumed);
-    }
-  }
-  return {};
-}
-Result<void> Flag::Parse(std::vector<std::string>&& arguments) const {
-  CF_EXPECT(Parse(arguments));
-  return {};
-}
-
 Flag::Flag(std::string name, Flag::Style style) : style_(style) {
   ValidateAlias(name);
   aliases_.emplace_back(std::move(name));
--- a/base/cvd/cuttlefish/flag_parser/flag.cpp
+++ b/base/cvd/cuttlefish/flag_parser/flag.cpp
@@ -19,142 +19,77 @@
 #include <algorithm>
 #include <functional>
 #include <iostream>
-#include <iterator>
-#include <optional>
 #include <ostream>
 #include <string>
+#include <unordered_set>
 #include <utility>
 #include <vector>
 
+#include "absl/cleanup/cleanup.h"
 #include "absl/log/check.h"
-#include "absl/log/log.h"
-#include "absl/strings/match.h"
+#include "absl/strings/str_join.h"
 #include "absl/strings/str_replace.h"
+#include "absl/strings/strip.h"
 
+#include "cuttlefish/common/libs/utils/contains.h"
 #include "cuttlefish/result/result.h"
 
 namespace cuttlefish {
 
 namespace {
 
-struct Separated {
-  std::vector<std::string> args_before_mark;
-  std::vector<std::string> args_after_mark;
-};
-Separated SeparateByEndOfOptionMark(std::vector<std::string> args) {
-  std::vector<std::string> args_before_mark;
-  std::vector<std::string> args_after_mark;
-
-  auto itr = std::find(args.begin(), args.end(), "--");
-  bool has_mark = (itr != args.end());
-  if (!has_mark) {
-    args_before_mark = std::move(args);
-  } else {
-    args_before_mark.insert(args_before_mark.end(), args.begin(), itr);
-    args_after_mark.insert(args_after_mark.end(), itr + 1, args.end());
+// Removes "--" or "-" from the beginning. Returns true if it removed something.
+bool ConsumeDashes(std::string_view& flag_arg) {
+  if (!absl::ConsumePrefix(&flag_arg, "-")) {
+    return false;
   }
-
-  return Separated{
-      .args_before_mark = std::move(args_before_mark),
-      .args_after_mark = std::move(args_after_mark),
-  };
+  absl::ConsumePrefix(&flag_arg, "-");
+  return true;
 }
 
-Result<void> ConsumeFlagsImpl(const std::vector<Flag>& flags,
-                                     std::vector<std::string>& args) {
-  for (const auto& flag : flags) {
-    CF_EXPECT(flag.Parse(args));
-  }
-  return {};
-}
-
-Result<void> ConsumeFlagsImpl(const std::vector<Flag>& flags,
-                                     std::vector<std::string>&& args) {
-  for (const auto& flag : flags) {
-    CF_EXPECT(flag.Parse(args));
-  }
-  return {};
+bool MatchNormalized(std::string_view str1, std::string_view str2) {
+  const std::string n1 = absl::StrReplaceAll(str1, {{"-", "_"}});
+  const std::string n2 = absl::StrReplaceAll(str2, {{"-", "_"}});
+  return n1 == n2;
 }
 
 }  // namespace
 
-std::ostream& operator<<(std::ostream& out, const FlagAlias& alias) {
-  switch (alias.mode) {
-    case FlagAliasMode::kFlagExact:
-      return out << alias.name;
-    case FlagAliasMode::kFlagPrefix:
-      return out << alias.name << "*";
-    case FlagAliasMode::kFlagConsumesFollowing:
-      return out << alias.name << " *";
-    default:
-      LOG(FATAL) << "Unexpected flag alias mode " << (int)alias.mode;
-  }
-  return out;
+Flag Flag::StringFlag(std::string name) {
+  return Flag(std::move(name), Style::String);
 }
 
-Flag& Flag::UnvalidatedAlias(const FlagAlias& alias) & {
-  aliases_.push_back(alias);
-  return *this;
-}
-Flag Flag::UnvalidatedAlias(const FlagAlias& alias) && {
-  aliases_.push_back(alias);
-  return *this;
+Flag Flag::BoolFlag(std::string name) {
+  return Flag(std::move(name), Style::Bool);
 }
 
-void Flag::ValidateAlias(const FlagAlias& alias) {
-  using absl::StartsWith;
-
-  CHECK(StartsWith(alias.name, "-")) << "Flags should start with \"-\"";
-  if (alias.mode == FlagAliasMode::kFlagPrefix) {
-    CHECK(absl::EndsWith(alias.name, "=")) << "Prefix flags must end with '='";
-  }
-
-  CHECK(!HasAlias(alias)) << "Duplicate flag alias: " << alias.name;
-  if (alias.mode == FlagAliasMode::kFlagConsumesFollowing) {
-    CHECK(!HasAlias({FlagAliasMode::kFlagExact, alias.name}))
-        << "Overlapping flag aliases for " << alias.name;
-  } else if (alias.mode == FlagAliasMode::kFlagExact) {
-    CHECK(!HasAlias({FlagAliasMode::kFlagConsumesFollowing, alias.name}))
-        << "Overlapping flag aliases for " << alias.name;
-  }
-}
-
-Flag& Flag::Alias(const FlagAlias& alias) & {
+Flag& Flag::Alias(std::string alias) & {
   ValidateAlias(alias);
-  aliases_.push_back(alias);
+  aliases_.emplace_back(std::move(alias));
   return *this;
 }
-Flag Flag::Alias(const FlagAlias& alias) && {
-  ValidateAlias(alias);
-  aliases_.push_back(alias);
+Flag Flag::Alias(std::string alias) && { return std::move(Alias(alias)); }
+
+Flag& Flag::Help(std::string help) & {
+  help_ = std::move(help);
   return *this;
+}
+Flag Flag::Help(std::string help) && { return std::move(Help(help)); }
+
+Flag& Flag::Getter(std::function<std::string()> getter) & {
+  getter_ = std::move(getter);
+  return *this;
+}
+Flag Flag::Getter(std::function<std::string()> getter) && {
+  return std::move(Getter(getter));
 }
 
-Flag& Flag::Help(const std::string& help) & {
-  help_ = help;
-  return *this;
-}
-Flag Flag::Help(const std::string& help) && {
-  help_ = help;
-  return *this;
-}
-
-Flag& Flag::Getter(std::function<std::string()> fn) & {
-  getter_ = std::move(fn);
-  return *this;
-}
-Flag Flag::Getter(std::function<std::string()> fn) && {
-  getter_ = std::move(fn);
-  return *this;
-}
-
-Flag& Flag::Setter(std::function<Result<void>(const FlagMatch&)> setter) & {
+Flag& Flag::Setter(std::function<Result<void>(std::string_view)> setter) & {
   setter_ = std::move(setter);
   return *this;
 }
-Flag Flag::Setter(std::function<Result<void>(const FlagMatch&)> setter) && {
-  setter_ = std::move(setter);
-  return *this;
+Flag Flag::Setter(std::function<Result<void>(std::string_view)> setter) && {
+  return std::move(Setter(setter));
 }
 
 Flag& Flag::AddValidator(std::function<Result<void>()> validator) & {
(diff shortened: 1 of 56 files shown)


base/cvd/cuttlefish/flag_parser/flag.cpp
```diff
@@ -19,142 +19,95 @@
...
-
-Result<void> ConsumeFlagsImpl(const std::vector<Flag>& flags,
-                                     std::vector<std::string>&& args) {
-  for (const auto& flag : flags) {
-    CF_EXPECT(flag.Parse(args));
-  }
-  return {};
+bool MatchNormalized(std::string_view str1, std::string_view str2) {
+  std::string n1 = absl::StrReplaceAll(str1, {{"-", "_"}});
+  std::string n2 = absl::StrReplaceAll(str2, {{"-", "_"}});
```
Databean: nit^2: `const std::string`

base/cvd/cuttlefish/flag_parser/flag.cpp
```diff
@@ -19,142 +19,95 @@
...
+Flag& Flag::Help(std::string help) & {
+  help_ = std::move(help);
   return *this;
 }
-Flag Flag::Help(const std::string& help) && {
-  help_ = help;
-  return *this;
+Flag Flag::Help(std::string help) && {
+  return std::move(this->Help(std::move(help)));
+  help_ = std::move(help);
```
Databean: nit: dead line after the `return`.

Should the other `&&` functions use the same kind of `return this->...` style?

base/cvd/cuttlefish/flag_parser/flag.cpp
```diff
@@ -19,142 +19,95 @@
...
-    CHECK(!HasAlias({FlagAliasMode::kFlagExact, alias.name}))
-        << "Overlapping flag aliases for " << alias.name;
-  } else if (alias.mode == FlagAliasMode::kFlagExact) {
-    CHECK(!HasAlias({FlagAliasMode::kFlagConsumesFollowing, alias.name}))
-        << "Overlapping flag aliases for " << alias.name;
-  }
+Flag Flag::Alias(std::string alias) && {
+  ValidateAlias(alias);
+  long_aliases_.emplace_back(std::move(alias));
+  return *this;
```
Databean: Some of the other `&&` functions are changed to use `return std::move(*this);`, can that be done consistently throughout?

base/cvd/cuttlefish/flag_parser/flag.cpp
```diff
@@ -163,197 +116,230 @@ Flag& Flag::AddValidator(std::function<Result<void>()> validator) & {
...
+  long_aliases_.emplace_back(std::move(name));
+}
+
+void Flag::ValidateAlias(const std::string& alias) {
+  CHECK(!alias.starts_with("-")) << "Flag aliases should not start with \"-\"";
+  CHECK(!Contains(long_aliases_, alias)) << "Duplicate flag alias: " << alias;
+}
+void Flag::ValidateAlias(char alias) {
+  CHECK(!Contains(short_aliases_, alias)) << "Duplicate flag alias: " << alias;
+}
```
Databean, line 132: It looks possible to insert a duplicate alias here like `.Alias("c").Alias('c')` as long and short aliases are stored separately.

base/cvd/cuttlefish/flag_parser/flag.cpp
```diff
@@ -163,197 +116,230 @@ Flag& Flag::AddValidator(std::function<Result<void>()> validator) & {
...
+      CF_EXPECT(SetAndValidate("no"));
+      return 1;
+    }
+  }
+  for (char alias : short_aliases_) {
+    if (current_arg == std::string({alias})) {
+      CF_EXPECT(SetAndValidate("yes"));
+      return 1;
+    }
+  }
```
Databean: I'm not sure I understand the distinction between long and short aliases. It looks to me like short aliases are defined as being a single character, but otherwise work the same as long aliases: one or two dashes, alias name, then the flag suffix (following argument, equals, etc.)

It seems to me like this could just as easily be represented as single-character `std::string` long aliases.

"Short aliases" also brings to mind the distinction in [`getopt_long`](https://man7.org/linux/man-pages/man3/getopt_long.3.html), the "standard" GNU argument parser. In this implementation, "short" boolean options can be passed in groups. For example, for three boolean arguments "a", "b", "c" `-abc` is equivalent to `-a -b -c`. In `getopt*`, "short" options can only be passed with a single dash, and long options can only be passed with two dashes.

base/cvd/cuttlefish/host/libs/config/touchpad.cpp
```diff
@@ -68,10 +67,10 @@ ParseTouchpadConfigsFromArgs(std::vector<std::string>& args) {
...
   const std::vector<Flag> touchpad_flags = {
-      GflagsCompatFlag(kTouchpadFlag)
+    Flag::StringFlag(kTouchpadFlag)
           .Help(kTouchpadHelp)
-          .Setter([&](const FlagMatch& match) -> Result<void> {
-            repeated_touchpad_flag_values.push_back(match.value);
+          .Setter([&](std::string_view arg) -> Result<void> {
+            repeated_touchpad_flag_values.emplace_back(arg);
             return {};
           }),
```
Databean, line 75: Maybe we need a common utility for `std::vector<std::string>`s built from repeated flags.

base/cvd/cuttlefish/host/commands/tombstone_receiver/main.cpp
```diff
@@ -69,10 +69,10 @@ int TombstoneReceiverMain(int argc, char** argv) {
           .Help("File descriptor to an already created vsock server"));
 
   flags.emplace_back(HelpFlag(flags));
-  flags.emplace_back(UnexpectedArgumentGuard());
 
   std::vector<std::string> args(argv + 1, argv + argc);  // Skip argv[0]
-  auto parse_res = ConsumeFlags(flags, args);
+  auto parse_res =
+      ConsumeFlags(flags, args, {.fail_on_unexpected_argument = true});
```
Databean: nit: `Result<void>`

base/cvd/cuttlefish/host/commands/stop/main.cc
```diff
@@ -176,59 +174,30 @@ FlagVaules GetFlagValues(int argc, char** argv) {
...
-      GflagsCompatFlag("instance_nums", instance_nums)
-          .Help("Comma-separated list of instance numbers to stop."));
   flags.emplace_back(HelpFlag(flags));
   bool helpxml = false;
   flags.emplace_back(HelpXmlFlag(flags, std::cout, helpxml));
-  flags.emplace_back(UnexpectedArgumentGuard());
   std::vector<std::string> args(argv + 1, argv + argc);  // Skip argv[0]
-  auto parse_res = ConsumeFlags(flags, args);
+  auto parse_res =
+      ConsumeFlags(flags, args, {.fail_on_unexpected_argument = true});
```
Databean: nit: `Result<void>`

base/cvd/cuttlefish/host/commands/snapshot_util_cvd/parse.cc
```diff
@@ -126,8 +126,8 @@ Result<Parsed> Parse(std::vector<std::string>& args) {
                       .Help("Suspend/resume before/after taking the snapshot"));
   flags.push_back(HelpFlag(flags));
   flags.push_back(HelpXmlFlag(flags, std::cout, help_xml));
-  flags.push_back(UnexpectedArgumentGuard());
-  auto parse_res = ConsumeFlags(flags, args);
+  auto parse_res =
+      ConsumeFlags(flags, args, {.fail_on_unexpected_argument = true});
```
Databean: nit: `Result<void>`

base/cvd/cuttlefish/flag_parser/shared_fd_flag.cpp
```diff
@@ -17,31 +17,27 @@
...
+  return Flag::StringFlag(name).Setter(
+      [&out, name](std::string_view arg) -> Result<void> {
+        int raw_fd;
+        CF_EXPECTF(absl::SimpleAtoi(arg, &raw_fd),
+                   "Failed to parse value \"{}\" for fd flag \"--{}\"", arg,
+                   name);
+        out = SharedFD::Dup(raw_fd);
+        if (out->IsOpen()) {
+          close(raw_fd);
+        }
```
Databean: Not that the implementation before was correct, but should this report an error if `SharedFD::Dup` fails?

base/cvd/cuttlefish/flag_parser/flag.cpp
```diff
@@ -163,197 +116,230 @@ Flag& Flag::AddValidator(std::function<Result<void>()> validator) & {
...
-  if (!setter_ && !aliases_.empty()) {
-    return CF_ERRF("No setter for flag with alias {}", aliases_[0].name);
+std::string Flag::Name() const { return long_aliases_.front(); }
+
+std::string Flag::Synopsis() const {
+  std::vector<std::string> options;
+  for (const std::string& alias : long_aliases_) {
+    switch (style_) {
+      case Flag::Style::String:
+        options.emplace_back("--" + alias + "=VAL");
```
Databean: nit: `absl::StrCat`

base/cvd/cuttlefish/flag_parser/flag.cpp
```diff
@@ -163,197 +116,230 @@ Flag& Flag::AddValidator(std::function<Result<void>()> validator) & {
...
-        }
-        CF_EXPECTF(SetAndValidate({alias.name, arg.substr(alias.name.size())}),
-                   "Processing \"{}\" failed", arg);
-        return FlagProcessResult::kFlagConsumed;
-      default:
-        return CF_ERRF("Unknown flag alias mode: {}", (int)alias.mode);
+  for (char alias : short_aliases_) {
+    switch (style_) {
+      case Flag::Style::String:
+        options.emplace_back(std::string("-") + alias + " VAL");
```
Databean: nit: `absl::StrCat`

base/cvd/cuttlefish/flag_parser/flag.cpp
```diff
@@ -163,197 +116,230 @@ Flag& Flag::AddValidator(std::function<Result<void>()> validator) & {
...
+      parsing_end =
+          std::remove_if(it, parsing_end, [&consumed](const std::string&) {
+            // "Remove" the first `consumed` elements.
+            if (consumed) {
+              --consumed;
+              return true;
+            }
+            return false;
+          });
+      continue;
```
Databean: Would this be less messy if `args` was converted into a `std::list` where elements can be more easily removed in the middle?

base/cvd/cuttlefish/flag_parser/flag.cpp
```diff
@@ -163,197 +116,230 @@ Flag& Flag::AddValidator(std::function<Result<void>()> validator) & {
...
+      // Stop parsing at first unrecognized argument.
+      break;
     }
+    if (opts.fail_on_unknown_flag) {
+      auto unknown_flag_it = std::find_if(
+          it, parsing_end,
+          [](const std::string& arg) { return arg.starts_with("-"); });
+      CF_EXPECTF(unknown_flag_it == parsing_end, "Unknown flag \"{}\"",
+                 *unknown_flag_it);
+    }
```
Databean: This seems kind of redundant with `fail_on_unexpected_argument` and there are currently no callers to `InvalidFlagGuard`, maybe `fail_on_unknown_flag` can be removed?

base/cvd/cuttlefish/flag_parser/flag.cpp
jemoreira: done

base/cvd/cuttlefish/flag_parser/flag.cpp
jemoreira: done

base/cvd/cuttlefish/flag_parser/flag.cpp
jemoreira: done

Changed all && implementations to look like `return std::move(Method(param));`

base/cvd/cuttlefish/flag_parser/flag.cpp
jemoreira: done

base/cvd/cuttlefish/flag_parser/flag.cpp
jemoreira: done

base/cvd/cuttlefish/flag_parser/flag.cpp
jemoreira: removed the short aliases

base/cvd/cuttlefish/flag_parser/flag.cpp
jemoreira: short aliases were meant to resemble that, but we can't implement the grouped feature now (too many places use -flag instead of --flag). I just removed the short aliases distinction altogether.

base/cvd/cuttlefish/flag_parser/flag.cpp
jemoreira: done

base/cvd/cuttlefish/flag_parser/shared_fd_flag.cpp
jemoreira: done

base/cvd/cuttlefish/host/commands/snapshot_util_cvd/parse.cc
jemoreira: done

base/cvd/cuttlefish/host/commands/stop/main.cc
jemoreira: done

base/cvd/cuttlefish/host/commands/tombstone_receiver/main.cpp
jemoreira: done

base/cvd/cuttlefish/host/libs/config/touchpad.cpp
jemoreira: I agree, but can't be named GflagsCompatFlag because that one already exists. The best way is to rename most of those overloads according to how they parse the flags and have two differently named that accept a vector. That's too far out of scope for this PR.

base/cvd/cuttlefish/flag_parser/flag.cpp
jemoreira: Changed to move the arguments not consumed to the front of the vector instead of removing the consumed arguments. It should look simpler now.

base/cvd/cuttlefish/flag_parser/flag.cpp
Databean: is it done?

jemoreira: It was, but lost it in the rebase. It's done now.

base/cvd/cuttlefish/flag_parser/flag_parser_test.cpp
```diff
@@ -328,32 +318,35 @@ TEST(FlagParser, InvalidIntFlag) {
...
+  auto consume_check_unexpected = [](std::vector<std::string> args) {
+    return ConsumeFlags({}, std::move(args),
+                        {.fail_on_unexpected_argument = true});
+  };
+  ASSERT_THAT(consume_check_unexpected({}), IsOk());
+  ASSERT_THAT(consume_check_unexpected({"positional"}), IsError());
+  ASSERT_THAT(consume_check_unexpected({"positional", "positional2"}),
+              IsError());
+  ASSERT_THAT(consume_check_unexpected({"-flag"}), IsError());
+  ASSERT_THAT(consume_check_unexpected({"-"}), IsError());
```
Databean: Should these be `EXPECT_THAT`? They all seem independent.

base/cvd/cuttlefish/flag_parser/gflags_compat_test.cpp
```diff
@@ -428,7 +423,8 @@ TEST(FlagParser, OptionalStringFlag_EmptyOptEmtpyFlag) {
   std::optional<std::string> value;
 
   // With EmptyString option
-  auto flag_empty_string = GflagsCompatFlag("myflag", value, CoerceToNullopt::EmptyString);
+  auto flag_empty_string =
+      GflagsCompatFlag("myflag", value, CoerceToNullopt::EmptyString);
```
Databean, line 344: Should this file be split into two files the same way the flag parser was split? Maybe something for another PR?

base/cvd/cuttlefish/flag_parser/gflags_compat.cpp
```diff
@@ -346,57 +241,35 @@ Flag HelpFlag(const std::vector<Flag>& flags, std::string text) {
...
+                 bool& print_xml, std::string text) {
   const std::string name = "helpxml";
-  auto setter = [name, &out, &value, text,
-                 &flags](const FlagMatch& match) -> Result<void> {
-    bool print_xml = false;
-    CF_EXPECT(GflagsCompatBoolFlagSetter(name, print_xml, match));
+  auto setter = [name, &out, &print_xml, text,
+                 &flags](std::string_view arg) -> Result<void> {
+    print_xml = CF_EXPECTF(ParseFlagValue<bool>(arg),
+                           "Unexpected value for \"--helpxml\": \"{}\"", arg);
```
Databean: nit: single quotes to avoid backslashing?

base/cvd/cuttlefish/flag_parser/flag.cpp
```diff
@@ -19,341 +19,274 @@
...
-        break;
-      case Flag::FlagProcessResult::kFlagConsumedWithFollowing:
-        args.erase(args.begin(), args.begin() + 2);
+      CF_EXPECTF(!opts.constrained_matching || !consumed,
+                 "Multiple flags match \"{}\"", *match_it);
+      consumed = flag_consumed;
+      if (!opts.constrained_matching) {
+        // Only match other flags when checking for ambiguous matches
         break;
+      }
```
Databean: The reason the previous implementation only checked for ambiguous matches in `ConsumeFlagsConstrained` is that that `ConsumeFlags` would scan the entire argument list for matches for a particular flag before moving to the next flag, so it wasn't possible to determine in advance if something else would consume the flag. If we're doing a single pass only, it might as well always check for ambiguous matches.

On a related subject, the behavioral change to do a single pass might affect some existing `Flag` setter and validator implementations. Before this change, if two flags were provided in the same `ConsumeFlags` call, the later member in `std::vector<Flag>` could assume that any arguments handled by the earlier `Flag` would already be processed. In the new implementation if there is a condition involving the values of two flags, it has to be checked in both setters/validators or after flag processing is done.

base/cvd/cuttlefish/flag_parser/flag.cpp
```diff
@@ -19,341 +19,274 @@
...
+      it = arguments.erase(it, it + consumed);
     }
   }
   return {};
 }
 Result<void> Flag::Parse(std::vector<std::string>&& arguments) const {
-  CF_EXPECT(Parse(static_cast<std::vector<std::string>&>(arguments)));
+  CF_EXPECT(Parse(arguments));
   return {};
 }
```
Databean: Not sure the `Flag::Parse` interface is that valuable. Maybe callers should be updated to use `ConsumeFlags` instead if the real work is handled by `Match` anyway now.

Databean (approved):

base/cvd/cuttlefish/flag_parser/flag.cpp
jemoreira: replaced Flag::Parse with calls to ConsumeFlags.

base/cvd/cuttlefish/flag_parser/flag.cpp
jemoreira: Changed to always check for ambiguity. Made the check even stronger by detecting ambiguity between the flags themselves and not just when they parse a set of arguments.

There seem to be no cases where a flag's setter or validator assumes another flag has parsed. In fact there are cases where setters and validators assume flags could have been parsed in any orders (load and start subcommands), so this change should be safe.

Also removed the option to stop parsing (without failing) when a subcommand is found. This was only used to enforce that selectors appear before the subcommand, which is not really necessary.

base/cvd/cuttlefish/flag_parser/flag_parser_test.cpp
jemoreira: done

base/cvd/cuttlefish/flag_parser/gflags_compat_test.cpp
jemoreira: done

base/cvd/cuttlefish/flag_parser/gflags_compat.cpp
jemoreira: done

PR #2683, opened 2026-06-08
Enforce project style include grouping with clang-format

Databean:
Uses [IncludeCategories](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#includecategories) to group includes, and [IncludeBlocks](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#includeblocks) to actively rearrange them to a style better matching the [style guide](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes).

Third-party libraries are also converted to double-quotes includes where possible.

Only files currently covered by clang-format tests are modified.

Bug: b/512215781

### DIFF
--- a/base/cvd/cuttlefish/result/result_type.h
+++ b/base/cvd/cuttlefish/result/result_type.h
@@ -15,7 +15,7 @@
 
 #pragma once
 
-#include <android-base/result.h>  // IWYU pragma: export
+#include "android-base/result.h"  // IWYU pragma: export
 
 #include "cuttlefish/result/error_type.h"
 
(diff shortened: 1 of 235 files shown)


cjreynol (approved):

base/cvd/cuttlefish/result/result_type.h
```diff
@@ -15,7 +15,7 @@
 
 #pragma once
 
-#include <android-base/result.h>  // IWYU pragma: export
```
jemoreira, line 18: it shouldn't drop the IWYU comments

Databean: Good catch, fixed.

Databean:
Thanks for reviewing!

PR #2688, opened 2026-06-10
Add <filesystem> include to graphics_flags.cc

Databean:
Motivation: the IsGoogler call uses std::filesystem.


Databean:
@jmacnak points out there is no `IsGoogler` call here.

PR #2689, opened 2026-06-10
Lens facing

ser-io:


jemoreira (approved):

ser-io:
Added 4  v4l2 methods which were required to pass v4l2-compliance, they are no-op though.

PR #2690, opened 2026-06-10
Clarify proxy port vs. non-proxy port + show the non-proxy port to the user

dxapd:
It seemed as though the naming was conflating the proxy port with the actual operator port, so I renamed the proxy port to suit what it actually was. Also changed streamer.cpp to show the non-proxy port to the user instead of the proxy port.

### DIFF
--- a/base/cvd/cuttlefish/host/libs/config/cuttlefish_config.h
+++ b/base/cvd/cuttlefish/host/libs/config/cuttlefish_config.h
@@ -191,6 +191,10 @@ class CuttlefishConfig {
   void set_extra_kernel_cmdline(const std::string& extra_cmdline);
   std::vector<std::string> extra_kernel_cmdline() const;
 
+  // The port for the signaling (operator) server.
+  void set_sig_server_port(int port);
+  int sig_server_port() const;
+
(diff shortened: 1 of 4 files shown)


base/cvd/cuttlefish/host/libs/config/cuttlefish_config.h
```diff
@@ -191,6 +191,10 @@ class CuttlefishConfig {
   void set_extra_kernel_cmdline(const std::string& extra_cmdline);
   std::vector<std::string> extra_kernel_cmdline() const;
 
+  // The port for the signaling (operator) server.
+  void set_sig_server_port(int port);
+  int sig_server_port() const;
+
```
jemoreira, line 197: This is a constant, it doesn't need to be in the config.

dxapd: got it, I'll remove this config declaration. if I want to avoid hardcoding "1443" in the string in streamer.cpp, where should i declare the constant? i don't believe i saw this constant defined anywhere

Databean (approved):

PR #2692, opened 2026-06-10
Update cvd monitor to use inotify

Databean:
This replaces the 50ms polling loop with an event-driven approach using inotify to wait for log file changes, implementing FIONREAD event coalescing and a 20 FPS rate-limiting throttle to protect terminal bandwidth during log storms.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/command_handler.cc
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/command_handler.cc
@@ -16,7 +16,11 @@
 
 #include "cuttlefish/host/commands/cvd/cli/commands/monitor/command_handler.h"
 
+#include <errno.h>
 #include <fcntl.h>
+#include <poll.h>
+#include <sys/inotify.h>
+#include <sys/types.h>
 #include <unistd.h>
 
 #include <chrono>
@@ -70,6 +74,16 @@ void ClearLastNLines(int n) {
   }
 }
 
+void UpdateFileAndWatch(const SharedFD& inotify_fd, const std::string& path,
+                        SharedFD& fd, int& watch) {
+  if (!fd->IsOpen()) {
+    fd = SharedFD::Open(path, O_RDONLY);
+  }
+  if (fd->IsOpen() && watch == -1) {
+    watch = inotify_fd->InotifyAddWatch(path, IN_MODIFY);
+  }
+}
+
 }  // namespace
 
 CvdMonitorCommandHandler::CvdMonitorCommandHandler(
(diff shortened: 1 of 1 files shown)


base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/command_handler.cc
```diff
@@ -97,15 +98,36 @@ Result<void> CvdMonitorCommandHandler::Handle(const CommandRequest& request) {
...
+  int launcher_watch = -1;
+  int logcat_watch = -1;
+  std::chrono::steady_clock::time_point last_draw_time =
+      std::chrono::steady_clock::now();
+
   while (true) {
     if (!kernel_fd->IsOpen()) {
       kernel_fd = SharedFD::Open(kernel_log, O_RDONLY);
+      if (kernel_fd->IsOpen() && kernel_watch == -1) {
+        kernel_watch = inotify_fd->InotifyAddWatch(kernel_log, IN_MODIFY);
```
jemoreira: This could fail and then it won't be attempted again. Is that ok?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/command_handler.cc
```diff
@@ -97,15 +98,36 @@ Result<void> CvdMonitorCommandHandler::Handle(const CommandRequest& request) {
...
+      }
     }
     if (!launcher_fd->IsOpen()) {
       launcher_fd = SharedFD::Open(launcher_log, O_RDONLY);
+      if (launcher_fd->IsOpen() && launcher_watch == -1) {
+        launcher_watch = inotify_fd->InotifyAddWatch(launcher_log, IN_MODIFY);
+      }
     }
```
jemoreira: nit: These 3 blocks are very similar, consider reusing some code:
```
if (!kernel_fd->IsOpen()) {
  std::tie(kernel_fd, kernel_fd) = TryOpenAndWatch(kernel_log);
}
```

base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/command_handler.cc
```diff
@@ -121,9 +143,35 @@ Result<void> CvdMonitorCommandHandler::Handle(const CommandRequest& request) {
...
+
+    // Enforce a maximum framerate (max 20 FPS / min 50ms between draws)
+    // so we don't saturate SSH bandwidth or CPU during heavy, continuous
+    // logging.
+    std::chrono::steady_clock::time_point now =
+        std::chrono::steady_clock::now();
+    std::chrono::steady_clock::duration elapsed = now - last_draw_time;
+    std::chrono::milliseconds min_frame_time(50);
+    if (elapsed < min_frame_time) {
+      std::this_thread::sleep_for(min_frame_time - elapsed);
```
jemoreira, line 156: nit: This read+sleep+read again could be simplified by making the inotify fd nonblocking (`inotify_init1`) and then always sleeping at the beginning of the loop followed by reading from the inotify fd until it runs out of events. That would avoid the need for the ioctl and the extra read before sleeping. It would only attempt to create watches every 50ms this way, but that's already the case because of this sleep anyways.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/command_handler.cc
Databean: Updated to retry adding watches. If any of the watches are missing, the waiting logic will wait at most 1 second before retrying.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/command_handler.cc
Databean: Created a function in an anonymous namespace for this. `std::tie` ended up not being that helpful since it needs to know the current state of both the fd and the watch.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/command_handler.cc
Databean: Made the inotify file descriptor non-blocking and changed the draining operation to run in a loop.

Databean:
Thanks for reviewing!

jemoreira (approved):

PR #2695, opened 2026-06-11
Refactor DownloadFileWithBackup to a freestanding function

Databean:
This makes the BuildApi interface smaller, reducing the burden on interface implementations. The extracted logic is also now shared between all implementations.

- Remove DownloadFileWithBackup from BuildApi interface.
- Implement it as a freestanding function accepting BuildApi& in build_api.cpp.
- Remove implementations from AndroidBuildApi and CachingBuildApi.
- Add unit tests for the new freestanding function.


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #2696, opened 2026-06-11
More descriptive value names when printing flag help messages

jemoreira:
Prints `--instance-name=NAME[,NAME...]` instead of `--instance-name=VAL` by default. Allows customizing it.

Assisted-by: Gemini

### DIFF
--- a/base/cvd/cuttlefish/flag_parser/flag.cpp
+++ b/base/cvd/cuttlefish/flag_parser/flag.cpp
@@ -144,6 +144,8 @@ const std::string& Flag::Description() const { return help_; }
 
 std::string Flag::CurrentValue() const { return getter_(); }
 
+const std::string& Flag::ValueNameHint() const { return value_name_hint_; }
+
 Flag::Flag(std::string name, Flag::Style style) : style_(style) {
   ValidateAlias(name);
   value_name_hint_ = DefaultValueNameHint(name);
--- a/base/cvd/cuttlefish/flag_parser/flag.cpp
+++ b/base/cvd/cuttlefish/flag_parser/flag.cpp
@@ -30,9 +30,11 @@
 
 #include "absl/cleanup/cleanup.h"
 #include "absl/log/check.h"
+#include "absl/strings/ascii.h"
 #include "absl/strings/str_cat.h"
 #include "absl/strings/str_join.h"
 #include "absl/strings/str_replace.h"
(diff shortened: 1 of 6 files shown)


base/cvd/cuttlefish/flag_parser/gflags_compat.cpp
```diff
@@ -199,6 +199,11 @@ std::string XmlEscape(const std::string& s) {
   return absl::StrReplaceAll(s, {{"<", "&lt;"}, {">", "&gt;"}});
 }
 
+Flag WithVectorNameValueHint(Flag flag) {
+  std::string hint = flag.ValueNameHint();
+  return std::move(flag).ValueNameHint(hint + "[," + hint + "...]");
```
Databean: nit: `absl::StrCat`

base/cvd/cuttlefish/flag_parser/flag.cpp
```diff
@@ -53,6 +55,15 @@ bool MatchNormalized(std::string_view str1, std::string_view str2) {
   return n1 == n2;
 }
 
+std::string DefaultValueNameHint(std::string_view name) {
+  std::vector<std::string_view> parts =
+      absl::StrSplit(name, absl::ByAnyChar("-_"), absl::SkipEmpty());
+  if (parts.empty()) {
+    return "VAL";
+  }
```
Databean, line 67: It seems a bit unintuitive to make the "default" the last word, but then only sometimes use this "default" because some `GflagsCompatFlag` overloads override it to the empty string.

What do you think of always defaulting to `"VAL"` but invoking this "find the last word" logic in some of the `GflagsCompatFlag` implementations instead?

Databean (approved):

base/cvd/cuttlefish/flag_parser/gflags_compat.cpp
jemoreira: done

base/cvd/cuttlefish/flag_parser/flag.cpp
jemoreira: No `GflagsCompatFlag` overload should set this to the empty string. The only overloads that change it still use the default to construct a better one.

Flag was the most central location I could find, making it part of `GflagsCompatFlag` would require some repetition.

PR #2709, opened 2026-06-12
podcvd supports GPU acceleration

0405ysj:
When this PR and https://github.com/google/android-cuttlefish/pull/2708 is being merged, GPU acceleration would be available on podcvd.

Tested via `podcvd create --gpu_mode=gfxstream`.

Context: b/491255968

### DIFF
--- a/container/debian/podcvd-setup
+++ b/container/debian/podcvd-setup
@@ -148,6 +148,44 @@ setup_rootless_podman() {
     sudo -H -u "$username" podman system migrate
 }
 
+setup_nvidia_gpu() {
+    if ! command -v nvidia-smi >/dev/null 2>&1; then
+        echo "nvidia-smi not found. NVIDIA GPU is likely not available."
+        return 1
+    fi
+    if ! command -v nvidia-ctk >/dev/null 2>&1; then
+        echo "nvidia-ctk not found. Please install nvidia-container-toolkit."
+        return 1
+    fi
+    local VERSION=$(nvidia-smi --query-gpu=driver_version --format=csv,noheader | head -n 1)
+    [ -n "$VERSION" ] || return 1
+    local CDI_CONFIG_PATH=/etc/cdi/nvidia-podcvd.yaml
+    sudo mkdir -p /etc/cdi
+    sudo nvidia-ctk cdi generate --output=$CDI_CONFIG_PATH 2>/dev/null || return 1
+    sudo yq -y '.kind = "android.com/gpu-podcvd"' -i $CDI_CONFIG_PATH 2>/dev/null || return 1
+
+    # Some nvidia libraries were missing from the configuration file created by
+    # executing `nvidia-ctk cdi generate`. This leads an error to launch
+    # Cuttlefish instance with GPU acceleration. (e.g. vulkan) This step is
+    # required to fulfill missing nvidia libraries so that container instance
+    # can have all necessary libraries.
+    for lib in /usr/lib/*-linux-gnu/libnvidia*${VERSION}; do
+        [ -e "$lib" ] || continue
+        if grep -qF "$lib" "$CDI_CONFIG_PATH"; then
+            continue
+        fi
+        sudo yq -y "
+        .containerEdits.mounts += [
+            {
+            \"hostPath\": \"$lib\",
+            \"containerPath\": \"$lib\",
+            \"options\": [\"ro\", \"nosuid\", \"nodev\", \"rbind\", \"rprivate\"]
+            }
+        ]
+        " -i $CDI_CONFIG_PATH || return 1
+    done
+}
+
 USER_CONFIG="/etc/podcvd.users"
 username="${1:-${SUDO_USER:-$(id -un)}}"
 if [ -z "$username" ] || ! id "$username" >/dev/null 2>&1; then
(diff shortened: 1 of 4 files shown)


0405ysj:
The reason for setting this PR as a draft for a moment is, there's no good guard codes on `setup_nvidia_gpu`, such as `nvidia-container-toolkit` not installed on the machine. When nvidia drivers exist and it's ready to use for containers, I verified it's working but needs few more lines of codes to get review and merge it. Of course, I need to add all reviewers required when it's ready.

container/debian/podcvd-setup
```diff
@@ -148,6 +148,38 @@ setup_rootless_podman() {
...
+        echo "nvidia-ctk not found. Please install nvidia-container-toolkit."
+        return 1
+    fi
+    local VERSION=$(nvidia-smi --query-gpu=driver_version --format=csv,noheader | head -n 1)
+    [ -n "$VERSION" ] || return 1
+    local CDI_CONFIG_PATH=/etc/cdi/nvidia-podcvd.yaml
+    sudo mkdir -p /etc/cdi
+    sudo nvidia-ctk cdi generate --output=$CDI_CONFIG_PATH 2>/dev/null || return 1
+    sudo yq -y '.kind = "android.com/gpu-podcvd"' -i $CDI_CONFIG_PATH 2>/dev/null || return 1
+    for lib in /usr/lib/*-linux-gnu/libnvidia*${VERSION}; do
```
jmacnak, line 172: Is this due to the issue we were seeing before where it seemed like `nvidia-ctk cdi generate` was failing to detect some of the libraries (and thus missing some of the mounts in the resulting yaml file) so we are manually appending those here? If so, please add a comment explaining this as one would expect `nvidia-ctk cdi generate` to handle this.

jmacnak (commented):
> When this PR and https://github.com/google/android-cuttlefish/pull/2708 is being merged, GPU acceleration would be available on podcvd.

The first comment on the PR links to https://github.com/google/android-cuttlefish/pull/2708 . IIUC, `mesa-vulkan-drivers` is for the Intel/AMD/lavapipe-software vulkan drivers. Why are the those drivers needed to get Nvidia working?

0405ysj:
> IIUC, mesa-vulkan-drivers is for the Intel/AMD/lavapipe-software vulkan drivers. Why are the those drivers needed to get Nvidia working?

I saw logs indicating failed checks around vulkan, so I thought it's required.

```
assemble_cvd(627)  D 06-24 00:52:17   627   627 graphics_flags.cc:498]   5. FAILED: The host does not have the Vulkan loader installed. Please ensure the Vulkan loader is installed and available.
assemble_cvd(627)  D 06-24 00:52:17   627   627 graphics_flags.cc:498]   6. FAILED: The host does not have a Vulkan driver available. Please ensure a Vulkan driver is installed.
```

On the other hand, I was able to boot Cuttlefish instance with consuming GPU resources(checked via `nvidia-smi`) though those logs were appeared. Would you have any opinion whether I should keep `mesa-vulkan-drivers` or not?

According to your comment, `mesa-vulkan-drivers` is required for `--gpu_mode=gfxstream_guest_angle_host_lavapipe` at least?

https://github.com/google/android-cuttlefish/blob/fcc54ea052aeeb69afb3a22bd57531b52119488b/base/cvd/cuttlefish/host/commands/assemble_cvd/graphics_flags.cc#L279-L287

jmacnak:
>  5. FAILED: The host does not have the Vulkan loader installed. Please ensure the Vulkan loader is installed and available.

For this one, you want to install `libvulkan1` which is built from https://github.com/KhronosGroup/vulkan-loader . This is the vendor neutral library for opening Vulkan drivers and layers.

> 6. FAILED: The host does not have a Vulkan driver available. Please ensure a Vulkan driver is installed.

For this one, this is likely more related to the current PR. Besides the `/usr/lib/*-linux-gnu/libnvidia*` libraries, there should also be a Vulkan ICD manifest file for Nvidia (typically in `/usr/share/vulkan/icd.d` or `/etc/vulkan/icd.d`, see https://github.com/KhronosGroup/Vulkan-Loader/blob/main/docs/LoaderDriverInterface.md#driver-discovery-on-linux for the default search locations. The manifest file tells the Vulkan loader what version/features are supported and where the vendor's specific Vulkan library is in the `library_path` key.

jmacnak:
> On the other hand, I was able to boot Cuttlefish instance with consuming GPU resources(checked via nvidia-smi) though those logs were appeared. Would you have any opinion whether I should keep mesa-vulkan-drivers or not?

It looks like `mesa-vulkan-drivers` depends on `libvulkan1` so installing the former likely installed the piece we were missing.

Potentially we should like to the Vulkan loader docs page / or potentially just list `sudo apt-get install libvulkan1` around https://github.com/google/android-cuttlefish/blob/fcc54ea052aeeb69afb3a22bd57531b52119488b/base/cvd/cuttlefish/host/commands/assemble_cvd/graphics_flags.cc#L178.

0405ysj:
I think I can switch into `libvulkan1` from `mesa-vulkan-drivers` on Containerfile. I naively thought installing `libvulkan1` only resolves check 5 not 6, but I could see both were passed after installing `libvulkan1` only. I applied it into this PR. Thanks for the comment!

container/debian/podcvd-setup
0405ysj: Yes, exactly. Somehow `nvidia-ctk cdi generate` misses to detect some nvidia libraries, so `vulkaninfo` or `eglinfo` didn't work properly without these lines.

I added explanation comment why this step is required to launch Cuttlefish.

jmacnak (approved):

PR #2710, opened 2026-06-12
Fix ho client for group start/stop requests

ser-io:


frontend/src/libhoclient/host_orchestrator_client.go
```diff
@@ -464,7 +464,10 @@ func (c *HostOrchestratorClientImpl) Powerbtn(groupName, instanceName string) er
 }
 
 func (c *HostOrchestratorClientImpl) Stop(groupName, instanceName string) error {
-	path := fmt.Sprintf("/cvds/%s/%s/:stop", groupName, instanceName)
+	path := fmt.Sprintf("/cvds/%s/:stop", groupName)
```
jemoreira, line 471: Could these be two functions? `StopGroup` and `StopInstance`?

ser-io: Good call! Changed the names.

jemoreira (approved):

PR #2714, opened 2026-06-13
Always print selectors before subcommand flags

jemoreira:


base/cvd/cuttlefish/host/commands/cvd/cli/commands/command_handler.cpp
```diff
@@ -55,6 +55,9 @@ Result<std::string> CvdCommandHandler::DetailedHelp(
   cvd_common::Args args = request.SubcommandArguments();
   CF_EXPECT(ConsumeFlags(flags, args));
 
+  // Make sure the flags are in alphabetical order
+  std::sort(flags.begin(), flags.end());
```
Databean, line 59: Should this happen before `ConsumeFlags`? Might as well immediately run it after declaring the variable.

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/command_handler.cpp
jemoreira: Before the flags refactor the order of flags could influence how they were parsed. It shouldn't matter today, but I still prefer to consume the flags first, given that sorting is only done for the help output, just in case.

PR #2720, opened 2026-06-16
Skip TAP device validation for cvdalloc launches.

dxapd:
Some recent changes tightened tap device validation in the current path, but unfortunately tightening up tap device validation isn't really compatible with how cvdalloc owns the resources it creates, so currently this halts device boot.. Ideally, we will be able to trust cvdalloc to validate its own resources at some point.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/run_cvd/validate.cpp
+++ b/base/cvd/cuttlefish/host/commands/run_cvd/validate.cpp
@@ -37,7 +37,8 @@ namespace cuttlefish {
 static Result<void> TestTapDevices(
     const CuttlefishConfig::InstanceSpecific& instance) {
 #ifdef __linux__
-  if (InSandbox()) {
+  if (InSandbox()
+  || instance.use_cvdalloc()) { // cvdalloc owns its own resources. those resources can't be validated this way.
     return {};
   }
   auto wifi = instance.wifi_tap_name();


base/cvd/cuttlefish/host/commands/run_cvd/validate.cpp
```diff
@@ -37,7 +37,8 @@ namespace cuttlefish {
 static Result<void> TestTapDevices(
     const CuttlefishConfig::InstanceSpecific& instance) {
 #ifdef __linux__
-  if (InSandbox()) {
+  if (InSandbox()
+  || instance.use_cvdalloc()) { // cvdalloc owns its own resources. those resources can't be validated this way.
```
Databean, line 41: Can some equivalent logic be invoked in [`Cvdalloc::WaitForAvailability`](https://github.com/google/android-cuttlefish/blob/b0fe18997989ec37a6b695682cbaba1917681a94/base/cvd/cuttlefish/host/commands/run_cvd/launch/cvdalloc.cpp#L84) to validate from this side the tap devices were created successfully?

dxapd: imo something like that would need to be semantically different. my understanding is that since cvdalloc is supposed to be the component that knows and cares about "allocating tap devices", the expectation from run_cvd would be something more like "ValidateCvdallocOk" (or perhaps something better worded than that). that way, the main process is just asking "did allocation succeed?".

theoretically, it is defensible to have the main process hold cvdalloc accountable more granularly, but that's a little bit too much of split implementation for my taste.

something like ValidateCvdallocOk could be rolled into this PR or a follow-up, doesn't matter too much - what do you think?

Databean: I think it's reasonable to tie "the TAP devices given to crosvm are valid" to the logic used to launch crosvm. Admittedly it could be closer now. Maybe a followup PR can include moving this logic somewhere in `CrosvmBuilder`.

Databean (approved):

base/cvd/cuttlefish/host/commands/run_cvd/validate.cpp
dxapd: I see, that makes sense. I think what I would like to see is just "i want xyz tap devices" (the requirement itself) and "i want to make sure xyz tap devices exist" (the verification) to be in the same/similar place. I suppose that it's already not too clear right now where the TAP device requirements/names are coming from in this file.

PR #2723, opened 2026-06-17
Foreground mode for `cvd start` (`--daemon=false`)

Databean:
When run with `--daemon=false`, `cvd start` will launch the device in the background but immediately attach a log monitor (`cvd monitor`) to the console, providing real-time boot status. Upon receiving an interrupt signal (Ctrl-C), it gracefully shuts down the launched instance group.

Assisted-by: Jetski:gemini-2.5-pro
Bug: b/519304405

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/monitor.cc
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/monitor.cc
@@ -18,8 +18,9 @@
 
 #include <errno.h>
 #include <fcntl.h>
-#include <poll.h>
+#include <sys/eventfd.h>
 #include <sys/inotify.h>
+#include <sys/poll.h>
 #include <sys/types.h>
 #include <unistd.h>
 
@@ -28,6 +29,7 @@
 #include <memory>
 #include <string>
 #include <thread>
+#include <vector>
 
 #include "absl/strings/str_cat.h"
 
@@ -40,15 +42,7 @@
 #include "cuttlefish/result/result.h"
 
 namespace cuttlefish {
-
 namespace {
-
-void ClearLastNLines(int n) {
-  if (n > 0) {
-    // Move cursor up N lines and clear to end of screen
-    std::cout << AnsiCursorUp(n) << kAnsiClearScreenAfterCursor << std::flush;
-  }
-}
 
 void UpdateFileAndWatch(const SharedFD& inotify_fd, const std::string& path,
                         SharedFD& fd, int& watch) {
@@ -62,7 +56,17 @@ void UpdateFileAndWatch(const SharedFD& inotify_fd, const std::string& path,
 
 }  // namespace
 
+void ClearLastNLines(int n) {
+  if (n > 0) {
+    // Move cursor up N lines and clear to end of screen
+    std::cout << AnsiCursorUp(n) << kAnsiClearScreenAfterCursor << std::flush;
+  }
+}
 Result<void> MonitorLogs(const LocalInstance& instance) {
+  return MonitorLogs(instance, SharedFD());
+}
+
+Result<void> MonitorLogs(const LocalInstance& instance, SharedFD stop_eventfd) {
   CF_EXPECT(isatty(0), "The monitor command requires an interactive terminal.");
 
   const std::string kernel_log =
@@ -122,19 +126,34 @@ Result<void> MonitorLogs(const LocalInstance& instance) {
     }
     last_draw_time = std::chrono::steady_clock::now();
 
+    // Setup poll_fds
+    std::vector<PollSharedFd> poll_fds = {
+        PollSharedFd{.fd = inotify_fd, .events = POLLIN, .revents = 0},
+    };
+    if (stop_eventfd->IsOpen()) {
+      poll_fds.push_back(
+          PollSharedFd{.fd = stop_eventfd, .events = POLLIN, .revents = 0});
+    }
+
     // Block until file changes occur. If any watch failed or is missing, use
     // a 1-second fallback timeout to awake and retry.
-    // NOLINTNEXTLINE(misc-include-cleaner)
-    PollSharedFd poll_fd = {inotify_fd, POLLIN, 0};
     int timeout_ms = -1;
     if (dir_watch == -1 || kernel_watch == -1 || launcher_watch == -1 ||
         logcat_watch == -1) {
       timeout_ms = 1000;
     }
 
-    // NOLINTNEXTLINE(misc-include-cleaner)
-    if (SharedFD::Poll(&poll_fd, 1, timeout_ms) > 0 &&
-        (poll_fd.revents & POLLIN)) {
+    int poll_res = SharedFD::Poll(poll_fds, timeout_ms);
+
+    if (stop_eventfd->IsOpen() && (poll_fds[1].revents & POLLIN)) {
+      // Stop requested via eventfd
+      eventfd_t val;
+      stop_eventfd->EventfdRead(&val);
+      ClearLastNLines(total_lines_drawn);
+      return {};
+    }
+
+    if (poll_res > 0 && (poll_fds[0].revents & POLLIN)) {
       // Exhaustively drain all available events from the non-blocking
       // descriptor to coalesce rapid file modifications.
       char buf[4096]
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/monitor.cc
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/monitor.cc
@@ -34,6 +34,7 @@
 #include "absl/strings/str_cat.h"
 
 #include "cuttlefish/common/libs/fs/shared_fd.h"
+#include "cuttlefish/common/libs/utils/files.h"
 #include "cuttlefish/host/commands/cvd/cli/commands/monitor/ansi_codes.h"
 #include "cuttlefish/host/commands/cvd/cli/commands/monitor/display.h"
 #include "cuttlefish/host/commands/cvd/cli/utils.h"
@@ -47,7 +48,9 @@ namespace {
 void UpdateFileAndWatch(const SharedFD& inotify_fd, const std::string& path,
                         SharedFD& fd, int& watch) {
   if (!fd->IsOpen()) {
-    fd = SharedFD::Open(path, O_RDONLY);
+    if (FileExists(path)) {
+      fd = SharedFD::Open(path, O_RDONLY);
+    }
   }
   if (fd->IsOpen() && watch == -1) {
     watch = inotify_fd->InotifyAddWatch(path, IN_MODIFY);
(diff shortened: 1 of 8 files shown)


base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/monitor.h
```diff
@@ -16,11 +16,16 @@
...
 #include "cuttlefish/host/commands/cvd/instances/local_instance.h"
 #include "cuttlefish/result/result.h"
 
 namespace cuttlefish {
 
+void ClearLastNLines(int n);
+
 Result<void> MonitorLogs(const LocalInstance& instance);
 
+Result<void> MonitorLogs(const LocalInstance& instance, SharedFD stop_eventfd);
```
jemoreira, line 31: nit: Add some comments explaining what the eventfd is for. It's not clear from the header alone whether the monitor reads the fd to determine when to stop or it writes to it when it decides to stop something.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
```diff
@@ -370,8 +379,24 @@ Result<void> CvdStartCommandHandler::Handle(const CommandRequest& request) {
...
   }
 
-  CF_EXPECT(ConsumeDaemonModeFlag(subcmd_args));
+  bool is_daemon = CF_EXPECT(ConsumeDaemonModeFlag(subcmd_args));
   subcmd_args.push_back("--daemon=true");
+  if (!is_daemon) {
+    bool has_stats_flag = false;
+    for (const std::string& arg : subcmd_args) {
+      if (absl::StartsWith(arg, "--report_anonymous_usage_stats") ||
+          absl::StartsWith(arg, "-report_anonymous_usage_stats")) {
```
jemoreira: Can this be done with the flag instead? Flag presence can be determined by using `std::optional<bool>` instead of just `bool` with `GflagsCompatFlag`.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
```diff
@@ -317,16 +331,11 @@ CvdStartCommandHandler::CvdStartCommandHandler(
...
-static Result<void> ConsumeDaemonModeFlag(cvd_common::Args& args) {
+static Result<bool> ConsumeDaemonModeFlag(cvd_common::Args& args) {
   bool daemon = true;
-  Flag flag =
-      GflagsCompatFlag("daemon", daemon)
-          .AddValidator([&daemon]() -> Result<void> {
-            CF_EXPECT(!!daemon, "`cvd start` must always run in daemon mode.");
-            return {};
-          });
+  Flag flag = GflagsCompatFlag("daemon", daemon);
```
jemoreira: This flag should be returned from the `Flags` method so it's documented by `cvd help start`.

This flag is reported by cvd_internal_start, but removed from the list by `GetCvdInternalFlags`. I think it should continue to be removed but this one returned from `BuildOwnFlags`.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
```diff
@@ -393,41 +418,112 @@ Result<void> CvdStartCommandHandler::Handle(const CommandRequest& request) {
...
-    if (!interrupt_res.ok()) {
-      LOG(ERROR) << "Failed to stop subprocesses: " << interrupt_res.error();
-      LOG(ERROR) << "Devices may still be executing in the background, run "
-                    "`cvd reset` to ensure a clean state";
-    }
+  SharedFD memfd;
+  SharedFD stop_eventfd;
+  if (!is_daemon) {
+    memfd = SharedFD::MemfdCreate("cvd_internal_start_output");
+    CF_EXPECT(memfd->IsOpen(), "Failed to create memfd for subprocess output");
```
jemoreira: nit: Include `StrError()` in the error message for these two fds.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
```diff
@@ -549,6 +645,12 @@ Result<void> CvdStartCommandHandler::LaunchSingleInstance(
     }
   };
 
+  LOG(INFO)
```
jemoreira: This is repeated from line 468, it could go out of sync. Is there a way to avoid the repetition? Could it be shown at creation time instead of start time?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/monitor.cc
```diff
@@ -97,8 +101,28 @@ Result<void> MonitorLogs(const LocalInstance& instance, SharedFD stop_eventfd) {
       std::chrono::steady_clock::time_point();
 
   while (true) {
+    if (using_assemble_log) {
+      if (FileExists(launcher_log)) {
+        if (launcher_watch != -1) {
+          inotify_fd->InotifyRmWatch(launcher_watch);
+          launcher_watch = -1;
+        }
+        launcher_fd = SharedFD::Open(launcher_log, O_RDONLY);
```
jemoreira, line 113: nit: log if `Open` fails?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/monitor.cc
```diff
@@ -132,11 +133,30 @@ Result<void> MonitorLogs(const LocalInstance& instance, SharedFD stop_eventfd) {
...
+    if (logcat_fd->IsOpen() && logcat_fd->LSeek(0, SEEK_END) > 0) {
+      logcat_ready = true;
+    }
+
+    size_t total_content = 30;
+
+    if (using_assemble_log) {
+      display.DrawFile(launcher_fd, kLogNameAssembleCvd, total_content);
+    } else if (!kernel_fd->IsOpen()) {
+      display.DrawFile(launcher_fd, kLogNameLauncher, total_content);
```
jemoreira: This code is really hard to read and understand, especially this section where it asks about the kernel logs and draws the launcher. Consider rewriting roughly like this:
```
launcher_name = using_assemble_cvd? kLogNameAssembleCvd: kLogNameLauncher;
launcher_lines = total_content;
kernel_lines = 0
logcat_lines = 0;
if (kernel_fd->IsOpen()) {
  launcher_lines = total_content/3;
  kernel_lines = total_content - launcher_lines;
}
if (logcat_ready) {
  kernel_lines = total_content / 3;
  logcat_lines = total_content - launcher_lines - kernel_lines;
}
// This assumes DrawFile won't do anything if given 0 lines
display.DrawFile(launcher_fd, launcher_name, launcher_lines);
display.DrawFile(kernel_fd, kLogNameKernel, kernel_lines);
display.DrawFile(logcat_fd, kLogNameLogcat, logcat_lines);
```

jemoreira (commented):
IMHO this is adding more complication to the already complicated start command handler and the next PR is doing something similar to create. All this only to support a legacy flow that we had moved away from. I think it would be better if this was implemented in a different binary that called `cvd create`, `cvd start`, `cvd monitor` as needed and handled the user interactions instead of adding more "modes" to the existing commands.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/monitor.cc
Databean: The `SharedFD`s in an error state are passed down to `DrawFile`, which internally prints the `StrError()` messages for them within the ASCII boxes normally used for showing log contents.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/monitor.h
Databean: (AI) added a comment.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
Databean: (AI) added it to the flags struct.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
Databean: (AI) added it to the flags struct.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
Databean: (AI) done.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
Databean: (AI) removed this entirely.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/monitor.cc
Databean: (AI) done.

Databean:
> IMHO this is adding more complication to the already complicated start command handler and the next PR is doing something similar to create.

Completely agree with this, it's a mess.

> All this only to support a legacy flow that we had moved away from.

I personally also miss the flow of one command to start a device, and having it exit when I press ctrl-c. The two PRs together bring back this flow.

> I think it would be better if this was implemented in a different binary that called `cvd create`, `cvd start`, `cvd monitor` as needed and handled the user interactions instead of adding more "modes" to the existing commands.

I would rather not introduce more layers on top of the `cvd` executable in the command line interface, only below. `cvd` is already supposed to be the "user-facing" command line interface.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.h
```diff
@@ -63,6 +64,8 @@ class CvdStartCommandHandler : public CvdCommandHandler {
   SubprocessWaiter subprocess_waiter_;
   struct {
     std::vector<std::string> host_substitutions;
+    std::optional<bool> daemon;
```
jemoreira: based on how it's used this could be bool with default value of true

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
```diff
@@ -370,8 +371,18 @@ Result<void> CvdStartCommandHandler::Handle(const CommandRequest& request) {
...
+  bool is_daemon = own_flags_.daemon.value_or(true);
   subcmd_args.push_back("--daemon=true");
+  if (own_flags_.report_anonymous_usage_stats) {
+    subcmd_args.push_back("--report_anonymous_usage_stats=" +
+                          *own_flags_.report_anonymous_usage_stats);
+  } else if (!is_daemon) {
+    // If the user did not specify the anonymous usage stats flag, default it
+    // to 'n' to ensure that cvd_internal_start does not block waiting for
+    // input on stdin.
+    subcmd_args.push_back("--report_anonymous_usage_stats=n");
```
jemoreira, line 385: Wouldn't the correct approach be to ask the user when not in daemon mode?

jemoreira (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
Databean: I both don't want to interrupt users and don't want to reimplement the consent flow.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.h
Databean: Done.

jemoreira (approved):

PR #2724, opened 2026-06-17
cvd: Add --reuse flag to cvd create

Databean:
Adds `--reuse_if_possible` flag to `cvd create`. If this flag is provided, `cvd create` will check if there is an existing instance group with matching host and product paths. If a matching group is found and it is not in an active state (preparing, starting, running, stopping), it will delegate to `cvd start` for that group and skip the rest of the creation logic.

Exposes `ExpandProductPaths` in `cvd` utils to share it between `cvd create` and `CreationAnalyzer` for consistent product path expansion.

Assisted-by: Jetski:Gemini-Next
Bug: b/519304405
TAG=agy
CONV=34ddc7b8-a695-4937-be99-b796087dd987


base/cvd/cuttlefish/host/commands/cvd/utils/common.cpp
```diff
@@ -146,4 +147,18 @@ std::string AssemblyDirFromHome(const std::string& group_home_dir) {
   return to_ret;
 }
 
+std::vector<std::string> ExpandProductPaths(const std::string& product_path,
+                                            size_t num_instances) {
+  const std::vector<std::string_view> split = absl::StrSplit(product_path, ',');
+  std::vector<std::string> expanded;
+  expanded.reserve(num_instances);
+  for (size_t i = 0; i < std::min(split.size(), num_instances); ++i) {
+    expanded.push_back(std::string(split[i]));
```
jemoreira: nit: `expanded.emplace_back(split[i])`

base/cvd/cuttlefish/host/commands/cvd/utils/common.h
```diff
@@ -56,4 +56,7 @@ std::string AssemblyDirFromHome(const std::string& group_home_dir);
 // variables followed by the current directory.
 Result<std::string> AndroidHostPath(const cvd_common::Envs& env);
 
+std::vector<std::string> ExpandProductPaths(const std::string& product_path,
```
jemoreira: nit: This function looks like it will always be used for dealing with user provided values (a vectorized flag?), so it should go somewhere under the `cli` directory.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
```diff
@@ -236,6 +250,47 @@ std::vector<Flag> BuildSelectorFlagsForCreateHelp(
   };
 }
 
+bool IsActiveState(cvd::InstanceState state) {
```
jemoreira: why not use `LocalInstance::IsActive`?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
```diff
@@ -236,6 +250,47 @@ std::vector<Flag> BuildSelectorFlagsForCreateHelp(
...
+         state == cvd::INSTANCE_STATE_STOPPING;
+}
+
+bool IsGroupReuseable(const LocalInstanceGroup& group) {
+  for (const LocalInstance& instance : group.Instances()) {
+    if (IsActiveState(instance.State())) {
+      return false;
+    }
+  }
+  return true;
```
jemoreira: nit: This whole thing could be `return std::ranges::any_of(group.Instances(), &LocalInstance::IsActive);`

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
```diff
@@ -275,18 +330,49 @@ Result<void> CvdCreateCommandHandler::Handle(const CommandRequest& request) {
...
+    const std::vector<std::string> target_products =
+        ExpandProductPaths(target_product, num_instances_parser_.NumInstances());
+
+    const std::vector<LocalInstanceGroup> groups =
+        CF_EXPECT(instance_manager_.FindGroups({}));
+
+    std::optional<LocalInstanceGroup> matched_group;
+    for (const LocalInstanceGroup& g : groups) {
+      const auto match_res = MatchPaths(g, target_host, target_products);
+      if (!match_res.ok()) {
```
jemoreira: nit: Log the errors?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
```diff
@@ -275,18 +330,49 @@ Result<void> CvdCreateCommandHandler::Handle(const CommandRequest& request) {
...
+    if (matched_group) {
+      if (IsGroupReuseable(*matched_group)) {
+        LOG(INFO) << "Reusing existing instance group: "
+                  << matched_group->GroupName();
+        CF_EXPECT(StartGroup(*matched_group, subcmd_args, envs, instance_manager_));
+        return {};
+      } else {
+        LOG(WARNING) << "Found matching instance group '"
+                     << matched_group->GroupName()
+                     << "' but it is in active state. Proceding with creation of a new group.";
```
jemoreira: This can cause the number of groups to grow indefinitely and end up exhausting the available instance ids (even with cvdalloc, the fixed port ranges we use still impose a limit). Even if the user hits Ctrl-C while in not-daemon mode the stop operation may fail and the original group end up active in the background. It'd be better to unconditionally attempt to start the existing group and let it fail with files in use or some other error. The user can always run `cvd reset -y` to fix that.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
```diff
@@ -275,18 +330,49 @@ Result<void> CvdCreateCommandHandler::Handle(const CommandRequest& request) {
...
+    std::optional<LocalInstanceGroup> matched_group;
+    for (const LocalInstanceGroup& g : groups) {
+      const auto match_res = MatchPaths(g, target_host, target_products);
+      if (!match_res.ok()) {
+        continue;
+      }
+      if (*match_res) {
+        matched_group = g;
+        break;
+      }
```
jemoreira: Why choose the first one if two groups match? What if the first one is still running and the second one is stopped? The user may expect to reuse the second one.

The real problem here is that the flag doesn't let you specify which group to reuse and can lead to ambiguous situations.

We already have the selector flag to specify which group to start, so why not have `cvd --group_name=mygroup create --replace_existing [args]` instead?

I assume this feature is only ever going to be used by `launch_cvd`, so `launch_cvd` could call `cvd create --group_name=_launch_cvd_{checksum(host_path, target_path)} [args]`. The group is removed first if it already exists (it would be nice to reuse the same directories under /var/tmp/cvd for the new group). This approach would also allow reusing the group even if the number of instances changes, which is probably what a user running `launch_cvd` expects. The command should probably fail if the existing group has active instances.

base/cvd/cuttlefish/host/commands/cvd/utils/common.h
Databean: Moved to `cli/utils.h`.

base/cvd/cuttlefish/host/commands/cvd/utils/common.cpp
Databean: Rewrote this and used `emplace_back`.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
Databean: Now reported via `CF_EXPECT`.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
Databean: Changed the semantics to pass through selectors and fail if there is ambiguity.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
Databean: Changed the semantics to fail if the chosen group is already running.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
Databean: Done.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
Databean: Good point, changed.

base/cvd/cuttlefish/host/commands/cvd/cli/selector/creation_analyzer.cpp
```diff
@@ -82,31 +83,23 @@ Result<GroupCreationInfo> AnalyzeCreation(const CreationAnalyzerParam& params) {
...
-  while (group_directories.product_out_paths.size() < params.num_instances) {
-    // Use the first product path when more instances are required than product
-    // paths provided. This supports creating multiple identical instances from
-    // a single set of images.
-    group_directories.product_out_paths.emplace_back(
-        group_directories.product_out_paths[0]);
+    product_out_val = group_directories.host_artifacts_path.value_or("");
   }
+  const std::vector<std::string> expanded =
+      ExpandProductPaths(product_out_val, params.num_instances);
```
jemoreira, line 101: nit: If ExpandProductPaths were written like this:
```
std::vector<std::string> expanded = absl::StrSplit(product_path, ',');
CF_EXPECT(expanded.size() <= num_instances, ...);
expanded.resize(num_instances, expanded[0]);
```
then this block could be:
```
const auto it = params.envs.find(kAndroidProductOut);
std::string product_out_val = it != params.envs.end()? it->second : group_directories.host_artifacts_path.value_or("");
const std::vector<std::string> expanded =
      ExpandProductPaths(product_out_val, params.num_instances);
```

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
```diff
@@ -236,12 +250,73 @@ std::vector<Flag> BuildSelectorFlagsForCreateHelp(
...
 
+Result<std::optional<LocalInstanceGroup>>
+CvdCreateCommandHandler::TryReuseGroup(
+    const selector::SelectorOptions& selectors, const cvd_common::Envs& envs) {
+  InstanceDatabase::Filter filter =
+      selector::BuildFilterFromSelectors(selectors);
+
+  const std::vector<LocalInstanceGroup> groups =
+      CF_EXPECT(instance_manager_.FindGroups(filter));
+  if (groups.empty()) {
```
jemoreira, line 296: nit: these `if`s could be implemented as guards:
```
if (groups.empty()) {
  return std::nullopt;
}
CF_EXPECTF(groups.size() == 1, ...);
...
```

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
```diff
@@ -236,12 +250,73 @@ std::vector<Flag> BuildSelectorFlagsForCreateHelp(
...
+
+    auto target_product_it = envs.find(kAndroidProductOut);
+    CF_EXPECT(target_product_it != envs.end());
+
+    const std::vector<std::string> target_products = ExpandProductPaths(
+        target_product_it->second, num_instances_parser_.NumInstances());
+    bool match = CF_EXPECT(
+        MatchPaths(groups[0], target_host_it->second, target_products));
+    CF_EXPECT(std::move(match),
+              "Trying to change the host paths on an existing group");
```
jemoreira: nit: I don't think it's necessary to fail on this check. `launch_cvd` doesn't fail on an equivalent situation today and it would allow reusing the group while having a different number of instances. This could be implemented by removing the old group and creating a new one with the same group directory under /var/tmp/cvd/{uid}.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.h
```diff
@@ -44,13 +45,15 @@ class CvdCreateCommandHandler : public CvdCommandHandler {
     std::string product_path;
     bool start;
     std::string config_file;
+    bool reuse;
   };
 
   std::vector<Flag> ConfigFileModeFlags();
   std::vector<Flag> FlagModeFlags(const cvd_common::Envs& env,
                                   const selector::SelectorOptions&);
+  Result<std::optional<LocalInstanceGroup>> TryReuseGroup(
```
jemoreira: nit: Rename as `FindReusableGroup`. The current name made me think this function would also start the devices within the group.

jemoreira (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/selector/creation_analyzer.cpp
Databean: `AnalyzeCreation` currently considers it a non-fatal case if `expanded.size() > num_instances` and runs `LOG(WARNING)`, while using `CF_EXPECT` on that comparison makes it fatal. I am not prepared to change that behavior right now.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
Databean: I would prefer to submit first with more restrictive checks, and grant additional functionality later if this becomes a problem.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.h
Databean: Done.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/create.cpp
Databean: Done.

PR #2729, opened 2026-06-18
Split vulkan check into vulkan loader and vulkan driver checks

jmacnak:
Bug: b/525390484
Test: `cvd create` and inspect launcher.log

### DIFF
--- a/base/cvd/cuttlefish/host/graphics_detector/graphics_detector_vk_loader.cpp
+++ b/base/cvd/cuttlefish/host/graphics_detector/graphics_detector_vk_loader.cpp
@@ -1,0 +1,35 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


base/cvd/cuttlefish/host/graphics_detector/graphics_detector_vk_loader.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include <dlfcn.h>
+
+#include "cuttlefish/host/graphics_detector/expected.h"
+
+namespace gfxstream {
+
+gfxstream::expected<Ok, std::string> PopulateVulkanLoaderAvailability(
+    ::gfxstream::proto::GraphicsAvailability* availability) {
+  void* libvulkan = dlopen("libvulkan.so.1", RTLD_LAZY | RTLD_LOCAL);
```
3405691582, line 27: If libvulkan.so.1 is present in a sandbox but non-standard location, then we would still need to fiddle with LD_LIBRARY_PATH to have this work, right?

jmacnak: Correct.

3405691582: It would be nice not to have to fiddle with LD_LIBRARY_PATH to deal with this, but I can't think of a good way to do this off the top of my head other than defining a few good paths, which isn't flexible, or using Linux-specifc functionality.

3405691582 (approved):

PR #2735, opened 2026-06-22
Restart monitor view when files are deleted.

Databean:
When cvd create `--daemon=false` runs, the monitor view can open the log files from the previous run. The newly running device deletes the old log files and starts writing new ones, but the monitor holds references to the old files, and waits forever for them to update.

This change adds delete event monitoring to the inotify watches, and reads through through the inotify event list to scan for delete events. Delete events trigger resetting the monitor view state so it can re-open the log files and create new watches.

Bug: b/526626064
Test: launch device, exit, launch new device


rmuthiah (approved):

Databean:
Thanks for reviewing!

PR #2736, opened 2026-06-22
Apply misc-include-cleaner to cuttlefish/common/transport

Databean:
Also uses more `const` and less `auto`.

Bug: b/523396865
Test: bazel test //...


cjreynol (approved):

Databean:
Thanks for reviewing!

PR #2748, opened 2026-06-24
Podcvd supports cache subcommand

0405ysj:
`cache` is the last piece for supporting all subcommands under podcvd. However, it was a bit complicated to support this functionality, as it wasn't available to create hard links for cached artifacts since directories were under different filesystems(separated bind mounts).

So this PR suggests to add copy-based fallback mechanism when creating hard links is not applicable, and then support `podcvd cache`. Copy-based fallback has worse performance than hard links, but better performance than not enabling any caching mechanism.

Context: b/491256959

### DIFF
--- a/base/cvd/cuttlefish/common/libs/utils/files.cpp
+++ b/base/cvd/cuttlefish/common/libs/utils/files.cpp
@@ -115,46 +115,89 @@ Result<bool> AreHardLinked(const std::string& source,
           CF_EXPECT(FileInodeNumber(destination)));
 }
 
-Result<std::string> CreateHardLink(const std::string& target,
-                                   const std::string& hardlink,
-                                   const bool overwrite_existing) {
-  if (FileExists(hardlink)) {
-    if (CF_EXPECT(AreHardLinked(target, hardlink))) {
-      return hardlink;
+Result<bool> AreFilesIdentical(const std::string& path1,
+                               const std::string& path2) {
+  struct stat st1;
+  struct stat st2;
+  CF_EXPECTF(stat(path1.c_str(), &st1) == 0, "stat failed for {}", path1);
+  CF_EXPECTF(stat(path2.c_str(), &st2) == 0, "stat failed for {}", path2);
+  if (st1.st_size != st2.st_size) {
+    return false;
+  }
+
+  SharedFD fd1 = SharedFD::Open(path1, O_RDONLY);
+  SharedFD fd2 = SharedFD::Open(path2, O_RDONLY);
+  CF_EXPECTF(fd1->IsOpen(), "Failed to open \"{}\"", path1);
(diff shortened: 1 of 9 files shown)


k311093 (approved):

base/cvd/cuttlefish/common/libs/utils/files.cpp
```diff
@@ -115,46 +115,89 @@ Result<bool> AreHardLinked(const std::string& source,
...
+  CF_EXPECTF(fd2->IsOpen(), "Failed to open \"{}\"", path2);
+
+  char buf1[4096];
+  char buf2[4096];
+  while (true) {
+    auto r1 = fd1->Read(buf1, sizeof(buf1));
+    auto r2 = fd2->Read(buf2, sizeof(buf2));
+    CF_EXPECTF(r1 >= 0, "Read failed for \"{}\"", path1);
+    CF_EXPECTF(r2 >= 0, "Read failed for \"{}\"", path2);
+    if (r1 != r2) {
```
jemoreira, line 140: Identical files don't necessarily always return the same number of bytes when reading. A valid approach would be to read as much as possible until 4096 bytes are read or no more bytes are left to read on both files and then compare.

base/cvd/cuttlefish/common/libs/utils/files.cpp
```diff
@@ -115,46 +115,89 @@ Result<bool> AreHardLinked(const std::string& source,
...
+
+Result<std::string> LinkOrCopy(const std::string& target,
+                               const std::string& destination,
+                               const bool overwrite_existing) {
+  if (FileExists(destination)) {
+    if (CF_EXPECT(AreHardLinked(target, destination))) {
+      return destination;
     }
     if (!overwrite_existing) {
+      if (CF_EXPECT(AreFilesIdentical(target, destination))) {
```
jemoreira, line 161: This shouldn't check that files are identical. Cuttlefish sometimes modifies artifacts (de-sparsing images), but those should not be counted as cache misses.

Also, comparing files by content is costly. Every usage of `CreateHardLink` before and of `LinkOrCopy` now seems to require the `overwrite_existing` behavior, so making this function fail when the file exists and is not a hard link is probably the best approach.

0405ysj: Agree, I misunderstood the role of `overwrite_existing`. I wrote https://github.com/google/android-cuttlefish/pull/2781 as a draft for the fix. Will send you to review this PR. Thanks!

base/cvd/cuttlefish/common/libs/utils/files.cpp
0405ysj: `AreFilesIdentical` will be removed at https://github.com/google/android-cuttlefish/pull/2781 with applying your another [comment](https://github.com/google/android-cuttlefish/pull/2748/changes#r3483261040).

PR #2749, opened 2026-06-24
Disable crosvm sandboxing inside a sandbox.

3405691582:
The sandbox may not have the seccomp policy directory set up; this may be redundant anyway.


Databean:
Part of the original design of [go/cf-snowglobe](http://go/cf-snowglobe) which described host executable sandboxing was that `crosvm` specifically would be excluded from the Cuttlefish sandbox because it provides its own sandbox capabilities. That is invalidated if we're turning it off in the sandbox mode.

3405691582:
I see -- some sandbox environments might need this. I'll convert this to draft for now while I think about other options.

PR #2751, opened 2026-06-24
cvd selectors: Prompt once to select instance

jemoreira:
As opposed to asking for a group first.

It also takes into consideration the --group_name flag to narrow the options when present.

And adds more colors to make the selection options even more obvious:

Screenshot_20260624_150115

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/selector/selector.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/selector/selector.cpp
@@ -38,33 +38,19 @@ namespace cuttlefish {
 namespace selector {
 namespace {
 
-enum class DisplayBehavior {
-  LabelGroup,
-  LabelInstance,
-};
-
-std::string GroupDisplay(const std::vector<LocalInstanceGroup>& groups,
-                         const DisplayBehavior behavior) {
+std::string GroupDisplay(const std::vector<LocalInstanceGroup>& groups) {
   std::stringstream result;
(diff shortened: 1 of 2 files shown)


base/cvd/cuttlefish/host/commands/cvd/cli/selector/selector.cpp
```diff
@@ -90,7 +92,8 @@ Result<LocalInstanceGroup> PromptUserForGroup(
     const InstanceManager& instance_manager) {
   const std::vector<LocalInstanceGroup> groups =
       CF_EXPECT(instance_manager.FindGroups({}));
-  std::cout << GroupDisplay(groups);
+  const TerminalColors colors(isatty(1));
```
cjreynol: Will you update the other `TerminalColors` instantiation to match this one.  Line 97 uses `isatty(2)` for `stderr`.

cjreynol (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/selector/selector.cpp
jemoreira: good catch.

I made it ensure both stdin and stderr are terminals to since it needs to provide and receive information from the user. Used stderr to allowing piping the command's output. Finally, the selection menu is always shown in a tty so I dropped a few calls to `isatty` to avoid the unnecessary ioctls.

cjreynol (approved):

PR #2754, opened 2026-06-25
Way of downloading debs on container should be compatible to trixie

0405ysj:
Postsubmit is broken, as the way of registering apt repository to download debs weren't compatible with trixie. This PR is the fix for that.

Context: b/491256959


k311093 (approved):

0405ysj:
Thanks Jaeman for reporting this issue!

PR #2770, opened 2026-06-26
Configure Kokoro to do podcvd E2E tests

0405ysj:
I originally implemented to trigger podcvd E2E tests from GitHub action, but some limitations were observed.
- (Critical reason) A lot of test cases weren't able to fetch required artifacts. For example, when the branch is git_main.
- GitHub runner has very limited computing resources. I'm expecting that as the root cause of current flakiness of bugreport test. (90 seconds locally, about 300 seconds on GitHub runner)
- Looking forward to make test cases non-exclusive on Kokoro-based podcvd testing, as podcvd can deal with concurrent instance creation.

So I suggest to move podcvd E2E test into Kokoro. GitHub action should be kept during transition period for keeping podcvd stable, and will be cleaned up afterwards.

Context: b/527644057

### DIFF
--- a/tools/buildutils/cw/Containerfile
+++ b/tools/buildutils/cw/Containerfile
@@ -1,15 +1,9 @@
-FROM mirror.gcr.io/library/debian:bookworm-20250811 AS base
+FROM mirror.gcr.io/library/debian:13 AS base
 
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt update -y && apt upgrade -y
 RUN apt install -y sudo devscripts
-
-# Download newer version of golang with keep using bookworm for building debian
-# packages.
-RUN echo "deb http://deb.debian.org/debian bookworm-backports main" > /etc/apt/sources.list.d/backports.list
-RUN apt update
-RUN apt install -t bookworm-backports -y golang
 
 COPY ./tools/buildutils/installbazel.sh /installbazel.sh
 RUN /installbazel.sh && rm /installbazel.sh
--- a/tools/testutils/runcvde2etests.sh
+++ b/tools/testutils/runcvde2etests.sh
@@ -6,18 +6,23 @@ REPO_DIR="$(realpath "$(dirname "$0")/../..")"
 OUTPUT_DIR="$(pwd)"
 CREDENTIAL_SOURCE="${CREDENTIAL_SOURCE:-}"
 
-bazel_test_tag_filter_arg="--test_tag_filters=-requires_gpu"
-while getopts "g" opt; do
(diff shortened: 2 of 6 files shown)


0405ysj:
Marked this PR as draft for now, as yet I cannot trigger Kokoro for podcvd testing.

0405ysj:
Marking as draft for a moment as it seems tools/buildutils/installbazel.sh is broken regardless of this PR.

0405ysj:
Hello all, could I get review on this PR? Now it's ready.

tools/buildutils/cw/Containerfile
```diff
@@ -1,16 +1,10 @@
-FROM mirror.gcr.io/library/debian:bookworm-20250811 AS base
+FROM mirror.gcr.io/library/debian:13 AS base
```
0405ysj, line 1: @ser-io Would upgrading this container image into trixie be okay for now? I vaguely remember I asked few months ago, and now I'm curious if the blocker is still existing or not.

0405ysj:
@jemoreira @jmacnak Gentle ping. Could I get review for this PR? (Sergio seems to be OOO) Or, I'm considering to get review from SEO folks.

tools/testutils/runcvde2etests.sh
```diff
@@ -53,6 +58,20 @@ cd "${REPO_DIR}/e2etests"
 # those tests
 trap gather_test_results EXIT
 
+test_tags="-requires_gpu"
+if [[ "${gpu_enabled}" == true ]]; then
+  test_tags="requires_gpu"
+fi
+podcvd_args=""
+if [[ "${podcvd_enabled}" == true ]]; then
+  test_tags+=",podcvd"
```
jmacnak, line 67: IIUC, `test_tag_filters` is a logical OR for positive tags and AND-NOT for negative tags. 

Should we be making this `test_tags="podcvd"`?

jmacnak (approved):

tools/testutils/runcvde2etests.sh
0405ysj: I think it's okay for the current. From bazel doc, it sounds `--test-tag-filters=a,b,-c,-d` is equal to `(a OR b) AND 
(NOT c) AND (NOT d)`. So it shouldn't have no problem unless doing a test with both enabling podcvd & GPU. This can be mattered if we keep this state for a long time, however my next plan is to define a negative tag named like `podcvd_excluded` and enabling more tests afterwards.

https://bazel.build/docs/user-manual?hl=en#test-tag-filters

PR #2774, opened 2026-06-26
cvd logs: Allow printing logs without --print

jemoreira:
by making `cvd logs foo` equivalent to `cvd logs --print=foo`

Assisted-by: Jetski:Gemini-Next

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/logs.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/logs.cpp
@@ -7,6 +7,7 @@
 #include <cstring>
 #include <iostream>
 #include <memory>
+#include <optional>
 #include <string>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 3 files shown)


base/cvd/cuttlefish/host/commands/cvd/cli/commands/logs.cpp
```diff
@@ -198,12 +206,12 @@ Result<void> CvdLogsHandler::HandlePrint(const CommandRequest& request) {
...
   for (const std::string& filename : log_filenames) {
     const std::string basename = android::base::Basename(filename);
-    if (basename == print_target_flag_) {
+    if (basename == *print_target_flag_) {
       CF_EXPECT(PrintLog(filename, pager_));
       return {};
     }
   }
-  return CF_ERRF("Not found `{}` logs", print_target_flag_);
+  return CF_ERRF("Not found `{}` logs", *print_target_flag_);
```
Databean: Could this use a special error in the case that `print_target_flag_` ends up with a present optional holding an empty string?

Databean: nit: `"Could not find '{}' in the logs directory"` maybe?

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/logs.cpp
jemoreira: done

jemoreira: done

PR #2776, opened 2026-06-26
Ignore ancient builds with higher safe levels

Databean:
When searching for the latest build on a branch for a particular safe level, it is possible to encounter very old builds. This happens if a build used to have more validation, and it since has dropped running that validation.

To find the combination of "latest" and "most safe", the fetcher will now discard builds from consideration if they are more than a week older than the latest build at any safe level.

Bug: b/524908123
Test: Use build example build from b/524908123#comment19

### DIFF
--- a/base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
+++ b/base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
@@ -16,12 +16,17 @@
 #include "cuttlefish/host/libs/web/android_build_api.h"
 
 #include <dirent.h>
+#include <time.h>
 #include <unistd.h>
 
+#include <algorithm>
 #include <chrono>
+#include <iomanip>
 #include <memory>
 #include <optional>
+#include <ranges>
 #include <set>
+#include <sstream>
 #include <string>
 #include <string_view>
 #include <thread>
@@ -31,6 +36,7 @@
 #include <variant>
 #include <vector>
 
+#include "absl/log/check.h"
 #include "absl/log/log.h"
 #include "android-base/file.h"
 #include "json/value.h"
@@ -82,6 +88,15 @@ Result<Json::Value> GetResponseJson(const HttpResponse<Json::Value>& response,
             "Response was successful, but contains error information.  Check "
             "log file for full response.");
   return response.data;
+}
+
+Result<std::chrono::system_clock::time_point> ParseTime(std::string_view str) {
+  std::stringstream stream = std::stringstream(std::string(str));
+  tm time_tm;
+  stream >> std::get_time(&time_tm, "%Y-%m-%dT%H:%M:%S.");
+  CF_EXPECTF(!!stream, "Failed to parse time '{}'", str);
+
+  return std::chrono::system_clock::from_time_t(mktime(&time_tm));
 }
 
 }  // namespace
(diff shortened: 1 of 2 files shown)


base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
```diff
@@ -261,8 +266,24 @@ Result<std::vector<std::string>> AndroidBuildApi::Headers() {
   return headers;
 }
 
+static Result<std::chrono::system_clock::time_point> ParseTime(
```
jemoreira: nit: We usually use anonymous namespaces

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
```diff
@@ -287,9 +308,29 @@ Result<std::optional<std::string>> AndroidBuildApi::LatestBuildId(
...
+    candidates.emplace_back(CandidateBuild{
+        .build_id = build["buildId"].asString(),
+        .creation_time = CF_EXPECT(ParseTime(completion)),
+    });
+  }
+  if (candidates.empty()) {
+    return std::nullopt;
   }
-  return std::nullopt;
+  // Drop candidate builds older than 1 week
```
jemoreira: nit: This comment isn't accurate, it will keep older candidates as long as the latest one is also old.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
```diff
@@ -287,9 +308,29 @@ Result<std::optional<std::string>> AndroidBuildApi::LatestBuildId(
...
+  auto creation = [](const auto& cd) { return cd.creation_time; };
+  std::chrono::system_clock::time_point latest =
+      std::ranges::max(std::views::transform(candidates, creation));
+  auto recent = [latest](const auto& cd) {
+    // NOLINTNEXTLINE(misc-include-cleaner): <chrono>
+    return latest - cd.creation_time < std::chrono::weeks(1);
+  };
+  std::ranges::filter_view recent_candidates(candidates, recent);
```
jemoreira: nit: You can use `find_if` instead of "filter and then get the first one", it does the same thing.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
```diff
@@ -287,9 +308,29 @@ Result<std::optional<std::string>> AndroidBuildApi::LatestBuildId(
...
+  CF_EXPECT(!std::ranges::empty(recent_candidates));
```
jemoreira: nit: This should be a `CHECK` instead of `CF_EXPECT`. There is at least one created at `latest`.

jemoreira (approved):

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
Databean: Changed to an anonymous namespace.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
Databean: Changed to "Drop candidate builds 1 week older than the latest"

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
Databean: Used `std::find_if`.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
Databean: Changed to `CHECK`.

Databean:
Thanks for reviewing!

PR #2781, opened 2026-06-29
Do not check if files are identical before copying cached artifacts

0405ysj:
Unlike creating hardlink, it's expensive to check whether files are identical or not before copying artifacts. With this change, it throws an error rather than check files when copy-based fallback is used(e.g. `podcvd`), `overwrite_existing` is turned off(currently turned on for all cases logically), and restarting CF groups from the same CF host with same artifacts(unlikely happening from `podcvd`).

Context: https://github.com/google/android-cuttlefish/pull/2748/changes#r3483261040

### DIFF
--- a/base/cvd/cuttlefish/common/libs/utils/files.cpp
+++ b/base/cvd/cuttlefish/common/libs/utils/files.cpp
@@ -115,41 +115,6 @@ Result<bool> AreHardLinked(const std::string& source,
           CF_EXPECT(FileInodeNumber(destination)));
 }
 
-Result<bool> AreFilesIdentical(const std::string& path1,
-                               const std::string& path2) {
-  struct stat st1;
-  struct stat st2;
-  CF_EXPECTF(stat(path1.c_str(), &st1) == 0, "stat failed for {}", path1);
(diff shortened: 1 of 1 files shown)


base/cvd/cuttlefish/common/libs/utils/files.cpp
```diff
@@ -158,12 +123,9 @@ Result<std::string> LinkOrCopy(const std::string& target,
       return destination;
     }
     if (!overwrite_existing) {
-      if (CF_EXPECT(AreFilesIdentical(target, destination))) {
-        return destination;
-      }
       return CF_ERRF(
           "Cannot link/copy from \"{}\" to \"{}\", the second file already "
-          "exists and is different from the first",
+          "exists and is not hardlinked to the first",
```
jemoreira: nit: "... is not a hard link to"

jemoreira (approved):

base/cvd/cuttlefish/common/libs/utils/files.cpp
0405ysj: Done!

PR #2791, opened 2026-06-29
web ui: Better handling when device has no audio

jemoreira:
When the device registers with the signaling server it includes information about its supported audio streams. The client uses this information to decide whether to add audio capabilities to the webRTC connection.

When no audio streams are reported by the device, the web UI won't show the volume UP/DOWN buttons and the button the enable/disable audio is "disabled": shows a different cursor, doesn't react to mouse events and displays a message when hovered over.

Screenshot_20260629_115123

Screenshot_20260629_141712

### DIFF
--- a/base/cvd/cuttlefish/host/frontend/webrtc/html_client/js/app.js
+++ b/base/cvd/cuttlefish/host/frontend/webrtc/html_client/js/app.js
@@ -185,11 +185,14 @@ class DeviceControlApp {
 
     this.#showDeviceUI();
 
+    const hasAudio = deviceHasAudio(this.#deviceConnection.description);
+
     // Enable non-ADB buttons, these buttons use data channels to communicate
     // with the host, so they're ready to go as soon as the webrtc connection is
     // established.
     this.#getControlPanelButtons()
         .filter(b => !b.dataset.adb)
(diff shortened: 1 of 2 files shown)


base/cvd/cuttlefish/host/frontend/webrtc/html_client/js/app.js
```diff
@@ -383,22 +386,38 @@ class DeviceControlApp {
     this.#deviceConnection.onStreamChange(stream => this.#onStreamChange(stream));
 
     // Set up audio
+    const hasAudio = deviceHasAudio(this.#deviceConnection.description);
```
Databean: Is it ever possible that a device has a microphone but no speakers or speakers but no microphone? Should this logic separately distinguish "has microphone" and "has speakers"?

Databean (approved):

base/cvd/cuttlefish/host/frontend/webrtc/html_client/js/app.js
jemoreira: It's not impossible, but very unlikely. I did consider it, the problem is that the streams in the device description don't include the direction, so this information is not available until the connection is established, which happens after the buttons are set up. I'd prefer not to handle this case until really necessary (when we have a device with only one).

PR #2804, opened 2026-07-01
Fallback to looking at the parent process cmdline if parent process exe is locked down

Databean:
When the parent process is `sshd` running as root, it seems to be impossible to readlink /proc/getppid()/exe to find what the parent process is. However, in practice it looks like /proc/getppid()/cmdline is available so it is still possible to examine that.

Bug: b/530299540
Test: Run with only fallback logic

### DIFF
--- a/base/cvd/cuttlefish/host/commands/start/main.cc
+++ b/base/cvd/cuttlefish/host/commands/start/main.cc
@@ -15,7 +15,6 @@
 
 #include <unistd.h>
 
-#include <iostream>
 #include <optional>
 #include <sstream>
 #include <unordered_set>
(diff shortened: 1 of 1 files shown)


base/cvd/cuttlefish/host/commands/start/main.cc
```diff
@@ -210,10 +209,18 @@ Result<void> LinkLogs2InstanceDir(
...
-  return exe_path->ends_with("/cvd");
+  if (exe_path.ok()) {
+    return exe_path->ends_with("/cvd");
+  }
+  const std::string cmdline_path = absl::StrCat(ppid_path, "/cmdline");
+  Result<std::string> cmdline_res = ReadFileContents(cmdline_path);
+  CHECK(cmdline_res.ok()) << cmdline_res.error();
+  std::vector<std::string> cmdline = absl::StrSplit(*cmdline_res, '\0');
+  CHECK(!cmdline.empty());
+  return cmdline[0] == "cvd" || cmdline[0].ends_with("/cvd");
```
jemoreira, line 223: This breaks having a symlink to cvd with a different name, for example `~/bin/local_cvd` pointing to a locally built cvd. This use case doesn't work now either because apparently cvd looks at its own name to make decisions on what subcommand is being executed, but landing this check will add to that.

jemoreira (approved):

base/cvd/cuttlefish/host/commands/start/main.cc
Databean: The `readlink` logic would still most likely catch that case and is evaluated first.

PR #2812, opened 2026-07-02
Improve the LaunchSingleInstance in cvd start

SuperStrongDinosaur:
This change improves the robustness of the single-instance boot (LaunchSingleInstance) in `cvd start`:                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     Refine Bypass Validation Checks:                                                                                                                                                                                                          
       - Prevented bypass if the user requests foreground mode (`--daemon=false`), since `LaunchSingleInstance` does not implement foreground log tailing.                                                                                       
       - Prevented bypass if any configuration-altering flags are passed (only `--daemon` is considered safe), ensuring `assemble_cvd` is not skipped when config changes are requested.                                                                                             
       - Enforce that the main instance itself cannot bypass the full boot, and that the main instance must be in `RUNNING` state before a secondary instance is allowed to use the fast-track bypass.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
@@ -323,25 +323,47 @@ Result<std::vector<Flag>> GetCvdInternalStartFlags(
   return flags;
 }
 
+bool CanBypassToSingleInstance(const LocalInstance& instance,
+                               const LocalInstanceGroup& group,
+                               const std::vector<std::string>& subcmd_args) {
+  if (instance.State() != cvd::INSTANCE_STATE_STOPPED) {
+    return false;
+  }
+  if (group.StartTime() == TimeStamp{}) {
+    return false;
+  }
+
+  std::vector<std::string> args_copy = subcmd_args;
+  bool daemon = true;
+  std::vector<Flag> safe_flags = {
+      GflagsCompatFlag("daemon", daemon),
+  };
+  const Result<void> res = ConsumeFlags(safe_flags, args_copy);
+  if (!res.ok() || !daemon || !args_copy.empty()) {
+    return false;
+  }
+
+  const std::vector<LocalInstance>& instances = group.Instances();
+  if (instances.empty()) {
+    return false;
+  }
+  const LocalInstance& main_instance = instances[0];
+  if (instance.Id() == main_instance.Id()) {
+    return false;
+  }
+  if (main_instance.State() != cvd::INSTANCE_STATE_RUNNING) {
+    return false;
+  }
+
+  return true;
+}
+
 }  // namespace
 
 CvdStartCommandHandler::CvdStartCommandHandler(
     InstanceManager& instance_manager)
     : instance_manager_(instance_manager) {
   own_flags_.daemon = true;
-}
-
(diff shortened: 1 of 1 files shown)


base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
```diff
@@ -601,9 +619,11 @@ Result<void> CvdStartCommandHandler::LaunchSingleInstance(
   run_cvd_envs[kAndroidSoongHostOut] = group.HostArtifactsPath();
   run_cvd_envs[kCvdMarkEnv] = "true";
 
+  const cvd_common::Args run_cvd_args = {"--daemon=true"};
```
Databean: `--daemon` is not an argument to `run_cvd`. It's only an argument to `assemble_cvd` and by extension `cvd_internal_start`.

The flag is defined here:
https://github.com/google/android-cuttlefish/blob/b76c144d0071b2e6dc4a39a62d9165cca8902c84/base/cvd/cuttlefish/host/commands/assemble_cvd/flags/daemon.cc#L29

The value of `--daemon` is stored in `CuttlefishConfig`:
https://github.com/google/android-cuttlefish/blob/b76c144d0071b2e6dc4a39a62d9165cca8902c84/base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc#L1029

This is read by `run_cvd`:
https://github.com/google/android-cuttlefish/blob/b76c144d0071b2e6dc4a39a62d9165cca8902c84/base/cvd/cuttlefish/host/commands/run_cvd/boot_state_machine.cc#L266-L268

SuperStrongDinosaur: done, removed this flag and updated description

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
```diff
@@ -325,27 +326,47 @@ Result<std::vector<Flag>> GetCvdInternalStartFlags(
...
-      GflagsCompatFlag("report_anonymous_usage_stats", report_anonymous),
   };
   auto res = ConsumeFlags(safe_flags, args_copy);
-  if (!res.ok()) {
-    return true;
+  if (!res.ok() || !daemon || !args_copy.empty()) {
+    return false;
+  }
+
+  auto instances = group.Instances();
```
Databean: nit: the [style guide](https://google.github.io/styleguide/cppguide.html#Type_deduction) writes

> Use type deduction only if it makes the code clearer to readers who aren't familiar with the project, or if it makes the code safer. Do not use it merely to avoid the inconvenience of writing an explicit type.

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
Databean: Thanks!

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
SuperStrongDinosaur: done

PR #2813, opened 2026-07-03
Update podcvd skill with more richer context

0405ysj:
Context: b/528346866


ikicha (approved):

0405ysj:
Bypassing the merge queue, as this change only affects on podcvd skill, and no checks cares about it.

PR #2814, opened 2026-07-03
Support ps subcommand on podcvd

0405ysj:
Suggested output format of `cvd ps`:
```
$ cvd ps 2>/dev/null
GROUPS   NAMES   ID   STATUS    CREATED               ADB_SERIAL       WEB_ACCESS                                                   
cvd_1    1       1    Running   2026-07-10 07:56:55   localhost:6520   https://localhost:1443/devices/cvd_1-1-1/files/client.html
```

Suggested output format of `podcvd ps`:
```
GROUPS   NAMES   ID   STATUS    CREATED               ADB_SERIAL           WEB_ACCESS
cvd_1    1       1    Running   2026-07-13 01:16:15   192.168.112.1:6520   https://192.168.112.1:11443/devices/cvd_1-1-1/files/client.html
cvd_2    1       1    Running   2026-07-13 01:16:27   192.168.112.2:6520   https://192.168.112.2:11443/devices/cvd_2-1-1/files/client.html
```

Context: b/530777084

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/ps.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/ps.cpp
@@ -50,9 +50,13 @@ usage: cvd ps [--help]
 )";
 
 const std::map<PsColumns, std::string_view> kHeaders = {
-    {PsColumns::kNames, "NAMES"},   {PsColumns::kId, "ID"},
-    {PsColumns::kStatus, "STATUS"}, {PsColumns::kCreated, "CREATED"},
-    {PsColumns::kPorts, "PORTS"},   {PsColumns::kWebAccess, "WEB_ACCESS"},
+    {PsColumns::kGroups, "GROUPS"},
+    {PsColumns::kNames, "NAMES"},
+    {PsColumns::kId, "ID"},
+    {PsColumns::kStatus, "STATUS"},
+    {PsColumns::kCreated, "CREATED"},
+    {PsColumns::kAdbSerial, "ADB_SERIAL"},
+    {PsColumns::kWebAccess, "WEB_ACCESS"},
 };
 
 void PrintTable(const std::map<PsColumns, std::string_view>& headers,
@@ -131,14 +135,14 @@ PsRow CvdPsCommandHandler::InstanceToRow(const LocalInstanceGroup& group,
                                          LocalInstance& instance) const {
   PsRow row;
 
-  row[PsColumns::kNames] =
-      fmt::format("{}/{}", group.GroupName(), instance.Name());
+  row[PsColumns::kGroups] = group.GroupName();
+  row[PsColumns::kNames] = instance.Name();
   row[PsColumns::kId] = std::to_string(instance.Id());
 
   // Fetch live status
   const Result<Json::Value> status_json_res = instance.FetchStatus();
   std::string status_str = "Unknown (Fetch Failed)";
-  std::string ports_str = "-";
+  std::string adb_serial_str = "-";
   std::string web_access_str = "-";
 
   if (status_json_res.ok()) {
@@ -147,7 +151,8 @@ PsRow CvdPsCommandHandler::InstanceToRow(const LocalInstanceGroup& group,
 
     if (instance.IsActive()) {
       if (status_json.isMember("adb_port")) {
-        ports_str = fmt::format("adb:{}", status_json["adb_port"].asInt());
+        adb_serial_str =
+            fmt::format("localhost:{}", status_json["adb_port"].asInt());
       }
       if (status_json.isMember("web_access")) {
         web_access_str = status_json["web_access"].asString();
@@ -157,7 +162,7 @@ PsRow CvdPsCommandHandler::InstanceToRow(const LocalInstanceGroup& group,
 
   row[PsColumns::kStatus] = status_str;
   row[PsColumns::kCreated] = Format(group.StartTime());
-  row[PsColumns::kPorts] = ports_str;
+  row[PsColumns::kAdbSerial] = adb_serial_str;
   row[PsColumns::kWebAccess] = web_access_str;
 
   return row;
--- a/container/src/podcvd/internal/cvd.go
+++ b/container/src/podcvd/internal/cvd.go
@@ -91,16 +91,18 @@ func UpdateCvdGroupJsonRaw(data any, podcvdHomeDir, ipAddr string) {
 	}
 }
 
+func updateIPAndPortString(data, ipAddr string) string {
+	operatorIpAndPortOnHost := fmt.Sprintf("%s:%d", ipAddr, portOperatorHttpsOnHost)
+	for _, host := range []string{"0.0.0.0", "localhost", "127.0.0.1"} {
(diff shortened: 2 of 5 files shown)


rickgonzalez:
Sharing an explanation of this PR I did during dev/test of Lenzon - Hope you find it helpful.
https://www.lenzon.ai/viewer/cmrb01fsy0000jcwzlwurvng2?voice=google-chirp3

0405ysj:
@dxapd Could I get the review for the modification on `cvd ps` as we talked on b/530777084? Now I'm ready for the review. Thanks!

0405ysj:
@k311093 @ikicha Could I get review for the modification on `podcvd` code for `ps` subcommand support? Thanks!

container/src/podcvd/internal/cvd.go
```diff
@@ -91,16 +94,114 @@ func UpdateCvdGroupJsonRaw(data any, podcvdHomeDir, ipAddr string) {
...
+		end := lineLen
+		if i < len(cols)-1 && cols[i+1].start < lineLen {
+			end = cols[i+1].start
+		}
+		rowMap[col.name] = updateIPAndPortString(strings.TrimSpace(line[col.start:end]), ipAddr)
+	}
+	return rowMap
+}
+
+func parsePsOutput(stdout, ipAddr string) ([]map[string]string, []psColumn, error) {
```
dxapd, line 179: Is re-string parsing the human-readable output of `cvd ps` back into a data structure the best way to do this? The purpose of `cvd ps` is to grab all of the groups' information from instance_manager_ and convert it into human-readable output based on kHeaders. This seems to add a redundant step, when you could:
- call `cvd fleet` and use standard JSON parsing to put it into a data structure, and then use `kHeaders` to format your output, 
- or possibly call into instance_manager_ yourself (not sure if this one is possible from podcvd).

But before that, do podcvd commands need to be formatted a specific way? If not, you may be able to just directly feed the output of `cvd ps` into stdout.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/ps.cpp
```diff
@@ -50,9 +51,13 @@ usage: cvd ps [--help]
 )";
```
dxapd, line 50: Have you tested the changes in this file? Could you share what the output of `cvd ps` looks like after the changes?

0405ysj: What I've described at PR description is the output of `cvd ps`. I copied & pasted from my machine after execution, not a hand-written one.

https://github.com/google/android-cuttlefish/pull/2814#issue-4800802660

dxapd: 👍

container/src/podcvd/internal/cvd.go
0405ysj: > But before that, do podcvd commands need to be formatted a specific way? If not, you may be able to just directly feed the output of cvd ps into stdout.

The only requirement is `podcvd ps` should have same format as `cvd ps`. When multiple Cuttlefish instance groups exist, each group is in separated container instances. When executing `cvd ps` from containers, `podcvd ps` needs to parse the data in advance, and then reconstruct it.

Leaving a representative example why not parsing the human-readable output can be problematic, difference on the padding.
```
$ podman exec -it <container_name_for_cvd_1> cvd ps
GROUPS   NAMES   ID   STATUS    CREATED               ADB_SERIAL           WEB_ACCESS
cvd_1    1       1    Running   2026-07-13 01:16:15   192.168.112.1:6520   https://192.168.112.1:11443/devices/cvd_1-1-1/files/client.html
$ podman exec -it <container_name_for_cvd_11> cvd ps
GROUPS    NAMES   ID   STATUS    CREATED               ADB_SERIAL            WEB_ACCESS
cvd_11    1       1    Running   2026-07-13 01:16:27   192.168.112.11:6520   https://192.168.112.11:11443/devices/cvd_2-1-1/files/client.html
```

0405ysj: Also I should have considered potential format changes in the future. In case of having long-running Cuttlefish instance group via podcvd, its container instance has older version of cvd and there can have another running container instance with newer version of cvd perhpas. Even for such case, `podcvd ps` should provide information as possible as `cvd ps` does.

From `podcvd` modification, you could see any header row(e.g. `GROUPS   NAMES   ID ...`) isn't hardcoded there.

dxapd: Yeah, I guess what I want to understand is just this: If I (or someone else) makes a modification to `cvd ps` (e.g. new column, or renaming a column), will that break `podcvd ps`? If so, is there any kind of test that will immediately start failing if someone updates `cvd ps` but forgets to update `podcvd ps`?

Re: Your point about the instance having an older version of cvd, I would say that that's actually a good argument _against_ re-parsing the output of `cvd ps` with regex - You might end up with an incompatibility between what's running on podcvd and some instance running some different, unknown version of cvd.

At that point, (assuming you can't access instance_manager_) it seems better to call `cvd fleet`, parse the json output into a data structure, and then construct the table yourself using kHeaders. That way, since json is standardized and generic, both podcvd and cvd will be aligned on the _semantic_ part (kHeaders), and the machinery on both sides doesn't need to get tangled up - you can output it in whatever format you want. 

Part of my intention with kHeaders was to make it as difficult as possible to make a change in one part of the command, and accidentally break a different piece by forgetting to update it - while building a row, you just need to grab the corresponding data and put it into the necessary kHeader column. 

Ultimately, it's on the podcvd side, so it's up to you. But if we do need another level of parsing logic, I just wanna be sure we've done what we can in terms of how ergonomic it is to add new stuff without breaking things.

0405ysj: There's currently no tests for keeping the behavior of `podcvd ps` as stable. I planned to reuse cvd E2E test for testing podcvd as well.(b/527644057) When it's completed and add a test around `cvd ps`, both `cvd ps` & `podcvd ps` could be kept under the test. But no progress on that yet due to the bandwidth.

I think it's better to couple arguments; 'parsing `cvd ps` with regex' and 'intention of using kHeaders on `cvd ps`'. Because underlying reasons are same. The reason why I decided parsing `cvd ps` rather than grabbing information from `cvd fleet` was to avoid danger of structure change(e.g. adding `GROUPS` column into `cvd ps`) with forgetting corresponding change on `podcvd ps`.

I'm leaning on your suggestion retrieving info from `cvd fleet` if I can have a test immediately, but not now. So I want to submit this PR as-is for now and refactor afterwards when I can keep it stable with tests. (of course, todo comment should be there for a moment)

dxapd: Hmm, sounds reasonable if you've thought of that and add a todo comment then, yeah

dxapd (approved):

container/src/podcvd/internal/cvd.go
0405ysj: Thanks for the review! I also added TODO comment right above of `PrintPsOutputs` function.

PR #2815, opened 2026-07-06
Return a Result<siginfo_t> from Subprocess::Wait

Databean:
The return value was only ever 0 or -1 based on the result of waitid. Using a Result avoids the output argument in favor of using the return value to produce the outcome of the function call.

There were some callers that ignored the return value of Wait that now check it for errors as well.

Bug: b/531812844

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/utils/subprocess_waiter.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/utils/subprocess_waiter.cpp
@@ -39,17 +39,13 @@ Result<siginfo_t> SubprocessWaiter::Wait() {
   CF_EXPECT(!interrupted_, "Interrupted");
   CF_EXPECT(subprocess_.has_value());
 
-  siginfo_t infop{};
-
   interrupt_lock.unlock();
 
   // This blocks until the process exits, but doesn't reap it.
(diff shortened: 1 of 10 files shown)


base/cvd/cuttlefish/host/commands/cvd/utils/subprocess_waiter.cpp
```diff
@@ -38,17 +37,13 @@ Result<siginfo_t> SubprocessWaiter::Wait() {
   CF_EXPECT(!interrupted_, "Interrupted");
   CF_EXPECT(subprocess_.has_value());
 
-  siginfo_t infop{};
-
   interrupt_lock.unlock();
 
   // This blocks until the process exits, but doesn't reap it.
-  auto result = subprocess_->Wait(&infop, WEXITED | WNOWAIT);
-  CF_EXPECTF(result != -1, "Lost track of subprocess pid: {}", StrError(errno));
```
dxapd, line 48: Did you intend to drop these prints?

Databean: Added more detail to the error message in `Subprocess::Wait` to compensate.

dxapd (approved):

Databean:
Thanks for reviewing!

PR #2819, opened 2026-07-07
podcvd: Support trusted TLS for operator on container instances

0405ysj:
Accessing operator living in container instances by chrome browser(or any other browsers perhaps) is so annoying, as certification procedure is not prepared so that the browser displays the warning message in advance of accessing the operator. This is even worse experience than operator working with `cvd`, as the lifecycle of operator is not permanent, but bound to the container's lifecycle.

Context: b/532008913

### DIFF
--- a/container/debian/control
+++ b/container/debian/control
@@ -14,6 +14,7 @@ Depends: acl,
          adb,
          android-sdk-platform-tools-common,
          crun,
+         mkcert,
          podman (>= 4.4),
          yq,
 Pre-Depends: ${misc:Pre-Depends}
--- a/container/debian/cuttlefish-podcvd.cuttlefish-podcvd.init
+++ b/container/debian/cuttlefish-podcvd.cuttlefish-podcvd.init
@@ -33,6 +33,31 @@ podcvd_cidr=${podcvd_cidr:-192.168.112.0/20}
 
 podcvd_ifname="podcvd"
 
+gen_cert() {
+    local CA_DIR="/etc/cuttlefish-podcvd/ca"
+    local PUB_DIR="/etc/cuttlefish-podcvd/pub"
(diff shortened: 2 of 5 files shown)


0405ysj:
@ikicha @k311093  As Jorge is OOO this week, would it be better to get review & submit PR in advance and get some feedbacks later?

ikicha (commented):
Not sure if installing this self signed certifiacte in system root ca and chrome by default is common pattern or okay. Even though we don't install it by default, it becomes just one time consent with new cert, right?

0405ysj:
> it becomes just one time consent with new cert, right?

The answer was no from what I probably saw few days ago, I still need to check this once more.

jemoreira (commented):
IIUC this is creating a new local certificate authority and a certificate for all the IPs podcvd might use, configuring chrome to trust that certificate and sharing the certificate with all podcvd containers. Please correct me if I got something wrong.

Is it possible to make all browsers, not just chrome, trust the certificate? I think it should be possible to install a local certificate authority in a way that everything in the local system trusts it, not just chrome. Then it should not be necessary to add the certificate anywhere since it was signed with the trusted CA and therefore trusted as well. Did you try this approach?

0405ysj:
> IIUC this is creating a new local certificate authority and a certificate for all the IPs podcvd might use, configuring chrome to trust that certificate and sharing the certificate with all podcvd containers. Please correct me if I got something wrong.

That's correct.

> Is it possible to make all browsers, not just chrome, trust the certificate?

Let me check with other browsers on my machines if I can. I've verified only within chrome browser so far. I expect all chromium browser and any other browsers follows `update-ca-certificates`, but not firefox.

jemoreira (approved):

PR #2824, opened 2026-07-08
Attempt suppressing a trace on boot failure.

3405691582:
When a boot fails, we print a trace saying that run_cvd failed. This isn't strictly necessary or informative, so attempt to suppress this by just returning a CF_ERR instead of CF_EXPECT'ing that the executable returned successfully -- because we already know it didn't.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
@@ -448,6 +448,9 @@ Result<void> CvdStartCommandHandler::LaunchDevice(
     // run_cvd processes may be still running in background
     // the order of the following operations should be kept
     CF_EXPECT(CvdResetGroup(group));
+
+    GatherVmBootCompleteMetrics(group);
+    return CF_ERR("Device launch failed.");
   }
   CF_EXPECT(CheckProcessExitedNormally(infop));
   GatherVmBootCompleteMetrics(group);


3405691582:
I'm not sure if this will have the intended effect, because I'm not sure how to trigger this myself, but I'm assuming you'd have a slightly better idea here than I would.

Databean:
I cherry-picked this change and additionally changed `host/commands/start/main.cpp` to immediately `return -1;` and it still produces this error:
```
Device launch failed, cleaning up
5. main.cc:152 | CvdMain | 
4. cvd.cpp:84 | HandleCvdCommand | 
3. cvd.cpp:69 | HandleCommand | 
2. create.cpp:363 | Handle | 
1. create.cpp:129 | StartGroup | 
 | cuttlefish/host/commands/cvd/cli/commands/start.cpp:564
 v Result<void> cuttlefish::CvdStartCommandHandler::LaunchDevice(Command, LocalInstanceGroup &, const cvd_common::Envs &, const CommandRequest &)
Device launch failed.
```
This PR cuts a level off the stack trace, but ultimately still produces the stack trace, which is still relatively noisy.

Databean:
I've uploaded https://www.github.com/google/android-cuttlefish/pull/2826 to hide the stack traces created by `cvd`, it doesn't yet affect stack traces produced by `assemble_cvd` / `run_cvd` / etc.

3405691582:
> I cherry-picked this change and additionally changed `host/commands/start/main.cpp` to immediately `return -1;`

(this was a good idea, I should have just done this.)

> I've uploaded https://www.github.com/google/android-cuttlefish/pull/2826 to hide the stack traces created by `cvd`,

SGTM, I'll close this.

PR #2825, opened 2026-07-08
Start enforcing BUILD.bazel file formatting within bazel on some files

Databean:
We currently have a [github action](https://github.com/google/android-cuttlefish/blob/7e18adc6cb543f4804fffd168e4d86cdfd747425/.github/workflows/presubmit.yaml#L20-L74) that enforces BUILD.bazel file formatting. It would be better if this validation happened within bazel itself, so developers can locally run `bazel test //...` to validate `BUILD.bazel` files without needing to go to GitHub.

This has the downside that `buildifier` will sometimes try to fix files in-place, which does not work and fails with this opaque error:
```
FAIL: //:cf_build_test (Exit 1) (see <path redacted>/test.log)
INFO: From Testing //:cf_build_test:
==================== Test output for //:cf_build_test:
buildifier: open MODULE.bazel: read-only file system
================================================================================
```

Opening `<path redacted>/test.log` does not have more information besides what is shown in the error.

Bug: b/532697459


base/cvd/cuttlefish/bazel/rules.bzl
```diff
@@ -17,14 +17,30 @@ load("@cc_compatibility_proxy//:proxy.bzl", "cc_binary", "cc_library", "cc_test"
...
 LINKOPTS = BUILD_VAR_LINKOPTS
 
+def _cf_build_test_implementation(name, srcs, **kwargs):
+    native.filegroup(
+        name = name + "_starlark_files",
+        srcs = (srcs or []) + ["BUILD.bazel"],
+        tags = ["lint-with-buildifier"],
+    )
+    buildifier_test(
+        name = name,
```
cjreynol: Per our discussion, maybe we append something like "_LINT_TEST" to the name?

Or, if we are going to copy the same name around then why not omit it as an argument entirely?  Name them all `cf_build_lint_test`.

base/cvd/MODULE.bazel
```diff
@@ -62,6 +59,7 @@ include("//cuttlefish/host/commands/append_squashfs_overlay:append_squashfs_over
 include("//cuttlefish/host/commands/vhost_user_input:vhost_user_input.MODULE.bazel")
 include("//cuttlefish/host/commands/vhost_user_media:vhost_user_media.MODULE.bazel")
 include("//toolchain:bazel.MODULE.bazel")
+include("//toolchain:buildifier.MODULE.bazel")
```
cjreynol, line 62: Feel free to do this in a later PR, but let's update the GitHub action to be `buildifier` instead of [`buildozer`](https://github.com/google/android-cuttlefish/blob/7e18adc6cb543f4804fffd168e4d86cdfd747425/.github/workflows/presubmit.yaml#L29).

cjreynol (approved):

base/cvd/cuttlefish/bazel/rules.bzl
Databean: Added `_LINT_TEST`.

base/cvd/MODULE.bazel
Databean: Will do in a different PR.

Databean:
Thanks for reviewing!

PR #2826, opened 2026-07-09
Report `cvd` stack traces failing `main` only to the log file

Databean:
Bug: b/507554710

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/main.cc
+++ b/base/cvd/cuttlefish/host/commands/cvd/main.cc
@@ -39,6 +39,7 @@
 #include "cuttlefish/flag_parser/flag.h"
 #include "cuttlefish/flag_parser/gflags_compat.h"
 #include "cuttlefish/host/commands/cvd/cli/log_files.h"
+#include "cuttlefish/host/commands/cvd/cli/utils.h"
 #include "cuttlefish/host/commands/cvd/cvd.h"
 #include "cuttlefish/host/commands/cvd/utils/common.h"
 #include "cuttlefish/host/commands/cvd/version/version.h"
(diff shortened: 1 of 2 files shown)


base/cvd/cuttlefish/host/commands/cvd/main.cc
```diff
@@ -226,11 +229,17 @@ int main(int argc, char** argv) {
...
   std::vector<std::string> all_args(argv, argv + argc);
 
-  cuttlefish::InitializeLogs(all_args);
+  std::optional<std::string> log_file = cuttlefish::InitializeLogs(all_args);
 
   cuttlefish::Result<void> result = cuttlefish::CvdMain(std::move(all_args));
   if (result.ok()) {
     return 0;
+  } else if (log_file.has_value()) {
+    VLOG(0) << result.error();
```
3405691582, line 238: Probably not for now but: this should be abstracted further (so that logging color behavior is always consistent), and we should honor NO_COLOR.

3405691582 (approved):

base/cvd/cuttlefish/host/commands/cvd/main.cc
Databean: Good point, added to b/507940443.

Databean:
This was failing `create_with_gce_metadata_credentials_test` because the test validates that there is specific content in the error message:

https://github.com/google/android-cuttlefish/blob/cb4277389b0e1108105d19244656777d598b98d8/e2etests/orchestration/create_with_gce_metadata_credentials_test/main_test.go#L63-L69

I've additionally restricted the logic to the scenario where `stderr` is an interactive terminal to hopefully preserve host orchestrator interactions with `cvd`.

FYI @ser-io

PR #2829, opened 2026-07-09
Fix xts test result assertion.

ser-io:
Fail when result is different than "pass".

Bug: b/533039176

### DIFF
--- a/e2etests/cvd/common/common.go
+++ b/e2etests/cvd/common/common.go
@@ -629,7 +629,7 @@ func RunXts(t *testing.T, cuttlefishArgs FetchAndCreateArgs, xtsArgs XtsArgs) {
 				testname := fmt.Sprintf("%s#%s", xtsTestCase.Name, xtsTest.Name)
 				t.Run(testname, func(t *testing.T) {
 					log.Printf("%s result: %s", testname, xtsTest.Result)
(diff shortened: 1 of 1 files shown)


e2etests/cvd/common/common.go
```diff
@@ -629,7 +629,7 @@ func RunXts(t *testing.T, cuttlefishArgs FetchAndCreateArgs, xtsArgs XtsArgs) {
 				testname := fmt.Sprintf("%s#%s", xtsTestCase.Name, xtsTest.Name)
 				t.Run(testname, func(t *testing.T) {
 					log.Printf("%s result: %s", testname, xtsTest.Result)
-					if xtsTest.Result == "failed" {
+					if xtsTest.Result != "pass" {
```
jmacnak: Could you run one of the CTS test suites that has like an `ASSUMPTION_FAILURE`? I just want to make sure that "skipped" tests do not show up as failures.

jmacnak (approved):
(pre-emptively approving assuming no issues or minor changes for skipped tests)

e2etests/cvd/common/common.go
ser-io: Changes the assertion `strings.HasPrefix(xtsTest.Result, "fail")` failing with "fail" and "failed" instances only.

PR #2833, opened 2026-07-09
Don't show monitor errors for a missing `assemble_cvd.log`

Databean:
This file is expected to be missing until the device starts launching.

Bug: b/533115150

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/monitor.cc
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/monitor.cc
@@ -41,6 +41,7 @@
 #include "cuttlefish/host/commands/cvd/cli/utils.h"
 #include "cuttlefish/host/commands/cvd/instances/local_instance.h"
 #include "cuttlefish/host/libs/log_names/log_names.h"
+#include "cuttlefish/io/in_memory.h"
 #include "cuttlefish/io/shared_fd.h"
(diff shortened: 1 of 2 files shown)


base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/monitor.cc
```diff
@@ -165,7 +166,12 @@ Result<void> MonitorLogs(const LocalInstance& instance, SharedFD stop_eventfd) {
       kernel_lines = total_content / 3;
       logcat_lines = total_content - launcher_lines - kernel_lines;
     }
-    display.DrawFile(SharedFdIo(launcher_fd), launcher_name, launcher_lines);
+    if (!FileExists(assemble_log)) {
+      display.DrawFile(*InMemoryIo("Waiting for assemble_cvd.log creation"),
```
cjreynol, line 170: NIT: is this message useful, or can we just leave that file display blank (matching it not existing) until the device is populated?

cjreynol (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/monitor.cc
Databean: Since we're claiming the space already to fit the scrolling pane view, we might as well put a single line in it explaining the current state.

PR #2834, opened 2026-07-10
Move the TerminalColors type to ansi_codes

Databean:
This places it closer to related code.

Bug: b/507940443


cjreynol (approved):

Databean:
Thanks for reviewing!

PR #2836, opened 2026-07-10
Define a `should_color` library and use it

Databean:
Honors NO_COLOR, FORCE_COLOR, and defaults to whether the terminal is
interactive.

Bug: b/533162913

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/main.cc
+++ b/base/cvd/cuttlefish/host/commands/cvd/main.cc
@@ -32,6 +32,7 @@
 #include "android-base/file.h"
 #include "fmt/format.h"
 
+#include "cuttlefish/ansi_codes/should_color.h"
 #include "cuttlefish/ansi_codes/terminal_colors.h"
 #include "cuttlefish/common/libs/utils/environment.h"
 #include "cuttlefish/common/libs/utils/files.h"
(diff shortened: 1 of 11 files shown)


base/cvd/cuttlefish/host/commands/cvd/main.cc
```diff
@@ -168,7 +169,7 @@ Result<void> CvdMain(cvd_common::Args all_args) {
  * is red.
  */
 std::string ColoredUrl(const std::string& url) {
-  if (!isatty(STDERR_FILENO)) {
+  if (!ShouldColorStderr()) {
```
3405691582, line 172: FORCE_COLOR will mean that we emit ANSI color codes to non-ttys. I think this is okay -- force color means force color -- but it also is a little surprising.

3405691582 (approved):
There's more to think about and potentially iterate on here (e.g., bold or reverse video styles under NO_COLOR and the comment about ttys), but this is probably fine for the moment.

base/cvd/cuttlefish/host/commands/cvd/main.cc
Databean: If I understand correctly the intended use case is something like forcing color codes while piping into `less` which would look like a non-tty to our program but which may ultimately still render color codes correctly to the terminal.

Databean:
Thanks for reviewing!

PR #2837, opened 2026-07-10
Remove the `cvd_common::[Args,Envs]` aliases.

Databean:
https://google.github.io/styleguide/cppguide.html#Aliases writes

> Don't put an alias in your public API just to save typing in the implementation; do so only if you intend it to be used by your clients.

> When defining a public alias, document the intent of the new name, including whether it is guaranteed to always be the same as the type it's currently aliased to, or whether a more limited compatibility is intended. This lets the user know whether they can treat the types as substitutable or whether more specific rules must be followed, and can help the implementation retain some degree of freedom to change the alias.

These aliases were defined without documentation, and most of the API of the underlying types was being used.

This change replaces uses of the aliases with their referenced types.

Bug: b/533168205


cjreynol (approved):

Databean:
Thanks for reviewing!

PR #2845, opened 2026-07-10
Add dynamic V4L2 test pattern controls to emulated_camera_mplane

changyeon-jo:
Implement support for dynamic video test patterns and their associated
V4L2 control handlers. Supported patterns include:
- Pattern 0 (Legacy Pulse): Uniform color-cycling.
- Pattern 1 (SMPTE + Bouncing Box): Full color SMPTE bars with an animated
  inverse-color bouncing box overlay, adapted for the 3-sink layout and
  640x480 resolution.
- Pattern 2 (Animated Julia Set): CPU-intensive fractal pattern simulation
  for high-load host load scenarios.

Supports V4L2_CID_TEST_PATTERN and V4L2_CID_IMAGE_PROC_CLASS via the
VirtioMediaIoctlHandler implementation.

Bug: b/490699808


base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -194,24 +173,144 @@ impl VirtioMediaDeviceSession for EmulatedCameraSession {
 impl EmulatedCameraSession {
     fn write_pattern<WY: std::io::Write, WU: std::io::Write, WV: std::io::Write>(
         iteration: u64,
+        pattern: u32,
```
ser-io: Create a new abstraction for multi planar patterns and avoid construct:

```
if pattern == 1 {
} else if pattern == 2 {
} else if pattern == N {
}
```

Wondering whether you can have something like:

```
self.pattern.write()
```

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -194,24 +173,144 @@ impl VirtioMediaDeviceSession for EmulatedCameraSession {
...
-            writer_y.write_all(&[y]).map_err(|_| libc::EIO)?;
-        }
-        for _ in 0..(WIDTH * HEIGHT / 4) {
-            writer_u.write_all(&[u]).map_err(|_| libc::EIO)?;
-        }
-        for _ in 0..(WIDTH * HEIGHT / 4) {
-            writer_v.write_all(&[v]).map_err(|_| libc::EIO)?;
+        let sequence = iteration;
+
+        if pattern == 1 {
```
ser-io: Have one commit per pattern.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -387,15 +469,67 @@ where
...
-// Use an offset for virtio-media custom camera class control id values.
-const CID_OFFSET: u32 = bindings::V4L2_CID_CAMERA_CLASS_BASE + 0x100;
-const CID_LENS_FACING: u32 = CID_OFFSET + 1;
-
 const PIXELFORMAT: u32 = PixelFormat::from_fourcc(b"YM12").to_u32();
 const WIDTH: u32 = 640;
 const HEIGHT: u32 = 480;
 const FRAME_RATE: u32 = 30;
 
+const SMPTE_COLOR_WHITE: (u8, u8, u8) = (180, 128, 128);
```
ser-io: implement each pattern in its own file.

ser-io (changes requested):

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -831,81 +965,205 @@ where
...
+        }
+    }
+
+    fn s_ctrl(
+        &mut self,
+        _session: &mut Self::Session,
+        id: u32,
+        value: i32,
+    ) -> IoctlResult<bindings::v4l2_control> {
+        if id == V4L2_CID_TEST_PATTERN {
```
ser-io: why do you need a control to dynamically change the patterns? can we initialize the camera device with a given pattern at beginning and use such pattern for the lifetime of the device in order to avoid the new control and avoid this complexity.

You can pass the the new pattern as a new flag when creating the new device. You can have something like:
`cvd create --media="type=v4l2_emulated_camera_mplane,pattern=SMPTE". See as an example of how to add a new media property:https://github.com/google/android-cuttlefish/pull/2689/changes/67a401173ef33b686294f6c04ad6690506e1203a

ser-io (changes requested):

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: Done!

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: Thanks for the suggestion. Test pattern generations are separated from this file.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: Done.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: I'd prefer being able to toggle between test pattens without restarting a CF instance or framework service. This is similar to what you can achieve with v4l2-ctl.

changyeon-jo: Sorry, my comment was incomplete - I wanted this to work similar to [vivid](https://www.kernel.org/doc/html/v4.8/media/v4l-drivers/vivid.html), which is a test driver and you can change a test pattern via v4l2-ctl.

ser-io (changes requested):
Please move commit: "Implement real-time FFmpeg host video streaming for emulated_camera_mplane" into its own PR.


Also, create a brand new virtio-media device for video streaming. Don't extend emulated_camera_mplane for this.

changyeon-jo:
@ser-io 

Re:
> Please move commit: "Implement real-time FFmpeg host video streaming for emulated_camera_mplane" into its own PR.

I can separate FFmpeg commit from PR. 

Re:
> Also, create a brand new virtio-media device for video streaming. Don't extend emulated_camera_mplane for this.

May I ask why you prefer having this in separate device? IMHO, it's also "emulating" a camera data stream from a media source and therefore does not sound out of this device's scope.

If you intend to make this as the one strictly generating test patterns, shouldn't we need to rename this like "test_pattern_generator_mplane" or something similar? :)

ser-io:
> @ser-io
> 
> Re:
> 
> > Please move commit: "Implement real-time FFmpeg host video streaming for emulated_camera_mplane" into its own PR.
> 
> I can separate FFmpeg commit from PR.
> 
> Re:
> 
> > Also, create a brand new virtio-media device for video streaming. Don't extend emulated_camera_mplane for this.
> 
> May I ask why you prefer having this in separate device? IMHO, it's also "emulating" a camera data stream from a media source and therefore does not sound out of this device's scope.
> 
> If you intend to make this as the one strictly generating test patterns, shouldn't we need to rename this like "test_pattern_generator_mplane" or something similar? :)

Cuttlefish attaches virtio-media devices using vhost-user protocol, which allows the flexibility to have device emulation per host process. Having said that, the case for streaming emulated frames and using FFmpeg for host video streaming are quite different in terms of how the frames are populated which present a case for different binaries. Given we already have the vhost-user protocol in place, it's cleaner to have two different binaries with their own specific flags.

The current name emulated_camera_mplane is based on existing [EmulatedCamera](https://cs.android.com/android/platform/superproject/main/+/main:hardware/google/camera/devices/EmulatedCamera/) which this device aims to replace.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/pattern/legacy_pulse.rs
```diff
@@ -0,0 +1,26 @@
+use super::{FramePattern, HEIGHT, WIDTH};
+use std::io::Write;
+
+pub struct LegacyPulse;
```
ser-io: why LegacyPulse? can we name it just "Pulse"

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -271,6 +274,8 @@ pub struct EmulatedCamera<Q: VirtioMediaEventQueue, HM: VirtioMediaHostMemoryMap
     /// same time. It will fails if we allow simultaneous sessions to be active, so we need this
     /// artificial limitation to make it pass fully.
     active_session: Option<u32>,
+    /// Currently selected test pattern.
```
ser-io: nit: add the new `current_pattern` property at the end.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -309,6 +315,61 @@ where
     }
 }
 
+const V4L2_CID_IMAGE_PROC_CLASS: u32 = 0x009f0000;
```
ser-io: For an easier review let's have this commit having the new abstraction only and have the follow-up commit handling the new controls. I want to have a clear separation between the two code changes.

ser-io (changes requested):

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
ser-io: I understand, in that case let's first add the capability to start the device with a given pattern, adding a new flag to the v4l2_emulated_camera_mplane binary.  For instance we can do something like:


```
cvd create \
 --media="type=v4l2_emulated_camera_mplane,pattern=FOO" 
 --media="type=v4l2_emulated_camera_mplane,pattern=BAR"
```

After having the commits satisfying this capability then add the commit that allows changing the patterns via v4l2 control.

ser-io: Discard the previous message. I can do this later on. Just be aware that the starting pattern will be determined by a flag later on.

changyeon-jo:
Thanks for explanation, @ser-io. Since this replaces the camera framework team's EmulateCamera, it sounds like you're aiming to emulate the Android camera device's attributes and characteristics, rather than just the data stream. Do you plan to support [FULL](https://developer.android.com/reference/kotlin/android/hardware/camera2/CameraMetadata#info_supported_hardware_level_full) or [LEVEL_3](https://developer.android.com/reference/kotlin/android/hardware/camera2/CameraMetadata#info_supported_hardware_level_3)? :)

Also, will this device's characteristics be configurable similar to hardware/google/camera/devices/EmulatedCamera/hwl/configs/emu_camera_back.json?

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/main.rs
```diff
@@ -24,6 +24,7 @@ use virtio_media::protocol::VirtioMediaDeviceConfig;
 use vm_memory::{GuestMemoryAtomic, GuestMemoryMmap};
 
 mod device;
+pub mod pattern;
```
ser-io: why pub? I don't think it needs to be used by another crate yet.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/pattern/mod.rs
```diff
@@ -14,3 +16,21 @@ pub trait FramePattern: Send + Sync {
         sink_v: WV,
     ) -> Result<(), i32>;
 }
+
+pub enum CameraPattern {
```
ser-io: Why do we need this wrapper? Let's use `Pulse` implementation directly and have the switching logic explicit in the `write_pattern` method.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -197,26 +195,14 @@ impl VirtioMediaDeviceSession for EmulatedCameraSession {
...
-            writer_y.write_all(&[y]).map_err(|_| libc::EIO)?;
-        }
-        for _ in 0..(WIDTH * HEIGHT / 4) {
-            writer_u.write_all(&[u]).map_err(|_| libc::EIO)?;
-        }
-        for _ in 0..(WIDTH * HEIGHT / 4) {
-            writer_v.write_all(&[v]).map_err(|_| libc::EIO)?;
-        }
-        Ok(())
+        use crate::pattern::FramePattern;
```
ser-io: Declare the import (use) with others at the top.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -197,26 +195,14 @@ impl VirtioMediaDeviceSession for EmulatedCameraSession {
...
+        crate::pattern::CameraPattern::Pulse
```
ser-io: nit: Don't need to use full path. I think `CameraPattern::Pulse.write` would be ok.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -298,6 +305,61 @@ where
     }
 }
 
+const V4L2_CID_IMAGE_PROC_CLASS: u32 = 0x009f0000;
```
ser-io: I can see these are standard `const`  https://docs.kernel.org/userspace-api/media/v4l/ext-ctrls-image-process.html. Are not they already defined in the in the Rust v4l2r::bindings port?

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -298,6 +305,61 @@ where
...
+        id: CID_LENS_FACING,
+        type_: bindings::v4l2_ctrl_type_V4L2_CTRL_TYPE_INTEGER,
+        name: "LENS_FACING",
+        minimum: LensFacing::Front as i32,
+        maximum: LensFacing::External as i32,
+        step: 1,
+        default_value: LensFacing::Front as i32,
+        flags: bindings::V4L2_CTRL_FLAG_READ_ONLY,
+    },
+    ControlDef {
```
ser-io: why do you need two different controls to switch current pattern?

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
ser-io, line 1: Add to the commit message the actual v4l2 commands we can now use to play with these new controls.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -152,17 +164,10 @@ impl Buffer {
...
-                };
-                // SAFETY: clock_gettime is a standard POSIX libc call with a valid pointer.
-                unsafe {
-                    libc::clock_gettime(libc::CLOCK_MONOTONIC, &mut ts);
-                }
+                let elapsed = get_start_time().elapsed();
                 self.v4l2_buffer.set_timestamp(bindings::timeval {
-                    tv_sec: ts.tv_sec as bindings::__time_t,
-                    tv_usec: (ts.tv_nsec / 1000) as bindings::__time_t,
+                    tv_sec: elapsed.as_secs() as bindings::__time_t,
```
ser-io: The External Camera HAL is ok with these values as they are always in the future ( from the guest HAL POV) and no longer drop the buffers.  I don't think this trivial sync solution is the way to go at the moment. Let's sync offline to understand the problem you're having.

ser-io (changes requested):
Thanks for changing the commits to self-contained smaller commits. Highly appreciated.

ser-io:
> Thanks for explanation, @ser-io. Since this replaces the camera framework team's EmulateCamera, it sounds like you're aiming to emulate the Android camera device's attributes and characteristics, rather than just the data stream. Do you plan to support [FULL](https://developer.android.com/reference/kotlin/android/hardware/camera2/CameraMetadata#info_supported_hardware_level_full) or [LEVEL_3](https://developer.android.com/reference/kotlin/android/hardware/camera2/CameraMetadata#info_supported_hardware_level_3)? :)
> 
> Also, will this device's characteristics be configurable similar to hardware/google/camera/devices/EmulatedCamera/hwl/configs/emu_camera_back.json?

Correct! It should be [FULL](https://cs.android.com/android/platform/superproject/main/+/main:device/google/cuttlefish/shared/camera/config/standard.mk) however this is still in the very early stages, priority is to migrate existing workflow (mostly from auto targets) that can now use the data streams and the existing external camera hal.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: Yes, they are currently missing in `v4l2r` crate at the moment. We should add them to v4l2r crate eventually.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: This change was applied to address a race condition with a video playback. Excluded from this device for now - let's sync up offline.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: One of them (`V4L2_CID_IMAGE_PROC_CLASS`) is actually a V4L2 'Control Class'. It's a standard read-only V4L2 convention used to group image processing controls together in the menu structure. The actual functional control used to switch the pattern is just the single `V4L2_CID_TEST_PATTERN`.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: Done.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/main.rs
changyeon-jo: `pub` is moved.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/pattern/mod.rs
changyeon-jo: Removed.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: Yeah, this is something I can't guide `jetski`/`gemini` not to do. I don't know why they prefer using fully qualified names. :(

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: Thanks for catching this. I assumed that `cargo fmt`/`rustfmt` would correct this.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
ser-io: I can build using `bindings::V4L2_CID_TEST_PATTERN` and `bindings::V4L2_CID_IMAGE_PROC_CLASS`, I have linux headers version: 6.18. `cat /usr/include/linux/version.h`. which version do you have?

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/pattern/mod.rs
ser-io, line 1: When introducing new files this repo we need to add the LICENSE at the top.

```
// Copyright 2026, The Android Open Source Project
...
```

Add the license to all the new files introduced in this PR.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -16,10 +16,11 @@ use std::collections::VecDeque;
 use std::io::Result as IoResult;
 use std::io::Seek;
 use std::io::SeekFrom;
-use std::io::Write;
 use std::os::fd::AsFd;
 use std::os::fd::BorrowedFd;
 
+use std::io::Write;
```
ser-io: nit: why this was moved here? it should be with other "std::io" declarations.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -301,6 +310,61 @@ where
...
+    type_: u32,
+    name: &'static str,
+    minimum: i32,
+    maximum: i32,
+    step: u32,
+    default_value: i32,
+    flags: u32,
+}
+
+const CONTROLS: [ControlDef; 3] = [
```
ser-io: Let's remove "CID_LENS_FACING" from this list, CID_LENS_FACING struct is managed by `lens_facing_query_ext_ctrl` method, the new `ControlDef` instance is not aligned with that method.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -301,6 +310,61 @@ where
     }
 }
 
+const V4L2_CID_IMAGE_PROC_CLASS: u32 = 0x009f0000;
+const V4L2_CID_TEST_PATTERN: u32 = 0x009f0903;
+
+struct ControlDef {
```
ser-io: why not use `bindings::v4l2_query_ext_ctrl` directly?

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -301,6 +310,61 @@ where
...
+    },
+    ControlDef {
+        id: V4L2_CID_TEST_PATTERN,
+        type_: bindings::v4l2_ctrl_type_V4L2_CTRL_TYPE_MENU,
+        name: "Test Pattern",
+        minimum: 0,
+        maximum: 0,
+        step: 1,
+        default_value: 0,
+        flags: 0,
```
ser-io: this control is write and read, why not declaring those flags?

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -301,6 +310,61 @@ where
...
+    },
+    ControlDef {
+        id: V4L2_CID_IMAGE_PROC_CLASS,
+        type_: bindings::v4l2_ctrl_type_V4L2_CTRL_TYPE_CTRL_CLASS,
+        name: "Image Processing Controls",
+        minimum: 0,
+        maximum: 0,
+        step: 0,
+        default_value: 0,
+        flags: bindings::V4L2_CTRL_FLAG_READ_ONLY | bindings::V4L2_CTRL_FLAG_WRITE_ONLY,
```
ser-io: can you write to class control?

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -265,6 +271,8 @@ pub struct EmulatedCamera<Q: VirtioMediaEventQueue, HM: VirtioMediaHostMemoryMap
     active_session: Option<u32>,
     /// Lens facing configuration.
     lens_facing: LensFacing,
+    /// Currently selected test pattern.
+    current_pattern: u32,
```
ser-io: Use an enum rather than "u32". Have the first value (0) named undefined as this struct needs to be proto compatible, and defined Pulse as 1.

ser-io (changes requested):

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -830,16 +894,95 @@ where
...
+        let requested_id: u32 = unsafe { std::mem::transmute(id) };
+
+        if let Some(ctrl) = lookup_control(requested_id, flags) {
+            if ctrl.id == CID_LENS_FACING {
                 return Ok(self.lens_facing_query_ext_ctrl());
             }
-        } else if id == CID_LENS_FACING {
-            return Ok(self.lens_facing_query_ext_ctrl());
+
+            let mut name = [0 as ::std::os::raw::c_char; 32];
```
ser-io: define a method in ControlDef for this.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -830,16 +894,95 @@ where
...
+                step: ctrl.step as u64,
+                default_value: ctrl.default_value as i64,
+                flags: ctrl.flags,
+                elem_size: 4,
+                elems: 1,
+                nr_of_dims: 0,
+                dims: [0; 4],
+                ..Default::default()
+            })
+        } else {
```
ser-io, line 1479: Avoid these `else` clauses as they are not helpful in this context.  Return ` return Err(libc::EINVAL);` directly.

ser-io (changes requested):

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: In V4L2, a control implicitly defaults to being Read/Write when the flags field is explicitly 0. Setting standard flags is only strictly necessary when restricting access (like READ_ONLY). I've left this as 0 to align with the core C driver conventions

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: Good catch. The class control should definitely not be writable. I've updated the flags to strictly bindings::V4L2_CTRL_FLAG_READ_ONLY.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: Done

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: Done

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: We need ControlDef because v4l2_query_ext_ctrl uses a C-style byte array ([i8; 32]) for the name field, which cannot be ergonomically or safely initialized from string literals inside a static Rust const array.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: Done

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: Done

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: Done

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/pattern/mod.rs
changyeon-jo: Done

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: Thanks for pushing this back. It was my mistake that I only check v4l2r crate.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/pattern/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::io::Write;
+
+pub const WIDTH: u32 = 640;
```
ser-io: Merge this commit with the next commit Pulse implementation commit: [084680b3deba349e9c046d86ea141aa077585ae7](https://github.com/google/android-cuttlefish/pull/2845/changes/084680b3deba349e9c046d86ea141aa077585ae7) in order to have the trait and its first implementation in a single commit.

This commits has elements that are not needed as a standalone commit but make sense if merged with the next commit, for example: the const declarations, importing `mod pattern` in src/main.rs and Adding "src/pattern/pulse.rs" in the BUILD file which doesn't exist yet.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -21,6 +21,7 @@ use std::io::Write;
 use std::os::fd::AsFd;
 use std::os::fd::BorrowedFd;
 
+use std::str::FromStr;
```
ser-io, line 23: nit: Add `use std::str::FromStr;` to the "std" imports block. It shouldn't be part of the third party imports block.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -199,19 +217,23 @@ impl VirtioMediaDeviceSession for EmulatedCameraSession {
 impl EmulatedCameraSession {
     fn write_pattern<WY: Write, WU: Write, WV: Write>(
         iteration: u64,
+        pattern: TestPattern,
```
ser-io: Let's be consistent here, let's use "pattern: Pattern" or "test_pattern: TestPattern"

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -830,16 +923,79 @@ where
...
-        let id: u32 = unsafe { std::mem::transmute(id) };
-        // If V4L2_CTRL_FLAG_NEXT_CTRL present returns the first control with a higher ID.
+        let requested_id: u32 = unsafe { std::mem::transmute(id) };
+
         if flags.contains(QueryCtrlFlags::NEXT) {
-            if id < CID_LENS_FACING {
+            if requested_id < CID_LENS_FACING {
+                return Ok(self.lens_facing_query_ext_ctrl());
+            }
+            if let Some(ctrl) = CONTROLS.iter().find(|ctrl| ctrl.id > requested_id) {
```
ser-io: Let's rather implement a new method for each control, the same as we currently do for lens_facing_query_ext_ctrl.

The idea of a lookup in a list and then doing `to_v4l2_query_ext_ctrl` where we have

```

        let (max_val, default_val) = if self.id == bindings::V4L2_CID_TEST_PATTERN {
            (1, 1) // Pulse is 1
        } else {
            (self.maximum as i64, self.default_value as i64)
        };
```

is not extensible. When adding a new control we shouldn't be forced to dig into and modify the body of an existing function.

ser-io (dismissed):

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/pattern/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// advertised. A short write would leave stale bytes in the guest buffer.
+    fn assert_plane_sizes(pattern: &dyn FramePattern) {
+        for iteration in [0u64, 1, 255, 4096] {
+            let mut plane_y = Vec::new();
+            let mut plane_u = Vec::new();
+            let mut plane_v = Vec::new();
+            pattern
+                .write(iteration, &mut plane_y, &mut plane_u, &mut plane_v)
+                .expect("writing into a Vec cannot fail");
+            assert_eq!(plane_y.len(), (WIDTH * HEIGHT) as usize);
```
chihchiachen: Now emulated_camera_mplane supports resolution configuration in the runtime. So we should test all supported resolutions here.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/pattern/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+/// and held as a `&dyn FramePattern`.
+pub trait FramePattern {
+    /// Writes the frame identified by `iteration` into the three plane sinks.
+    ///
+    /// Implementations must write exactly `WIDTH * HEIGHT` bytes to `sink_y` and
+    /// `WIDTH * HEIGHT / 4` bytes to each of `sink_u` and `sink_v`, which is the plane
+    /// size the device advertised to the guest.
+    ///
+    /// Returns a raw `errno` value on failure.
+    fn write(
```
chihchiachen, line 33: Now `emulated_camera_mplane` supports resolution configuration in the runtime. So write() should take W/H as args.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/pattern/pulse.rs
```diff
@@ -0,0 +1,43 @@
...
+        &self,
+        iteration: u64,
+        width: u32,
+        height: u32,
+        sink_y: &mut dyn Write,
+        sink_u: &mut dyn Write,
+        sink_v: &mut dyn Write,
+    ) -> Result<(), i32> {
+        let sequence = iteration;
+        let y = (sequence % 256) as u8;
```
chihchiachen: You'd need to pass controls to write() and apply the gain to y.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -886,7 +870,12 @@ where
         let buffer = host_buffer.v4l2_buffer.clone();
 
         if session.streaming {
-            session.process_queued_buffers(&mut self.evt_queue, &self.controls, self.width, self.height)?;
+            session.process_queued_buffers(
```
chihchiachen: I saw lots of reformatting changes that are irrelevant to this commit.

Can you group all of them into commit 807fcb7b3c335df4ab6b6676145f1e540a85fd3b?

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -1277,12 +1473,55 @@ where
...
+                || event.id == bindings::V4L2_CID_IMAGE_PROC_CLASS)
         {
             Ok(())
         } else {
             Err(libc::EINVAL)
         };
     }
 }
+
+#[cfg(test)]
```
chihchiachen, line 1813: This file is already bulky.
Do you think we should move the unit tests into a separate file (say test.rs)? (Feel free to push back.)

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/pattern/pulse.rs
changyeon-jo: Done. Updated `FramePattern::write` to accept `controls: &CameraControls` and applied gain scaling to luma in `Pulse::write`. Added a unit test `pulse_applies_gain_scaling` to verify.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: Done. Re-generated Commit 1 (`Run cargo fmt on baseline emulated_camera_mplane`) so all formatting diffs across the baseline codebase are grouped together in Commit 1.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: Done. Extracted the device unit tests into `src/device/tests.rs` and added it to `BUILD.bazel`.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
changyeon-jo: If you referred to changes renaming "session" as"_session", they are to fix unused variable warnings. I prefer keeping them to get a clean build outcome but also don't mind removing them. :)

chihchiachen: Quite the opposite, I prefer to fix all unused variable warnings.

I saw "session" is not used in g_fmt(), s_fmt(), and try_fmt() and should be renamed to "_session".
Should we fix them in commit 807fcb7 as well?

changyeon-jo: I prefer to keep https://github.com/google/android-cuttlefish/commit/807fcb7b3c335df4ab6b6676145f1e540a85fd3b as pure `cargo fmt` commit. This is subtle but, if you prefer, we can create a separate commit.

chihchiachen: OK - let's address those warnings later, in a separate commit. (No rush!)

changyeon-jo: Update: Separated this into two distinct baseline commits:
1. Commit 1 (`Run cargo fmt on baseline emulated_camera_mplane`): Pure formatting diffs across baseline files.
2. Commit 2 (`Clean up unused session parameters and obsolete constants in emulated_camera_mplane`): Prefixes unused `session` parameters with `_session` and removes obsolete constants (`WIDTH`, `HEIGHT`, `default_fmt`).

The subsequent feature commits are now completely free of extraneous diffs.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -1360,19 +1589,21 @@ where
...
         // Process controls. Class controls are write-only headers and must fail on read.
-        for (_idx, ctrl) in ctrl_array.iter_mut().enumerate() {
+        for (idx, ctrl) in ctrl_array.iter_mut().enumerate() {
             match ctrl.id {
-                bindings::V4L2_CID_USER_CLASS | bindings::V4L2_CID_CAMERA_CLASS => {
-                    ctrls.error_idx = ctrls.count;
+                bindings::V4L2_CID_USER_CLASS
+                | bindings::V4L2_CID_CAMERA_CLASS
+                | bindings::V4L2_CID_IMAGE_PROC_CLASS => {
+                    ctrls.error_idx = idx as u32;
```
chihchiachen: This and L1628 seems to cause v4l2-compliance regression:

[According to the Linux V4L2 specification]
For VIDIOC_G_EXT_CTRLS, if a control is write-only, the driver must fail the operation with EACCES and set error_idx to count (indicating that the get operation is invalid for this control type), not the 0-based array index idx.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/BUILD.bazel
```diff
@@ -15,6 +15,7 @@ rust_binary(
     name = "emulated_camera_mplane_binary",
     srcs = [
         "src/device.rs",
+        "src/device/tests.rs",
```
chihchiachen: Sorry for pointing you to the wrong direction.
Moving out the unit tests doesn't look as great as I thought.
Can we move the unit tests back to device.rs?

To reduce the size of device.rs, I can refactor it into a few files later.
For now, I think it's better the collocate the test cases and the tested code.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/pattern/smpte.rs
```diff
@@ -0,0 +1,138 @@
...
+];
+
+/// SMPTE color bars overlaid with an inverse-color box bouncing around the frame.
+pub struct SmpteBars;
+
+impl FramePattern for SmpteBars {
+    fn write(
+        &self,
+        iteration: u64,
+        _controls: &CameraControls,
```
chihchiachen, line 52: Is it reasonable to ignore `gain` for SMPTE? If yes, can you add a comment to provide the justification?

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/pattern/julia_set.rs
```diff
@@ -0,0 +1,81 @@
...
+const PERIOD_FRAMES: u64 = (std::f32::consts::TAU / ANGLE_STEP) as u64;
+
+/// Animated Julia set, deliberately expensive to render.
+pub struct JuliaSet;
+
+impl FramePattern for JuliaSet {
+    fn write(
+        &self,
+        iteration: u64,
+        _controls: &CameraControls,
```
chihchiachen, line 48: Is it reasonable to ignore `gain` for JuliaSet? If yes, can you add a comment to provide the justification?

chihchiachen:
One v4l2-compliance test case still fails. 
Please check out the kokoro logs and fix the regression.
Thanks!

chihchiachen (approved):

PR #2847, opened 2026-07-11
Create a //cuttlefish/process library

Databean:
The logic in `common/libs/utils/subprocess*` is substantial enough for a dedicated directory.

Bug: b/533572315


base/cvd/cuttlefish/process/BUILD.bazel
```diff
@@ -6,13 +6,29 @@ package(
...
+        "@abseil-cpp//absl/strings",
+    ],
+)
+
 cf_cc_library(
     name = "managed_stdio",
     srcs = ["managed_stdio.cc"],
     hdrs = ["managed_stdio.h"],
     deps = [
+        ":command_subprocess",
```
cjreynol: NIT: It is not applied universally, but we have generally been sticking to full target paths

cjreynol (approved):

base/cvd/cuttlefish/process/BUILD.bazel
Databean: Good catch, this was updated by `buildozer` and I missed it. Updated to use an absolute path.

Databean:
Thanks for reviewing!

PR #2852, opened 2026-07-14
Support linting with depend-on-what-you-use and default-enable it

Databean:
This overlaps with layering_check (already enabled) but additionally validates that all build dependencies are actually necessary. Unlike other linters, errors are reported at `bazel build` rather than `bazel test` time.

Validation is enabled by default on all targets, and disabled for targets that are currently non-compliant.

Bug: b/534499070


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #2867, opened 2026-07-17
Move RecursivelyRemoveDirectory out of utils:files

Databean:
This chips away at making utils:files smaller.

Bug: b/536109448

### DIFF
--- a/base/cvd/cuttlefish/common/libs/utils/files.cpp
+++ b/base/cvd/cuttlefish/common/libs/utils/files.cpp
@@ -24,7 +24,6 @@
 #include <dirent.h>
 #include <errno.h>
 #include <fcntl.h>
-#include <ftw.h>
 #include <libgen.h>
 #include <sched.h>
 #include <stddef.h>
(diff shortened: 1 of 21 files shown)


base/cvd/cuttlefish/common/libs/utils/BUILD.bazel
```diff
@@ -269,6 +269,18 @@ cf_cc_library(
     ],
 )
 
+cf_cc_library(
```
cjreynol: If we are going to cut down this file, can we put it in something like `cuttlefish/files`.  I find it useful knowing where to go for certain classes of operations.

cjreynol (approved):

base/cvd/cuttlefish/common/libs/utils/BUILD.bazel
Databean: Moved to `//cuttlefish/files:recursively_remove_directory`.

Databean:
Thanks for reviewing!

PR #2872, opened 2026-07-19
Enable trixie backports

grandpaul:
We enable backports on Trixie so that we can install newer kernel. 
Also we fix the unit tests because the download scripts needs to be updated.

### DIFF
--- a/gigabyte-ampere-cuttlefish-installer/preseed/preseed.cfg
+++ b/gigabyte-ampere-cuttlefish-installer/preseed/preseed.cfg
@@ -299,7 +299,7 @@ d-i partman-partitioning/default_label string gpt
 # (default: false).
 d-i apt-setup/cdrom/set-first boolean false
 # Backport selection
-# d-i apt-setup/services-select multiselect backports
+d-i apt-setup/services-select multiselect backports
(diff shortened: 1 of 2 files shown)


gigabyte-ampere-cuttlefish-installer/preseed/preseed.cfg
```diff
@@ -299,7 +299,7 @@ d-i partman-partitioning/default_label string gpt
 # (default: false).
 d-i apt-setup/cdrom/set-first boolean false
 # Backport selection
-# d-i apt-setup/services-select multiselect backports
+d-i apt-setup/services-select multiselect backports
```
dimorinny, line 302: By doing so, it becomes much harder to predict which kernel version we'll end up with, given how frequently backports are updated. For example, we might make an unrelated change to the installer, but simply rebuilding it later could trigger an unexpected major kernel upgrade. In fact, I booted a QEMU VM using this installer and it actually installed kernel 7.0.13 (since 11 Jul 2026), rather than 6.18 as expected.

I'm still figuring out what a good approach would look like, but in the meantime, do you think this is a real concern?

dimorinny (changes requested):

PR #2873, opened 2026-07-20
Add support for Guest Lavapipe GPU mode

jmacnak:
Based on @valentineburley arsp/4898478.

This introduces a couple of new guest config values to determine if
various vulkan apexes are supported in order to work around the fact
that Cuttlefish at HEAD currently has both gfxstream and swiftshader
in a single apex (aosp/2818460). After this change is released, the
lavapipe vulkan apex can be introduced and the pre-existing gfxstream
plus swiftshader apex can be split.

Bug: b/530070181
Bug: b/536966381

```
Test: bazel run //cuttlefish/package:cvd -- create \
      (on main with pre-existing vulkan apex)

Test: bazel run //cuttlefish/package:cvd -- create \
      --gpu_mode=guest_swiftshader
      (on main with pre-existing vulkan apex)

Test: bazel run //cuttlefish/package:cvd -- create \
      --gpu_mode=gfxstream_guest_angle
      (on main with pre-existing vulkan apex)

Test: bazel run //cuttlefish/package:cvd -- create \
      --gpu_mode=gfxstream_guest_angle
      (on main with separated vulkan apexes from ag/41081278)

Test: bazel run //cuttlefish/package:cvd -- create \
      (on main with separated vulkan apexes from ag/41081278)

Test: bazel run //cuttlefish/package:cvd -- create \
      --gpu_mode=guest_lavapipe
      (on main with separated vulkan apexes from ag/41081278)

Test: bazel run //cuttlefish/package:cvd -- create \
      --config_file=spec.json
       (where spec.json specifies older udc image)
```

### DIFF
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/bootconfig_args.cpp
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/bootconfig_args.cpp
@@ -325,6 +325,26 @@ Result<std::unordered_map<std::string, std::string>> BootconfigArgsFromConfig(
               ? "com.android.hardware.graphics.composer.drm_hwcomposer"
               : "com.android.hardware.graphics.composer.ranchu";
 
+  const GpuMode gpu_mode = instance.gpu_mode();
+  if (gpu_mode != GpuMode::None) {
+    if (IsGfxstreamMode(gpu_mode)) {
+      if (instance.has_vulkan_gfxstream_apex()) {
+        bootconfig_args["androidboot.vendor.apex.com.google.cf.vulkan"] =
+            "com.google.cf.vulkan.gfxstream";
+      }
+    } else if (gpu_mode == GpuMode::GuestLavapipe) {
+      if (instance.has_vulkan_lavapipe_apex()) {
+        bootconfig_args["androidboot.vendor.apex.com.google.cf.vulkan"] =
+            "com.google.cf.vulkan.lavapipe";
+      }
+    } else if (gpu_mode == GpuMode::GuestSwiftshader) {
+      if (instance.has_vulkan_swiftshader_apex()) {
+        bootconfig_args["androidboot.vendor.apex.com.google.cf.vulkan"] =
+            "com.google.cf.vulkan.swiftshader";
+      }
+    }
+  }
+
   if (instance.vhal_proxy_server_port()) {
     bootconfig_args["androidboot.vhal_proxy_server_port"] =
         std::to_string(instance.vhal_proxy_server_port());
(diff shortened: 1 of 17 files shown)


base/cvd/cuttlefish/host/commands/assemble_cvd/bootconfig_args.cpp
```diff
@@ -325,6 +325,26 @@ Result<std::unordered_map<std::string, std::string>> BootconfigArgsFromConfig(
...
+        bootconfig_args["androidboot.vendor.apex.com.google.cf.vulkan"] =
+            "com.google.cf.vulkan.lavapipe";
+      }
+    } else if (gpu_mode == GpuMode::GuestSwiftshader) {
+      if (instance.has_vulkan_swiftshader_apex()) {
+        bootconfig_args["androidboot.vendor.apex.com.google.cf.vulkan"] =
+            "com.google.cf.vulkan.swiftshader";
+      }
+    }
+  }
```
Databean, line 346: Should there be some default value for this in bootconfig if `guest_config` doesn't set anything? Or is every build that needs this config set guaranteed to have the guest config member defined?

base/cvd/cuttlefish/host/libs/vm_manager/gem5_manager.cpp
```diff
@@ -244,7 +244,19 @@ Gem5Manager::ConfigureGraphics(
 
   std::unordered_map<std::string, std::string> bootconfig_args;
 
-  if (instance.gpu_mode() == GpuMode::GuestSwiftshader) {
+  if (instance.gpu_mode() == GpuMode::GuestLavapipe) {
+    LOG(INFO) << "We are in Lavapipe mode";
```
Databean: nit: `VLOG(0)`?

Databean (approved):

base/cvd/cuttlefish/host/commands/assemble_cvd/bootconfig_args.cpp
jmacnak: Right now, there is a single apex that has both the gfxstream and swiftshader vulkan drivers. IIUC, you do not need to specify the bootconfig arg when there is only a single option (as we are not currently setting a value AFAICT). The assumption is that if the guest config does not report having the specific apex, you are using an older build which is either pre-aosp/2818460 and does not have a vulkan apex or has only the apex from aosp/2818460. After this PR is released and available, we can submit ag/41081278 to break apart the existing one and add the lavapipe apex.

base/cvd/cuttlefish/host/libs/vm_manager/gem5_manager.cpp
jmacnak: Done.

base/cvd/cuttlefish/host/commands/assemble_cvd/bootconfig_args.cpp
valentineburley: Yes, that was my experience as well. As long as there's only a single vulkan apex, it gets picked automatically. The bootconfig only becomes necessary once a second option is introduced.

PR #2878, opened 2026-07-21
Restructure cvdalloc to use nftables + Test cvdalloc using the host sandbox

dxapd:
This is a huge PR, but thankfully it has two mostly separate logical pieces.

First, the changes to cvdalloc itself. In the process of migrating cvdalloc to use nftables, I thought it would be good to restructure it to enforce rule cleanup using RAII.

I also cleaned up some other stuff in cvdalloc that I found while going along (e.g. sanitizing the environment variables when elevating).

Second, sandboxed testing for cvdalloc. This is building on #2959 to do snapshot-based diff testing of the host's state during cvdalloc runs, so we can more accurately test that cvdalloc makes the right changes to the state of the host system. I moved all of that code into base/cvd to prevent weird dependency issues from cropping up.

These live in separate folders, so they can be reviewed (mostly) separately, but I think they should go in together. Mostly since I think it gives credence to the code itself being correct, but at that point you also need to trust the testing harness :)


dxapd:
Can't add him as a reviewer for some reason, but also tagging @sferrogoo here.

base/cvd/cuttlefish/host/commands/cvdalloc/privilege.cpp
```diff
@@ -157,4 +162,43 @@ int DropPrivileges(uid_t orig) {
...
+
+Result<ScopedPrivileges> ScopedPrivileges::Elevate() {
+  uid_t orig = getuid();
+  // The child processes we exec run with elevated privilege (CAP_NET_ADMIN via
+  // ambient caps) but with AT_SECURE=0, so the dynamic linker won't scrub their
+  // environment for us. Sanitize with an allowlist.
+#if defined(__linux__)
+  // On Linux, only sanitize when this exec actually gained privilege (e.g. via
+  // file caps), as signalled by AT_SECURE.
+  const bool should_sanitize = getauxval(AT_SECURE) != 0;
```
3405691582, line 177: You should probably call this something other than "sanitize", since that has a distinctly different meaning.

3405691582:
It would be better to split this pr up for ease of review, for example, you could land some of the RAII changes incrementally. I can still review like this if you'd like though, I know GitHub does not exactly make this easy...

base/cvd/cuttlefish/host/commands/cvdalloc/privilege.cpp
```diff
@@ -157,4 +162,43 @@ int DropPrivileges(uid_t orig) {
...
+#else
+  // Elsewhere we can't rely on AT_SECURE, so sanitize unconditionally.
+  const bool should_sanitize = true;
+#endif
+  if (should_sanitize) {
+    CF_EXPECTF(clearenv() == 0, "Couldn't clear environment: {}",
+               StrError(errno));
+    CF_EXPECTF(setenv("PATH", kTrustedPath, /*overwrite=*/1) == 0,
```
3405691582, line 185: Why are we substituting the PATH like this?

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -74,82 +116,106 @@ Result<void> Teardown(int id, std::string_view ethernet_bridge_name,
...
   };
 
-  CF_EXPECT(
-      Allocate(id, kCvdallocEthernetBridgeName, kCvdallocWirelessBridgeName));
+  // Allocate resources, then signal run_cvd that the instance is ready.
+  std::vector<NftRule> rules = CF_EXPECT(Allocate(
+      nft, id, kCvdallocEthernetBridgeName, kCvdallocWirelessBridgeName));
   CF_EXPECT(cvdalloc::Post(sock));
 
+  // Block until run_cvd signals that the instance is tearing down.
```
3405691582, line 173: redundant comment.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -74,82 +116,106 @@ Result<void> Teardown(int id, std::string_view ethernet_bridge_name,
...
   absl::Cleanup teardown = [id]() {
     LOG(INFO) << "cvdalloc: teardown started";
-    // TODO: b/471069557 - diagnose unused
     Result<void> unused =
         Teardown(id, kCvdallocEthernetBridgeName, kCvdallocWirelessBridgeName);
   };
 
-  CF_EXPECT(
-      Allocate(id, kCvdallocEthernetBridgeName, kCvdallocWirelessBridgeName));
+  // Allocate resources, then signal run_cvd that the instance is ready.
```
3405691582, line 168: redundant comment.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -74,82 +116,106 @@ Result<void> Teardown(int id, std::string_view ethernet_bridge_name,
...
+  std::vector<NftRule> rules = CF_EXPECT(Allocate(
+      nft, id, kCvdallocEthernetBridgeName, kCvdallocWirelessBridgeName));
   CF_EXPECT(cvdalloc::Post(sock));
 
+  // Block until run_cvd signals that the instance is tearing down.
   LOG(INFO) << "cvdalloc: waiting to teardown";
-
   CF_EXPECT(cvdalloc::Wait(sock, cvdalloc::kSemNoTimeout));
+
+  // Release resources and acknowledge completion.
```
3405691582, line 177: redundant comment.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -74,82 +116,106 @@ Result<void> Teardown(int id, std::string_view ethernet_bridge_name,
...
-  }
+// Removes the static nftables environment, which atomically cleans up every
+// chain and rule the tables contain.
+Result<void> CvdallocTeardown(Nftables& nft) {
+  LOG(INFO) << "cvdalloc: running teardown";
+  CF_EXPECT(TeardownFirewall(nft));
+  return {};
+}
 
+// Allocates per-instance network resources, signals readiness to run_cvd,
```
3405691582, line 155: redundant comment.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -74,82 +116,106 @@ Result<void> Teardown(int id, std::string_view ethernet_bridge_name,
...
-      LOG(ERROR) << "cvdalloc: couldn't drop privileges: " << StrError(errno);
-    }
-  };
-
-  r = BeginElevatedPrivileges();
-  if (r == -1) {
-    return CF_ERRF("Couldn't elevate permissions: {}", StrError(errno));
-  }
-
+  // Release all resources if we bail out before the normal teardown below.
```
3405691582, line 161: redundant comment.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -74,82 +116,106 @@ Result<void> Teardown(int id, std::string_view ethernet_bridge_name,
...
 
-  auto sock = SharedFD::Dup(absl::GetFlag(FLAGS_socket));
-  if (!sock->IsOpen()) {
-    return CF_ERRNO("cvdalloc: socket is closed: " << sock->StrError());
-  }
-  int r = TEMP_FAILURE_RETRY(close(absl::GetFlag(FLAGS_socket)));
-  if (r == -1) {
-    return CF_ERRNO("close: " << StrError(errno));
-  }
+// Removes the static nftables environment, which atomically cleans up every
```
3405691582, line 147: redundant comment.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -74,82 +116,106 @@ Result<void> Teardown(int id, std::string_view ethernet_bridge_name,
...
+Result<SharedFD> AdoptSocket(int socket_fd) {
+  SharedFD sock = SharedFD::Dup(socket_fd);
+  CF_EXPECT(sock->IsOpen(), "cvdalloc: socket is closed: " << sock->StrError());
+  CF_EXPECTF(TEMP_FAILURE_RETRY(close(socket_fd)) != -1, "close: {}",
+             StrError(errno));
+  return sock;
+}
 
-  int id = absl::GetFlag(FLAGS_id);
+// Installs the static nftables environment (tables, chains, and the static
```
3405691582, line 139: redundant comment.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -74,82 +116,106 @@ Result<void> Teardown(int id, std::string_view ethernet_bridge_name,
...
-
-Result<int> CvdallocMain(int argc, char* argv[]) {
-  std::vector<char*> args = absl::ParseCommandLine(argc, argv);
-
-  if (absl::GetFlag(FLAGS_id) == 0 || absl::GetFlag(FLAGS_socket) == 0) {
-    Usage();
-    /* No need to dump a trace for usage. */
-    return 1;
-  }
+// Adopts the socket file descriptor passed on the command line: duplicates it
```
3405691582, line 129: redundant comment.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
```diff
@@ -39,33 +42,72 @@
...
 ABSL_FLAG(int, id, 0, "Id");
 ABSL_FLAG(int, socket, 0, "Socket");
+ABSL_FLAG(bool, setup, false,
+          "Set up nftables tables, chains, and static bridge NAT rules");
+ABSL_FLAG(bool, teardown, false, "Tear down nftables tables");
 
 namespace cuttlefish {
 namespace {
 
+// The mutually exclusive modes in which cvdalloc can run.
```
3405691582, line 52: Have we changed something behaviorally about cvdalloc? I don't understand what these "modes" are about. Previously, the lifecycle of cvdalloc was such that it would match the lifecycle of the instance and manage its resources accordingly. Has that changed?

3405691582 (commented):
As mentioned, you should split this up, but I have some quick comments to start off with if you like.

dxapd:
> It would be better to split this pr up for ease of review, for example, you could land some of the RAII changes incrementally. I can still review like this if you'd like though, I know GitHub does not exactly make this easy...

Happy to split this up (or make it easier to review in any other way) if you have any suggestions on how to do so, but I wasn't sure how to split it up without leaving cvdalloc in an odd incremental state. (and at this point It might require a significant amount of work, but I'm not sure. so i'm happy to make reviewing easier in any other way as well e.g. writing something up)

If you want to look only the cvdalloc part (and not look at the testing part), you can also minimize the "host_tests" folder. That part of the work is logically separate (though I put it in the same PR for the reason I mentioned in the OP)

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
dxapd: It's not so much cvdalloc that's changed so much as nftables is a bit different from iptables. In iptables, everybody manages the same data structures in the kernel. On the other hand, with nftables you need to manage your own tables and chains. This needs to be done at install/uninstall time, and I wanted that code to be in cvdalloc (to for example centralize the names of related constants), + separated from the work that the host init script does.

base/cvd/cuttlefish/host/commands/cvdalloc/privilege.cpp
dxapd: What is the distinctly different meaning? I'm using sanitize here in the same vein as "sanitizing input".

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
dxapd: This one can be removed.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
dxapd: this one can also be removed.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
dxapd: this one can also be removed.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
dxapd: this one too

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
dxapd: this one too.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
dxapd: this one too.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
dxapd: this one too.

base/cvd/cuttlefish/host/commands/cvdalloc/cvdalloc.cpp
dxapd: i think i'd like to keep this comment. the other ones are indeed redundant, but this function takes an extra split second to parse at a glance.

base/cvd/cuttlefish/host/commands/cvdalloc/privilege.cpp
dxapd: mostly hygiene, since cvd is using execvpe() to execute a privileged binary. the appropriate hygiene is also done on the other end (searching for the nft etc. binaries in hardcoded paths), so it's not a huge problem.

3405691582:
> > It would be better to split this pr up for ease of review, for example, you could land some of the RAII changes incrementally. I can still review like this if you'd like though, I know GitHub does not exactly make this easy...
> 
> Happy to split this up (or make it easier to review in any other way) if you have any suggestions on how to do so, but I wasn't sure how to split it up without leaving cvdalloc in an odd incremental state. (and at this point It might require a significant amount of work, but I'm not sure. so i'm happy to make reviewing easier in any other way as well e.g. writing something up)

One way, just on cursory inspection: 1. introduce your Nftables library and the testing infrastructure to drive against it 2. introduce a new nftables-oriented alloc driver that's basically the same as the netlink driver, but with nftables for the IptableConfig change, 3. privilege changes/RAII 4. bridge changes in cvdalloc (they seem behavioral) 5. other changes for enabling nftables (detecting alloc driver) 

There might be finer-grained or alternative ways of doing it, I didn't look too finely at things. This also then has the advantage of having a switch on the alloc driver, so the inner cvdalloc behaviors can be switched too if needed.

This shouldn't be too much work, but it does require some git slicing-and-dicing, which you may or may not feel comfortable with doing. I generally prefer to land changes that can be rebased into main as well (this changeset would need to be squashed).

PR #2883, opened 2026-07-22
Add button for adding device panel vertically into view panel

k311093:
In the multiple device scanerio with landscape device screen, stacking device panel horizontally is not a good experience to use. (it need to scroll page right side to see the device).

Add a toggle button to adding device panel vertically to support this case.

b/534778982

### DIFF
--- a/frontend/src/operator/webui/src/app/app.component.html
+++ b/frontend/src/operator/webui/src/app/app.component.html
@@ -2,6 +2,14 @@
   <mat-toolbar class="cloud-android-toolbar">
     <button mat-icon-button id="device-pane-toggle" (click)="snav.toggle()"><mat-icon>menu</mat-icon></button>
     <h1 class="cloud-android-title">Cloud Android</h1>
+    <mat-slide-toggle
+      color="primary"
+      [checked]="(displaysService.addVertically$ | async)!"
+      (change)="displaysService.toggleAddVertically($event.checked)"
+      class="add-vertically-toggle"
+      >
+      Add Device Panel<br />Vertically
+    </mat-slide-toggle>
   </mat-toolbar>
 
   <mat-sidenav-container class="drawer-container">
--- a/frontend/src/operator/webui/src/app/view-pane/view-pane.component.ts
+++ b/frontend/src/operator/webui/src/app/view-pane/view-pane.component.ts
@@ -14,6 +14,7 @@ import {
   KtdGridLayout,
   KtdGridLayoutItem,
   ktdTrackById,
+  ktdGridCompact,
 } from '@katoid/angular-grid-layout';
 import {DisplayInfo} from '../../../../intercept/js/server_connector'
 
@@ -237,16 +238,19 @@ export class ViewPaneComponent implements OnInit, OnDestroy, AfterViewInit {
     item: DeviceGridItem,
     layout: DeviceGridItem[]
   ): DeviceGridItem {
-    const lastMaxY =
-      layout.length !== 0 ? Math.max(...layout.map(obj => obj.y)) : 0;
-    const lastRowLayout = layout.filter(obj => obj.y === lastMaxY);
-    const lastMaxX =
-      layout.length !== 0
-        ? Math.max(...lastRowLayout.map(obj => obj.x))
-        : -item.w;
+    const lastItemY = Math.max(...layout.map(obj => obj.y), 0);
+    const bottomBorder = Math.max(...layout.map(obj => obj.y + obj.h), 0);
+    const lastRowLayout = layout.filter(obj => obj.y === lastItemY);
+    const rightBorder = Math.max(...lastRowLayout.map(obj => obj.x + obj.w), 0)
 
-    item.x = lastMaxX + item.w * 2 <= this.cols ? lastMaxX + item.w : 0;
-    item.y = lastMaxX + item.h * 2 <= this.cols ? lastMaxY : lastMaxY + item.h;
+    if (this.displaysService.addVertically$.value
+        || rightBorder + item.w > this.cols) {  // Stack vertically
+      item.x = 0;
+      item.y = bottomBorder;
+    } else {
+      item.x = rightBorder;
+      item.y = lastItemY;
+    }
 
     item.placed = true;
 
@@ -294,8 +298,22 @@ export class ViewPaneComponent implements OnInit, OnDestroy, AfterViewInit {
         }
         layoutById.set(id, item);
     });
+
     return Array.from(layoutById, ([, value]) => value);
-  }, []), map(items => items.filter(item => item.visible)));
+  }, []), map(items => {
+    const visibleItems = items.filter(item => item.visible);
+    return this.packItems(visibleItems);
+  }));
+
+  private packItems(items: DeviceGridItem[]): DeviceGridItem[] {
+    const vertCompacted = ktdGridCompact(items, 'vertical', this.cols);
+    const bothCompacted = ktdGridCompact(vertCompacted, 'horizontal', this.cols);
+    const compactedMap = new Map(bothCompacted.map(item => [item.id, item]));
+    return items.map(item => {
+      const compacted = compactedMap.get(item.id);
+      return compacted ? { ...item, ...compacted } : item;
+    });
+  }
 
   forceShowDevice(deviceId: string) {
     this.displaysService.onDeviceDisplayInfo({
--- a/frontend/src/operator/webui/src/app/app.component.scss
+++ b/frontend/src/operator/webui/src/app/app.component.scss
@@ -11,14 +11,33 @@
 }
 
 .cloud-android-toolbar {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  position: relative;
   background: #073042;
   color: #eee;
   box-shadow: 0px 4px 10px rgba(0, 0, 0, 0.5);
-  text-align: center;
 }
 
 .cloud-android-title {
-  margin: 0 auto;
+  position: absolute;
+  left: 50%;
+  transform: translateX(-50%);
+  margin: 0;
+}
+
+.add-vertically-toggle {
+  margin-left: auto;
+  font-size: 13px;
+  line-height: 1.2;
+  color: #fff;
+
+  ::ng-deep .mdc-label {
+    color: #fff;
+    white-space: normal;
+    text-align: left;
+  }
 }
 
 .drawer-container {
--- a/frontend/src/operator/webui/src/app/app.component.ts
+++ b/frontend/src/operator/webui/src/app/app.component.ts
@@ -13,7 +13,7 @@ import {BUILD_VERSION} from '../environments/version'
 })
 export class AppComponent {
   readonly version = BUILD_VERSION;
-  constructor(private displaysService: DisplaysService) {}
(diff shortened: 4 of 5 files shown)


frontend/src/operator/webui/src/app/app.component.html
```diff
@@ -2,6 +2,14 @@
...
     <button mat-icon-button id="device-pane-toggle" (click)="snav.toggle()"><mat-icon>menu</mat-icon></button>
     <h1 class="cloud-android-title">Cloud Android</h1>
+    <mat-slide-toggle
+      color="primary"
+      [checked]="(displaysService.addVertically$ | async)!"
+      (change)="displaysService.toggleAddVertically($event.checked)"
+      class="add-vertically-toggle"
+      >
+      Add Device Panel<br />Vertically
+    </mat-slide-toggle>
```
0405ysj, line 12: Do you have any screenshots for it? I cannot intuitively imagine how operator WebUI looks like.

k311093: The button looks like this

image

If it is enabled, device tile would be added vertically instead of horizontally.

frontend/src/operator/webui/src/app/view-pane/view-pane.component.ts
```diff
@@ -237,16 +237,25 @@ export class ViewPaneComponent implements OnInit, OnDestroy, AfterViewInit {
...
+      const lastMaxY =
+        layout.length !== 0 ? Math.max(...layout.map(obj => obj.y)) : 0;
+      const lastRowLayout = layout.filter(obj => obj.y === lastMaxY);
+      const lastMaxX =
+        layout.length !== 0
+          ? Math.max(...lastRowLayout.map(obj => obj.x))
+          : -item.w;
+
+      item.x = lastMaxX + item.w * 2 <= this.cols ? lastMaxX + item.w : 0;
+      item.y = lastMaxX + item.h * 2 <= this.cols ? lastMaxY : lastMaxY + item.h;
```
0405ysj: `lastMaxY` instead of `lastMaxX`?

frontend/src/operator/webui/src/app/view-pane/view-pane.component.ts
```diff
@@ -237,16 +237,25 @@ export class ViewPaneComponent implements OnInit, OnDestroy, AfterViewInit {
...
-        ? Math.max(...lastRowLayout.map(obj => obj.x))
-        : -item.w;
-
-    item.x = lastMaxX + item.w * 2 <= this.cols ? lastMaxX + item.w : 0;
-    item.y = lastMaxX + item.h * 2 <= this.cols ? lastMaxY : lastMaxY + item.h;
+    if (this.displaysService.addVertically$.value) {
+      const lastMaxY =
+        layout.length !== 0
+          ? Math.max(...layout.map(obj => obj.y + obj.h))
+          : 0;
```
0405ysj: I'm not certain what `lastMaxX` or `lastMaxY` are representing.

My naive thought around x,y,w,h is:
- left: x
- right: x+w
- top: y
- bottom: y+h

However I can see `y+h` & `y` is used together for calculating `lastMaxY`.

frontend/src/operator/webui/src/app/view-pane/view-pane.component.ts
```diff
@@ -237,16 +237,25 @@ export class ViewPaneComponent implements OnInit, OnDestroy, AfterViewInit {
...
+    } else {
+      const lastMaxY =
+        layout.length !== 0 ? Math.max(...layout.map(obj => obj.y)) : 0;
+      const lastRowLayout = layout.filter(obj => obj.y === lastMaxY);
+      const lastMaxX =
+        layout.length !== 0
+          ? Math.max(...lastRowLayout.map(obj => obj.x))
+          : -item.w;
+
+      item.x = lastMaxX + item.w * 2 <= this.cols ? lastMaxX + item.w : 0;
```
0405ysj: As following comment of the above comment, I don't understand how the calculation `item.w * 2` is coming from.

I can only make a rough guess that it could be implemented with only assuming homogeneous(e.g. 2 identical CF phone instances) scenario. But probably heterogeneous scenario(e.g. 1 CF phone + 1 CF tablet) should be considered too?

URL for above comment: https://github.com/google/android-cuttlefish/pull/2883/changes#r3635124511

frontend/src/operator/webui/src/app/view-pane/view-pane.component.ts
```diff
@@ -294,7 +303,21 @@ export class ViewPaneComponent implements OnInit, OnDestroy, AfterViewInit {
         }
         layoutById.set(id, item);
     });
-    return Array.from(layoutById, ([, value]) => value);
+
+    const updatedLayout = Array.from(layoutById, ([, value]) => value);
+
+    if (this.displaysService.addVertically$.value) {
```
0405ysj: Why only vertical display needs sort?

0405ysj (commented):
I'm not certain if I'm the right person to review, as I don't have much experience on Angular.

frontend/src/operator/webui/src/app/app.component.html
0405ysj: Thanks for providing the screenshot!

frontend/src/operator/webui/src/app/view-pane/view-pane.component.ts
k311093: `* 2` is coming from zoom ratio (0.5) of initially added device.

Changed to use the constant to calculate it.

frontend/src/operator/webui/src/app/view-pane/view-pane.component.ts
k311093: `lastMaxY` for upper if statement was for Y coordinate for new item.
For `lastMaxY` in the else block it is for the last item's coordination for calculate the place of new item.

Changed names for describe it's usage more.

frontend/src/operator/webui/src/app/view-pane/view-pane.component.ts
k311093: `lastMaxX` was right because the statement was for checking new item exceeds view panel width, but `item.h` was not right. Changed to `item.w`.

frontend/src/operator/webui/src/app/view-pane/view-pane.component.ts
```diff
@@ -237,16 +237,25 @@ export class ViewPaneComponent implements OnInit, OnDestroy, AfterViewInit {
...
+      const lastItemY =
+        layout.length !== 0 ? Math.max(...layout.map(obj => obj.y)) : 0;
+      const lastRowLayout = layout.filter(obj => obj.y === lastItemY);
+      const lastItemX =
+        layout.length !== 0
+          ? Math.max(...lastRowLayout.map(obj => obj.x))
+          : -item.w;
+
+      item.x = lastItemX + item.w * (1.0 / this.defaultDisplayZoom) <= this.cols ? lastItemX + item.w : 0;
+      item.y = lastItemX + item.w * (1.0 / this.defaultDisplayZoom) <= this.cols ? lastItemY : lastItemY + item.h;
```
0405ysj: For better visibility, please consider to remove `ternary operator` with complicated/duplicated condition.

Or, what about the code looks like below?

```
maxBottom= ...
if (this.displaysService.addVertically$.value) {
  stackVertical = true
} else {
  maxRight = ...
  stackVertical = maxRight + item.w * (1.0 / this.defaultDisplayZoom) <= this.col
}
if (stackVertical) {
  item.x = 0
  item.y = maxBottom + item.h
} else {
  item.x = maxRight + item.w
  item.y = maxBottom
}
```

k311093: Refactored code for better readability.

frontend/src/operator/webui/src/app/view-pane/view-pane.component.ts
k311093: It does not needs to be, but meanwhile, found a problem that panels collides and go far away from its original opsition. Added packItems() which packs vertical, horizontal for panels to avoid that problem.

frontend/src/operator/webui/src/app/view-pane/view-pane.component.ts
0405ysj: Thanks for the explanation!

frontend/src/operator/webui/src/app/view-pane/view-pane.component.ts
0405ysj: Seems this comment is outdated, marking as resolved.

frontend/src/operator/webui/src/app/view-pane/view-pane.component.ts
```diff
@@ -237,16 +238,31 @@ export class ViewPaneComponent implements OnInit, OnDestroy, AfterViewInit {
...
+        ? Math.max(...lastRowLayout.map(obj => obj.x + obj.w))
+        : 0;
 
-    item.x = lastMaxX + item.w * 2 <= this.cols ? lastMaxX + item.w : 0;
-    item.y = lastMaxX + item.h * 2 <= this.cols ? lastMaxY : lastMaxY + item.h;
+    if (this.displaysService.addVertically$.value) {
+      stackVertical = true;
+    } else {
+      stackVertical =
+        layout.length !== 0 && rightBorder + item.w > this.cols;
```
0405ysj: `layout.length !== 0` is not necessary here, as there should be no diff whether stacking horizontal or vertical when no item exists. So I think we can shrink like:

```
stackVertical = this.displaysService.addVertically$.value && rightBorder + item.w > this.cols;
```

frontend/src/operator/webui/src/app/view-pane/view-pane.component.ts
```diff
@@ -237,16 +238,31 @@ export class ViewPaneComponent implements OnInit, OnDestroy, AfterViewInit {
     item: DeviceGridItem,
     layout: DeviceGridItem[]
   ): DeviceGridItem {
-    const lastMaxY =
+    let stackVertical = false;
+    const lastItemY =
       layout.length !== 0 ? Math.max(...layout.map(obj => obj.y)) : 0;
```
0405ysj: nit: `const lastItemY = Math.max(...layout.map(obj => obj.y), 0);`?

0405ysj (approved):

frontend/src/operator/webui/src/app/view-pane/view-pane.component.ts
k311093: Removed unnecessary condition for the statement.

frontend/src/operator/webui/src/app/view-pane/view-pane.component.ts
k311093: Removed ternary operators for this block and changed to use default value.

frontend/src/operator/webui/src/app/view-pane/view-pane.component.ts
0405ysj: Would we still need conditional statement?

k311093: It could be simplified if we remove `stackVertical`. Removed and merged into one if statement.

PR #2885, opened 2026-07-22
Fix file already exist error while running substitute in Docker

k311093:
In the OverlayFS, even if unlink() is succeeded, looks like there is a delay between unlink call and the file is actually disappears.

Change to use rename call which ensures atomic operation for substitution.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
+++ b/base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
@@ -17,7 +17,9 @@
 
 #include <unistd.h>
 
+#include <cerrno>
 #include <optional>
+#include <random>
 #include <string>
 #include <string_view>
 #include <vector>
@@ -26,6 +28,7 @@
 #include "absl/strings/match.h"
 #include "absl/strings/strip.h"
 #include "android-base/file.h"
+#include "fmt/format.h"
 #include "google/protobuf/text_format.h"
 
 #include "cuttlefish/common/libs/utils/environment.h"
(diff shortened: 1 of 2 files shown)


base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
```diff
@@ -63,10 +65,18 @@ Result<void> Substitute(const std::string& target,
 
   CF_EXPECT(EnsureDirectoryExists(android::base::Dirname(full_link_name)));
 
-  int unlink_res = unlink(full_link_name.c_str());
-  CF_EXPECTF(unlink_res == 0 || errno == ENOENT, "{}", StrError(errno));
+  std::random_device rd;
+  const std::string tmp_link_name =
+      fmt::format("{}.tmp_{}_{}", full_link_name, getpid(), rd());
```
Databean: Can this use [`SharedFD::Mkstemp`](https://github.com/google/android-cuttlefish/blob/5b9b17e395a89e2f8c4e68940802b1d2d92e119a/base/cvd/cuttlefish/common/libs/fs/shared_fd.cpp#L545) (internally [`mkstemp`](https://man7.org/linux/man-pages/man3/mkstemp.3.html)? My understanding is that `glibc` implements this with some retry logic in case of collisions: https://codebrowser.dev/glibc/glibc/sysdeps/posix/tempname.c.html#try_tempname_len

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
Databean: On second thought, creating a temporary file doesn't help as you have to `unlink` it to `symlink` it anyway, coming back to the original problem.

Might help to implement some retry logic here instead of using `mkstemp`, but not necessary.

k311093: Thanks for the suggestion! Added retry loop when creating link is failed with EEXIST.

beroklove7-prog (commented):
https://source.android.com/docs/whatsnew/site-updates?year=2026#aosp-changes

PR #2886, opened 2026-07-22
1.55.1 update

rmuthiah:


ser-io (approved):

base/debian/changelog
```diff
@@ -1,3 +1,10 @@
+cuttlefish-common (1.55.1) unstable; urgency=medium
+
+  * Start development on v1.55.1 by @rmuthiah in
```
Databean: Should the other commits be mentioned?

Databean (approved):

base/debian/changelog
rmuthiah: Good point, done.

PR #2887, opened 2026-07-22
Enable multi queue emulation when needed for hwui on skiavk

jmacnak:


base/cvd/cuttlefish/host/graphics_detector/graphics_detector_vk.cpp
```diff
@@ -45,6 +45,21 @@ gfxstream::expected<Ok, vk::Result> PopulateVulkanAvailabilityImpl(
     for (const auto& ext : exts) {
       outPhysicalDevice->add_extensions(std::string(ext.extensionName));
     }
+
+    const auto queueFamilies = physicalDevice.getQueueFamilyProperties();
+    for (const auto& queueFamily : queueFamilies) {
+      auto* outQueueFamily = outPhysicalDevice->add_queue_families();
+      if (queueFamily.queueFlags & vk::QueueFlagBits::eCompute) {
```
ser-io: check this [TotW #141](https://g3doc.corp.google.com/devtools/library_club/g3doc/totw/141.md). wondering whether we could rather do.

```
outQueueFamily->set_supports_compute((queueFamily.queueFlags & vk::QueueFlagBits::eCompute) != 0);
```

base/cvd/cuttlefish/host/commands/assemble_cvd/graphics_flags.cc
```diff
@@ -751,11 +773,21 @@ Result<void> SetGfxstreamFlags(
...
   }
 
-  // SwiftShader currently only supports a single queue. SkiaVK requests
-  // a second queue used for transfers.
-  if (gpu_mode == GpuMode::GfxstreamGuestAngleHostSwiftshader &&
-      hwui_renderer == GuestHwuiRenderer::kSkiaVk) {
-    features["VulkanVirtualQueue"] = true;
+  if (hwui_renderer == GuestHwuiRenderer::kSkiaVk) {
+    // SkiaVK requires a second graphics queue used for AHB transfers.
+    bool needs_multi_queue_emulation = false;
```
ser-io: I think using ternary operator makes things easier to read

```
bool needs_multi_queue_emulation = gpu_mode == GpuMode::GfxstreamGuestAngleHostSwiftshader ? true : !HasMultipleGraphicsQueues(availability);
```

ser-io (changes requested):

base/cvd/cuttlefish/host/commands/assemble_cvd/graphics_flags.cc
jmacnak: Updated.

base/cvd/cuttlefish/host/graphics_detector/graphics_detector_vk.cpp
jmacnak: `queueFlags` is a type safe wrapper from https://github.com/KhronosGroup/Vulkan-Headers/blob/d314eae73fdc90847bb8304a86ee7c6a8ee023b6/include/vulkan/vulkan_enums.hpp#L26 which doesn't support the int comparison. It has the bool conversion operator as explicit so switched to the explicit cast to bool.

ser-io (approved):

PR #2889, opened 2026-07-22
Enforce depend-on-what-you-use for allocd

Databean:
Bug: b/534499070


3405691582 (approved):

Databean:
Thanks for reviewing!

PR #2892, opened 2026-07-23
Integrate notifications into MonitorSource

Databean:
This moves setting up inotify logic into FileMonitorSource and simplifies the monitor loop a bit. It also prepares for alternate MonitorSource implementations that are not based on files.

Bug: b/537428390


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #2893, opened 2026-07-23
Enforce depend-on-what-you-use on cuttlefish/process

Databean:
Bug: b/534499070


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #2895, opened 2026-07-23
Test for logspam

jmacnak:


e2etests/cvd/common/common.go
```diff
@@ -166,7 +166,7 @@ type FetchAndCreateArgs struct {
 }
 
 // Performs `cvd fetch <args>`.
-func (tc *TestContext) CVDFetch(args FetchArgs) error {
+func (tc *TestContext) CVDFetch(args FetchArgs) (CommandOutput, error) {
```
Databean, line 169: https://google.github.io/styleguide/go/decisions#returning-errors writes

> If a function returns an error, callers must treat all non-error return values as unspecified unless explicitly documented otherwise.

A lot of the call sites look like
```
	if res, err := tc.CVDFetch(cuttlefishArgs.Fetch); err != nil {
		t.Fatalf("cvd fetch failed with %v, stderr:%s", err, res.Stderr)
	}
```

It would be valid here to document the properties of `res` during an error, but what do you think of instead including the relevant parts of `res` into `err` either here or in `RunCmd` so that all the callers don't need to duplicate the error handling?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
```diff
@@ -521,7 +521,7 @@ Result<void> CvdStartCommandHandler::Handle(const CommandRequest& request) {
 
   listener_handle.reset();
 
-  if (isatty(0)) {
+  if (isatty(0) || own_flags_.force_terminal_based_output) {
```
Databean: I'm surprised that `isatty(0)` is `true` in the go e2e tests, but it would also be possible to force it to be not a terminal with `StdinPipe()`.

rmuthiah (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
```diff
@@ -714,6 +714,9 @@ std::vector<Flag> CvdStartCommandHandler::BuildOwnFlags() {
           .Help("Run the start command in the background as a daemon. "
                 "If false, the command runs in the foreground and monitors "
                 "logs."),
+      GflagsCompatFlag("force_terminal_based_output",
+                       own_flags_.force_terminal_based_output)
+          .Help("Force terminal based output (useful for testing)."),
```
Databean: What do you think of defining a flag more closely tied to the relevant behavior, e.g. `--print_instance` or `--print_group` that has a default value set to `isatty(0)`?

Databean (approved):

e2etests/cvd/common/common.go
jmacnak: Switched back to just `t.Fatalf` as the underlying `runCmdWithContextEnv` does not actually populate the `CommandOutput` on error.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
jmacnak: `isatty(0)` is `false` in the e2e tests but the goal is to test effectively the behavior when it is `true`, i.e. we want to test the "human" readable output.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp
jmacnak: WDYT about `print_group_format` with `json`, `human`, or `none` ?

Databean: Sounds good.

Databean (approved):

PR #2897, opened 2026-07-23
Migrate from android::base::expected to tl::expected

Databean:
On today's edition of "why did you do this..."

Cost-benefit analysis:

- (+) Fewer source drift issues from AOSP libbase vs google-internal libbase.
- (+) More of `std::expected` is available, like `and_then`.
- (+) Does not automatically import libfmt headers.
- (-) Have to migrate from `.ok()` extension method to standard `.has_value()`.
- (-) Linting takes longer.
- (?) More extension methods are available.

Reference links:

- https://github.com/TartanLlama/expected
- https://tl.tartanllama.xyz/en/latest/api/expected.html
- https://registry.bazel.build/modules/tl-expected

Bug: b/522805341


jemoreira (approved):

jmacnak (approved):

Databean:
Thanks for reviewing!

I overestimated the linting time, it doesn't look like there is much difference.

PR #2902, opened 2026-07-24
Replace most uses of `Trace()` in tests with result matchers

Databean:
Bug: b/538646001


jmacnak (approved):

Databean:
Thanks for reviewing!

PR #2903, opened 2026-07-24
Ensure kernel requested gets applied

rmuthiah:
Security kernels don't get installed unless dist-upgrade is run.

Also delete the old kernels so when the NVIDIA
drivers get installed, they don't install against
the older kernel.

Bug: b/538688547

### DIFF
--- a/tools/baseimage/pkg/gce/scripts/install_kernel_main.sh
+++ b/tools/baseimage/pkg/gce/scripts/install_kernel_main.sh
@@ -34,7 +34,7 @@ version=$(sudo chroot /mnt/image/ /usr/bin/dpkg -s linux-image-cloud-${arch} | g
 echo "START VERSION: ${version}"
 
 sudo chroot /mnt/image /usr/bin/apt-get update
-sudo chroot /mnt/image /usr/bin/apt-get upgrade -y
+sudo chroot /mnt/image /usr/bin/apt-get dist-upgrade -y
 
 version=$(sudo chroot /mnt/image/ /usr/bin/dpkg -s linux-image-cloud-${arch} | grep ^Depends: | \
   cut -d: -f2 | cut -d" " -f2 )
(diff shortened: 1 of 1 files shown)


tools/baseimage/pkg/gce/scripts/install_kernel_main.sh
```diff
@@ -52,6 +52,20 @@ if [ "${version}" != "${linux_image_deb}" ]; then
   exit 1
 fi
 
+# Remove old kernel packages, keeping only the target kernel and the
```
ser-io, line 55: The NVIDIA driver is installed in the "Create base image" stage meaning it would be run in a freshly booted machine, and not in the same machine this script just ran, hence picking the latest installed kernel.

In order to keep this script simple, can we avoid having this code?

ser-io (changes requested):

tools/baseimage/pkg/gce/scripts/install_kernel_main.sh
```diff
@@ -34,7 +34,7 @@ version=$(sudo chroot /mnt/image/ /usr/bin/dpkg -s linux-image-cloud-${arch} | g
 echo "START VERSION: ${version}"
 
 sudo chroot /mnt/image /usr/bin/apt-get update
-sudo chroot /mnt/image /usr/bin/apt-get upgrade -y
+sudo chroot /mnt/image /usr/bin/apt-get dist-upgrade -y
```
ser-io, line 37: is it there a bug to better understand this change?

ser-io (changes requested):

tools/baseimage/pkg/gce/scripts/install_kernel_main.sh
rmuthiah: Done - b/538688547

tools/baseimage/pkg/gce/scripts/install_kernel_main.sh
rmuthiah: This is correct. However - the legacy kernel will linger on the freshly booted image even if its not the active kernel. Then when the linux image headers version check occurs prior to the nvidia dkms build step - the result is non-deterministic, there will be two kernel headers returned. This is fine if the kernel that is updated to is ABI compatible (e.g. 6.1 bpo 50 to 6.1 bpo 60) but not fine if a major kernel bump is done (6.12 to 6.18+deb13).

ser-io: Sounds good. Let's remove it.

ser-io (approved):

PR #2906, opened 2026-07-24
Hold kernel package during base image builds

rmuthiah:
Needed so that base image updates are only done as needed.

Bug: b/538688546

### DIFF
--- a/tools/baseimage/pkg/gce/scripts/create_base_image_main.sh
+++ b/tools/baseimage/pkg/gce/scripts/create_base_image_main.sh
@@ -26,9 +26,12 @@ sudo apt-get upgrade -y
 # Avoids blocking "Default mirror not found" popup prompt when pbuilder is installed.
 echo "pbuilder        pbuilder/mirrorsite     string  https://deb.debian.org/debian" | sudo debconf-set-selections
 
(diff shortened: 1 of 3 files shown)


tools/baseimage/pkg/gce/scripts/create_base_image_main.sh
```diff
@@ -26,10 +26,7 @@ sudo apt-get upgrade -y
 # Avoids blocking "Default mirror not found" popup prompt when pbuilder is installed.
 echo "pbuilder        pbuilder/mirrorsite     string  https://deb.debian.org/debian" | sudo debconf-set-selections
 
-kmodver_begin=$(sudo chroot /mnt/image/ /usr/bin/dpkg -s linux-image-cloud-${arch} | grep ^Depends: | \
```
ser-io, line 29: No need to remove the assertion, it's our own back up to make sure the kernel is not updated.

ser-io (changes requested):

tools/baseimage/pkg/gce/scripts/create_base_image_main.sh
rmuthiah: Good point, thanks. I've re-added that check.

ser-io (approved):

PR #2908, opened 2026-07-27
gigabyte-ampere-cuttlefish-installer: utils: download-ci-cf.sh: use JSON API

grandpaul:
The previous way to download the artifacts doesn't work. We fix it by using the latest JSON API. And also make the script more verbose on error.

### DIFF
--- a/gigabyte-ampere-cuttlefish-installer/utils/download-ci-cf.sh
+++ b/gigabyte-ampere-cuttlefish-installer/utils/download-ci-cf.sh
@@ -6,9 +6,23 @@ set -o errexit
 
 URL=https://ci.android.com/builds/latest/branches/aosp-android-latest-release/targets/aosp_cf_arm64_only_phone-userdebug/view/BUILD_INFO
 RURL=$(curl -Ls -o /dev/null -w %{url_effective} ${URL})
-echo $RURL
+echo "RURL = ${RURL}"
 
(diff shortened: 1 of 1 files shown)


gigabyte-ampere-cuttlefish-installer/utils/download-ci-cf.sh
```diff
@@ -6,9 +6,23 @@ set -o errexit
...
 URL=https://ci.android.com/builds/latest/branches/aosp-android-latest-release/targets/aosp_cf_arm64_only_phone-userdebug/view/BUILD_INFO
 RURL=$(curl -Ls -o /dev/null -w %{url_effective} ${URL})
-echo $RURL
+echo "RURL = ${RURL}"
 
-FILENAME=$(wget -nv -O - ${RURL%/view/BUILD_INFO}/ | grep aosp_cf_arm64_only_phone-img- | sed 's/.*\(aosp_cf_arm64_only_phone-img-[0-9]*[.]zip\).*/\1/g')
+BUILD_ID=$(echo "${RURL}" | sed -n 's/.*\/builds\/submitted\/\([^\/]*\)\/.*/\1/p')
+echo "BUILD_ID = ${BUILD_ID}"
+
+if [ x"${BUILD_ID}" = x"" ]; then
```
3405691582: use [[ (https://google.github.io/styleguide/shellguide.html#s6.3-tests)
plus the "x" trick shouldn't be necessary with bash comparisons

rmuthiah (approved):

PR #2917, opened 2026-07-28
gigabyte-ampere-cuttlefish-installer: enable nested virtualization

dimorinny:
* Pins 6.18 kernel and compatible Nvidia driver.
* Appends `kvm.mode=nested` into Kernel command line so server can be booted into `VHE+NV2` when possible.

Bug: 537008147


rmuthiah (approved):

grandpaul (approved):
Looks good to me.

PR #2921, opened 2026-07-29
Add `-e` and `--monitor_verbosity` to `cvd monitor`

Databean:
`cvd monitor` can now filter messages to only show a certain verbosity or higher.

Bug: b/540119286


base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/command_handler.cc
```diff
@@ -95,7 +107,28 @@ bool CvdMonitorCommandHandler::RequiresDeviceExists() const { return true; }
 
 Result<std::string> CvdMonitorCommandHandler::DetailedHelp(
     const CommandRequest& request) {
-  return kDetailedHelpText;
+  std::vector<Flag> flags = CF_EXPECT(Flags(request));
```
jemoreira: nit: Consider overriding `Flags` and `Description` instead of `DetailedHelp` to make the output more consistent with other commands.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/command_handler.cc
```diff
@@ -95,7 +107,28 @@ bool CvdMonitorCommandHandler::RequiresDeviceExists() const { return true; }
...
+    CF_EXPECTF(flags_.verbosity_ != MonitorVerbosity::kUnknown,
+               "Unknown verbosity '{}'", verbosity);
+    return {};
+  };
+  auto set_error = [this](std::string_view) -> Result<void> {
+    flags_.verbosity_ = MonitorVerbosity::kError;
+    return {};
+  };
+  return std::vector<Flag>{
+      Flag::StringFlag("monitor_verbosity")
```
jemoreira: nit: Why not just `verbosity`? `cvd monitor --monitor_verbosity` sounds more verbose than `cvd monitor --verbosity`. After all the flag is still doing roughly the same job as the `--verbosity` flag of other commands.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/file_monitor_source.cc
```diff
@@ -52,49 +51,65 @@ struct FileData {
...
 
-  absl::Cord accumulated_data;
+  std::string accumulated_data;
   size_t offset = ret.total_size;
-  size_t newline_count = 0;
 
-  while (offset > 0 && newline_count < n + 1) {
-    static constexpr size_t kChunkSize = 4096;
+  // TODO(schuffelen): make FileMonitorSource stateful to avoid re-reading
+  while (offset > 0 && ret.total_size - offset < (1 << 24)) {
```
jemoreira: nit: Please use a named constant for this `1<<24` and the `1<<20` below.

jemoreira (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/command_handler.cc
Databean: The `--verbosity` flag is silently being consumed at a different level as a `cvd`-wide flag, so it can't be re-parsed here. This tripped me up for a while.

https://github.com/google/android-cuttlefish/blob/5b9b17e395a89e2f8c4e68940802b1d2d92e119a/base/cvd/cuttlefish/host/commands/cvd/main.cc#L56-L66

Are you suggesting that this should access the `--verbosity` value given to that other flag? There's the `CF_CONSOLE_SEVERITY` / `CF_FILE_SEVERITY` environment variables or there might be some other global state accessible through the values given to `LogToStderrAndFiles`.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/file_monitor_source.cc
Databean: `kChunkSize` is already a named constant, but I can move it outside this function. Also added `kMaximumScrollback` for the other number.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/command_handler.cc
Databean: Done.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/command_handler.cc
jemoreira: I'm saying that `--monitor_verbosity` and `--verbosity` are doing the same thing and therefore should only be one, given that other commands use --verbosity, that's the one we should have.

Accessing value given to the global flag SGTM.

jemoreira (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/monitor/command_handler.cc
Databean: Accessed the global through the environment variable.

Databean:
FYI I've renamed most of what was originally called "verbosity" to "severity" and reused some of the severity logic from `tee_logging.h`.

PR #2425, opened 2026-04-20
Introduce emulated camera device generating single planar (MJPEG) for…

ser-io:
…mat frames.

- Android reference v4l2 Camera HAL only support MJPEG for now https://cs.android.com/android/platform/superproject/main/+/main:hardware/interfaces/camera/device/default/ExternalCameraDevice.cpp;l=38;drc=1fb9ba494c43ee7c2fd4d09a3513f0f98ed5c01f

Test:

```
bazel run emulated_camera_splane:emulated_camera_splane -- --socket-path /tmp/media.sock
```

Run the VM:

```
cargo run -- --log-level=debug run --cpus 4 --mem 4096 \
  --rwdisk /path/to/debian-12.img \
  --params "root=/dev/vda1" \
  --vhost-user media,socket=/tmp/media.sock \
  /path/to/bzImage

```

Run v4l2-compliance:

```
v4l2-compliance -d0 -s

```

Bug: b/502639876

PR #2935, opened 2026-07-30
Add new virtio-media V4l2 stream proxy virtual device

bridadan:
This PR introduces the `v4l2_stream` device, a `vhost-user-media` backend that enables streaming video from a host-side  source (via a FIFO/named pipe) into a guest VM using the `virtio-media` protocol.

It allows video data generated on the host (e.g., from a video file, synthetic test source, or any other producer of raw video frames) to be streamed directly into the guest VM's V4L2 framework.

The first commit copies over the `emulated_camera_mplane` crate into the new `v4l2_stream` crate, and includes minimal changes to get it building and launchable. The second commit modifies the new crate with the new functionality.

b/472497998


ser-io (changes requested):
Let's avoid forking emulated_camera_mplane as the first commit, it makes the second commit very hard to review. Feel free to introduce the new device in the first commit.

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream/src/main.rs
```diff
@@ -49,33 +37,78 @@ struct CmdLineArgs {
     verbosity: log::LevelFilter,
     /// Lens facing configuration: FRONT, BACK, or EXTERNAL.
     #[clap(long, value_name = "LENS_FACING", default_value = "EXTERNAL")]
-    lens_facing: String,
+    lens_facing: Option<String>,
```
ser-io: Don't implement lens_facing for this device now as it's not needed. This would allow us to rmove the relevant ctrl handling methods making the PR easier to review

ser-io (changes requested):

base/cvd/cuttlefish/host/libs/config/media.h
```diff
@@ -33,10 +33,18 @@ constexpr const char kMediaHelp[] =
     "    'v4l2_emulated_camera_mplane': emulated media capture device "
     "(multi-plane)\n"
     "    'v4l2_proxy': proxy a host V4L2 device into the guest\n"
-    "    'v4l2_stream': v4l2 stream device\n"
-    " 'lens_facing': optional, supported values: 'FRONT', 'BACK', 'EXTERNAL'\n"
+    "    'v4l2_stream': stream video from a host named pipe into the guest\n"
```
ser-io: As discussed in the design doc, I think "v4l2_stream_proxy" better describe this device as the device does not generate/create the stream itself; it acts as a passive proxy, reading a pre-existing stream from the host and passing it to the guest VM implementing the v4l2 protocol.

Having said that. I'd also like to suggest new parameter names for this device. The would be:

```
input_path
input_width
input_height
input_fps
input_format
```

The most important concern is Forward Compatibility, if decide to support scaling or format translation in the future (e.g., taking a 1080p MJPEG input and outputting 720p YUV to the guest), we can easily introduce output_width and output_format parameters without any naming conflicts or ambiguity.

I also like that these new names are specific to proxy-like devices and they won't be misused by other devices.

ser-io (changes requested):

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream/test_stream.sh
```diff
@@ -0,0 +1,164 @@
+#!/bin/bash
```
ser-io: Let's implement a proper cvd e2e tests under https://github.com/google/android-cuttlefish/tree/main/e2etests/cvd/media_tests rather than having this bash script.

Also, add the new test in its own commit as the last commit of the PR.

ser-io (changes requested):

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
```diff
@@ -57,6 +66,97 @@ use virtio_media::protocol::V4l2Ioctl;
...
+    pub fn from_str(s: &str) -> Option<Self> {
+        match s {
+            "MJPEG" => Some(Self::Mjpeg),
+            "Z16" => Some(Self::Z16),
+            "YUV420" => Some(Self::Yuv420),
+            "YUV420M" => Some(Self::Yuv420M),
+            "NV12" => Some(Self::Nv12),
+            "NV12M" => Some(Self::Nv12M),
+            _ => None,
+        }
```
ser-io, line 65: Your test generates format YUV420M. Can you support this format for now and add new formats when needed. 

If you need to support all this formats now. In this specific commit can you just support format: YUV420M and add support for other formats in other follow-up commits. I'd like to review this PR as minimal as possible.

ser-io (changes requested):

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
bridadan: Since you only want to support one format for now, would you also like me to remove the `format` option from the CLI?

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream/test_stream.sh
bridadan: Does the test need to be a new folder within `media_tests`, or can it be a new function within the existing v4l2compliance test?

bridadan: And would you prefer I not include this script at all? The interactive mode can be useful outside of testing, but I realize it may be odd to include it here

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
ser-io: yes, that would be definitely better, thanks.

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream/test_stream.sh
ser-io: I see, let's remove the script for now and skip the e2e tests implementation as I think this script is an utility to try-out the new device rather than an e2e test as the existing ones.

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
```diff
@@ -0,0 +1,1152 @@
+// Copyright 2026, The Android Open Source Project
```
ser-io, line 1: Similiar to https://github.com/google/android-cuttlefish/pull/2425/changes/f62529f8b84f5d7d26582c6c6b5379c9cc187f06. Let's have the first commit with the code of new Rust binary only.

The vhost user binaries could be attached to any crosvm VM instance, so it's ok to have the first commit only with code related to the vhost user binary. 

Then, we can have the next commit integrating "cvd" with the new `v4l2_stream_proxy` binary.

ser-io (changes requested):

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
```diff
@@ -0,0 +1,1152 @@
...
+    streaming: bool,
+}
+
+impl VirtioMediaDeviceSession for V4l2StreamSession {
+    fn poll_fd(&self) -> Option<BorrowedFd<'_>> {
+        None
+    }
+}
+
+/// V4l2 stream device used for testing Android camera stack.
```
ser-io, line 235: nit: update this comment tailored to the v4l2_stream_proxy device.

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
```diff
@@ -0,0 +1,1152 @@
...
+        if self.config.format.is_multiplanar() {
+            QueueType::VideoCaptureMplane
+        } else {
+            QueueType::VideoCapture
+        }
+    }
+
+    fn default_fmt(&self, queue: QueueType) -> v4l2_format {
+        let plane_sizes = self.config.format.plane_sizes(self.config.input_width, self.config.input_height);
+        if self.config.format.is_multiplanar() {
```
ser-io: should we support only multiplanar for now. then  make the case for single plane in a different PR?

base/cvd/cuttlefish/host/libs/config/media.h
```diff
@@ -32,10 +32,18 @@ constexpr const char kMediaHelp[] =
...
-    " 'lens_facing': optional, supported values: 'FRONT', 'BACK', 'EXTERNAL'\n"
+    "    'v4l2_stream_proxy': stream video from a host named pipe into the guest\n"
+    "Supported keys for v4l2_emulated_camera_splane, "
+    "v4l2_emulated_camera_mplane, and v4l2_proxy:\n"
+    "    'lens_facing': optional, supported values: 'FRONT', 'BACK', 'EXTERNAL'\n"
+    "Supported keys for v4l2_stream_proxy:\n"
+    "    'input_path': required for v4l2_stream_proxy, path to the host named pipe\n"
+    "    'input_width': required for v4l2_stream_proxy, width of the video stream in pixels\n"
+    "    'input_height': required for v4l2_stream_proxy, height of the video stream in pixels\n"
+    "    'input_fps': required for v4l2_stream_proxy, frames per second (e.g., 30 or 30000/1001)\n"
```
ser-io: as these new properties are only relevant to the new `v4l2_stream_proxy` type. let's group this new fileds into its own block. we can append a new block  like the following:

```
<EMPTY LINE>
"v4l2_stream_proxy properties:\n"
 "    'input_path': [required] path to the host named pipe"
 "    'input_width': [required] width of the video stream in pixels"
 ...
```


You keep the new "v4l2_stream_proxy" entry under the   "Supported types:\n" block.

e2etests/cvd/media_tests/v4l2compliance/main_test.go
```diff
@@ -73,3 +80,104 @@ func TestEmulatedCameraV4l2Compliance(t *testing.T) {
 		})
 	}
 }
+
+func TestV4l2StreamProxyCompliance(t *testing.T) {
```
ser-io: Let's create a new directory for  v4l2 stream proxy device e2e tests. Let's move this new test to file: /media_tests/v4l2_stream_proxy/v4l2_compliance/main_test.go

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
```diff
@@ -0,0 +1,1152 @@
...
+                        plane_idx: 0,
+                        plane_offset: 0,
+                    }
+                } else {
+                    WorkerState::Idle { fifo_file: file }
+                }
+            }
+            Err(e) => {
+                log::error!("Failed to open FIFO: {:?}", e);
+                std::thread::sleep(std::time::Duration::from_millis(500));
```
ser-io: why is this needed? sleep calls lead to flakiness, let's explain why it's needed and whether the timtout value should be part of the "config: &Config"

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
```diff
@@ -0,0 +1,1152 @@
...
+    buffers: Vec<Buffer>,
+    /// Queue of buffers awaiting processing.
+    queued_buffers: VecDeque<usize>,
+    /// Current sequence number of the generated frames.
+    sequence: u32,
+    /// Time when the last frame was completed.
+    last_frame_time: Instant,
+}
+
+enum WorkerCmd {
```
ser-io: let's move `WorkerCmd`, `WorkerHandle`, `WorkerState` and other worker related functionalties into its own file and package. I'd like to have a clear view of the public interface of the `worker` package and how the `VirtioMediaDevice` implementation interacts with it.

ser-io (changes requested):

e2etests/cvd/media_tests/v4l2compliance/main_test.go
bridadan: Do you in general want to keep the emulated tests separated from the stream proxy tests? Would you want to do something like:

- All existing tests, move to `/media_tests/v4l2_emulated_camera`
    - e.g. `/media_tests/v4l2_emulated_camera/cts`, `/media_tests/v4l2_emulated_camera/v4l2_compliance`
- This new test, move to `/media_tests/v4l2_stream_proxy`
    - e.g. `/media_tests/v4l2_stream_proxy/v4l2_compliance`

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
bridadan: As per your previous request, the only format supported is YUV420 multiplanar. I can go ahead remove all of the logic that detects if it is single or multiplanar and just assume everything is multiplanar for now.

base/cvd/cuttlefish/host/libs/config/media.h
bridadan: The current format is:

```
Supported types:
    <all the types>
Supported keys for v4l2_emulated_camera_splane, v4l2_emulated_camera_mplane, and v4l2_proxy:
    lens_facing: ...
Supported keys for v4l2_stream_proxy:
    <all the other ones>
```

To clarify, are you requesting the following format instead?
```
Supported types:
    <all the existing types> 
    'v4l2_stream_proxy': stream video from a host named pipe into the guest
 
v4l2_stream_proxy properties:
    <input_*>
   
Supported keys:
    lens_facing: ...
```

One thing to note: `lens_facing` support in `v4l2_stream_proxy` was removed based on your previous comment. If the second version is what you're requesting, how would you like to note that `lens_facing` isn't supported?

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
```diff
@@ -0,0 +1,1152 @@
...
+
+        if poll_fds[1].revents().unwrap_or(PollFlags::empty()).contains(PollFlags::POLLIN) {
+            let plane_size = plane_sizes[plane_idx];
+            let remaining = plane_size - plane_offset;
+            let read_chunk = std::cmp::min(local_buf.len(), remaining);
+
+            match fifo_file.read(&mut local_buf[..read_chunk]) {
+                Ok(0) => {
+                    log::info!("FIFO EOF, writer disconnected. Re-opening...");
+                    std::thread::sleep(std::time::Duration::from_millis(100));
```
bridadan: Note to self, this sleep should be removed

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
bridadan: The sleep was added here to prevent a fast loop in case of continuous errors, which should keep system load down. But agreed it can lead to inconsistency. I just didn't have a better idea at the time of implementation.

All of the sleeps in this file are in the error path (except for the one I noted below). Perhaps we should just remove them and see how things play out during testing? I agree they cause some concern while reading the code. We can always revisit once we see some real error behavior in the wild.

e2etests/cvd/media_tests/v4l2compliance/main_test.go
ser-io: Tests that are aimed to act on an specific virtio-media device should have it's own directory. In this case we should have `/media_tests/v4l2_emulated_camera_mplane` and `/media_tests/v4l2_stream_proxy` for the v4l2 compliance tests. However the cts tests should stay under `/media_tests/cts` (for now) as they don't target an specific virtio-media device but the phone as a whole. W

In this PR, let's focus only on the new `/media_tests/v4l2_stream_proxy/` directory. I'll take care of `/media_tests/v4l2_emulated_camera_mplane` later.

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
bridadan: "its own file and package".

Do you mean just its own file? Or do you mean I need to create a whole other crate? I would push back on the separate crate as its only used by the v4l2_stream_proxy crate right now.

base/cvd/cuttlefish/host/libs/config/media.h
ser-io: Yes, let's do:

```
Supported types:
    <all the existing types> 
    'v4l2_stream_proxy': stream video from a host named pipe into the guest
 
v4l2_stream_proxy properties:
    <input_*>
  
```

Skip "lens_facing" property in this commit/PR as it was not introduced here and it's confusing. I'll add it later in a follow up PR.

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
ser-io: Not another create. In Rust the term is module, looking forward to encapsulate the worker logic into it's own module separated from the v4l2 device implementation. We can have this module implementation into its own file.

```
mod worker {
}
```

bridadan: Clarified offline that a new file alone is sufficient

tools/testutils/ffmpeg_v4l2_stream_proxy.sh
```diff
@@ -0,0 +1,120 @@
...
+  echo "Detected parameters: Width=$WIDTH, Height=$HEIGHT, FPS=$FPS"
+  
+  FFMPEG_ARGS=(-re -stream_loop -1 -i "$VIDEO_FILE")
+  echo "Streaming from $VIDEO_FILE to $FIFO_PATH..."
+else
+  FFMPEG_ARGS=(-re -f lavfi -i "testsrc=size=${WIDTH}x${HEIGHT}:rate=${FPS}")
+  echo "Streaming testsrc (${WIDTH}x${HEIGHT} @ ${FPS}fps) to $FIFO_PATH..."
+fi
+
+echo "Suggested CVD config: --media=v4l2_stream_proxy:input_path=$FIFO_PATH:input_width=$WIDTH:input_height=$HEIGHT:input_fps=$FPS"
```
ser-io, line 111: To avoid confusion with actual cvd canonical configuration we can say something like:

"Use the following flag to attach the v4l2_stream_proxy device to the CF instance: --"

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/README.md
```diff
@@ -0,0 +1,70 @@
+# V4L2 Stream Proxy Host Tool
+
+This directory contains the `v4l2_stream_proxy` vhost-user device. To test this device, you can use the provided helper script to stream video from the host into the guest.
```
ser-io: nit: the "helper script" is to start the stream. not to stream it from host into the guest.

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/worker.rs
```diff
@@ -0,0 +1,335 @@
...
+) {
+    log::info!("Worker thread started for FIFO: {:?}", config.input_path);
+
+    let plane_sizes = config.format.plane_sizes(config.input_width, config.input_height);
+    let mut local_buf = [0u8; 4096];
+    
+    let mut state = WorkerState::Unopened;
+
+    while !matches!(state, WorkerState::Stopped) {
+        state = state.transition(
```
ser-io: The `handle_*` functions inside the `impl WorkerState {` are static functions, not methods which is confusing. The only method in that `impl` block is `transition` which receive all the parameters to then dispatch relevant parameters the the relevant handle_ function.

To make this code easier to follow let's move the `match` inside  `transition` right here, removing this indirection and then move  all the "handle_*" functions out of the `impl WorkerState {` so it's clear they are static function in this module removing the  `impl WorkerState {` as a whole.

ser-io (changes requested):

ser-io (approved):

tools/testutils/ffmpeg_v4l2_stream_proxy.sh
bridadan: Ah I missed this feedback, I will tackle in a follow up

bridadan:
I removed the e2e test. It was failing because the guest requires enabling the external camera provider hal, which is disabled by default

bridadan:
> I removed the e2e test. It was failing because the guest requires enabling the external camera provider hal, which is disabled by default

This was not correct, there is no dependency on the external camera provider hal since this only tests the v4l2 layer (Linux kernel stuff, not Android HAL). The issue was related to how the e2e tests were executed, which I've now fixed.

PR #2936, opened 2026-07-30
Report on progress of substitution at `cvd fetch` time

Databean:
Bug: b/540960886


jemoreira (approved):
What about substitutions made at start time? Shouldn't those be reported too?

Databean:
Substitutions made at start time invoke this code path as well:
https://github.com/google/android-cuttlefish/blob/05aa7d570c15bb3a864e2fb9c8ad2ba3235203fe/base/cvd/cuttlefish/host/commands/cvd/cli/commands/start.cpp#L415-L416

PR #2937, opened 2026-07-30
Add a "--host_substitution=only" mode

Databean:
This skips downloading `cvd-host_package.tar.gz` and only downloads the img-*.zip file. The files normally provided from `cvd-host_package.tar.gz` are instead loaded from `/usr/lib/cuttlefish-common` or the equivalent from the development directory.

It is not yet possible to boot a device from only these files.

Bug: b/540976680

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/fetch/symlink_host_package.cc
+++ b/base/cvd/cuttlefish/host/commands/cvd/fetch/symlink_host_package.cc
@@ -1,0 +1,71 @@
+//
+// Copyright (C) 2026 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


base/cvd/cuttlefish/host/commands/cvd/fetch/symlink_host_package.cc
```diff
@@ -0,0 +1,72 @@
...
+      return {};
+    }
+    std::string_view local_path(path);
+    CF_EXPECTF(absl::ConsumePrefix(&local_path, common_dir),
+               "Unexpected prefix in : '{}'", local_path);
+
+    const std::string full_target_path =
+        absl::StrCat(target_dir, "/", local_path);
+
+    CF_EXPECT(EnsureDirectoryExists(android::base::Dirname(full_target_path)));
```
jemoreira: nit: The call to `Dirname` could be avoided if this is done when `IsDirectory(path)` is `true`. It would also avoid repeating the call for every file in the directory. I'm assuming directories are visited before its contents, disregard if that's not the case.

jemoreira (approved):

base/cvd/cuttlefish/host/commands/cvd/fetch/symlink_host_package.cc
Databean: Good point, it is true that directories are visited first. Updated.

PR #2938, opened 2026-07-31
Don't run DWYU when layering_check is disabled

Databean:
Both will fail in the same places, so avoid having to redundantly disable both layering_check and DWYU checks.

Bug: b/534499070


dxapd (approved):

Databean:
Thanks for reviewing!

PR #2940, opened 2026-07-31
Cherry-Pick Fix file already exist error while running substitute in Docker into version-1.55-dev

k311093:
In the OverlayFS, even if unlink() is succeeded, looks like there is a delay between unlink call and the file is actually disappears.

Change to use rename call which ensures atomic operation for substitution.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
+++ b/base/cvd/cuttlefish/host/commands/cvd/fetch/substitute.cc
@@ -17,8 +17,10 @@
 
 #include <unistd.h>
 
+#include <cerrno>
 #include <functional>
 #include <optional>
+#include <random>
 #include <string>
 #include <string_view>
 #include <vector>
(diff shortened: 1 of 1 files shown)


0405ysj (approved):
Does it need to modify `base/cvd/cuttlefish/host/commands/cvd/fetch/BUILD.bazel` as https://github.com/google/android-cuttlefish/pull/2885/changes? Otherwise, LGTM.

0405ysj:
Also could we import this patch under `version-1.56-dev` branch as well unless there's a reason not to add this change?

k311093:
> Does it need to modify `base/cvd/cuttlefish/host/commands/cvd/fetch/BUILD.bazel` as https://github.com/google/android-cuttlefish/pull/2885/changes? Otherwise, LGTM.

In this branch `base/cvd/cuttlefish/host/commands/cvd/fetch/BUILD.bazel` already has `@fmt` dependency.

rmuthiah (approved):

PR #2932, opened 2026-07-30
Update to v1.55.2.

3405691582:
Note: I am not bothering to reproduce the GitHub changelog format here, because reasons.

PR #2941, opened 2026-07-31
gigabyte-ampere-cuttlefish-installer: utils: download-ci-cf.sh: use J…

rmuthiah:
…SON API

The previous way to download the artifacts doesn't work. We fix it by using the latest JSON API. And also make the script more verbose on error.

Needed to resolve a arm64 CI fetch error.


rmuthiah:
already in #2932 - thanks @3405691582

PR #2942, opened 2026-07-31
cuttlefish: delete more virgl

gurchetansingh:
Many people upstream call virgl "notoriously broken" and the urge to delete it upstream is growing irresitable.

"VirGL is no longer considered maintained any more. Unless someone
 steps up to maintain it in the long run, the code is likely to star
 bit-rotting and might eventually be removed. We recommend users
 start moving over to other solutions sooner rather than later"

https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/41240 https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/41259 https://lists.freedesktop.org/archives/mesa-announce/2026-May/000849.html

The Android GPU virtualization teams already did their part [1], deleting virgl from AOSP in 2024.

There's some dead code around here host side, so let's delete that to leave no room for mis-interpretations.

Cuttlefish users are advised to continue investing into gfxstream Vulkan.

[1] https://www.youtube.com/watch?v=81M62lI4mhc

### DIFF
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/proto/guest_config.proto
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/proto/guest_config.proto
@@ -22,13 +22,12 @@ enum GpuMode {
   GPU_MODE_UNKNOWN = 0;
   GPU_MODE_AUTO = 1;
   GPU_MODE_CUSTOM = 2;
-  GPU_MODE_DRM_VIRGL = 3;
-  GPU_MODE_GFXSTREAM = 4;
-  GPU_MODE_GFXSTREAM_GUEST_ANGLE = 5;
-  GPU_MODE_GFXSTREAM_GUEST_ANGLE_HOST_LAVAPIPE = 6;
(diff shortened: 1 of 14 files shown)


gurchetansingh:
CC: @jmacnak @adelva1984

base/cvd/cuttlefish/host/commands/assemble_cvd/proto/guest_config.proto
```diff
@@ -22,13 +22,12 @@ enum GpuMode {
   GPU_MODE_UNKNOWN = 0;
   GPU_MODE_AUTO = 1;
   GPU_MODE_CUSTOM = 2;
-  GPU_MODE_DRM_VIRGL = 3;
```
jmacnak, line 25: Please keep the enum value but add a `[deprecated = true];` annotation just out of an abundance of caution and to potentially get more useful errors instead of just text proto parsing failures.

jmacnak (approved):

PR #2946, opened 2026-07-31
Enforce build_test, DWYU, and misc-include-cleaner on casimir_control_server

Databean:
Also split up some build targets

Bug: b/523396865
Bug: b/532697459
Bug: b/534499070


base/cvd/cuttlefish/host/commands/casimir_control_server/.clang-tidy
```diff
@@ -0,0 +1 @@
+../../../../clang_tidy_configs/with_include_cleaner
```
dxapd, line 1: assuming this is a symlink since it doesn't clearly show that in the review view.

dxapd (approved):

base/cvd/cuttlefish/host/commands/casimir_control_server/.clang-tidy
dxapd: (and it looks like it is, in the file view)

Databean: Yes, this is a symlink.

Databean:
Thanks for reviewing!

PR #2948, opened 2026-07-31
Compile shaders at build time

Databean:
Sorry, not sure how to test this.

Bug: b/533517706


jmacnak (approved):
> Sorry, not sure how to test this.

Should be okay as long as the detector tests complete successfully in the kokoro_gpu run.

Databean:
Thanks, sounds good. `kokoro_gpu` did pass, but it looks like the vulkan toolchain prebuilt has issues on arm64. Will investigate later.

Databean:
`rules_vulkan` downloads SDK prebuilts from https://vulkan.lunarg.com/sdk/home , but it looks like there is no published `arm64` version.

PR #2953, opened 2026-08-03
Add an optional A/B pvmfw partition to the OS composite disk

dimorinny:
The image comes from the Android build when available. Targets that don't build protected VM firmware are unaffected.

Bug: 536088107

### DIFF
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/disk/android_composite_disk_config.cc
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/disk/android_composite_disk_config.cc
@@ -52,6 +52,7 @@ constexpr struct {
   std::string_view init_boot = "init_boot";
   std::string_view metadata = "metadata";
   std::string_view misc = "misc";
+  std::string_view pvmfw = "pvmfw";
   std::string_view super = "super";
   std::string_view userdata = "userdata";
   std::string_view vbmeta = "vbmeta";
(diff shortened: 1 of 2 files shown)


base/cvd/cuttlefish/host/commands/assemble_cvd/flags/pvmfw_image.cc
```diff
@@ -0,0 +1,76 @@
...
+#include "cuttlefish/files/file_exists.h"
+#include "cuttlefish/host/commands/assemble_cvd/flags/system_image_dir.h"
+#include "cuttlefish/host/commands/assemble_cvd/flags_defaults.h"
+#include "cuttlefish/result/result.h"
+
+DEFINE_string(pvmfw_image, CF_DEFAULTS_PVMFW_IMAGE,
+              "Location of the cuttlefish protected VM firmware image. If not "
+              "given it is taken from pvmfw.img in the directory specified by "
+              "-system_image_dir, when that file is present. If given, the "
+              "file must exist.");
```
Databean: Are there scenarios where this file is built separately from the Android build system and a custom version needs to be tested? Would rather avoid adding a new flag if it's not going to be used.

Databean (approved):

base/cvd/cuttlefish/host/commands/assemble_cvd/flags/pvmfw_image.cc
dimorinny: I'm agree. Don't have any specific case in mind to override the pvmfw image. Will propagate into the flag if ever required.

Databean (approved):

PR #2959, opened 2026-08-04
Add sandboxed e2e tests for host resources.

dxapd:
Add a way to run tests for resource allocation. This will be adapted for cvdalloc soon.

Currently, I don't run the code in the new directory in kokoro. I plan to try and do so if possible w/ the cvdalloc stuff, but the main use of this right now is for local testing.


e2etests/host_resources/common/init_network_ns.go
```diff
@@ -0,0 +1,55 @@
...
+	for _, p := range parts {
+		have[p] = true
+	}
+	for _, e := range []string{"/usr/sbin", "/usr/bin", "/sbin", "/bin"} {
+		if !have[e] {
+			parts = append(parts, e)
+		}
+	}
+	os.Setenv("PATH", strings.Join(parts, ":"))
+}
```
Databean, line 50: Can the callers to these executables use absolute paths rather than setting up `PATH`? Are the locations of these executables unpredictable?

e2etests/host_resources/common/init_network_ns.go
Databean, line 1: The file is called `network_ns.go`, which gives me the impression that it is creating the network namespace, but if I understand correctly it is doing the initial setup inside the namespace. Maybe `inside_network_ns.go` to be less ambiguous?

e2etests/host_resources/common/sandbox.go
```diff
@@ -0,0 +1,167 @@
...
+	initScript string
+	closed     bool
+}
+
+type InitEnv struct {
+	NumCvdAccounts int
+}
+
+func NewSandbox(t *testing.T) *Sandbox {
+	t.Helper()
```
Databean: Using [`Helper()`](https://pkg.go.dev/testing#T.Helper) in these files seems like it will make failures harder to understand.

> When printing file and line information, that function will be skipped.

e2etests/host_resources/common/sandbox.go
```diff
@@ -0,0 +1,167 @@
...
+	// spawn the process that will keep the sandbox alive
+	cmd := exec.Command("unshare", "--user", "--map-root-user", "--net", "--mount", "sleep", "infinity")
+	if err := cmd.Start(); err != nil {
+		t.Fatalf("cannot create rootless user+net namespace: %v", err)
+	}
+	s.keeper = cmd
+	s.pid = cmd.Process.Pid
+
+	if err := s.waitReady(); err != nil {
+		s.Close()
```
Databean: If `t.Cleanup(s.Close)` is moved before this block, would all these error handlers need to call `s.Close()` manually?

e2etests/host_resources/common/init_filesystem_ns.go
Databean, line 1: The file is called `filesystem_ns.go`, which gives me the impression that it is creating the filesystem namespace, but if I understand correctly it is doing the initial setup inside the namespace. Maybe `inside_filesystem_ns.go` to be less ambiguous?

e2etests/host_resources/common/sandbox.go
```diff
@@ -0,0 +1,167 @@
...
+		t.Fatalf("namespace not usable on this host: %v", err)
+	}
```
Databean, line 63: What's the distinction separating methods that report errors through `t.Fatal*` versus returning `error`?  `NewSandbox` and the `check*` methods use `t.Fatal*` but these others return `error`.

e2etests/host_resources/common/sandbox.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func (s *Sandbox) Run(args ...string) (CommandOutput, error) {
+	full := s.nsenterArgs(args...)
+	cmd := exec.Command(full[0], full[1:]...)
+	var outBuf, errBuf bytes.Buffer
+	cmd.Stdout = &outBuf
+	cmd.Stderr = &errBuf
+	log.Printf("[sandbox] running: %s", strings.Join(args, " "))
+	err := cmd.Run()
```
Databean, line 167: Should this use the [context](https://pkg.go.dev/testing#T.Context) from the test environment so that the commands are halted if the test exits?

e2etests/host_resources/common/sandbox.go
```diff
@@ -0,0 +1,167 @@
...
+	Stdout string
+	Stderr string
+}
+
+type Sandbox struct {
+	t          *testing.T
+	keeper     *exec.Cmd
+	pid        int
+	tempdir    string
+	initScript string
```
Databean: If I understand correctly, this refers to the `/etc/init.d/cuttlefish-host-resources` script. This isn't a great separation of concerns: the Sandbox type is aware both of namespacing concerns and is involved in managing the host-resources script directly. It would be easier to consider the Sandbox type purely as managing namespaces, and it should expose enough operations that the test can use it to start operations on the init script inside.

e2etests/host_resources/common/sandbox.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	s.initScript = full
+
+	// spawn the process that will keep the sandbox alive
+	cmd := exec.Command("unshare", "--user", "--map-root-user", "--net", "--mount", "sleep", "infinity")
+	if err := cmd.Start(); err != nil {
+		t.Fatalf("cannot create rootless user+net namespace: %v", err)
+	}
+	s.keeper = cmd
+	s.pid = cmd.Process.Pid
```
Databean: Rather than using `unshare`, is it possible to use [`SysProcAttr`](https://pkg.go.dev/syscall#SysProcAttr) with flags like [`CLONE_NEWNET`](https://pkg.go.dev/syscall#pkg-constants)?

It doesn't look like there is a convenient alternative for `nsenter` though. A kind of awkward way to avoid `nsenter` would be to have `bash` as the root process of the namespace and keep file descriptors open to run commands inside by writing to the shell's stdin and reading results from its stdout, but that seems more convoluted.

e2etests/host_resources/common/state.go
Databean, line 1: The functions in this file that are member functions of `Sandbox` feel like they could just as easily be free functions that accept a `Sandbox`, as they don't interact with any of its private members.

http://go/UnhealthyIdioms#violation-of-single-responsibility-principle

It doesn't make much difference in go semantics since visibility is package-level so even non-member functions can access private struct members, but as a reader it draws clearer distinctions about the meaning of the Sandbox type.

e2etests/host_resources/common/nft.go
Databean, line 1: This file defines some `parse*` functions that internally define more complex structures with json annotations and return simpler types with easily accessible public fields.

What do you think instead of defining the public types around the json schema, keeping the fields private, and then defining accessor methods to produce the values the callers need to access? It would make it easier to connect the code to the command output.

The downside is that it makes it harder to describe the ideal state in test cases like `TestStaticResourcesInit` which also rely on being able to construct the same types. Maybe a hybrid of the two approaches is to try to extract the json schema types out of the `parse*` methods and implement conversions between the two types of structures.

There's also [`UnmarshalerFrom`](https://pkg.go.dev/encoding/json/v2#UnmarshalerFrom) and with [`jsontext.Decoder`](https://pkg.go.dev/encoding/json/jsontext#Decoder) but that looks like a bigger hassle because of how low-level it is.

e2etests/host_resources/common/init_filesystem_ns.go
dxapd: makes sense, though i think the best prefix here would be `init` or something similar

e2etests/host_resources/common/sandbox.go
dxapd: it appears that that is mostly what unshare does, plus a bunch of setup:

https://github.com/util-linux/util-linux/blob/master/sys-utils/unshare.c#L939

e2etests/host_resources/common/nft.go
dxapd: my thinking was actually to let this file fall on the messier side so the test files are super clear, but let me see if i can straighten things out in here. i'll try a few things

e2etests/host_resources/common/sandbox.go
dxapd: yeah the host resources stuff can be moved to another file

e2etests/host_resources/common/init_network_ns.go
dxapd: this is mostly for the benefit of the init script (which runs them expecting them to be on PATH)

e2etests/host_resources/common/nft.go
dxapd: okay so I misunderstood this comment at the time I wrote the above. I don't think I want to implement public types based on the schema with accessor methods, but I do think there's a bunch of duplication I can cut down on. moved the json tags to the struct declarations themselves + tried using your hybrid idea for rules. I think that should consolidate the complexity nicely

e2etests/host_resources/common/state.go
dxapd: yep

e2etests/host_resources/common/sandbox.go
dxapd: yeah that looks better ty

e2etests/host_resources/common/sandbox.go
dxapd: none, i consolidated the fatals into the setup functions

e2etests/host_resources/common/sandbox.go
dxapd: ye (also used PID namespace to make sure stuff gets cleaned up)

e2etests/host_resources/common/init_network_ns.go
dxapd: same as other comment (yes, but init_)

e2etests/host_resources/common/sandbox.go
dxapd: yep

e2etests/host_resources/common/state.go
```diff
@@ -0,0 +1,154 @@
...
+	for i := range nft.Chains {
+		nft.Chains[i].Handle = 0
+	}
+	for i := range nft.Rules {
+		nft.Rules[i].Handle = 0
+	}
+	sort.Slice(nft.Tables, func(i, j int) bool { return nft.Tables[i].less(nft.Tables[j]) })
+	sort.Slice(nft.Chains, func(i, j int) bool { return nft.Chains[i].less(nft.Chains[j]) })
+	sort.Slice(nft.Rules, func(i, j int) bool { return nft.Rules[i].less(nft.Rules[j]) })
+	out.Nft = nft
```
Databean, line 120: The [documentation for `sort.Slice`](https://pkg.go.dev/sort#Slice) writes

> Note: in many situations, the newer [`slices.SortFunc`](https://pkg.go.dev/slices#SortFunc) function is more ergonomic and runs faster.

I think the more ergonomic part is that the callback function takes the actual elements rather than indexes into the slice? Then this could pass `less` or an equivalent compare function as a function reference rather than needing to define anonymous functions here that index into the slice.

e2etests/host_resources/common/state.go
```diff
@@ -0,0 +1,154 @@
...
+	} else {
+		s.t.Logf("snapshot: ip addr failed: %v", err)
+	}
+	for key, path := range map[string]string{
+		"net.ipv4.ip_forward":          "/proc/sys/net/ipv4/ip_forward",
+		"net.ipv6.conf.all.forwarding": "/proc/sys/net/ipv6/conf/all/forwarding",
+	} {
+		if out, err := s.Run("cat", path); err == nil {
+			hs.Sysctls[key] = strings.TrimSpace(out.Stdout)
+		}
```
Databean: Should this report the error if `err != nil`?

e2etests/host_resources/common/state.go
```diff
@@ -0,0 +1,154 @@
...
+)
+
+type HostState struct {
+	Nft     NftRuleset
+	Links   []Link
+	Addrs   []Addr
+	Sysctls map[string]string
+}
+
+func Snapshot(s *Sandbox) HostState {
```
Databean: The `TestStaticResourcesInit` function does a before-and-after comparison by calling `Snapshot` twice and comparing the results. There's a failure mode here that if any part of the `Snapshot` generation is broken, that field would be empty before and after and the test would not actually cover that field.

It looks like `Sysctls` is left out of `Normalize` so that it can be tested against fixed test case values: that makes it more resilient against unrelated setup issues.

Rather than reporting errors only through logging, can `Snapshot`  report errors through a second return value and fail the test on failing to collect a field?

e2etests/host_resources/common/nft.go
Databean: Thanks, this looks better.

e2etests/host_resources/common/state.go
dxapd: yea makes sense

e2etests/host_resources/common/state.go
dxapd: yea i think that makes sense. i can fail the test if any of these fail during construction

e2etests/host_resources/common/state.go
dxapd: resolved below

e2etests/host_resources/static_resources_init_test/main_test.go
```diff
@@ -0,0 +1,161 @@
...
+				return a.Table < b.Table
+			}
+			return a.Name < b.Name
+		}),
+		cmpopts.IgnoreFields(common.NftTable{}, "Handle"),
+		cmpopts.IgnoreFields(common.NftChain{}, "Handle"),
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
```
Databean, line 100: Are there more test cases planned in the future? With only one case, the test could be written without this loop.

Databean (approved):

PR #2960, opened 2026-08-04
Proxy --acpi-table, --device-tree-overlay, and --file-backed-mapping to crosvm run

erjiaqing:
This CL Introduce new command line arguments and cuttlefish config properties to allow proxying user provided ACPI tables, device tree overlays, and file-backed mappings to `crosvm run` command.

Assisted-by: Antigravity:Gemini-Next
Bug: 510177847

### DIFF
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
@@ -618,6 +618,12 @@ Result<CuttlefishConfig> InitializeCuttlefishConfiguration(
   std::vector<bool> smt_vec = CF_EXPECT(GET_FLAG_BOOL_VALUE(smt));
   std::vector<std::string> crosvm_binary_vec =
       CF_EXPECT(GET_FLAG_STR_VALUE(crosvm_binary));
+  std::vector<std::string> crosvm_acpi_table_vec =
+      CF_EXPECT(GET_FLAG_STR_VALUE(crosvm_acpi_table));
+  std::vector<std::string> crosvm_device_tree_overlay_vec =
+      CF_EXPECT(GET_FLAG_STR_VALUE(crosvm_device_tree_overlay));
+  std::vector<std::string> crosvm_file_backed_mapping_vec =
+      CF_EXPECT(GET_FLAG_STR_VALUE(crosvm_file_backed_mapping));
   std::vector<std::string> seccomp_policy_dir_vec =
       CF_EXPECT(GET_FLAG_STR_VALUE(seccomp_policy_dir));
   std::vector<std::string> qemu_binary_dir_vec =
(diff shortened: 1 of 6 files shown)


base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
```diff
@@ -618,6 +618,12 @@ Result<CuttlefishConfig> InitializeCuttlefishConfiguration(
   std::vector<bool> smt_vec = CF_EXPECT(GET_FLAG_BOOL_VALUE(smt));
   std::vector<std::string> crosvm_binary_vec =
       CF_EXPECT(GET_FLAG_STR_VALUE(crosvm_binary));
+  std::vector<std::string> crosvm_acpi_table_vec =
+      CF_EXPECT(GET_FLAG_STR_VALUE(crosvm_acpi_table));
+  std::vector<std::string> crosvm_device_tree_overlay_vec =
+      CF_EXPECT(GET_FLAG_STR_VALUE(crosvm_device_tree_overlay));
+  std::vector<std::string> crosvm_file_backed_mapping_vec =
+      CF_EXPECT(GET_FLAG_STR_VALUE(crosvm_file_backed_mapping));
```
Databean, line 629: The suggested arguments of `addr=NUM,size=NUM,path=PATH,offset=NUM,rw,sync,align,ram` will interact badly with `GET_FLAG_STR_VALUE`, which separates arguments to different by splitting on commas.

Specifically,
```
cvd create --crosvm_file_backed_mapping=addr=NUM,size=NUM --num_instances=2
```
will result in two invocations of crosvm:
```
$ crosvm --file-backed-mapping=addr=NUM
$ crosvm --file-backed-mapping=size=NUM
```
which I assume is not the intended behavior.

The easy option is to put this in `CuttlefishConfig` rather than `CuttlefishConfig::InstanceSpecific` and declare that all instances launched concurrently must use the same value.

Another route we've gone is e.g. `extra_bootconfig_args_base64` which expects the arguments to be base64-encoded so that `GET_FLAG_STR_VALUE` can correctly divide the arguments between instances.

erjiaqing: Thanks for the insight.

For our usecase, we are passing the opendice handover to the memory, which does not make sense to have multiple instances using the same handover file. So I added the `crosvm_file_backed_mapping_base64`. To allow multiple instances using different buffer.

Databean (approved):

Databean (approved):

Databean:
This change is [failing](https://github.com/google/android-cuttlefish/actions/runs/34396717563/job/102637741069?pr=2960) some clang-tidy lints:
```
  ./cuttlefish/host/libs/config/cuttlefish_config.h:665:32: error: parameter 'acpi_table' is const-qualified in the function declaration; const-qualification of parameters only has an effect in function definitions [readability-avoid-const-params-in-decls,-warnings-as-errors]
    665 |     void set_crosvm_acpi_table(const std::string acpi_table);
        |                                ^~~~~
  ./cuttlefish/host/libs/config/cuttlefish_config.h:666:41: error: parameter 'device_tree_overlay' is const-qualified in the function declaration; const-qualification of parameters only has an effect in function definitions [readability-avoid-const-params-in-decls,-warnings-as-errors]
    666 |     void set_crosvm_device_tree_overlay(const std::string device_tree_overlay);
        |                                         ^~~~~
  ./cuttlefish/host/libs/config/cuttlefish_config.h:667:41: error: parameter 'file_backed_mapping' is const-qualified in the function declaration; const-qualification of parameters only has an effect in function definitions [readability-avoid-const-params-in-decls,-warnings-as-errors]
    667 |     void set_crosvm_file_backed_mapping(const std::string file_backed_mapping);
        |                                         ^~~~~
```

See https://clang.llvm.org/extra/clang-tidy/checks/readability/avoid-const-params-in-decls.html

It is possible to reproduce this test by running `bazel test //...` locally.

PR #2961, opened 2026-08-04
gigabyte-ampere-cuttlefish-installer: wait for the vm to finish booting

dimorinny:
The qemu test jobs treat the serial console login prompt as "vm is ready" and ssh into the guest right away. The prompt only means getty started: the guest is still booting, so the ssh handshake gets reset (kex_exchange_identification: read: Connection reset by peer) and, once that is past, dns is not up yet for the first command needing it.

The console is polled every 30s, so the ssh always lands at the same point after boot and whether it works is luck. Four of twelve job instances across six recent runs failed this way, on both jobs and regardless of which kernel the installer had put on the vm.

Wait until systemd reports the boot finished before using the vm.

Bug: 537008147

### DIFF
--- a/.github/workflows/gigabyte-ampere-cuttlefish-installer.yaml
+++ b/.github/workflows/gigabyte-ampere-cuttlefish-installer.yaml
@@ -56,7 +56,7 @@ jobs:
         nvidia_gpu: ["true", "false"]
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


dimorinny:
Succeeded 10/10 times here: https://github.com/dimorinny/android-cuttlefish/commits/ci-ssh-wait-trial

1/10 runs had to wait more then 1 iteration in order to connect, so this confirms the fix works and useful.

rmuthiah (approved):

PR #2964, opened 2026-08-04
Fail more quickly if the deb argument is wrong.

3405691582:
If there's a mistake or issue with the linux_image_deb argument, then this only gets detected later in the process after several minutes. This hopefully should check the argument earlier, failing a little more quicker.

### DIFF
--- a/tools/baseimage/pkg/gce/scripts/install_kernel_main.sh
+++ b/tools/baseimage/pkg/gce/scripts/install_kernel_main.sh
@@ -27,13 +27,21 @@ arch=$(uname -m)
 [ "${arch}" = "aarch64" ] && arch=arm64
 
 sudo apt-get update
+sudo chroot /mnt/image /usr/bin/apt-get update
+
+if ! sudo chroot /mnt/image /usr/bin/apt-get install --dry-run \
+    "${linux_image_deb}" > /dev/null; then
+  echo "CREATE IMAGE FAILED!!!"
+  echo "Package not found: ${linux_image_deb}"
+  exit 1
+fi
+
 sudo apt-get upgrade -y
 
 version=$(sudo chroot /mnt/image/ /usr/bin/dpkg -s linux-image-cloud-${arch} | grep ^Depends: | \
   cut -d: -f2 | cut -d" " -f2 )
 echo "START VERSION: ${version}"
 
-sudo chroot /mnt/image /usr/bin/apt-get update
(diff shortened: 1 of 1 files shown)


tools/baseimage/pkg/gce/scripts/install_kernel_main.sh
```diff
@@ -27,13 +27,21 @@ arch=$(uname -m)
...
 
 sudo apt-get update
+sudo chroot /mnt/image /usr/bin/apt-get update
+
+if ! sudo chroot /mnt/image /usr/bin/apt-get install --dry-run \
+    "${linux_image_deb}" > /dev/null; then
+  echo "CREATE IMAGE FAILED!!!"
+  echo "Package not found: ${linux_image_deb}"
+  exit 1
+fi
```
ser-io, line 37: good idea, let's just make this logic part of the Golang binary. I think we can do a check just by pinging the debian endpoint:

```
        url := "https://packages.debian.org/[trixie|bookworm]/{linux_image_deb}"
        resp, err := http.Head(url)
        if err == nil && resp.StatusCode == http.StatusOK {
            fmt.Println("Exists (200 OK)")
        } else {
            fmt.Printf("Does not exist or error: %v (Status: %d)\n", err, resp.StatusCode)
        }
```

This way we would fail way faster as we won't need to create VM/disks and all of that stuff.

ser-io (changes requested):

tools/baseimage/pkg/gce/scripts/install_kernel_main.sh
3405691582: we could do something like that, but that becomes brittle if the path changes or the apt/package mechanism changes. 

also, `https://packages.debian.org/trixie/foo` returns 200 OK.

ser-io: We can rather use the `/search` endpoint:

```
curl -s "https://packages.debian.org/search?keywords=linux-image-6.12.101&suite=trixie&section=main" | grep -q "linux-image-6.12.100+deb13-cloud-amd64" && echo "Available" || echo "Not Available"
```

Given how ad-hoc this script is and how un-frequent is meant to be run I think is fine to rely on `https://packages.debian.org/search` in order to exit right away without engage on GCP resources. Verifying wether the package is valid via "chroot" and "apt" in this case would give us a better error message, however I think the delay for failing would almost be the same as we still need to create the vms, disks and so on which is the what takes the most when running this scripts.

PR #2967, opened 2026-08-04
Avoid dist-upgrade for now.

ser-io:


rmuthiah (changes requested):
Can you keep the deletion of the legacy kernel ? That's still needed for the NVIDIA driver build step.

ser-io:
> Can you keep the deletion of the legacy kernel ? That's still needed for the NVIDIA driver build step.

sounds good, I'll repharse the commit message/ PR then as it won't be a full revert.

rmuthiah (approved):

PR #2968, opened 2026-08-04
tools: baseimage: Add pre-flight existence checks

rmuthiah:
for GCE images and builder resources

Add pre-flight existence verification before starting image creation builds in create_gce_base_image and create_gce_fixed_kernel:
- Check that required source images exist in their respective projects before beginning.
- Check that the target image-name does not already exist in the target project.
- In GceHelper.BuildImage, check that builder VM instances and attached disks do not already exist before attempting resource creation.

Assisted-by: Antrigravity:Gemini-Next

### DIFF
--- a/tools/baseimage/cmd/create_gce_fixed_kernel/main.go
+++ b/tools/baseimage/cmd/create_gce_fixed_kernel/main.go
@@ -90,10 +90,30 @@ func main() {
 			slices.Collect(maps.Keys(sourceImageMap[arch.GceArch()])))
 	}
 
+	h, err := gce.NewGceHelper(project, zone)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	sourceImage := sourceImageMap[arch.GceArch()][debianVersion]
+	if srcExists, err := h.ImageExists(debianSourceImageProject, sourceImage); err != nil || !srcExists {
+		if err != nil {
+			log.Fatal(err)
+		}
+		log.Fatalf("source image %q does not exist in project %q", sourceImage, debianSourceImageProject)
+	}
+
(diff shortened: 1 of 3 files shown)


3405691582 (commented):
Throughout: coalesce the `if err != nil {} if exists {}` pattern into just returning `err != nil || exists` or similar. Not really useful to have separate branches here.

tools/baseimage/pkg/gce/gce.go
```diff
@@ -261,6 +299,22 @@ func (h *GceHelper) BuildImage(project, zone string, opts BuildImageOpts) error
 	insName := opts.ImageName
 	attachedDiskName := fmt.Sprintf("%s-attached-disk", insName)
 
+	insExists, err := h.InstanceExists(project, zone, insName)
```
ser-io: The GCE api already throw this errors we catch them and fail when the image and disk already exist. I think we don't need to take this responsibility in this scripts.

ser-io (changes requested):

tools/baseimage/pkg/gce/gce.go
rmuthiah: Good point. Removed the instance checks. Still kept the disk checks since disk creation occurs after the instance gets created.

tools/baseimage/cmd/create_gce_fixed_kernel/main.go
ser-io: use new "sourceImage" variable here.

ser-io (approved):

tools/baseimage/cmd/create_gce_fixed_kernel/main.go
rmuthiah: Thanks for the review! Done.

PR #2971, opened 2026-08-05
Extract RealPath from utils:files

Databean:
Bug: b/542760101


base/cvd/cuttlefish/common/libs/utils/files.cpp
```diff
@@ -222,18 +222,6 @@ std::string AbsolutePath(std::string_view path) {
...
-  char* res;
-  do {
-    res = realpath(path.c_str(), buffer.data());
-  } while (res == nullptr && errno == EINTR);
-  CF_EXPECTF(res != nullptr, "Could not get real path for path \"{}\": {}",
-             path, StrError(errno));
-  return std::string(buffer.data());
-}
-
 off_t FileSize(const std::string& path) {
```
dxapd, line 225: curious why realpath specifically on its own. is there an intention to make realpath more complex, or to break up this file in general, or to make it easier to access from somewhere else, or something else entirely?

wondering since if it's the second one, it might be better to break up the file in to broader chunks.

dxapd (approved):

base/cvd/cuttlefish/common/libs/utils/files.cpp
Databean: This is trying to follow [go/code-for-blaze#onesymbol](http://go/code-for-blaze#onesymbol) which writes

> Best Practice: Create individual source files (and associated `BUILD` targets) for every 'top-level entity'. This applies to whatever language specific top-level entities exist (Java `class`, Proto `message`, C++ `struct`...).

dxapd: interesting. i do personally prefer many small files arranged hierarchically but i'm surprised to see it codified

Databean:
Thanks for reviewing!

PR #2972, opened 2026-08-05
gigabyte-ampere-cuttlefish-installer: resolve the build id via status json

dimorinny:
The /builds/latest/ redirect the script followed is rate limited and returns 403 once the quota is exhausted, failing the download. Read the id from the target's status.json instead, which is not on that quota, and build the artifact URLs from it directly.

Bug: 537008147


rmuthiah (approved):

dimorinny:
History note, I removed the corresponding branch in my fork which caused this PR to be closed. Re-uploaded exactly the same change here: https://github.com/google/android-cuttlefish/pull/2975

PR #2978, opened 2026-08-06
Add a `base_dir` attribute for the package_files rule

Databean:
In practice all members of a package_rule were in the same directory. This makes it explicit and reduces some duplication in the `package_file_symlink_to_package_file` and `package_file_to_src` attributes.

Bug: b/429012834


base/cvd/cuttlefish/package/BUILD.bazel
```diff
@@ -2,157 +2,160 @@ load("//cuttlefish/package:rules.bzl", "package_files")
 
 package_files(
     name = "cvd",
```
jmacnak, line 4: optional drive-by: WDYT about renaming this to `common` and adding an alias of `cvd` -> `common` to keep the existing development flow working while trying to match the package name?

jmacnak (approved):

base/cvd/cuttlefish/package/BUILD.bazel
Databean: Good point, done.

Databean:
Thanks for reviewing!

PR #2981, opened 2026-08-06
Use bazelisk for amd64 as well

0405ysj:


Databean:
@monkey-jsun has been working on building/running the host tools on riscv64, and it looks like the bazelisk `.deb`s published on https://www.github.com/bazelbuild/bazelisk/releases are currently only for `amd64` and `arm64`, is this going to cause breakage?

0405ysj:
> @monkey-jsun has been working on building/running the host tools on riscv64, and it looks like the bazelisk `.deb`s published on https://www.github.com/bazelbuild/bazelisk/releases are currently only for `amd64` and `arm64`, is this going to cause breakage?

I think it's out of scope of this PR? So far, `installbazel.sh` have supported `amd64` and `arm64` only. I don't expect we can use this script for riscv64 without any modification.

Databean:
The possibly breaking change is that this PR updates `base/debian/control` to depend on a package that is not in debian upstream, and for which no riscv64 equivalent is published.

monkey-jsun:
My temporary solution for riscv64 is to give instruction on building riscv64 by yourself and bail out early.

https://github.com/google/android-cuttlefish/commit/7ed720238d1217082c18a60e61a90990114aa002

(ignore the polluted marker lines)

0405ysj:
Thanks for sharing how riscv64 will install bazel. I updated `base/debian/control`.

Databean:
Sorry, also comes to mind that @grandpaul is working on a strategy to make this build self-contained where `bazel` comes only from the repo and the source dependencies are vendored, does this conflict?

Can't think of issues beyond that.

Databean (approved):

0405ysj:
> Sorry, also comes to mind that @grandpaul is working on a strategy to make this build self-contained where `bazel` comes only from the repo and the source dependencies are vendored, does this conflict?
> 
> Can't think of issues beyond that.

No worries, checking everything is important than just merging PR, especially for this one. 

I don't know Paul's strategy yet though, I believe the new strategy wouldn't be keeping diverged bazel installation mechanism between arch.(bazel for amd64, bazelisk for arm64) Not yet certain if the direction should be applying bazelisk on amd64 or dropping bazelisk on arm64, depending on his answer.

0405ysj:
@grandpaul @Databean Gentle ping, would there any references how any change around `bazel` can be impacted by this PR? Otherwise I'm thinking about merging this PR soon.

PR #2983, opened 2026-08-06
Consolidate cf_cc_{binary,library,test} implementations

Databean:
This fixes the divergence that only `cf_cc_library` handled cases where both `srcs` and `hdrs` were empty, and makes the remaining divergences more explicit:

- Only `cf_cc_binary` interacts with linkopts
- Only `cf_cc_test` adds extra dependencies

It may come up later that `deps` is configurable for `binary` and `library` (required for allocd) but cannot be configurable for `test` (so that cf_cc_library can add the additional links).

Bug: b/543509210

### DIFF
--- a/base/cvd/cuttlefish/bazel/rules.bzl
+++ b/base/cvd/cuttlefish/bazel/rules.bzl
@@ -52,69 +52,54 @@ cf_build_test = macro(
     implementation = _cf_build_test_implementation,
 )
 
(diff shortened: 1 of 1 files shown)


3405691582 (approved):
I don't have a good Starlark instinct for potential subtleties or thoughts about how this would be best implemented, but it generally looks fine.

PR #2989, opened 2026-08-07
Implement ReaderSeeker on FileInstance

Databean:
`FileInstance::PRead` had no callers besides `SharedFdIo`. The `Seeker` methods are named differently so it doesn't conflict with `FileInstance::LSeek`.

Bug: b/540553865


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #2990, opened 2026-08-07
emulated_camera_mplane: Implement frame rate control

chihchiachen:
Decouple frame generation from the vhost-user command thread into a dedicated background worker thread to enforce a 30 FPS frame rate while maintaining non-blocking V4L2 ioctl behavior and v4l2-compliance.

Bug: b/534454250
Test: android.hardware.cts.CameraTest#testPreviewFpsRange

### DIFF
--- a/base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
+++ b/base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
@@ -20,6 +20,14 @@ use std::io::SeekFrom;
 use std::io::Write;
 use std::os::fd::AsFd;
 use std::os::fd::BorrowedFd;
+use std::str::FromStr;
+use std::sync::mpsc::{Receiver, Sender};
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
+use anyhow::{Context, Result as AnyhowResult};
+use vmm_sys_util::eventfd::{EventFd, EFD_NONBLOCK};
+use vmm_sys_util::poll::{PollContext, PollToken};
+use vmm_sys_util::timerfd::TimerFd;
 
 use v4l2r::PixelFormat;
 use v4l2r::QueueType;
@@ -55,7 +63,6 @@ use virtio_media::protocol::SgEntry;
 use virtio_media::protocol::V4l2Event;
 use virtio_media::protocol::V4l2Ioctl;
 use virtio_media::protocol::VIRTIO_MEDIA_MMAP_FLAG_RW;
-use std::str::FromStr;
 
 /// Rust equivalent of the V4L2_CTRL_ID2WHICH C preprocessor macro.
 /// Extracts the control class ID from a control ID by masking out the lower 16 bits
@@ -115,6 +122,7 @@ impl Default for Gain {
 }
 
 /// State of all camera controls.
+#[derive(Debug, Clone, PartialEq, Eq)]
 struct CameraControls {
     lens_facing: LensFacing,
     gain: Gain,
@@ -222,10 +230,20 @@ impl Buffer {
     }
 }
 
-/// Session data of [`EmulatedCamera`].
-pub struct EmulatedCameraSession {
-    /// Id of the session.
-    id: u32,
+/// Device-global properties shared across all sessions and worker threads.
+pub struct DeviceSharedState<Q: VirtioMediaEventQueue> {
+    /// Queue used to send events to the guest.
+    evt_queue: Q,
+    /// Camera controls (gain, lens facing).
+    controls: CameraControls,
+    /// Width of the video.
+    width: u32,
+    /// Height of the video.
+    height: u32,
+}
+
+/// Shared session state between ioctl handlers and the background frame worker.
+pub struct SessionSharedState {
     /// Current iteration of the pattern generation cycle.
     iteration: u64,
     /// Buffers currently allocated for this session.
@@ -234,6 +252,31 @@ pub struct EmulatedCameraSession {
     queued_buffers: VecDeque<usize>,
     /// Is the session currently streaming?
     streaming: bool,
+    /// Monotonic counter incremented on streamoff/reset to invalidate in-flight frames.
+    stream_count: u64,
+}
+
+/// Diagnostic message streamed from the background worker thread to the main session.
+#[derive(Debug)]
+enum DiagMsg {
+    Warning(String),
+    Error(anyhow::Error),
+}
+
+/// Session data of [`EmulatedCamera`].
+pub struct EmulatedCameraSession {
+    /// Id of the session.
+    id: u32,
+    /// Shared state between ioctl handlers and frame worker.
+    state: Arc<Mutex<SessionSharedState>>,
+    /// EventFd to wake up the background worker thread on state changes.
+    wakeup_evt: Arc<EventFd>,
+    /// EventFd to signal the background worker thread to exit.
+    exit_evt: Arc<EventFd>,
+    /// Background frame worker thread handle.
+    worker_handle: Option<std::thread::JoinHandle<AnyhowResult<()>>>,
+    /// Channel receiver for worker diagnostics.
+    diag_rx: Mutex<Receiver<DiagMsg>>,
 }
 
 impl VirtioMediaDeviceSession for EmulatedCameraSession {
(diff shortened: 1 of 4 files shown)


base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -274,59 +300,189 @@ impl EmulatedCameraSession {
...
     }
 }
 
+/// Write basic pattern into the queued buffers
+fn spawn_frame_worker<Q: VirtioMediaEventQueue + Send + 'static>(
+    session_id: u32,
+    shared: Arc<Mutex<SessionSharedState>>,
+    wakeup_evt: Arc<EventFd>,
+    evt_queue: Arc<Mutex<Q>>,
+) -> std::thread::JoinHandle<()> {
```
jmacnak: Could we update this to ensure that the worker thread initalized succesfully?

IIUC, you might be able to do something like:

```
fn spawn_frame_worker(...) -> Result<(), JoinHandle> {
   let (sender, receiver) = std::sync::mpsc::channel();
   
   let join_handle = std::thread::spawn(move || -> Result<(), ERROR_TYPE> {
       let mut time_fd = match TimerFd::new() {
          Ok(t) => t,
          Err(e) => {
              tx.sender(SOME_ERROR("Failed to create TimerFd for worker: {}" e)).unwrap();
              return;
           }
       }
       ...

       sender.send(Ok()).unwrap();

       ... rest of loop ...
   });

   match receiver.recv()? {
      ... handle thread init error ...
   }

   join_handle
}
```

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -234,6 +238,28 @@ pub struct EmulatedCameraSession {
     queued_buffers: VecDeque<usize>,
     /// Is the session currently streaming?
     streaming: bool,
+    /// Is the session stopped?
+    stopped: bool,
```
jmacnak: It looks like the session will never restart after a stop. Perhaps it would be a little easier to follow in `close_session` if we added a `exit_evt` and a `WorkerToken::Exit` enum.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -274,59 +300,189 @@ impl EmulatedCameraSession {
...
+                    Some(id) => id,
+                    None => continue,
+                };
+                let buffer = match guard.buffers.get_mut(buf_id) {
+                    Some(b) => b,
+                    None => continue,
+                };
+
+                for plane in &mut buffer.planes {
+                    let _ = plane.fd.as_file().seek(SeekFrom::Start(0));
```
bridadan: If this is a result, could you log a warning if the seek fails (or send it back over the channel as @jmacnak is suggesting)?

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
chihchiachen: Thank! Will fix it in the next patch.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
chihchiachen: Sure, I think it will be slightly cleaner with your approach. Let me fix it.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -429,33 +612,67 @@ where
...
     }
 
-    fn close_session(&mut self, session: Self::Session) {
+    fn close_session(&mut self, mut session: Self::Session) {
         // Nothing to cleanup when `close_session` is called for sessions without
         // allocated buffers, hence the early return.
-        if self.active_session != Some(session.id) {
-            return;
+        if self.active_session == Some(session.id) {
+            self.active_session = None;
```
bridadan: I don't think this check is exactly equivalent? But perhaps it doesn't matter here, I'm not sure what the invariant for `self.active_session` is supposed to be in this context. There's no early return path any more though, does that still need to be present?

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -274,49 +300,206 @@ impl EmulatedCameraSession {
...
+    exit_evt: Arc<EventFd>,
+    evt_queue: Arc<Mutex<Q>>,
+) -> std::result::Result<std::thread::JoinHandle<()>, i32> {
+    let (init_tx, init_rx) = std::sync::mpsc::channel();
+
+    let worker_handle = std::thread::spawn(move || {
+        let mut timer_fd = match TimerFd::new() {
+            Ok(t) => t,
+            Err(e) => {
+                log::error!("Failed to create TimerFd for worker thread: {}", e);
```
jmacnak: Less familiar with the media libs so feel free to push back on this:

Can these log messages be incorprated into the Error type? Separating the log from the error drops context and potentially makes it more difficult for readers to understand failures because they then have to try to scan logs and guess which logs are actual errors vs just informational.

Could this use something like

```
use thiserror::Error;

#[derive(Debug, Error)]
enum WorkerError {
  #[error("Failed to initialize timer: {0}")]
  FailedToInitializeTimer(#[from] io::Error),

  ...
}

#[derive(Debug, Error)]
enum EmulatedCameraError {
   #[error("Failed to initialize worker: {0}")]
  FailedToInitializeWorker(#[from] WorkerError),

  ...
}
```

or perhaps `anyhow::Error` if you never expect anyone to try to handle individual failures differently:

```
          let mut timer_fd = match TimerFd::new() {
            Ok(t) => t,
            Err(e) => {
              init_tx.send(Err(anyhow!("failed to initialize timer fd: {}", e)).unwrap();
              return;
            }

...

   let worker_handle = spawn_frame_worker(
            session_id,
            Arc::clone(&shared_state),
            Arc::clone(&wakeup_evt),
            Arc::clone(&exit_evt),
            Arc::clone(&self.evt_queue),
    ).context("Failed to spawn worker thread")?;
```

Both of these options effectively let you context and the errors become effectively

"Failed to do virtio-media: failed to start new emulate camera session: failed to start worker thread: failed to create timer fd"

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -274,49 +300,206 @@ impl EmulatedCameraSession {
...
+                let _ = init_tx.send(Err(e.errno()));
+                return;
+            }
+        };
+        let poll_ctx: PollContext<WorkerToken> = match PollContext::new() {
+            Ok(p) => p,
+            Err(e) => {
+                log::error!("Failed to create PollContext for worker thread: {}", e);
+                let _ = init_tx.send(Err(e.errno()));
```
jmacnak: Recommend `unwrap()` on these. We do not expect error passing from the worker thread to ever really fail so panic'ing seems appropriate here.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -274,49 +300,206 @@ impl EmulatedCameraSession {
...
+        }
+        let _ = init_tx.send(Ok(()));
+        let mut timer_armed = false;
+
+        loop {
+            let ready_events = match poll_ctx.wait() {
+                Ok(ev) => ev,
+                Err(e) => {
+                    log::error!("PollContext wait failed in worker thread: {}", e);
+                    break;
```
jmacnak: Again feel free to push back:

Should these non-initialization errors also be propagated back? Potentially you could do `std::thread::JoinHandle<Result<(), ERROR_TYPE>` potentially and then the owner of the worker thread could potentially check

```
if worker_handle.is_finished() {
  let error = worker_handle.join();
  ... propagate worker exiting unexpectedly early error ...
}
```

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -274,49 +300,206 @@ impl EmulatedCameraSession {
...
-            }, width, height);
-            evt_queue.send_event(V4l2Event::DequeueBuffer(DequeueBufferEvent::new(
-                self.id,
-                buffer.v4l2_buffer.clone(),
-            )));
-
-            self.iteration += 1;
+            // Re-acquire lock to verify stream_count and dispatch DequeueBuffer:
+            let mut guard = shared.lock().unwrap();
+            if guard.stream_count == stream_count {
```
bridadan, line 524: I don't fully understand why `stream_count` has to be checked instead of checking if `streaming == false`. Aren't they the same condition?

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -903,11 +1127,14 @@ where
...
-        session.queued_buffers.clear();
-        for buffer in session.buffers.iter_mut() {
+        let mut state = session.state.lock().unwrap();
+        state.streaming = false;
+        state.stream_count += 1;
+        state.queued_buffers.clear();
+        for buffer in state.buffers.iter_mut() {
             buffer.set_state(BufferState::New, self.width, self.height);
         }
+        let _ = session.wakeup_evt.write(1);
```
bridadan: Could we log or `unwrap()` these writes so they don't fail silently?  I see this pattern a few times in this file.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -634,6 +848,9 @@ where
         if SUPPORTED_SIZES.contains(&(req_width, req_height)) {
             self.width = req_width;
             self.height = req_height;
+            let mut state = session.state.lock().unwrap();
+            state.width = req_width;
+            state.height = req_height;
```
bridadan: It feels odd to have this duplicated between `self` and `session.state`. Can they be consolidated?

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -1196,12 +1423,16 @@ where
                         Ok(gain) => {
                             if self.controls.gain != gain {
                                 self.controls.gain = gain;
+                                {
+                                    let mut state = session.state.lock().unwrap();
+                                    state.controls.gain = gain;
```
bridadan: Same here, it feels odd to have this duplicated between `self` and `session.state`. Can they be consolidated?

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
chihchiachen: `stream_count` is needed to guard the edge case where the stream goes on->off->on during write_pattern().

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
chihchiachen: Good catch. Since `width` and `height` are moved to `struct SessionSharedState`, we should remove them from `struct EmulatedCameraSession`. Let me fix it.

chihchiachen: (Correction: I'll remove `width` and `height` from `struct EmulatedCamera`, not `struct EmulatedCameraSession`.)

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
chihchiachen: Thanks - I'll remove `controls` from `struct EmulatedCamera`.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
chihchiachen: It doesn't really harm, but let's remove the irrelevant change. Please check out my next patch.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
chihchiachen: Sure - let me fix them. Thanks.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
chihchiachen: How about moving the init work before the thread creation? This way we can return early from the init errors and bubble up the logs more easily. PTAL.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
chihchiachen: I'll take a approach that's similar to your the second suggestion. PTAL.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
chihchiachen: It turns out v4l2-compliance requires `control`/`width`/`height` to be global (not session-specific) states. Let's move those parameters to a new data struct `DeviceSharedState`.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
chihchiachen: It turns out v4l2-compliance requires `control`/`width`/`height` to be global (not session-specific) states. Let's move those parameters to a new data struct `DeviceSharedState`.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
chihchiachen: I used JoinHandle to propagate fatal errors and a mpsc channel to bubble up non-fatal errors/warnings. PTAL.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
chihchiachen: I added a mpsc channel to bubble up non-fatal errors/warnings. PTAL.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -275,58 +318,239 @@ impl EmulatedCameraSession {
...
+                    }
+                } else if !should_arm && timer_armed {
+                    if let Err(e) = timer_fd.clear() {
+                        send_err(anyhow::Error::from(e).context("Failed to disarm TimerFd"));
+                    } else {
+                        timer_armed = false;
+                    }
+                }
+
+                if !timer_triggered || !guard.streaming {
```
bridadan, line 466: Wouldn't we only proceed here if the timer was triggered? 
```suggestion
                if !timer_triggered {
```

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -877,25 +1159,25 @@ where
...
-        session.queued_buffers.push_back(buffer.index() as usize);
-
+        host_buffer.set_state(BufferState::Incoming, width, height);
         let buffer = host_buffer.v4l2_buffer.clone();
-
-        if session.streaming {
-            session.process_queued_buffers(&mut self.evt_queue, &self.controls, self.width, self.height)?;
-        }
+        state.queued_buffers.push_back(buffer.index() as usize);
+        session.wakeup_evt.write(1).unwrap();
```
bridadan, line 1167: This will wake up the worker thread. If we're already streaming, what prevents the worker thread from rendering another frame? If the guest calls `qbuf` shortly after a timer fires, wouldn't that lead to a much shorter frame interval for that second frame?

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
chihchiachen: With an extra check for `guard.streaming` here, we won't render frames for
- buffers pre-queued before STREAMON
- race condition between kernel timer and STREAMOFF (where `timer_triggered == true && guard.streaming == false` can happen after STREAMOFF)

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
chihchiachen: I think L466 ensures the second frame wouldn't be rendered until the next timer fires.

bridadan: 🤦 ok yeah that makes sense. It was too late in the day and I read that line wrong. Thanks for your patience!

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
bridadan: Got it, makes sense

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -877,25 +1159,25 @@ where
...
-        host_buffer.set_state(BufferState::Incoming, self.width, self.height);
-        session.queued_buffers.push_back(buffer.index() as usize);
-
+        host_buffer.set_state(BufferState::Incoming, width, height);
         let buffer = host_buffer.v4l2_buffer.clone();
-
-        if session.streaming {
-            session.process_queued_buffers(&mut self.evt_queue, &self.controls, self.width, self.height)?;
-        }
+        state.queued_buffers.push_back(buffer.index() as usize);
```
bridadan: Just want to double check the `buffer.index()` here is equivalent to the `buffer.index()` from the original code.

In the original code, `buffer` corresponds to the parameter passed to this function. In this PR, `buffer` is the cloned `v4l2_buffer` from the `host_buffer`. It looks like the types match up so I doubt its an issue, but just wanted to raise it just in case.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
```diff
@@ -275,58 +318,239 @@ impl EmulatedCameraSession {
...
+        // Periodically drain diagnostics from the running worker thread without blocking.
+        self.drain_diag();
+        // If the worker thread terminated unexpectedly (e.g. fatal poll_ctx error or panic),
+        // join the thread, flush any final diagnostics, and fail fast with EIO.
+        if self.worker_handle.as_ref().is_some_and(|h| h.is_finished()) {
+            self.join_worker();
+            self.drain_diag();
+            return Err(libc::EIO);
+        }
         Ok(())
```
bridadan: Nit: I'd reduce the double `drain_diag()` call with something like:

```suggestion
        // If the worker thread terminated unexpectedly (e.g. fatal poll_ctx error or panic),
        // join the thread, flush any final diagnostics, and fail fast with EIO.
        if self.worker_handle.as_ref().is_some_and(|h| h.is_finished()) {
            self.join_worker();
            self.drain_diag();
            Err(libc::EIO)
        } else {
            self.drain_diag();
            Ok(())
       }
```

bridadan (approved):

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
chihchiachen: Yes, the `buffer.index()` is the same. But I agree swapping L1163 and L1164 in this PR is confusing (which is done to meet Rust's borrowing rule). Let me refactor a bit to make it less confusing.

base/cvd/cuttlefish/host/commands/vhost_user_media/emulated_camera_mplane/src/device.rs
chihchiachen: Good idea, thanks! Let me fix it.

jmacnak (approved):

PR #3001, opened 2026-08-11
Create a UniqueFd counterpart to SharedFD

Databean:
Sometimes the machinery of std::shared_ptr is not necessary, and the semantics of std::unique_ptr are simpler.

Bug: b/545278508

### DIFF
--- a/base/cvd/cuttlefish/common/libs/fs/BUILD.bazel
+++ b/base/cvd/cuttlefish/common/libs/fs/BUILD.bazel
@@ -55,7 +55,7 @@ cf_cc_library(
     }),
     deps = [
         "//cuttlefish/common/libs/fs:file_instance",
-        "//cuttlefish/common/libs/utils:environment",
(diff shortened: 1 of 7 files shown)


jemoreira (commented):
What do you plan to do with functions that currently accept a SharedFD by value but don't really care whether it's a SharedFD or UniqueFD?
Have you considered adding a `BaseFD` class with private constructors/factory methods to be the parent of those two? That would allow code that doesn't care about the fd type to just borrow `BaseFD&`, and code that cares to use one of the specialized classes to share ownership of the fd with `SharedFD` or take ownership of it with `UniqueFD`.

Databean:
In my opinion code that would accept a `BaseFD&` should instead accept a `FileInstance&` or one of the pure virtual interfaces in `//cuttlefish/io/io.h` like `Reader&` or `ReaderWriterSeeker&`.

jemoreira (approved):

PR #3007, opened 2026-08-12
Use UniqueFd::Accept rather than SharedFD::Accept

Databean:
Bug: b/545278508


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #3013, opened 2026-08-13
Change RecvAll to operate on FileInstance, and return a Result

Databean:
Bug: b/545794593


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #3014, opened 2026-08-13
Migrate from ReadAll to ReadToString

Databean:
Uses a Result type instead of a pointer, and operates on a pure virtual Reader rather than a SharedFD.

Bug: b/545794593

### DIFF
--- a/base/cvd/cuttlefish/common/libs/fs/shared_buf.cc
+++ b/base/cvd/cuttlefish/common/libs/fs/shared_buf.cc
@@ -16,7 +16,6 @@
 
 #include "cuttlefish/common/libs/fs/shared_buf.h"
 
-#include <sstream>
 #include <string>
 #include <string_view>
 #include <vector>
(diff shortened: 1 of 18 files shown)


jemoreira (approved):
Consider renaming `ReadToString` as `ReadAll` in a different PR to make it clear it doesn't just read a line or a word. The `ToString` part is also not necessary since the function signature has that information.

Databean:
> The `ToString` part is also not necessary since the function signature has that information.

I disagree, consider an alternative function:
```c++
Result<std::vector<uint8_t>> ReadAll(Reader&);
```
This seems equally valid to me, but the two would conflict during overload resolution because the name and parameters are the same.

Would you accept something like
```c++
template<typename T>
Result<T> ReadAll(Reader&);
```
with a few pre-declared instantiations? I still have to confirm how this interacts with the `misc-include-cleaner` lint which I have seen struggle with some advanced uses.

PR #3016, opened 2026-08-13
host_orchestrator: reject archive entries that escape the extraction dir

0405ysj:
The user-artifacts extractor (ExtractArtifact -> untar/unzip) joined attacker-controlled archive entry names onto the destination directory with no containment check and created symlink entries verbatim, allowing a crafted .tar.gz/.zip to write files outside the extraction directory

Add isSafeToExtract() and use it in untar and unzip to verify each entry's resolved path stays within the destination, and reuse it to reject symlink entries whose resolved target escapes it. Add regression tests for the tar-traversal, tar-symlink and zip-traversal vectors.

### DIFF
--- a/tools/buildutils/installbazel.sh
+++ b/tools/buildutils/installbazel.sh
@@ -21,7 +21,7 @@ set -e
 function install_bazel_x86_64() {
   echo "Installing bazel"
   apt install apt-transport-https curl gnupg -y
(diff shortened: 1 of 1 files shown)


0405ysj:
Closing this PR. This commit will be handled via different way when we release a new one based on this branch.

PR #3017, opened 2026-08-13
base: Support CrosVM auto-detection of MTE support

ptosi:
Make the currently boolean `--mte` CVD flag accept the value `auto` to only enable guest MTE when supported.

Map it to CrosVM's new `--cpus mte=[auto]` flag (https://crrev.com/c/8137608) but keep supporting `--mte` when `true`.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd_flags.cpp
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd_flags.cpp
@@ -334,7 +334,9 @@ DEFINE_vec(use_sdcard, CF_DEFAULTS_USE_SDCARD ? "true" : "false",
 DEFINE_vec(protected_vm, fmt::format("{}", CF_DEFAULTS_PROTECTED_VM),
            "Boot in Protected VM mode");
 
-DEFINE_vec(mte, CF_DEFAULTS_MTE, "Enable MTE, one of {true, false}.");
+DEFINE_vec(mte, CF_DEFAULTS_MTE,
+           "Enable MTE, one of {true, false, auto}. Auto enables MTE only when "
+           "the host supports it.");
 
 DEFINE_vec(enable_pkvm, fmt::format("{}", CF_DEFAULTS_ENABLE_PKVM),
            "Provision the guest to run pKVM so it can host its own protected "
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd_flags.cpp
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd_flags.cpp
@@ -334,7 +334,7 @@ DEFINE_vec(use_sdcard, CF_DEFAULTS_USE_SDCARD ? "true" : "false",
 DEFINE_vec(protected_vm, fmt::format("{}", CF_DEFAULTS_PROTECTED_VM),
            "Boot in Protected VM mode");
 
-DEFINE_vec(mte, fmt::format("{}", CF_DEFAULTS_MTE), "Enable MTE");
+DEFINE_vec(mte, CF_DEFAULTS_MTE, "Enable MTE, one of {true, false}.");
 
 DEFINE_vec(enable_pkvm, fmt::format("{}", CF_DEFAULTS_ENABLE_PKVM),
            "Provision the guest to run pKVM so it can host its own protected "
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
@@ -99,6 +99,7 @@
 #include "cuttlefish/host/libs/config/gpu_mode.h"
 #include "cuttlefish/host/libs/config/host_tools_version.h"
 #include "cuttlefish/host/libs/config/instance_nums.h"
+#include "cuttlefish/host/libs/config/mte.h"
 #include "cuttlefish/host/libs/config/secure_hals.h"
 #include "cuttlefish/host/libs/config/vmm_mode.h"
 #include "cuttlefish/host/libs/vhal_proxy_server/vhal_proxy_server_eth_addr.h"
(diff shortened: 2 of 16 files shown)


base/cvd/cuttlefish/host/libs/config/cuttlefish_config.h
```diff
@@ -455,7 +455,9 @@ class CuttlefishConfig {
     std::vector<std::string> extra_bootconfig_args() const;
     bool record_screen() const;
     std::string gem5_debug_file() const;
-    bool mte() const;
+    bool mte_auto_enabled() const;
```
dimorinny: I would prefer to avoid splitting the MTE state across 3 methods and instead rely on a single `mte()` returning an enum with `on`/`auto`/`off` states.

We already do something similar for `guest_renderer_preload` (see `guest_renderer_preload.h`), and it would also let us reject unknown values instead of silently treating them as off.

What do you think?

base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
```diff
@@ -839,7 +839,12 @@ Result<CuttlefishConfig> InitializeCuttlefishConfiguration(
 
     instance.set_record_screen(record_screen_vec[instance_index]);
     instance.set_gem5_debug_file(gem5_debug_file_vec[instance_index]);
-    instance.set_mte(mte_vec[instance_index]);
+    const std::string& mte_val = mte_vec[instance_index];
+    if (mte_val == "auto") {
+      instance.auto_enable_mte();
+    } else {
+      instance.force_enable_mte(CF_EXPECT(ParseBool(mte_val, "mte")));
```
dimorinny: nit: do you think it would be useful to fail early with a clear message for non-arm64, which is obviously incompatible with MTE? For example, this is what we do with enable_pkvm_vec below.

base/cvd/cuttlefish/host/libs/vm_manager/crosvm_builder.h
```diff
@@ -35,9 +35,10 @@ class CrosvmBuilder {
                              int exit_code);
   void AddControlSocket(const std::string&, const std::string&);
 
-  Result<void> AddCpus(size_t cpus, const std::string& vcpu_config_path);
-  Result<void> AddCpus(const Json::Value&);
-  void AddCpus(size_t cpus);
+  Result<void> AddCpus(size_t cpus, const std::string& vcpu_config_path,
```
dimorinny, line 39: `AddCpus(size_t, bool mte_auto = false)` now overlaps with `AddCpus(size_t, const std::string&, bool = false)` — a `const char*` argument picks the bool overload, since `const char*` -> `bool` is a standard conversion and wins over the user-defined conversion to `std::string`. I tried it with `-Wall -Wextra` on both gcc and clang and there's no warning: the vcpu config path is silently dropped and `mte=[auto]` silently added.

Current callers pass `std::string` so nothing is broken today, but would it be worth an enum (or a small options struct) instead of a bare bool to close the trap?

base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd_flags.cpp
```diff
@@ -334,7 +334,7 @@ DEFINE_vec(use_sdcard, CF_DEFAULTS_USE_SDCARD ? "true" : "false",
 DEFINE_vec(protected_vm, fmt::format("{}", CF_DEFAULTS_PROTECTED_VM),
            "Boot in Protected VM mode");
 
-DEFINE_vec(mte, fmt::format("{}", CF_DEFAULTS_MTE), "Enable MTE");
+DEFINE_vec(mte, CF_DEFAULTS_MTE, "Enable MTE");
```
dimorinny: nit: the help string is still "Enable MTE", so the new auto value isn't documented anywhere user-facing — cvd surfaces these via `--helpxml`, so this is all a user sees.

`--guest_renderer_preload` just above is a nice model: "…one of {auto, enabled, disabled}. Auto will choose whether or not to disable based on the gpu mode and guest hwui renderer." Maybe something like "Enable MTE, one of {auto, true, false}. Auto enables MTE only when the host supports it."?

dimorinny (changes requested):

base/cvd/cuttlefish/host/commands/assemble_cvd/assemble_cvd_flags.cpp
ptosi: Done

base/cvd/cuttlefish/host/libs/vm_manager/crosvm_builder.h
ptosi: Wow, nice catch!

base/cvd/cuttlefish/host/libs/config/cuttlefish_config.h
ptosi: I've introduced an `enum`, thanks!

base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
ptosi: Done, rejecting both `true` and `auto`

base/cvd/cuttlefish/host/libs/config/cuttlefish_config_instance.cpp
```diff
@@ -1024,12 +1026,33 @@ void CuttlefishConfig::MutableInstanceSpecific::set_gem5_debug_file(
...
+    return Mte::kAuto;
+  }
+  auto bool_res = CF_EXPECT(ParseBool(str, "mte"));
+  return bool_res ? Mte::kOn : Mte::kOff;
+}
+
 static constexpr char kMte[] = "mte";
-void CuttlefishConfig::MutableInstanceSpecific::set_mte(bool mte) {
-  (*Dictionary())[kMte] = mte;
+void CuttlefishConfig::MutableInstanceSpecific::set_mte(Mte mte) {
```
dimorinny, line 1029: We usually use `static_cast<int>` to store the enum values in json config. For example:

```
static constexpr char kTargetArch[] = "target_arch";
void CuttlefishConfig::MutableInstanceSpecific::set_target_arch(
    Arch target_arch) {
  (*Dictionary())[kTargetArch] = static_cast<int>(target_arch);
}
Arch CuttlefishConfig::InstanceSpecific::target_arch() const {
  return static_cast<Arch>((*Dictionary())[kTargetArch].asInt());
}
```

dimorinny (approved):

base/cvd/cuttlefish/process/command.h
```diff
@@ -160,6 +161,18 @@ class Command {
...
+  Command& AddJoinedParameter(const Container& container,
+                              Separator separator = ",") & {
+    command_.push_back(absl::StrJoin(container, separator));
+    return *this;
+  }
+  template <typename Container, typename Separator = const char*>
+  Command AddJoinedParameter(const Container& container,
+                             Separator separator = ",") && {
+    return std::move(AddJoinedParameter(container, separator));
+  }
```
Databean, line 174: In this case the caller could just as easily call `absl::StrJoin`, I'm not sure the usage merits an extra method in `Command`.

Databean (approved):

PR #3025, opened 2026-08-14
Make FileInstance movable

Databean:
This removes the need for UniqueFd, as FileInstance can be used directly in cases where UniqueFd would be used otherwise.

Suggested by jemoreira@.

Bug: b/545794593

### DIFF
--- a/base/cvd/cuttlefish/common/libs/fs/file_instance.cc
+++ b/base/cvd/cuttlefish/common/libs/fs/file_instance.cc
@@ -77,6 +77,22 @@ constexpr size_t kPreferredBufferSize = 8192;
 
 }  // namespace
 
+FileInstance::FileInstance() : FileInstance(-1, 0) {}
+
+FileInstance::FileInstance(FileInstance&& other) : FileInstance() {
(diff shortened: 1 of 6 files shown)


base/cvd/cuttlefish/common/libs/fs/file_instance.cc
```diff
@@ -77,6 +77,22 @@ constexpr size_t kPreferredBufferSize = 8192;
 
 }  // namespace
 
+FileInstance::FileInstance() : FileInstance(-1, 0) {}
+
+FileInstance::FileInstance(FileInstance&& other) : FileInstance() {
+  std::swap(fd_, other.fd_);
```
jemoreira, line 83: Shouldn't this have `Close()` like `operator=` does?

jemoreira (approved):

base/cvd/cuttlefish/common/libs/fs/file_instance.cc
Databean: It's a constructor, it doesn't refer to anything yet.

PR #3030, opened 2026-08-14
Create a common wrapper function for stat(2)

Databean:
This guarantees that TEMP_FAILURE_RETRY is called, and puts the error formatting text in a common place.

Bug: b/546692445

### DIFF
--- a/base/cvd/cuttlefish/posix/stat.cc
+++ b/base/cvd/cuttlefish/posix/stat.cc
@@ -24,6 +24,7 @@
 #include <string_view>
 
 #include "cuttlefish/posix/strerror.h"
+#include "cuttlefish/posix/temp_failure_retry.h"  // IWYU pragma: keep
 #include "cuttlefish/result/expect.h"
 #include "cuttlefish/result/result_type.h"
 
--- a/base/cvd/cuttlefish/posix/stat.cc
+++ b/base/cvd/cuttlefish/posix/stat.cc
@@ -14,29 +14,34 @@
  * limitations under the License.
  */
 
-#include "cuttlefish/files/copy_with_attributes.h"
+#include "cuttlefish/posix/stat.h"
 
 #include <errno.h>
 #include <sys/stat.h>
-#include <sys/types.h>
+#include <unistd.h>
 
 #include <string>
+#include <string_view>
 
-#include "cuttlefish/files/copy.h"
-#include "cuttlefish/posix/stat.h"
 #include "cuttlefish/posix/strerror.h"
 #include "cuttlefish/result/expect.h"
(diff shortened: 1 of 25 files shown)


base/cvd/cuttlefish/posix/stat.cc
```diff
@@ -0,0 +1,47 @@
...
+
+#include "cuttlefish/posix/strerror.h"
+#include "cuttlefish/result/expect.h"
+#include "cuttlefish/result/result_type.h"
+
+namespace cuttlefish {
+
+Result<struct stat> Stat(const char* path) {
+  struct stat ret;
+  int success = TEMP_FAILURE_RETRY(stat(path, &ret));
```
3405691582, line 35: Optionally: TEMP_FAILURE_RETRY is a glibcism. Consider stuffing an ifndef define (x) for this somewhere for now?

3405691582 (approved):

base/cvd/cuttlefish/posix/stat.cc
Databean: I tried making a dedicated header for this, i.e.
```c++
// temp_failure_retry.h
#ifndef TEMP_FAILURE_RETRY
#define TEMP_FAILURE_RETRY(exp) ...
#endif
```
(yeah not actually posix as you point out, location is arbitrary)

But the [include-cleaner](https://clangd.llvm.org/design/include-cleaner) lint rejects it at the including file
```
error: included header temp_failure_retry.h is not used directly [misc-include-cleaner,-warnings-as-errors]
   27 | #include "cuttlefish/posix/temp_failure_retry.h"
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   28 | #include "cuttlefish/result/expect.h"
```
This can be worked around either at the include site
```c++
#include "cuttlefish/posix/temp_failure_retry.h"  // IWYU pragma: keep
```
or once within the header
```c++
#include <unistd.h>  // IWYU pragma: export
```
Getting all the `<unistd.h>` symbols for `temp_failure_retry.h` is misleading, so I think the best I can do is putting a "cuttlefish unistd.h" somewhere, if you think that is an improvement. I've added this as a second commit.

3405691582: I kind of prefer the former -- calling it unistd.h reifies the thought that this is part of libc, which it technicaly isn't. But if we're matching bionic, it's probably fine?

Databean: Switched to the first option, created `temp_failure_retry.h`

PR #3034, opened 2026-08-17
Support target configuration preset in Cuttlefish cvd load JSON

anshu10sep:
## Summary
Adds support for specifying a target configuration preset in Cuttlefish `cvd load` JSON configurations.

- Adds `optional string config = 11;` to `message Instance` in `load_config.proto`.
- Updates `cf_configs_instances.cpp` to emit `--config=<preset>` per instance when configured, allowing `cvd load` and cloud orchestrators to pass configuration presets natively in JSON.
- Adds comprehensive unit test coverage in `flags_parser_test.cc` for single-instance, multi-instance, and partial multi-instance scenarios.

## Test
Added unit tests in `flags_parser_test.cc`:
- `ConfigFlagsParserTest.ParseSingleInstanceConfig`
- `ConfigFlagsParserTest.ParseMultiInstanceConfig`
- `ConfigFlagsParserTest.ParseMultiInstanceConfigPartial`

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/parser/cf_configs_instances.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/parser/cf_configs_instances.cpp
@@ -34,7 +34,30 @@
 
 namespace cuttlefish {
 
+namespace {
+
 using cvd::config::EnvironmentSpecification;
+using cvd::config::Instance;
+
+std::string ConfigFlag(const Instance& instance) {
+  if (instance.has_config()) {
+    return instance.config();
+  }
+  return "";
+}
+
+std::vector<std::string> GenerateConfigFlags(
+    const EnvironmentSpecification& cfg) {
+  for (const auto& instance : cfg.instances()) {
+    if (instance.has_config() && !instance.config().empty()) {
+      return std::vector<std::string>{
+          GenerateInstanceFlag("config", cfg, ConfigFlag)};
+    }
+  }
+  return {};
+}
+
+}  // namespace
 
 Result<std::vector<std::string>> GenerateInstancesFlags(
     const EnvironmentSpecification& cfg) {
@@ -46,6 +69,7 @@ Result<std::vector<std::string>> GenerateInstancesFlags(
   res = MergeResults(std::move(res), CF_EXPECT(GenerateVmFlags(cfg)));
   res = MergeResults(std::move(res), GenerateConnectivityFlags(cfg));
   res = MergeResults(std::move(res), CF_EXPECT(GenerateMediaFlags(cfg)));
+  res = MergeResults(std::move(res), GenerateConfigFlags(cfg));
   return res;
 }
 
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/parser/flags_parser_test.cc
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/parser/flags_parser_test.cc
@@ -611,4 +611,85 @@ TEST(ConnectivityFlagsParserTest, ParseModemSimulatorSimTypeInvalidInt) {
       << "modem_simulator_sim_type flag is missing or wrongly formatted";
 }
 
+TEST(ConfigFlagsParserTest, ParseSingleInstanceConfig) {
+  const char* test_string = R""""(
+{
+    "instances" :
+    [
+        {
+          "config": "phone"
+        }
+    ]
+}
+  )"""";
+
+  Json::Value json_configs;
+  std::string json_text(test_string);
+
+  EXPECT_TRUE(ParseJsonString(json_text, json_configs))
+      << "Invalid Json string";
+  const Result<std::vector<std::string>> serialized_data =
+      LaunchCvdParserTester(json_configs);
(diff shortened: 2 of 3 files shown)


base/cvd/cuttlefish/host/commands/cvd/cli/parser/flags_parser_test.cc
```diff
@@ -611,4 +611,85 @@ TEST(ConnectivityFlagsParserTest, ParseModemSimulatorSimTypeInvalidInt) {
...
+  )"""";
+
+  Json::Value json_configs;
+  std::string json_text(test_string);
+
+  EXPECT_TRUE(ParseJsonString(json_text, json_configs))
+      << "Invalid Json string";
+  const Result<std::vector<std::string>> serialized_data =
+      LaunchCvdParserTester(json_configs);
+  EXPECT_THAT(serialized_data, IsOk());
```
Databean: This should be `ASSERT_THAT`. `EXPECT_THAT` doesn't exit the test, and `*serialized_data` will crash if `IsOk()` fails.

Same issue in the following tests.

base/cvd/cuttlefish/host/commands/cvd/cli/parser/cf_configs_instances.cpp
```diff
@@ -46,6 +53,18 @@ Result<std::vector<std::string>> GenerateInstancesFlags(
...
+
+  bool has_any_config = false;
+  for (const auto& instance : cfg.instances()) {
+    if (instance.has_config() && !instance.config().empty()) {
+      has_any_config = true;
+      break;
+    }
+  }
+  if (has_any_config) {
+    res.emplace_back(GenerateInstanceFlag("config", cfg, ConfigFlag));
```
cjreynol: For consistency, this line should also be in the form `MergeResults(std::move(res), ...);`

base/cvd/cuttlefish/host/commands/cvd/cli/parser/cf_configs_instances.cpp
```diff
@@ -46,6 +53,18 @@ Result<std::vector<std::string>> GenerateInstancesFlags(
   res = MergeResults(std::move(res), CF_EXPECT(GenerateVmFlags(cfg)));
   res = MergeResults(std::move(res), GenerateConnectivityFlags(cfg));
   res = MergeResults(std::move(res), CF_EXPECT(GenerateMediaFlags(cfg)));
+
+  bool has_any_config = false;
```
cjreynol: NIT: would you move this new block into a helper named something like `Generate<Field>Flags`, to match the other fields in this helper.  I understand this is a simple case with a string instead of a full message field, but it keeps `GenerateInstancesFlags` easy to read when it is (mostly) 1-to-1 with calls per proto field.

base/cvd/cuttlefish/host/commands/cvd/cli/parser/cf_configs_instances.cpp
```diff
@@ -36,6 +36,13 @@ namespace cuttlefish {
 
 using cvd::config::EnvironmentSpecification;
 
+static std::string ConfigFlag(const cvd::config::Instance& instance) {
```
cjreynol: NIT: would you put this definition and the `using cvd::config::EnvironmentSpecification;` statement on line 37 into an anonymous namespace block (`namespace { ... }  // namespace`) instead?

The real problem here is the existing `using` statement being exposed to the entire `cuttlefish` namespace, but since we are adding more logic it might as well go inside the anonymous namespace as well.

cjreynol (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/parser/flags_parser_test.cc
anshu10sep: Done. Updated to `ASSERT_THAT` across all config flags parser tests.

base/cvd/cuttlefish/host/commands/cvd/cli/parser/cf_configs_instances.cpp
anshu10sep: Done. Moved the using declarations and helper definitions into an anonymous namespace (`namespace { ... }`).

base/cvd/cuttlefish/host/commands/cvd/cli/parser/cf_configs_instances.cpp
anshu10sep: Done. Extracted the block into `GenerateConfigFlags(const EnvironmentSpecification& cfg)` helper.

base/cvd/cuttlefish/host/commands/cvd/cli/parser/cf_configs_instances.cpp
anshu10sep: Done. Updated to `res = MergeResults(std::move(res), GenerateConfigFlags(cfg));`.

cjreynol:
We should remove the merge commit from the PR

cjreynol:
Rebased the branch to remove the merge commit.  Once this merges I will backport the change to the `v1.55` branch.

PR #3036, opened 2026-08-17
Support seccomp policies baked into crosvm

Databean:
On Android, the seccomp policy files used for crosvm sandbox mode are not baked into the crosvm executable. We distribute them instead through `cvd-host_package.tar.gz`.

As we create our own crosvm build, we can bake these into the executable instead which is more convenient.

Bug: b/548053644


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #3037, opened 2026-08-17
Rename FileInstance to Fd

Databean:
An earlier commit introduced UniqueFd as a std::unique_ptr equivalent to std::shared_ptr for SharedFD. However, Making `FileInstance` movable makes `UniqueFd` redundant. Before using `FileInstance` everywhere, this is the best time to rename it something more ergonomic.

Bug: b/545278508


cjreynol (approved):

jemoreira (approved):

Databean:
Thanks for reviewing!

PR #3047, opened 2026-08-18
Support gs:// and https:// build sources in cvd fetch and cvd load

larsers:
Replaces #2238.

```
# a dist directory published to a bucket that artifacts are discovered from
# the listing, including the host package:
cvd fetch --default_build=gs://my-bucket/dist/

# artifacts published to any web server, e.g. by a CI system; the
# default build here names one zip, so the host package has its own URL:
cvd fetch \
  --default_build=https://ci.example.com/builds/42/aosp_cf_x86_64_phone-img-9099.zip \
  --host_package_build=https://ci.example.com/builds/42/cvd-host_package.tar.gz

# the same strings work in cvd load configs:
"disk": { "default_build": "gs://my-bucket/dist/" }
```

A URL build source is an artifact namespace, the same way an Android Build target or a local directory is one. `GcsBuild` and `HttpBuild` join the `Build` variant behind the existing `BuildApi` interface. A new `CompositeBuildApi` routes each build to its API, so the fetch logic has no URL-conditional branches. A URL ending in '/' names the directory holding a build's artifacts and for `gs://` the contents come from the bucket listing. A URL naming one object is a build of that artifact, where the existing `{selector}` syntax picks a member out of a zip and `#sha256=<hex>` pins the digest. Both forms are probed at `GetBuild` time, so a bad bucket or an expired signature fails before any bytes move. The tree already handles `gs://` prefixes this way in `LuciBuildApi`, this generalizes that to user supplied locations.

Credentials for `gs://` resolve independently of the Android Build credential: `cvd login` token, service account file, `~/.boto`, GCE metadata, then anonymous. Public buckets need no setup. `https://` sends no credentials and a pre-signed URL carries its own. Query strings are kept out of logs and records. Downloads are cached per artifact, keyed on the object generation or ETag. They are verified against `#sha256=` and the GCS `md5Hash`, and an interrupted download resumes with `Range`/`If-Range`. A plain `https://` directory has no listing, so its artifacts are named explicitly and are not cached.

The standalone fixes the feature depends on merged in #3070. This PR holds the remaining feature commits.

Unit tests cover the new code. `e2etests/cvd/cvd_url_build_tests` is a hermetic e2e suite driving `cvd fetch` and `cvd load` against a local TLS server so it needs no Android Build access.


jemoreira:
> Replaces #2238.
> 
> ```
> # a dist directory published to a bucket that artifacts are discovered from
> # the listing, including the host package:
> cvd fetch --default_build=gs://my-bucket/dist/
> 
> # artifacts published to any web server, e.g. by a CI system:
> cvd fetch \
>   --default_build=https://ci.example.com/builds/42/aosp_cf_x86_64_phone-img-9099.zip \
>   --host_package_build=https://ci.example.com/builds/42/cvd-host_package.tar.gz
> 
> # the same strings work in cvd load configs:
> "disk": { "default_build": "gs://my-bucket/dist/" }
> ```

There is a problem with this example (haven't seen the code yet): --default_build is expected to have everything, including the host packages, so if `--host_package` is not provided cvd should look for it in the default build.

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
```diff
@@ -497,7 +508,8 @@ Result<FetchResult> Fetch(const FetchFlags& flags,
                                target.builds, config, tracer);
     LOG(INFO) << "Starting fetch to \"" << target.directories.root << "\"";
     CF_EXPECT(FetchTarget(fetch_context, target.download_flags,
-                          flags.keep_downloaded_archives));
+                          flags.keep_downloaded_archives,
+                          target.builds.otatools_inferred));
```
jemoreira: nit: it should be possible to avoid passing `otatools_inferred` as an extra parameter all over the place by making the otatools build something like `std::optional<struct{Build build; bool inferred}>`. I believe `FetchContext` can just copy that optional and its `OtaToolsBuild` can return it too.

jemoreira (commented):
Reviewed the first 7 commits so far, those can be merged independently of the rest, so consider making another PR with just those.

base/cvd/cuttlefish/host/commands/cvd/fetch/build_api_credentials.cc
```diff
@@ -136,19 +138,67 @@ Result<std::unique_ptr<CredentialSource>> GetCredentialSource(
...
+      flags.credential_flags.credential_filepath,
+      flags.credential_flags.service_account_filepath, scope));
+}
+
+bool IsRunningOnGce() {
+  Result<std::string> product_name =
+      ReadFileContents("/sys/class/dmi/id/product_name");
+  return product_name.has_value() &&
+         absl::StrContains(*product_name, "Google Compute Engine");
+}
```
Databean, line 174: We already have [`DetectGceEnvironment`](https://github.com/google/android-cuttlefish/blob/97ac71d949284317023f99222036a6ffbb54a512/base/cvd/cuttlefish/host/libs/metrics/gce_environment.cc#L87) which instead tries to determine if `http://metadata.google.internal` is reachable. Can these two approaches be consolidated?

base/cvd/cuttlefish/host/commands/cvd/fetch/build_api_credentials_test.cpp
```diff
@@ -0,0 +1,265 @@
...
+std::string ServiceAccountJson() {
+  return fmt::format(
+      R"({{"client_email": "test@example.com", "private_key": "{}"}})",
+      kTestPrivateKey);
+}
+
+// The scope a service account asks for travels inside the base64url claim set
+// of the assertion it signs.
+std::string AssertionClaims(const std::string& request_body) {
+  std::string_view body = request_body;
```
Databean: Can this accept a `std::string_view` parameter instead since it immediately converts the parameter and never uses the original?

base/cvd/cuttlefish/host/libs/web/url_download.cpp
```diff
@@ -0,0 +1,182 @@
...
+        .url = download.url,
+        .headers = download.headers,
+    };
+    if (offset > 0) {
+      request.headers.push_back(fmt::format("Range: bytes={}-", offset));
+      if (!download.if_range.empty()) {
+        request.headers.push_back(
+            absl::StrCat("If-Range: ", download.if_range));
+      }
+    }
```
Databean, line 85: Downloading files with HTTP range requests is already implemented in [`ZipSourceFromUrl`](https://github.com/google/android-cuttlefish/blob/97ac71d949284317023f99222036a6ffbb54a512/base/cvd/cuttlefish/host/libs/zip/remote_zip.cc#L73), is it possible to reuse this?

There's also [`LazilyLoadedFile`](https://github.com/google/android-cuttlefish/blob/97ac71d949284317023f99222036a6ffbb54a512/base/cvd/cuttlefish/io/lazily_loaded_file.h) for data loaded piecemeal from a backing reader and cached in a source.

base/cvd/cuttlefish/host/libs/web/url_download.cpp
```diff
@@ -0,0 +1,182 @@
...
+  struct stat by_path = {};
+  if (stat(path.c_str(), &by_path) != 0) {
+    return false;
+  }
+  android::base::unique_fd raw(fd->UNMANAGED_Dup());
+  CF_EXPECTF(raw.get() >= 0, "Could not duplicate the descriptor of '{}'",
+             path);
+  struct stat by_fd = {};
+  CF_EXPECTF(fstat(raw.get(), &by_fd) == 0, "Could not read '{}' - {}", path,
+             strerror(errno));
```
Databean: Can the `Fd` type get an `Result<struct stat> Fstat()` function so that this can call `CF_EXPECT(fd->Fstat())` instead?

base/cvd/cuttlefish/host/libs/web/url_download.cpp
```diff
@@ -0,0 +1,182 @@
...
+    CF_EXPECT(FullDownload(http_client, download, path));
+    return {};
+  }
+
+  const std::string part_path = absl::StrCat(path, ".part");
+  for (int attempt = 0; attempt < kLockAttempts; attempt++) {
+    SharedFD part = SharedFD::Open(part_path, O_RDWR | O_CREAT, 0644);
+    CF_EXPECTF(part->IsOpen(), "Could not open '{}' - {}", part_path,
+               part->StrError());
+    CF_EXPECTF(part->Flock(LOCK_EX), "Could not lock '{}'", part_path);
```
Databean: Is there multithreading happening around downloading these files, or is this about multiple concurrent user invocations? What are these locks intended to protect against?

base/cvd/cuttlefish/host/libs/web/gcs_build_api_test.cpp
```diff
@@ -0,0 +1,481 @@
...
+  ASSERT_THAT(build, IsOk());
+
+  TemporaryDir target_directory;
+  std::string expected_path =
+      std::string(target_directory.path) + "/misc_info.txt";
+  EXPECT_THAT(api.DownloadFile(*build, target_directory.path, "misc_info.txt"),
+              IsOkAndValue(expected_path));
+
+  std::string contents;
+  ASSERT_TRUE(android::base::ReadFileToString(expected_path, &contents));
```
Databean: Sometimes this PR uses `android::base::ReadFileToString` and sometimes it uses [`ReadFileContents`](https://github.com/google/android-cuttlefish/blob/97ac71d949284317023f99222036a6ffbb54a512/base/cvd/cuttlefish/common/libs/utils/files.h#L58). Can this change use `ReadFileContents` throughout?

base/cvd/cuttlefish/host/commands/cvd/cli/parser/fetch_config_parser_test.cc
```diff
@@ -0,0 +1,366 @@
...
+}
+  )"""";
+
+  Json::Value json_configs;
+  std::string json_text(test_string);
+  ASSERT_TRUE(ParseJsonString(json_text, json_configs));
+
+  Result<std::vector<std::string>> flags = FetchCvdParserTester(json_configs);
+  ASSERT_THAT(flags, IsOk());
+  EXPECT_TRUE(flags->empty());
```
Databean: nit: `EXPECT_THAT(flags, IsOkAndValue(std::vector<std::string>{}));`

or maybe `EXPECT_THAT(flags, IsOkAndValue({}))`, not sure if that will work.

base/cvd/cuttlefish/host/commands/cvd/fetch/build_api_credentials_test.cpp
```diff
@@ -0,0 +1,265 @@
...
+  }
+
+  void TearDown() override {
+    setenv("HOME", previous_home_.c_str(), /* overwrite */ 1);
+  }
+
+  std::string WriteTempFile(const std::string& name,
+                            const std::string& contents) {
+    std::string path = fmt::format("{}/{}", temp_dir_.path, name);
+    std::ofstream file(path);
```
Databean: Elsewhere this PR uses `android::base::WriteStringToFile`, but here it uses `fstream`.

Can it use [`WriteNewFile`](https://github.com/google/android-cuttlefish/blob/97ac71d949284317023f99222036a6ffbb54a512/base/cvd/cuttlefish/common/libs/utils/files.h#L59)?

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
```diff
@@ -352,9 +405,17 @@ Result<void> FetchAndroidEfiLoaderTarget(FetchBuildContext& context) {
...
+                                 bool build_was_inferred) {
   FetchArtifact otatools = context.Artifact("otatools.zip");
-  CF_EXPECT(otatools.Download());
+  Result<void> downloaded = otatools.Download();
+  if (!downloaded.has_value() && build_was_inferred) {
+    LOG(WARNING) << "No otatools.zip in " << context
+                 << ", which was not requested but picked from the other "
+                    "builds, continuing without it";
+    return {};
+  }
```
Databean: Originally in the scenarios where otatools is implicitly requested, a later step in `assemble_cvd` depends on being able to invoke executables in otatools to combine images. If there is a mismatch on cases where they are inferred and cases where they are necessary, it should be removed from the set of inferred cases.

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
Databean: Originally in the scenarios where otatools is implicitly requested, a later step in `assemble_cvd` depends on being able to invoke executables in otatools to combine images. If there is a mismatch on cases where they are inferred and cases where they are necessary, it should be removed from the set of inferred cases, rather than this distinction of inferred vs not inferred.

base/cvd/cuttlefish/host/libs/web/url_namespace.cpp
```diff
@@ -0,0 +1,138 @@
...
+constexpr char kTargetFilesKind[] = "target_files";
+constexpr char kOtatoolsKind[] = "otatools";
+constexpr std::string_view kImgSuffix = "-img.zip";
+
+// Android Build names an archive "<product>-<kind>-<id>.zip"; a republished
+// artifact set drops the build id.
+bool NamesZipKind(std::string_view name, std::string_view kind) {
+  return absl::StrContains(name, absl::StrCat("-", kind, "-")) ||
+         absl::EndsWith(name, absl::StrCat("-", kind, ".zip"));
+}
```
Databean: It looks like this function distinguishes between two categories of builds, but the function name and return values of `true` and `false` do not describe the categories. Only the function comment describes the categories.

This should either return an `enum class` or the categories should be clear from the function name.

base/cvd/cuttlefish/host/libs/web/android_build_string.cpp
```diff
@@ -99,6 +161,49 @@ Result<DirectoryBuildString> ParseDirectoryBuildString(
   return result;
 }
 
+Result<BuildString> ParseUrlBuildString(std::string_view scheme,
+                                        std::string_view build_string) {
+  CF_EXPECTF(scheme != "http",
+             "Cleartext 'http://' build sources are not supported, use "
+             "'https://' instead.  Input: '{}'",
+             ScrubUrl(build_string));
```
Databean: This is redundant with the following `CF_EXPECTF` call.

base/cvd/cuttlefish/host/libs/web/composite_build_api.h
```diff
@@ -0,0 +1,51 @@
...
+                                   const std::string& artifact_name) override;
+
+  Result<SeekableZipSource> FileReader(
+      const Build& build, const std::string& artifact_name) override;
+
+ private:
+  BuildApi& ApiFor(const BuildString& build_string);
+  BuildApi& ApiFor(const Build& build);
+
+  BuildApi& android_;
```
Databean: It might be worth splitting up `AndroidBuildApi` at this point between `DeviceBuild` and `DirectoryBuild`, so it's not a special case handling two different types of `Build`s.

base/cvd/cuttlefish/host/libs/web/composite_build_api.h
```diff
@@ -0,0 +1,51 @@
...
+#include "cuttlefish/host/libs/web/android_build.h"
+#include "cuttlefish/host/libs/web/android_build_string.h"
+#include "cuttlefish/host/libs/web/build_api.h"
+#include "cuttlefish/host/libs/zip/libzip_cc/seekable_source.h"
+#include "cuttlefish/result/result.h"
+
+namespace cuttlefish {
+
+// Sends every build to the API that owns the source it names.
+class CompositeBuildApi : public BuildApi {
```
Databean: It kind of breaks the `BuildApi` interface to have many arguments that fail at runtime for particular `BuildApi` implementations. See the [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle).

What do you think of changing the `BuildApi` interface to be templated, like

```c++
template<typename BuildStringType, typename BuildType>
class BuildApi {
public:
  Result<BuildType> GetBuild(const BuildStringType&);
  // etc
};

class CompositeBuildApi : public BuildApi<BuildString, Build> {
  // ...
}

class HttpBuildApi : public BuildApi<HttpBuildString, HttpBuild> {
  // ...
}
```

Then `CompositeBuildApi` can still direct calls to the correct implementations, and the implementations don't have to consider invalid alternatives.

The most difficult part will be dealing with `CachingBuildApi`, which will probably need to be templated as well with more of the implementation moved into the header.

Databean (commented):
Haven't reviewed everything, but left some comments.

> The first six commits are standalone fixes the feature depends on and each is reviewable on its own.

Can they be separate PRs?

larsers:
> > Replaces #2238.
> > ```
> > # a dist directory published to a bucket that artifacts are discovered from
> > # the listing, including the host package:
> > cvd fetch --default_build=gs://my-bucket/dist/
> > 
> > # artifacts published to any web server, e.g. by a CI system:
> > cvd fetch \
> >   --default_build=https://ci.example.com/builds/42/aosp_cf_x86_64_phone-img-9099.zip \
> >   --host_package_build=https://ci.example.com/builds/42/cvd-host_package.tar.gz
> > 
> > # the same strings work in cvd load configs:
> > "disk": { "default_build": "gs://my-bucket/dist/" }
> > ```
> 
> There is a problem with this example (haven't seen the code yet): --default_build is expected to have everything, including the host packages, so if `--host_package` is not provided cvd should look for it in the default build.

It does work that way. With no `--host_package_build` the host package comes from the default build: `GetHostBuild` falls back to the first target's default build, and for the directory form the package is found in the listing, checked before any download starts (`CheckHostPackagePresent`). The https example names a single img zip as its default build, so that build holds no host package and the example passes one explicitly. I clarified the example comment.

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
larsers: Restructured in #3070 , which carries this commit: `Builds::otatools` is now `std::optional<OtaTools>` holding the build and the inferred flag, and `FetchContext::OtaToolsBuild()` returns both, so the extra parameters are gone...

larsers:
> Reviewed the first 7 commits so far, those can be merged independently of the rest, so consider making another PR with just those.

Split out as #3070. This PR rebases onto it after it merges, dropping to just the feature commits.

larsers:
> Haven't reviewed everything, but left some comments.
> 
> > The first six commits are standalone fixes the feature depends on and each is reviewable on its own.
> 
> Can they be separate PRs?

Yes. The first seven commits are now #3070, as requested. Each commit there is a standalone fix, so if one PR per fix is easier to review I can split them further. This branch gets rebased and pushed once that PR merges, and that push includes the fixes from the inline comments here.

base/cvd/cuttlefish/host/commands/cvd/fetch/build_api_credentials.cc
larsers: Hmm..I guess the detection code could share a home, but both checks are maybe worth keeping? 

Afaik, they answer different questions. `DetectGceEnvironment` queries the metadata server, which metrics needs anyway for the zone and project id, and it reports an error rather than "not GCE" when the request fails. The credential ladder only needs the static "is this machine GCE" fact to decide whether the metadata token endpoint is worth trying, and the DMI product name answers that without a network round trip or a timeout...

Wdyt about moving GCE detection out of host/libs/metrics into a shared library, keep both checks there, and let metrics use the DMI check to skip the HTTP probe on machines that are not GCE? 

I would prefer do that as a follow up to keep this PR contained...

base/cvd/cuttlefish/host/commands/cvd/fetch/build_api_credentials_test.cpp
larsers: Done.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
```diff
@@ -143,9 +144,15 @@ Result<Build> AndroidBuildApi::GetBuild(
 }
 
 Result<Build> AndroidBuildApi::GetBuild(const BuildString& build_string) {
-  Result<Build> result =
-      std::visit([this](auto&& arg) { return GetBuild(arg); }, build_string);
-  return CF_EXPECT(std::move(result));
+  if (const auto* device = std::get_if<DeviceBuildString>(&build_string)) {
```
jemoreira: nit: avoid the use of `auto` for simpler known types. https://google.github.io/styleguide/cppguide.html#Type_deduction

base/cvd/cuttlefish/host/libs/web/url_download.cpp
larsers: `ZipSourceFromUrl` is what I decided to use for random access...The FileReader on both new APIs goes through it, with a new overload that takes a known size to skip the HEAD request, and `CachingBuildApi::FileReader` wraps the source in `CacheZipSource`, which is built on `LazilyLoadedFile`. 

This file covers the other access pattern, downloading a whole artifact to disk, so one streaming GET instead of a ranged request per read chunk, resume across process restarts keyed on If-Range, a lock against concurrent invocations, and error response bodies never reaching the artifact path. Neither existing helper does that job, and one ranged request per libzip read would be slow for multi-gigabyte image zips.

It does reuse more than it did when you read it first time around... a download with nothing to resume against is now just a call to `HttpGetToFile`, and only the resume path keeps its own request loop, because `HttpGetToFile` hides the partial file that the offset and the lock both need...

base/cvd/cuttlefish/host/libs/web/url_download.cpp
larsers: Done... Fd has an `Fstat()` now, in its own commit, and this file uses Fd instead of SharedFD, so the dup is gone.

base/cvd/cuttlefish/host/libs/web/url_download.cpp
larsers: No threads... A fetch downloads sequentially. The locks are for concurrent cvd invocations. The artifact cache is shared, so anything that brings up several devices on one host at once, so a test harness, CI, or a service driving cvd, can have more than one fetch running against it. Cache paths are keyed on the object generation, so two fetches of the same build land on the same `.part` file.... So flock picks one downloader and the others wake up to either the finished file or a resumable part file. I can add a comment saying so.

base/cvd/cuttlefish/host/libs/web/gcs_build_api_test.cpp
larsers: Done...Changed to `ReadFileContents` throughout the new tests.

base/cvd/cuttlefish/host/commands/cvd/cli/parser/fetch_config_parser_test.cc
larsers: Fixed... `IsOkAndValue({})` does not compile because the matcher parameter cannot be deduced from a braced list, so it spells out `std::vector<std::string>{}`.

base/cvd/cuttlefish/host/commands/cvd/fetch/build_api_credentials_test.cpp
larsers: Done... `WriteNewFile` here and in the other new tests that wrote files through `android::base` or `fstream`.

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
larsers: I tried to understand where those executables come from... From what I can tell `BuildSuperImage` in `super_image_mixer.cc` runs `build_super_image` from `HostBinaryPath` and passes `--path=DefaultHostArtifactsPath("")`, and both of those resolve to the host package directory rather than the `otatools/` directory fetch extracts. 

As far as I can tell nothing in `assemble_cvd` reads that directory at all, and the `TODO` in `RebuildSuperImage` about `otatools/bin/merge_target_files` suggests the same. The rebuild also only runs when there is a system build. So in the kernel plus default case we infer an `otatools` build that nothing goes on to use, which looks like the mismatch I think you had in mind.

If that is right, this tolerance is not needed. The system case can stay strict, and this part can go away along with the inference. The tolerance is also indiscriminate... It cannot tell a missing `otatools.zip` from a download that failed for some other reason...

Is there any external consumer that depends on a kernel mix fetch producing the `otatools` directory? 

If not, I'll swap this commit in #3070 for one that drops the inference.

base/cvd/cuttlefish/host/libs/web/url_namespace.cpp
larsers: It is a predicate on artifact names rather than build categories... So both parts accept the same kind of archive, named with or without a build id. I've renamed it to `NameMatchesZipKind` with a comment that spells out the two accepted shapes. 

If you would rather have an enum for the two name shapes I can add one, though nothing downstream distinguishes them.

base/cvd/cuttlefish/host/libs/web/android_build_string.cpp
larsers: Done... Folded into the scheme allowlist check, keeping the cleartext http hint in its message.

base/cvd/cuttlefish/host/libs/web/composite_build_api.h
larsers: Sure...As this is the last place where one class handles two `Build` types. The change I made for your `BuildApi` interface comment already narrows it anyway... So the Android, GCS and HTTP APIs no longer implement the shared interface and each takes its own build type, and `AndroidBuildApi` handles its two through typed overloads. What is left deciding between build kinds is the composite's dispatch.

Splitting `DeviceBuild` and `DirectoryBuild` into separate classes means reworking code that predates this work... If you would rather have it here, I can do that instead?

base/cvd/cuttlefish/host/libs/web/composite_build_api.h
larsers: You're right... The invalid alternative part should probably not exist. I went one step simpler than the template, the leaf classes no longer implement `BuildApi` at all. Each exposes only methods in its own types, which already existed as private overloads, so `CompositeBuildApi` does the variant dispatch with `std::visit` and stays the only `BuildApi` the fetch code sees; and the caching wrap moved above the composite, so `CachingBuildApi` and the `BuildApi` interface are unchanged at the variant level and nothing needed templating. The invalid combinations are now unrepresentable at compile time.

If you would still prefer to have the templated interface, (e.g. so the caching layer can wrap each leaf separately again), I can switch to it...

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
larsers: Sure, happy to change it... But could you please clarify how you'd like `auto` handled in this codebase generally, so I can use it consistently rather than guess case by case?

base/cvd/cuttlefish/host/libs/web/android_build_string.h
```diff
@@ -49,7 +49,33 @@ struct DirectoryBuildString {
...
+  std::string url;
+  std::optional<std::string> filepath;
+  std::optional<std::string> sha256;
+
+  auto operator<=>(const GcsBuildString&) const = default;
+};
+
+std::ostream& operator<<(std::ostream& out, const GcsBuildString& build_string);
+
+// Names the same two forms over `https://`. Any credential travels in the URL
```
jemoreira: same as what? Single file and archive? Does the ending `/` also apply here?

base/cvd/cuttlefish/host/libs/web/android_build_string.cpp
```diff
@@ -26,18 +26,66 @@
...
+  return scheme;
+}
+
+bool IsSha256Fragment(std::string_view fragment) {
+  constexpr std::string_view kPrefix = "#sha256=";
+  constexpr size_t kHexDigits = 64;
+  if (!fragment.starts_with(kPrefix)) {
+    return false;
+  }
+  std::string_view digest = fragment.substr(kPrefix.size());
```
jemoreira: nit: `absl::ConsumePrefix` lets you combine the check and substring into a single function call.

base/cvd/cuttlefish/host/libs/web/android_build_string.cpp
```diff
@@ -26,18 +26,66 @@
...
+  return digest.size() == kHexDigits &&
+         digest.find_first_not_of("0123456789abcdefABCDEF") ==
+             std::string_view::npos;
+}
+
+// A fragment is a suffix by definition, so a '#' anywhere else is an error
+// rather than something silently kept in the URL.
+Result<std::pair<std::string_view, std::optional<std::string>>>
```
jemoreira: Any reason the fragment is returned as a string instead of a string_view?

base/cvd/cuttlefish/host/libs/web/android_build_string.cpp
```diff
@@ -26,18 +26,66 @@
...
+#include "cuttlefish/host/libs/web/http_client/scrub_secrets.h"
 #include "cuttlefish/result/result.h"
 
 namespace cuttlefish {
 
 namespace {
 
+// The "<scheme>" of a "<scheme>://" build string, which is what separates a
+// URL build source from a branch, a build id or a directory path.
+std::optional<std::string_view> UrlScheme(std::string_view build_string) {
```
jemoreira: nit: This can be implemented in one pass:
```
size_t separator = build_string.find_first_not_of(kSchemeCharacters);
if (separator == std::string_view::npos || !build_string.substr(separator).starts_with("://")) {
  return std::nullopt;
}
return build_string.substr(0, separator);
```

base/cvd/cuttlefish/host/libs/web/android_build_string.cpp
```diff
@@ -99,6 +147,51 @@ Result<DirectoryBuildString> ParseDirectoryBuildString(
...
+             "Unsupported URL scheme '{}'.  The supported URL schemes are "
+             "'gs://' and 'https://'.  Input: '{}'",
+             scheme, build_string);
+  // The comma-separated build flags split their values before parsing them,
+  // so a comma can only be rejected where the whole value is still visible.
+  CF_EXPECTF(build_string.find(',') == std::string_view::npos,
+             "URL build strings cannot contain a comma, because the "
+             "comma-separated build flags split their values on ','.  Input: "
+             "'{}'",
+             build_string);
```
jemoreira: I don't think it's the responsibility of this function to enforce no commas though. For example, if flag parsing changes and flag values could have commas this would be wrong (or at least inaccurate) in a pretty much unrelated file.

base/cvd/cuttlefish/host/libs/web/android_build_string.cpp
```diff
@@ -99,6 +147,51 @@ Result<DirectoryBuildString> ParseDirectoryBuildString(
...
+             "A URL build string cannot have characters after the closing "
+             "curly bracket.  Input: '{}'",
+             build_string);
+  auto [url, filepath] = CF_EXPECT(ParseFilepath(without_fragment));
+
+  size_t query = url.find('?');
+  bool object_form = !url.substr(0, query).ends_with('/');
+  CF_EXPECTF(object_form || query == std::string::npos,
+             "Query strings are only supported on URLs naming an object, not "
+             "on a '/'-terminated directory.  Input: '{}'",
```
jemoreira, line 182: why are query strings only supported on object form urls? A comment would be useful here.

base/cvd/cuttlefish/host/libs/web/url_namespace.h
```diff
@@ -0,0 +1,50 @@
...
+#include <vector>
+
+#include "cuttlefish/result/result.h"
+
+namespace cuttlefish {
+
+struct ParsedUrl {
+  std::string authority;  // bucket or host
+  std::string path;       // no leading '/', ends with '/' in directory form
+  bool directory_form;
```
jemoreira: nit: Isn't this equivalent to `path.ends_with('/')`? It could be an inline method of `ParsedUrl` instead.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
Databean: https://abseil.io/tips/232 has more guidance if https://google.github.io/styleguide/cppguide.html#Type_deduction is not enough.

base/cvd/cuttlefish/host/libs/web/url_namespace.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+}  // namespace
+
+Result<ParsedUrl> ParseUrl(std::string_view url) {
+  size_t scheme_end = url.find("://");
+  CF_EXPECTF(scheme_end != std::string_view::npos, "'{}' is not a URL.",
+             ScrubUrl(url));
+  std::string_view authority_and_path = url.substr(scheme_end + 3);
```
jemoreira: There is already a function that extracts the scheme, it should be reused here or that one should call `ParseUrl`.

base/cvd/cuttlefish/host/libs/web/url_namespace.h
```diff
@@ -0,0 +1,50 @@
...
+  std::string query;  // without the '?'
+};
+
+Result<ParsedUrl> ParseUrl(std::string_view url);
+
+// The product of a "<product>-img-<id>.zip" or "<product>-img.zip" artifact,
+// or "url" for names that do not follow the Android Build convention.
+std::string DeriveProduct(std::string_view basename);
```
jemoreira: nit: Can this return `std::optional<std::string>` instead of returning "url" for names that don't represent a product?

base/cvd/cuttlefish/host/libs/web/url_namespace.cpp
```diff
@@ -0,0 +1,131 @@
...
+  return ParsedUrl{
+      .authority = std::string(authority_and_path.substr(0, authority_end)),
+      .path = std::string(path),
+      .directory_form = path.empty() || path.ends_with('/'),
+      .query = query,
+  };
+}
+
+std::string DeriveProduct(std::string_view basename) {
+  if (!absl::EndsWith(basename, ".zip")) {
```
jemoreira: nit: `absl::ConsumeSuffix` will check and remove the `.zip` suffix in one step.

base/cvd/cuttlefish/host/libs/web/url_namespace.cpp
```diff
@@ -0,0 +1,131 @@
...
+    return kUrlProduct;
+  }
+  size_t infix = basename.find("-img-");
+  if (infix != 0 && infix != std::string_view::npos) {
+    return std::string(basename.substr(0, infix));
+  }
+  if (absl::EndsWith(basename, kImgSuffix) &&
```
jemoreira: nit: `absl::ConsumeSuffix`

base/cvd/cuttlefish/host/libs/web/url_namespace.cpp
```diff
@@ -0,0 +1,131 @@
...
+    return std::string(basename.substr(0, basename.size() - kImgSuffix.size()));
+  }
+  return kUrlProduct;
+}
+
+Result<std::string> ResolveUrlZipName(const std::vector<std::string>& names,
+                                      bool closed,
+                                      const std::optional<std::string>& object,
+                                      std::string_view kind) {
+  if (object.has_value()) {
```
jemoreira: Look like this function should have two overloads:
```
Result<std::string> ResolveUrlZipName(const std::string& object, std::string_view kind);

Result<std::string> ResolveUrlZipName(const std::vector<std::string>& names, bool closed, std::string_view kind);
```

base/cvd/cuttlefish/host/libs/web/url_namespace.cpp
```diff
@@ -0,0 +1,131 @@
...
+    if (NamesZipKind(*object, kind)) {
+      return *object;
+    }
+    // A build that is one archive has no other candidate, but its name must
+    // not claim to be a different kind of archive.
+    CF_EXPECTF(kind == kImgKind && absl::EndsWith(*object, ".zip") &&
+                   !NamesZipKind(*object, kTargetFilesKind) &&
+                   !NamesZipKind(*object, kOtatoolsKind),
```
jemoreira: This implies there are only 3 possible kinds: image, otatools and target files. Right?

In that case, kind should be an enum, not a string.

base/cvd/cuttlefish/host/libs/web/url_namespace.h
```diff
@@ -0,0 +1,50 @@
...
+
+// The name of the `kind` zip of a URL build, where `names` holds the contents
+// of a closed namespace and `object` the single object of a build that names
```
jemoreira: What are "contents of a closed namespace"?

If the namespace has to be "closed" why is `closed` a parameter to this function?

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
Databean: I tried with kernel and system mixed builds, and it doesn't look like either depends on otatools being there in practice.
```bash
# system mix
$ cvd fetch \
  --default_build=git_main/cf_x86_64_only_phone-trunk_staging-userdebug \      
  --host_package_build=git_main/cf_x86_64_only_phone-trunk_staging-userdebug \
  --system_build=git_main/gsi_x86_64-trunk_staging-userdebug \          
  --target_directory=$HOME/dl_mix
$ cd $HOME/dl_mix
$ rm -rf otatools
$ cvd create --product_path=$PWD --host_path=$PWD
$ cvd clear
# kernel mix
$ cvd fetch \
   --default_build=git_main/aosp_cf_x86_64_only_phone-trunk_staging-userdebug \
   --target_directory=$HOME/dl_kernel \
   --kernel_build=aosp_kernel-common-android-mainline/kernel_virt_x86_64
$ cd $HOME/dl_mix
$ rm -rf otatools
$ cvd create --product_path=$PWD --host_path=$PWD
$ cvd clear
```
so it seems fine to only download `otatools` when explicitly requested, and avoid tracking the inferred state.

base/cvd/cuttlefish/host/libs/web/android_build_api.cpp
larsers: OK, thanks for the clarification!

base/cvd/cuttlefish/host/libs/web/android_build_string.h
larsers: Well...Same as the GCS forms directly above. So a single file, or with a trailing `/` the prefix holding a build's artifacts. But sure... The trailing `/` means the same thing on `https://`. 

So the comment now states both forms itself instead of pointing at its neighbour.

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
larsers: OK, thanks for testing the combinations. I'll swap the tolerance commit in #3070 for one that drops the inference entirely, so `otatools` is only fetched when `--otatools_build` asks for it, and the inferred tracking goes away with it.

base/cvd/cuttlefish/host/libs/web/android_build_string.cpp
larsers: Done... that shape is nicer. I kept the check that the first character is a letter on top of it, since the allowed set contains digits and `+-.`, which cannot start a scheme, and it also rejects the empty scheme in `://x`.

base/cvd/cuttlefish/host/libs/web/android_build_string.cpp
larsers: The thought here was actually that this should be a rule for the URL build string format itself rather than knowledge about the flag code. These strings travel in comma separated lists, so the format reserves the comma and the parser rejects it as part of its own grammar.

Trying to reject it anywhere else does not really work. At the split site you can't tell a comma inside one URL apart from the separator between two values. After the split the pieces can still parse as something plausible. So without this check the failure just shows up later and is harder to figure out, or it does not show up at all.

I see what you mean though... The comment really does read like it is describing the flag parser, which I think is your point? 

I've rewritten it to state the format rule and the reason for it instead of describing how another file behaves.

If you would rather allow commas in the format I can just drop the check... It just makes the comma in a list mistake harder to debug.

base/cvd/cuttlefish/host/libs/web/android_build_string.cpp
larsers: Well, because the directory form resolves artifacts by joining each artifact name onto the URL, and a query string does not survive that join, the name would land after the query... 

A signed query also authorizes exactly one resource, so it can't cover a listing + every artifact under the prefix. 

I hear you... Added a comment now.

base/cvd/cuttlefish/host/libs/web/url_namespace.h
larsers: An empty path, as in a URL naming just the bucket or host, is the directory form too, so it is `path.empty() || path.ends_with('/')`. 

But I see what oyu mean. I've fixed it so the field is gone and it is an inline method.

base/cvd/cuttlefish/host/libs/web/url_namespace.cpp
larsers: Ah right, `UrlScheme` was actually file-private next to the build string parser. I fixed that now. It lives next to `ParseUrl` and they both use it, so the scheme is extracted and validated by just one function.

I looked at making the build string parser call `ParseUrl` instead but that didn't really work. At that point the string still has the fragment and the selector attached to it.

base/cvd/cuttlefish/host/libs/web/url_namespace.h
larsers: Done... It returns `std::optional` now and the call sites just apply the fallback using the constant that actually already held `"url"`.

I renamed that constant to something a bit more general too. It stands in for the target, the product and the cache key now instead of only the target.

base/cvd/cuttlefish/host/libs/web/url_namespace.cpp
larsers: Done.

base/cvd/cuttlefish/host/libs/web/url_namespace.cpp
larsers: Done.

base/cvd/cuttlefish/host/libs/web/url_namespace.cpp
larsers: You're right, I fixed that. It actually turned out a bit simpler than that. The `closed` flag really only existed so the https directory form could flow through to a specific error, since it has no listing.

Now that I split it, that error moved to the https call site instead. So the overloads are just one over the object and one over the listing. The `closed` flag is completely gone now, which I think answers your comment in the header about it too.

base/cvd/cuttlefish/host/libs/web/url_namespace.cpp
larsers: Done... an enum, with one mapping to the archive name spellings kept next to the matching, shaped like the existing `SafeLevel` formatter so the non URL name join uses the same mapping instead of a second copy.

base/cvd/cuttlefish/host/libs/web/url_namespace.h
larsers: I was basically trying to say that `names` is the complete object listing of the build, like you get from a `gs://` listing. The `https://` directory form does not have a listing at all so there is no list to search. That is a very different situation compared to a listing that just comes back empty. The `closed` parameter was there to keep track of that difference so both cases got their own specific error.

You are right that it was doing two jobs in one function...

I've fixed it by splitting it up. The case with no listing does not call this function anymore, and that error moved to the `https` call site instead. The `closed` flag is removed now and the comment just talks about the listing instead of a closed namespace.

base/cvd/cuttlefish/host/libs/web/android_build_string.cpp
larsers: No, not a good one at least... The digest ends up in the build string struct, which owns it, so a conversion happens somewhere, but the helper itself can hand back a view of its input. 

Fixed so it returns a `string_view` now, converted at the assignment.

base/cvd/cuttlefish/host/libs/web/android_build_string.cpp
larsers: Done.

base/cvd/cuttlefish/host/libs/web/composite_build_api.h
Databean: > the leaf classes no longer implement `BuildApi` at all

I'm confused, I'm not seeing this in the latest version uploaded.

larsers: Yeah sorry, that was my fault. The updated branch had not been pushed. I held it back so the #3070 commits could drop out after that PR merged, and then kept holding it longer than made sense.

base/cvd/cuttlefish/host/libs/web/android_build.h
```diff
@@ -53,10 +57,66 @@ struct DirectoryBuild {
...
+std::ostream& operator<<(std::ostream&, const GcsBuild&);
+
+// The same two forms over `https://`, where a pre-signed URL carries its
+// credential in the query string of `url`.
+struct HttpBuild {
+  static Result<HttpBuild> FromBuildString(const HttpBuildString& build_string);
+
+  std::string url;                    // object form, query included
+  std::string base;                   // directory form, ends with '/'
+  std::optional<std::string> object;  // set in the object form only
```
jemoreira: IIUC This can either have `url` and `object` empty with a real url in `base` or `base` empty with values in `url` and `object`. If that's the case, there are better ways to enforce this with the compiler:

Option 1: Use the same member variable for `url` and `base` and use the presence of the `object` to determine whether this is an object form or a directory form.

Option 2: Have `HttpObjectBuild` and `HttpDirBuild` inherit from the abstract struct `HttpBuild`.

Option 3: Use an std::variant to achieve something similar to Option 2 without needing pointers/reference everywhere `HttpBuild` is used.

base/cvd/cuttlefish/host/libs/web/android_build.cc
```diff
@@ -52,16 +66,86 @@ std::ostream& operator<<(std::ostream& out, const DirectoryBuild& build) {
              << "\", filepath=\"" << build.filepath.value_or("") << "\")";
 }
 
+Result<GcsBuild> GcsBuild::FromBuildString(const GcsBuildString& build_string) {
+  const ParsedUrl url = CF_EXPECT(ParseUrl(build_string.url));
+  const size_t basename = url.path.rfind('/') + 1;
```
jemoreira: Should this handle `rfind` returning `npos`?

base/cvd/cuttlefish/host/libs/web/zip_over_ranges.cpp
```diff
@@ -0,0 +1,97 @@
...
+    const HttpRequest& request) {
+  static constexpr std::string_view kPrefix = "Range: bytes=";
+  size_t start = 0;
+  size_t end = data_.size();
+  for (const std::string& header : request.headers) {
+    if (!absl::StartsWith(header, kPrefix)) {
+      continue;
+    }
+    *ranged_ = true;
+    const std::string range = header.substr(kPrefix.size());
```
jemoreira: nit: absl::ConsumePrefix

base/cvd/cuttlefish/host/libs/web/http_build_api.cpp
```diff
@@ -0,0 +1,150 @@
...
+      ConstructTargetFilepath(target_directory, artifact_name);
+  CF_EXPECT(EnsureDirectoryExists(target_directory));
+
+  if (IsArchiveMember(build, artifact_name)) {
+    if (!build.accept_ranges) {
+      return CF_ERRF(
+          "'{}' does not serve range requests, so '{}' cannot be read out of "
+          "it.",
+          build.id, artifact_name);
+    }
```
jemoreira: nit: `CF_EXPECTF(build.accept_ranges, ...);`

base/cvd/cuttlefish/host/libs/web/composite_build_api.h
```diff
@@ -0,0 +1,54 @@
...
+                                   const std::string& target_directory,
+                                   const std::string& artifact_name) override;
+
+  Result<SeekableZipSource> FileReader(
+      const Build& build, const std::string& artifact_name) override;
+
+ private:
+  AndroidBuildApi& android_;
+  GcsBuildApi& gcs_;
+  HttpBuildApi& http_;
```
jemoreira: Given that `CompositeBuildApi` is the only `BuildApi` to be used by the fetch code, could it own the concrete implementation objects instead of referencing them?

base/cvd/cuttlefish/host/commands/cvd/fetch/build_api_credentials.h
```diff
@@ -36,6 +36,19 @@ Result<std::unique_ptr<CredentialSource>> GetCredentialSourceFromFlags(
...
+// Reading the product name costs no HTTP round trip on a machine that is not
+// one.
+bool IsRunningOnGce();
+
+// Resolves credentials for Cloud Storage. Never opens the `credential_source`
+// file, which the Android Build path may consume from a pipe, and never
+// re-uses an Android Build token, which Cloud Storage rejects. A null result
+// means anonymous access, which public buckets serve.
+Result<std::unique_ptr<CredentialSource>> GetStorageCredentialSource(
+    HttpClient& http_client, const BuildApiFlags& flags, bool running_on_gce);
```
jemoreira, line 51: Why does this function need to be told it's running on GCE? Can't it just call `IsRunningOnGce` itself?

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
```diff
@@ -172,6 +174,40 @@ Result<Build> GetHostBuild(BuildApi& build_api,
       "aosp_cf_x86_64_only_phone-userdebug");
 }
 
+template <typename UrlBuild>
+Result<void> CheckObjectIsHostPackage(const UrlBuild& build,
```
jemoreira: nit: This function doesn't look specific to host packages. It could be renamed as `CheckObjectName` or `CheckObjectIsNamed`.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/load_configs.cpp
```diff
@@ -292,10 +292,12 @@ std::vector<HelpParagraph> LoadConfigsCommand::CommonCommandDescription() {
   description.emplace_back(
       "While most config file properties are self explanatory, the build "
       "properties (default_build, kernel.build, bootloader.build, etc) require "
-      "more explanation. These properties support two types of values:");
+      "more explanation. These properties support three types of values:");
 
   description.emplace_back(HelpParagraph::Raw(
       R"( - "@ab/<branch_or_build_id>[/<target>[{<filepath>}]]"
+ - "gs://<bucket>/<prefix>/" or "gs://<bucket>/<object>[{<filepath>}]", and the
```
jemoreira: nit: Please list the 4 variants separately, one per line

base/cvd/cuttlefish/host/libs/web/android_build.h
larsers: Done. 

I went with option 1. `HttpBuild` now has a single `url` member holding either the object URL, query included, or the directory URL ending in `/`, and `object` being set is what says which form it is. That is also how `GcsBuild` already works with `prefix` and `object`, and the cache key, digest and artifact checks were already branching on `object`. The signed URL stays exactly as given.

base/cvd/cuttlefish/host/libs/web/android_build.cc
larsers: Yeah you're right. The parsed path has no leading `/`, so an object at the bucket root has no `/` at all and `rfind` returns npos. The old code only worked because npos + 1 wraps to 0. Replaced it with a small `BasenameStart` helper that handles the no-slash case explicitly, used by both `FromBuildString` functions, and added a test for an object at the bucket root, which had no coverage.

base/cvd/cuttlefish/host/libs/web/http_build_api.cpp
larsers: Done. It reads build.accept_ranges == true because CF_EXPECTF binds its argument as decltype(x)&&, which a const member does not bind to... That is the same spelling the tree uses elsewhere for a bool.

base/cvd/cuttlefish/host/libs/web/composite_build_api.h
larsers: Done... `CompositeBuildApi` now takes the three APIs as `std::unique_ptr` and owns them, and `Downloaders` no longer keeps members of its own for them. I used `unique_ptr` rather than holding them by value because `AndroidBuildApi` has a deleted move constructor, so constructing it in place would have meant the composite knowing every leaf's dependencies.

base/cvd/cuttlefish/host/libs/web/zip_over_ranges.cpp
larsers: Done.

base/cvd/cuttlefish/host/commands/cvd/fetch/build_api_credentials.h
larsers: It is an explicit input so the credential ladder can be unit tested on any machine. The tests pin both branches.. So if the function read the DMI product name itself, the anonymous-access cases would flip on a GCE host and the metadata cases would be unreachable off one. The only production caller passes `IsRunningOnGce()`. 

I added a sentence to the comment saying so. If you would rather have a two-argument overload that calls the check itself, with the three-argument one kept for the tests, I can add that.

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc
larsers: Done... Renamed to `CheckObjectIsNamed` with a generic message. The host package specific part, the hint about `--host_package_build`, moved to the two call sites in `CheckHostPackagePresent`, added through the `CF_EXPECT` message

base/cvd/cuttlefish/host/commands/cvd/cli/commands/load_configs.cpp
larsers: Done... Each form is on its own line now.

base/cvd/cuttlefish/host/libs/web/http_build_api.cpp
jemoreira: A trick from JS that also works here is `!!build.accept_ranges`.

base/cvd/cuttlefish/host/commands/cvd/fetch/build_api_credentials.h
jemoreira: nit: Instead of having the only caller pass the function call, I think it would be cleaner to have an overload of this function without the `running_on_gce` parameter that passes `IsRunningOnGce()` to the one that has it. It seems the caller is also the only user of `IsRunningOnGce` so with this approach `IsRunningOnGce` doesn't need to be in the header.

base/cvd/cuttlefish/host/libs/web/digest.h
```diff
@@ -0,0 +1,40 @@
...
+std::string Sha256Hex(std::string_view data);
+
+// Fails unless `path` holds the hexadecimal SHA-256 `expected`, which is
+// compared without regard to case. `artifact_name` names the file in the error.
+Result<void> VerifySha256(const std::string& path, std::string_view expected,
+                          std::string_view artifact_name);
+
+// Checks the same against the base64 MD5 Cloud Storage reports for an object.
+Result<void> VerifyMd5(const std::string& path, std::string_view expected,
+                       std::string_view artifact_name);
```
Databean: nit: maybe call this `VerifyBase64Md5`? My default assumption is that md5 outputs are hex, not base64.

base/cvd/cuttlefish/host/libs/web/http_build_api.cpp
larsers: Done. It reads `!!build.accept_ranges` now.

base/cvd/cuttlefish/host/commands/cvd/fetch/build_api_credentials.h
larsers: Done. `GetStorageCredentialSource` has a two argument overload now that passes `IsRunningOnGce()` to the three arg one. `Downloaders` calls the 2-args overload, `IsRunningOnGce` moved into the anonymous namespace of the .cc, and the tests keep using the three arg overload.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/load_configs.cpp
```diff
@@ -292,10 +292,14 @@ std::vector<HelpParagraph> LoadConfigsCommand::CommonCommandDescription() {
...
       "properties (default_build, kernel.build, bootloader.build, etc) require "
-      "more explanation. These properties support two types of values:");
+      "more explanation. These properties support the following values:");
 
   description.emplace_back(HelpParagraph::Raw(
       R"( - "@ab/<branch_or_build_id>[/<target>[{<filepath>}]]"
+ - "gs://<bucket>/<prefix>/"
+ - "gs://<bucket>/<object>[{<filepath>}]"
+ - "https://<host>/<directory>/"
+ - "https://<host>/<object>[{<filepath>}]"
```
Databean: For parity with `@ab` can these start with `@` also like `@gs://` and `@https`? That should make parsing more consistent as well.

base/cvd/cuttlefish/host/libs/web/digest.cpp
```diff
@@ -0,0 +1,104 @@
...
+             "'{}' has sha256 '{}', but '{}' was expected.", artifact_name,
+             actual, expected);
+  return {};
+}
+
+Result<void> VerifyMd5(const std::string& path, std::string_view expected,
+                       std::string_view artifact_name) {
+  const std::vector<uint8_t> value = CF_EXPECT(DigestFile(path, EVP_md5()));
+  const std::string actual =
+      CF_EXPECT(EncodeBase64(value.data(), value.size()));
```
Databean: From https://en.wikipedia.org/wiki/MD5#MD5_hashes

> The 128-bit (16-byte) MD5 hashes (also termed _message digests_) are typically represented as a sequence of 32 [hexadecimal](https://en.wikipedia.org/wiki/Hexadecimal) digits.

If this data is stored in a non-typical format, can that be mentioned in the structure member name?

base/cvd/cuttlefish/host/libs/web/digest.cpp
```diff
@@ -0,0 +1,104 @@
...
+  CF_EXPECTF(actual == expected, "'{}' has md5 '{}', but '{}' was expected.",
+             artifact_name, actual, expected);
```
Databean: This could use `CF_EXPECT_EQ` which prints the values used for comparison.

base/cvd/cuttlefish/host/libs/web/digest.cpp
```diff
@@ -0,0 +1,104 @@
...
+std::string Sha256Hex(std::string_view data) {
+  uint8_t value[SHA256_DIGEST_LENGTH];
+  SHA256(reinterpret_cast<const uint8_t*>(data.data()), data.size(), value);
+  return HexDigest(value, sizeof(value));
+}
+
+Result<void> VerifySha256(const std::string& path, std::string_view expected,
+                          std::string_view artifact_name) {
+  const std::string actual = CF_EXPECT(Sha256File(path));
+  CF_EXPECTF(actual == absl::AsciiStrToLower(expected),
```
Databean, line 88: This could use `CF_EXPECT_EQ` which automatically writes an error message comparing the two values. Same for `VerifyMd5` below.

base/cvd/cuttlefish/host/libs/web/digest_test.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+constexpr char kAbcSha256[] =
+    "ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad";
+constexpr char kAbcMd5[] = "kAFQmDzST7DWlj99KOF/cg==";
+constexpr char kEmptySha256[] =
+    "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855";
+
+std::string WriteFile(const TemporaryDir& directory,
+                      std::string_view contents) {
+  std::string path = std::string(directory.path) + "/artifact";
```
Databean, line 43: nit: `absl::StrCat`, here and in other places that call `std::string` in this file.

base/cvd/cuttlefish/host/libs/web/android_build.h
```diff
@@ -53,10 +59,84 @@ struct DirectoryBuild {
 
 std::ostream& operator<<(std::ostream&, const DirectoryBuild&);
 
-using Build = std::variant<DeviceBuild, DirectoryBuild>;
+struct GcsObjectInfo {
+  std::optional<std::string> generation;
+  std::optional<std::string> md5;
```
Databean, line 64: Elsewhere it is mentioned GCS has these encoded as base64. From https://en.wikipedia.org/wiki/MD5#MD5_hashes

> The 128-bit (16-byte) MD5 hashes (also termed _message digests_) are typically represented as a sequence of 32 [hexadecimal](https://en.wikipedia.org/wiki/Hexadecimal) digits.

If this data is stored in a non-typical format, can that be mentioned in the structure member name? Same for `GcsBuild` below.

base/cvd/cuttlefish/host/libs/web/build_api_zip.cc
```diff
@@ -27,14 +27,16 @@
...
-Result<ReadableZip> OpenZip(BuildApi& build_api, const Build& build,
-                            const std::string& name) {
-  SeekableZipSource source = CF_EXPECT(build_api.FileReader(build, name));
-
+Result<ReadableZip> OpenZip(SeekableZipSource source) {
   SeekableZipSource buffered =
       CF_EXPECT(BufferZipSource(std::move(source), 1 << 26));
 
   return CF_EXPECT(ReadableZip::FromSource(std::move(buffered)));
 }
```
Databean, line 35: This doesn't have anything to do with `BuildApi` or anything else in the `web` directory. Additionally, `BufferZipSource` and `ReadableZip` are already public.

At best this could be a `BufferAndOpenZip` function somewhere in `host/libs/zip`.

base/cvd/cuttlefish/host/libs/web/http_build_api.cpp
```diff
@@ -0,0 +1,176 @@
...
+  }
+  const std::optional<std::string_view> length =
+      HeaderValue(response.headers, "content-length");
+  if (response.http_code != kPartialContent && length.has_value() &&
+      absl::SimpleAtoi(*length, &size)) {
+    return size;
+  }
+  return std::nullopt;
+}
```
Databean, line 88: `ServesRanges` and `ProbedSize` duplicate logic from [`GetSizeIfSupportsRangeRequests`](https://github.com/google/android-cuttlefish/blob/5a467328d69a4caa2715ad0da3bf1edd2b8c47ba/base/cvd/cuttlefish/host/libs/zip/remote_zip.cc#L111-L133). Is there a way to consolidate into one implementation?

base/cvd/cuttlefish/host/libs/web/http_build_api.cpp
```diff
@@ -0,0 +1,176 @@
...
+  HttpBuild build = CF_EXPECT(HttpBuild::FromBuildString(build_string));
+  // A directory of plain HTTPS URLs has nothing to list and nothing to probe,
+  // so its artifacts are only known to be absent when they answer 404.
+  if (build.object.has_value()) {
+    CF_EXPECT(ProbeObject(build));
+  }
+  return build;
+}
+
+Result<void> HttpBuildApi::ProbeObject(HttpBuild& build) {
```
Databean, line 105: https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs writes

>
> Prefer using return values over output parameters: they improve readability, and often provide the same or better performance. See [TotW #176](https://abseil.io/tips/176).

Can this function return a `Result<HttpBuild>` instead of a a `Result<void>`?

Also note [go/UnhealthyIdioms#missing-abstractions-sea-of-details](http://go/UnhealthyIdioms#missing-abstractions-sea-of-details) which writes

> Ad Hoc Workflow
> 
> - When a series of steps should be regularized in a workflow or state machine style, but is instead done in an ad hoc way. For example, proto buffers are passed from one stage to the next, with some parts being filled in by one step and another step expecting certain values in the proto to have meaning.

In this case, `HttpBuild` has both input members and output members.

base/cvd/cuttlefish/host/libs/web/url_download.cpp
Databean: > This file covers the other access pattern, downloading a whole artifact to disk

But it still uses `Range`.

> one ranged request per libzip read would be slow

`ZipSourceFromUrl` doesn't mandate using `libzip`. It produces a `SeekableZipSource`, which can be used to create a `SeekingZipSourceReader`, which implements the `ReaderSeeker` interface, giving access through this general pattern:
```
  Result<uint64_t> Read(void* data, uint64_t length) override;
  Result<uint64_t> PRead(void* buf, uint64_t count,
                         uint64_t offset) const override;

  Result<uint64_t> SeekSet(uint64_t offset) override;
  Result<uint64_t> SeekCur(int64_t offset) override;
  Result<uint64_t> SeekEnd(int64_t offset) override;
```
If you don't like the buffer sizes and seek offsets chosen by libzip, it's possible to call these methods without reimplementing their internals.

base/cvd/cuttlefish/host/libs/web/url_download.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+}  // namespace
+
+Result<bool> HoldsFileAt(Fd& fd, const std::string& path) {
+  // The descriptor is open before the lock says whose file it is, so comparing
+  // two paths would race with the rename that ends another download.
+  struct stat by_path = {};
+  if (stat(path.c_str(), &by_path) != 0) {
+    return false;
+  }
```
Databean, line 146: Can this use the [`Stat`](https://github.com/google/android-cuttlefish/blob/80d42d7c16f18d1c3b12d71dc0bb26926246bdd2/base/cvd/cuttlefish/posix/stat.h#L27) wrapper?

base/cvd/cuttlefish/host/libs/web/url_download.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    if (CF_EXPECT(HoldsFileAt(part, part_path))) {
+      CF_EXPECT(ResumeDownload(http_client, download, part, part_path, path));
+      return {};
+    }
+    // Another download of this artifact renamed the partial file away while
+    // this one waited for its lock.
+    if (FileExists(path)) {
+      return {};
+    }
```
Databean, line 176: Should the `.part` file be removed in this case? The invocation that wins the race to claim the lock file renames the `.part` file, but all the other invocations that lose the race still `open` the `.part` file with `O_CREAT` and then leave it lying around.

base/cvd/cuttlefish/host/libs/web/url_download.cpp
```diff
@@ -0,0 +1,181 @@
...
+    const HttpResponse<void> response =
+        CF_EXPECT(http_client.DownloadToCallback(request, callback));
+    if (!response.HttpSuccess()) {
+      // The body of an error response is not the artifact.
+      CF_EXPECT(RemoveFile(part_path));
+      return CF_ERRF("'{}' - {}:{}", ScrubUrl(download.url), response.http_code,
+                     response.StatusDescription());
+    }
+    if (offset == 0 || !download.size.has_value() ||
+        offset + written == *download.size) {
```
Databean, line 124: Rather than comparing against the size, it seems better to use the response code and distinguish `206 Partial Content` from `200 OK`. It is more semantically tied to the behavior of `If-Range` and if `download.size` is missing, there is no way to distinguish good and bad downloads.

Databean (commented):
18/63 files reviewed

base/cvd/cuttlefish/host/libs/web/zip_over_ranges.h
```diff
@@ -0,0 +1,56 @@
...
+#include "cuttlefish/host/libs/web/http_client/http_client.h"
+#include "cuttlefish/result/result.h"
+
+namespace cuttlefish {
+
+// Returns whether `headers` carry an `Authorization` header.
+bool HasAuthorization(const std::vector<std::string>& headers);
+
+// Serves one archive over HTTP range requests, as an object store does.
+class ZipOverRanges {
```
Databean, line 34: This looks very similar to the private `class HttpCallback` in `remote_zip_test.cc`, is there any possibility to share code there?

base/cvd/cuttlefish/host/libs/web/url_namespace.cpp
```diff
@@ -0,0 +1,151 @@
...
+  if (query_start != std::string_view::npos) {
+    query = std::string(path.substr(query_start + 1));
+    path = path.substr(0, query_start);
+  }
+  return ParsedUrl{
+      .authority = std::string(authority_and_path.substr(0, authority_end)),
+      .path = std::string(path),
+      .query = query,
+  };
+}
```
Databean, line 86: We're already importing libcurl which [has its own URL parsing functionality](https://curl.se/libcurl/c/libcurl-url.html).

Personally I would prefer a C++ wrapper over the libcurl functions than a reimplementation of its string parsing.

base/cvd/cuttlefish/host/libs/web/android_build_string_tests.cpp
```diff
@@ -202,4 +209,218 @@ TEST(BuildStringGflagsCompatFlagTests, MultiValueMixedWithEmptySuccess) {
...
+TEST(ParseBuildStringTests, UrlWithCommaFail) {
+  EXPECT_THAT(ParseBuildString("gs://bucket/file,name.zip"),
+              IsErrorAndMessage(HasSubstr("cannot contain a comma")));
+  EXPECT_THAT(
+      ParseBuildString("gs://bucket/f,ile.zip?X-Goog-Signature=SECRETVALUE"),
+      IsErrorAndMessage(AllOf(HasSubstr("cannot contain a comma"),
+                              Not(HasSubstr("SECRETVALUE")))));
+}
+
+TEST(ParseBuildStringTests, UrlErrorsWithoutQueryFail) {
```
Databean, line 362: This test is called `UrlErrorsWithoutQueryFail` but it looks like it fails when there is a query rather than without?

Databean (commented):
25/63 files reviewed

PR #3049, opened 2026-08-19
Podcvd can pass CVD_INVOKER env variable to cvd

0405ysj:
Context: b/535678395

### DIFF
--- a/container/src/podcvd/internal/container.go
+++ b/container/src/podcvd/internal/container.go
@@ -137,6 +137,7 @@ func (m *CuttlefishContainerManagerImpl) CreateAndStartContainer(ctx context.Con
 	for _, dev := range devices {
 		args = append(args, "--device", dev+":"+dev+":rwm")
 	}
+	args = append(args, "-e", "CVD_INVOKER=podcvd")
 	if hasNvidiaGPU() {
 		args = append(args,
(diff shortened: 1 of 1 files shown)


base/cvd/external_proto/cf_host.proto
```diff
@@ -42,6 +42,8 @@ message CuttlefishHost {
     CUTTLEFISH_HOST_INVOKER_ACLOUD = 2;
     // invoked by `host orchestrator`
     CUTTLEFISH_HOST_INVOKER_HOST_ORCHESTRATOR = 3;
+    // invoked by `podcvd`
+    CUTTLEFISH_HOST_INVOKER_PODCVD = 4;
```
cjreynol: Will you drop this change and the `metrics_conversion` change converting between enum types?

We cannot start transmitting until after we complete the latest round of reviews.  Technically we are still blocked by installation of `cuttlefish-metrics` in `podcvd` cases, but I would rather we be able to say the functionality is not active until we get the "all clear".

cjreynol (approved):

base/cvd/external_proto/cf_host.proto
0405ysj: Sure, I dropped any metric code and kept podcvd code. I'm not uncertain if you wish me to drop enum value as well, I believe you can deal with this later.

PR #3052, opened 2026-08-19
simplify ChromeOsStateImage::Reuse call

fkm3:
CF_EXPECT unwrapped the Result, then we put it back into a new Result that was always a success.


Databean (approved):

Databean:

PR #3053, opened 2026-08-19
verify instance nums match snapshot on restore

fkm3:
We currently don't support changing the instance nums between snapshot and restore. If you mess it up, you get errors about disks changes that might lead you to debug in the wrong direction, so, add an explicit check.

Bug: b/310035418


Databean:

Databean (approved):

PR #3055, opened 2026-08-19
run `adb connect` before `adb wait-for-device` when restoring a VM

fkm3:
Similar to https://r.android.com/3285054: In some environments adb_connector isn't running, so `wait-for-device` will get stuck. Additionally, even when it is running, adb_connector seems to use the IP 127.0.0.1, which doesn't match the IP 0.0.0.0 used by default in instance_.adb_host_port().


Databean (approved):

Databean:

PR #3056, opened 2026-08-19
Prepare to handle alternate network modes.

3405691582:
Code motion. The ifdef should probably be tweaked, but we retain behavior for now.


dxapd:
looks like some formatters are complaining

3405691582:
> looks like some formatters are complaining

This was already in the branch, not the commit; rebasing fixed that one. The other checks are flaking, I think.

dxapd (approved):

PR #3057, opened 2026-08-19
More clauses for avoiding tap devices.

3405691582:
cvdalloc defines an addressing scheme for configuring the RIL early because tap devices are not preconfigured. This scheme should also be used when tap devices are disabled or when (eventually) user-mode networking should be configured for crosvm.

To ensure that this addressing scheme doesn't conflict with qemu's usermode networking, we need to introspect the vmm, which necessitates adding an extra argument to ConfigureNetworkSettings.

Likewise, run_cvd tests if the existing tap devices are usable. This is unnecessary when using cvdalloc, but this is also unnecessary for user-mode networking for qemu or crosvm.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
@@ -1054,8 +1054,8 @@ Result<CuttlefishConfig> InitializeCuttlefishConfiguration(
     }
     instance.set_mobile_tap_name(iface_config.mobile_tap.name);
 
-    CF_EXPECT(ConfigureNetworkSettings(ril_dns_vec[instance_index],
-                                       const_instance, instance));
+    CF_EXPECT(ConfigureNetworkSettings(
+        ril_dns_vec[instance_index], tmp_config_obj, const_instance, instance));
 
     bool use_non_bridged_wireless =
         (NetworkInterfaceExists(iface_config.non_bridged_wireless_tap.name) ||
--- a/base/cvd/cuttlefish/host/commands/run_cvd/validate.cpp
+++ b/base/cvd/cuttlefish/host/commands/run_cvd/validate.cpp
@@ -29,6 +29,7 @@
 #include "allocd/alloc_utils.h"
 #include "cuttlefish/common/libs/utils/in_sandbox.h"
 #include "cuttlefish/host/libs/config/cuttlefish_config.h"
+#include "cuttlefish/host/libs/config/external_network_mode.h"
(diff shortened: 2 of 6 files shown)


base/cvd/cuttlefish/host/commands/run_cvd/validate.cpp
```diff
@@ -38,9 +38,9 @@ namespace cuttlefish {
 static Result<void> TestTapDevices(
     const CuttlefishConfig::InstanceSpecific& instance) {
 #ifdef __linux__
-  if (InSandbox() ||
-      instance.use_cvdalloc()) {  // cvdalloc owns its own resources. those
```
dxapd, line 42: are you sure you want to remove this comment?

base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
dxapd, line 1337: does this CF_EXPECT also need to change? i didn't read every line between here and the invocation of  ConfigureNetworkSettings, but at a quick glance it seems like the init will fail here anyways if `VmManagerIsCrosvm(config) && const_instance.external_network_mode() == ExternalNetworkMode::kSlirp;`

base/cvd/cuttlefish/host/commands/run_cvd/validate.cpp
3405691582: Sure; either that or we write more verbiage here to cover all the clauses. What would you prefer?

base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
3405691582: Yes it will need to change, but we'll address this in a later change/commit.

dxapd: sg

base/cvd/cuttlefish/host/commands/run_cvd/validate.cpp
dxapd: personally, i lean on the side of more verbiage since it's inferrable, but not immediately obvious, why we would need to skip testing tap devices. but up to you cuz it could go either way :D

dxapd (approved):

base/cvd/cuttlefish/host/commands/run_cvd/validate.cpp
3405691582: Let me come back to this in another pr.

PR #3058, opened 2026-08-19
Replace SharedFD::Creat with Fd::Creat, and have it return a Result

Databean:
Bug: b/540553865
Bug: b/545278508

### DIFF
--- a/base/cvd/cuttlefish/common/libs/fs/fd.h
+++ b/base/cvd/cuttlefish/common/libs/fs/fd.h
@@ -104,7 +104,7 @@ class Fd : public ReaderWriterSeeker {
   static Fd Accept(const Fd& listener, struct sockaddr* addr,
                    socklen_t* addrlen);
   static Fd Accept(const Fd& listener);
-  static Fd Creat(const std::string& pathname, mode_t mode);
+  static Result<Fd> Creat(const std::string& path, mode_t mode);
   static Fd Dup(int unmanaged_fd);
   static Result<Fd> Fifo(const std::string& pathname, mode_t mode);
   static Fd MemfdCreate(const std::string& name, unsigned int flags = 0);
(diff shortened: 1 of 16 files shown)


base/cvd/cuttlefish/common/libs/fs/fd.h
```diff
@@ -104,7 +104,7 @@ class Fd : public ReaderWriterSeeker {
   static Fd Accept(const Fd& listener, struct sockaddr* addr,
                    socklen_t* addrlen);
   static Fd Accept(const Fd& listener);
-  static Fd Creat(const std::string& pathname, mode_t mode);
+  static Result<Fd> Creat(const std::string& pathname, mode_t mode);
```
cjreynol: NIT: update first parameter to `path` to match the source file.

base/cvd/cuttlefish/host/libs/config/esp.cpp
```diff
@@ -262,7 +263,10 @@ bool LinuxEspBuilder::Build() const {
   auto builder = PrepareGrubESP(image_path_, *arch_);
 
   const auto tmp_grub_config = image_path_ + ".grub.cfg";
-  const auto config_file = SharedFD::Creat(tmp_grub_config, 0644);
+  const SharedFD config_file =
+      Fd::Creat(tmp_grub_config, 0644)
+          .transform([](Fd f) { return SharedFD(std::move(f)); })
```
cjreynol: Could we instead return the `Result<Fd>` from `Creat`, then have the following conditional on line 270 check `(!config_file.ok())`?

I (anecdotally) feel the pattern of manually handling a `Result` to convert it to a boolean rather than cascade failures is more common in our codebase.  My initial read of these lines had me asking if there was some special reason we needed a `SharedFD` returned here.

I also think returning a `Result` helps us more completely move away from needing the `IsOpen` calls.

cjreynol (approved):

base/cvd/cuttlefish/host/libs/config/esp.cpp
Databean: Used a `Result<SharedFD>` variable.

base/cvd/cuttlefish/common/libs/fs/fd.h
Databean: Good catch, fixed.

PR #3060, opened 2026-08-19
Include secure_env in the debian package, and disable the weaver HAL

Databean:
The Weaver HAL is missing from secure_env_only_oemlock, and it is nontrivial to include because it requires importing the WeaverTa dependency from AOSP.

Bug: b/548056221


jemoreira (approved):

Databean:
Added another commit which stubs out some suspend request responders to `secure_env`. I believe this was blocking the suspend operation in taking snapshots.

PR #3065, opened 2026-08-20
Update staticcheck go version and dav1d download url to fix github actions

Databean:

PR #3066, opened 2026-08-20
Support StrongBox KeyMint jcardsim JVM lookup in container images

victorhsieh:
When launching Cuttlefish with --enable_jcard_simulator=true, secure_env attempts to load libjvm.so to run the StrongBox KeyMint Java Card simulator applet. Inside rootless container environments like podcvd, this fails because the base image lacks a JRE runtime and JAVA_HOME is not set.

This commit resolves the issue by:
1. Installing openjdk-21-jre-headless in runner-base.
2. Setting ENV JAVA_HOME=/usr/lib/jvm/default-java so secure_env automatically locates libjvm.so via getenv("JAVA_HOME").

### DIFF
--- a/container/image/Containerfile
+++ b/container/image/Containerfile
@@ -26,11 +26,13 @@ RUN apt update
 RUN apt install -y --no-install-recommends \
     ca-certificates \
     curl \
+    default-jre-headless \
     libvulkan1 \
     mesa-utils \
     nginx \
     sudo
 RUN update-ca-certificates
+ENV JAVA_HOME=/usr/lib/jvm/default-java
 
 # When setting `auto` value, nginx spawns worker processes as the number of CPU
 # cores. It's discouraged in any multiple container scenarios, as it spawns too


container/image/Containerfile
```diff
@@ -29,8 +29,10 @@ RUN apt install -y --no-install-recommends \
     libvulkan1 \
     mesa-utils \
     nginx \
+    openjdk-21-jre-headless \
```
0405ysj: Do we have something at `/usr/lib/jvm/default-java` when installing this only?

victorhsieh:
Good catch!  I switched to default-jre-headless and verified it works (still, at least for me).

0405ysj (approved):

0405ysj:
FYI: Rebased PR as https://github.com/google/android-cuttlefish/pull/3065 is required to pass presubmit checks. It removes current flakiness on presubmit checks.

PR #3008, opened 2026-08-12
Include crosvm in the debian package

Databean:
Bug: b/402274999

PR #3072, opened 2026-08-21
Always build crosvm in `opt` mode

Databean:
This has two effects:

- Disables debug assertions crash over an IO safety issue in sandbox mode.
- Improves VM runtime performance.

With this change, we are closer to being able to package crosvm in `//cuttlefish/package:common`.

Bug: b/402274999

### DIFF
--- a/base/cvd/build_external/crosvm/BUILD.bazel
+++ b/base/cvd/build_external/crosvm/BUILD.bazel
@@ -1,3 +1,18 @@
+load("//build_external/crosvm:transition_opt.bzl", "build_in_opt")
+
 exports_files([
     "crosvm.config.toml",
 ])
+
+# Build crosvm as if bazel was invoked with `-c opt`. Without this, observed
+# some errors that looked like
+# ```
+# fatal runtime error: IO Safety violation: owned file descriptor already closed
+# ```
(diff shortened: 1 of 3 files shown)


Databean:
This is most of https://github.com/google/android-cuttlefish/pull/3008 but doesn't actually package it yet.

jmacnak:
> Disables debug assertions crash over an IO safety issue in sandbox mode.

Is there more info on this or why we think it is benign? This sounds pretty scary at least.

Databean:
> Is there more info on this or why we think it is benign? This sounds pretty scary at least.

I'm not really sure that it is benign :/ the error looks like

```
log_tee(460985)  D 08-21 16:20:58 460985 460985 log_tee.cpp:185] fatal runtime error: IO Safety violation: owned file descriptor already closed
```

This check is https://www.github.com/rust-lang/rust/pull/124210 which was added to stable in Rust 1.80. Crosvm upstream is [using Rust 1.88.0](https://chromium.googlesource.com/crosvm/crosvm/+/ffdfff5/rust-toolchain) so it is surprising we're seeing this error now.

I can't reproduce this with a `crosvm` executable manually created by `cargo build` either.

jemoreira (approved):

PR #3080, opened 2026-08-22
ci: Eliminate pseudo-TTY CRLF

rmuthiah:
When `podman exec -it` was executed in CI scripts, pseudo-terminal allocation (-t) caused stdout lines to be emitted with CRLF (\r\n). In `runner-special`, capturing this into bash passed carriage returns to `bazel test`, which failed with:
  invalid target name '... \r': target names may not end with carriage returns

In `run-cw-sharded-e2e-test`, this was previously worked around by `${t::-1}` string slice.

Furthermore, `runner-special` attempted to mitigate this using `sh -c "stty -onlcr && bazel query ..." 2>/dev/null`. On runner VMs where standard input was non-interactive, `stty` failed silently, short-circuiting the `&&` and skipping flaky tests altogether.

This change:
1. Removes `-it` from all `podman exec` calls in both `presubmit.yaml` and `run-cw-sharded-e2e-test/action.yaml`.
2. Removes the `${t::-1}` slicing workaround in `run-cw-sharded-e2e-test`.

Test: Verified `bazel query` and target loops parse cleanly without trailing \r.
Bug: b/550908119

### DIFF
--- a/.github/actions/run-cw-sharded-e2e-test/action.yaml
+++ b/.github/actions/run-cw-sharded-e2e-test/action.yaml
@@ -72,7 +72,6 @@ runs:
         # Run tests
         targets=( $(echo ${runner_tests}) )
         for t in "${targets[@]}"; do
-          t="${t::-1}"
           echo "running test: ${t}"
           sudo podman rm -f tester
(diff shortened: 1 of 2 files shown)


Databean (approved):
Thanks, had no idea what this was.

PR #3089, opened 2026-08-25
Use Fd instead of SharedFd in NativeFilesystem

Databean:
Also added more convenience versions of RemoveFile.

Bug: b/540553865
Bug: b/545278508


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #3090, opened 2026-08-25
Use Fd::Open in image_aggregator

Databean:
Bug: b/540553865
Bug: b/545278508


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #3092, opened 2026-08-25
SharedFD::Open -> Fd::Open in host/commands/cvd

Databean:
Bug: b/545278508


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #3095, opened 2026-08-25
SharedFD::Open -> Fd::Open in host/libs/config

Databean:
Bug: b/545278508

### DIFF
--- a/base/cvd/cuttlefish/host/libs/config/log_string_to_dir.cc
+++ b/base/cvd/cuttlefish/host/libs/config/log_string_to_dir.cc
@@ -15,15 +15,16 @@
  */
 #include "cuttlefish/host/libs/config/log_string_to_dir.h"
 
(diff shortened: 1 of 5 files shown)


base/cvd/cuttlefish/host/libs/config/log_string_to_dir.cc
cjreynol, line 28: Add `#include "cuttlefish/common/libs/fs/fd.h"`

cjreynol (approved):

base/cvd/cuttlefish/host/libs/config/log_string_to_dir.cc
Databean: Good catch, I'm surprised the linters didn't catch it.

Databean:
Thanks for reviewing!

PR #3096, opened 2026-08-25
Capture fetched build safety levels for metrics

cjreynol:
Not yet being transmitted, but available at the metrics message populating step `*conversion.cc::PopulateCuttlefishBuild` when we are ready to transmit.

Bug: 535685161

### DIFF
--- a/base/cvd/cuttlefish/host/libs/web/android_build_api.h
+++ b/base/cvd/cuttlefish/host/libs/web/android_build_api.h
@@ -63,6 +63,13 @@ class AndroidBuildApi : public BuildApi {
     std::string target;
     bool is_signed;
   };
+
+  struct CandidateBuild {
+    std::string build_id;
+    std::chrono::system_clock::time_point creation_time;
+    SafeLevel safe_level;
+  };
+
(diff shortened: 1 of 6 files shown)


base/cvd/cuttlefish/host/libs/web/android_build_api.h
```diff
@@ -63,15 +63,22 @@ class AndroidBuildApi : public BuildApi {
     std::string target;
     bool is_signed;
   };
+
+  struct CandidateBuild {
+    std::string build_id;
+    std::chrono::system_clock::time_point creation_time;
+    SafeLevel safe_level;
```
dxapd, line 70: seems reasonable, and I assume you're holding onto this + creation_time to transmit it in a later PR as you mentioned. but for my knowledge, what's a safe level? :)

dxapd (approved):

base/cvd/cuttlefish/host/libs/web/android_build_api.h
cjreynol: Explained and shared reference off GitHub.

PR #3097, opened 2026-08-25
base: cvd: allocd: Search /usr/sbin for dnsmasq and fix PID file path

grandpaul:
In Debian, dnsmasq is located in /usr/sbin/dnsmasq which is not typically in non-root users' $PATH. Add SearchForDnsmasq and DnsmasqPath fallback helpers similar to IptablesPath. Also fix StopDnsmasq to use CvdDir() instead of hardcoded /var/run/ and fix inverted file.is_open() logic.


dxapd (approved):
great find, thank you!!

PR #3104, opened 2026-08-27
Enable IPv6 ULA, SLAAC Router Advertisements, NAT66, and E2E validation

sferrogoo:
### Description
This change adds complete IPv6 support for Cuttlefish instances on the host and validates guest dual-stack and pure IPv6-only operation.

1. **IPv6 Unique Local Addresses (ULA)**:
   - Configures default subnets `fd00:cf:22::/64` on `cvd-wbr` and `fd00:cf:24::/64` on `cvd-ebr`.
   - Adds configuration toggles in `/etc/default/cuttlefish-host-resources`.

2. **Router Advertisements (SLAAC)**:
   - Configures `dnsmasq` to broadcast SLAAC Router Advertisements for autonomous in-guest address assignment.

3. **NAT66 & Forwarding**:
   - Implements native `nftables` masquerading and forwarding rules in `cuttlefish-host-resources.init`.
   - Adds `nftables` package dependency to `cuttlefish-base` in `debian/control`.

4. **Automated E2E Integration Testing**:
   - Adds `//cvd/network_tests:network_tests` verifying guest SLAAC acquisition, guest-to-host dual-stack communication, and 100% pure IPv6-only functionality after flushing IPv4.


sferrogoo:
@google/android-cuttlefish @adelva1984 @jemoreira PTAL

e2etests/cvd/network_tests/BUILD.bazel
0405ysj: The directory `e2etests/cvd/networking_tests` already exists. I assume there would be the better place for this test?

e2etests/cvd/network_tests/main_test.go
```diff
@@ -0,0 +1,159 @@
+// Copyright (C) 2026 The Android Open Source Project
```
dxapd, line 1: This appears to have been added in another directory "network_tests" at the same level as the "networking_tests" directory. Perhaps this would be better with either:

- a different directory name, if the goal of these tests is different from the goal of tests in networking_tests/
- rolled into networking_tests/ as a new test file or a new test case

e2etests/cvd/network_tests/main_test.go
```diff
@@ -0,0 +1,159 @@
...
+		if adbErr == nil {
+			break
+		}
+		t.Logf("Waiting for ADB connection (attempt %d/10)...", i+1)
+	}
+	if adbErr != nil {
+		t.Fatalf("Timed out waiting for ADB device connection: %v", adbErr)
+	}
+
+	// 1. Verify SLAAC IPv6 assignment on in-guest network interface
```
3405691582, line 58: factorize or split up the test further instead of doing this

e2etests/cvd/network_tests/main_test.go
```diff
@@ -0,0 +1,159 @@
...
+	t.Log("Verifying in-guest SLAAC IPv6 address assignment...")
+	var ip6Output e2etests.CommandOutput
+	var activeDev string
+	var gatewayIP string
+	for attempt := 1; attempt <= 15; attempt++ {
```
3405691582, line 63: why?

e2etests/cvd/network_tests/main_test.go
```diff
@@ -0,0 +1,159 @@
...
+		// Check all potential network interfaces (wireless and ethernet bridges)
```
3405691582, line 64: The following block is hard to reason about and this comment doesn't help.

e2etests/cvd/network_tests/main_test.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	c.RunCmd("adb", "shell", "su 0 sh -c \"ip -6 rule add pref 50 lookup main 2>/dev/null || true\"")
+	c.RunCmd("adb", "shell", "su 0 sh -c \"ip -4 rule add pref 50 lookup main 2>/dev/null || true\"")
+	c.RunCmd("adb", "shell", fmt.Sprintf("su 0 sh -c \"ip -6 route add %s dev %s 2>/dev/null || true\"", prefix, activeDev))
+	c.RunCmd("adb", "shell", fmt.Sprintf("su 0 sh -c \"ip -6 route add default via %s dev %s 2>/dev/null || true\"", gatewayIP, activeDev))
+
+	// 2. Verify guest-to-host IPv6 connectivity
+	t.Logf("Verifying dual-stack guest-to-host IPv6 connectivity (gateway %s on %s)...", gatewayIP, activeDev)
+	// Query detected router link-local address from neighbor table
+	neighOut, _ := c.RunCmd("adb", "shell", fmt.Sprintf("ip -6 neigh show dev %s", activeDev))
```
3405691582, line 116: factorize

e2etests/cvd/network_tests/main_test.go
```diff
@@ -0,0 +1,159 @@
...
+			fields := strings.Fields(line)
+			if len(fields) > 0 && strings.HasPrefix(fields[0], "fe80:") {
+				routerLL = fields[0]
+				break
+			}
+		}
+	}
+	t.Logf("Guest ULA: %s, Router LL: %s, Gateway ULA: %s", guestIP, routerLL, gatewayIP)
+
+	connCmd := fmt.Sprintf("su 0 toybox ping -6 -c 3 -I %s %s || su 0 toybox ping -6 -c 3 -I %s %s || su 0 toybox ping -6 -c 3 %s", guestIP, gatewayIP, activeDev, routerLL, gatewayIP)
```
3405691582, line 129: factorize

e2etests/cvd/network_tests/main_test.go
```diff
@@ -0,0 +1,159 @@
...
+	connOut, err := c.RunCmd("adb", "shell", connCmd)
+	if err != nil {
+		diag, _ := c.RunCmd("adb", "shell", fmt.Sprintf("su 0 ip -6 route show; su 0 ip -6 neigh show; su 0 toybox ping -6 -c 3 -I %s %s 2>&1 || true", guestIP, gatewayIP))
+		t.Fatalf("Dual-stack IPv6 connectivity failed: %v\nStdout: %s\nStderr: %s\nDiag:\n%s", err, connOut.Stdout, connOut.Stderr, diag.Stdout)
+	}
+	t.Logf("Dual-stack IPv6 connectivity successful:\n%s", connOut.Stdout)
+
+	// 3. Uninstall / Flush IPv4 (Simulate IPv6-Only environment)
```
3405691582, line 137: deserves a different test; you are testing too many variables within the one test.

e2etests/cvd/network_tests/main_test.go
3405691582, line 1: Does this test depend on the test host actually having the defaults updated?

How can you run the test on a v4-only host?

3405691582 (changes requested):

base/debian/control
```diff
@@ -50,6 +50,7 @@ Depends: adduser,
          iptables,
          jq,
          less,
+         nftables,
```
dxapd, line 53: this is a duplicate entry.

dxapd (changes requested):

e2etests/cvd/network_tests/main_test.go
sferrogoo: Done. Split the test into two separate flows: TestIPv6DualStackConnectivity and TestIPv6OnlyMode. Extracted the address discovery logic into a standalone pollIPv6Address() helper.

e2etests/cvd/network_tests/main_test.go
sferrogoo: This was a retry mechanism to workaround  several second delay after interface creation to process the host Router Advertisement and complete Duplicate Address Detection (DAD).    I refactored however now to be much cleaner - Replaced the ad-hoc retry loop with a standard deadline-based pollIPv6Address() helper with an explicit 30s timeout.

e2etests/cvd/network_tests/main_test.go
sferrogoo: Refactored to make it more clear: Logic now is: Depending on the device configuration (e.g. WiFi vs Ethernet bridge), the active interface may be eth1 or wlan0. Rather than iterating hardcoded device names, pollIPv6Address() now runs a single ip -6 -o addr show scope global and dynamically selects whichever interface received the Cuttlefish ULA prefix (fd00:cf:...).

e2etests/cvd/network_tests/main_test.go
sferrogoo: Done. Removed the raw neighbor table query and consolidated route installation into configureGuestIPv6Routes(), targeting the gateway discovered during SLAAC.

Let me know if you prefer more factorizing - like splitting out explicit test methods etc, happy to do so if it helps!

e2etests/cvd/network_tests/main_test.go
sferrogoo: Cleaned this ip a bit -  replaced the chained ping fallback string with pingIPv6Target(), which manages the ping command more cleanly.

e2etests/cvd/network_tests/main_test.go
sferrogoo: Done, Split the IPv4 flush and pure IPv6 routing assertions into a distinct TestIPv6OnlyMode test.

e2etests/cvd/network_tests/main_test.go
sferrogoo: You're right it would fail in ipv4 only. I tweaked a bit - Added checkHostIPv6Support(), which runs before fetching or launching the CVD. It checks whether the host cuttlefish bridge (cvd-ebr or cvd-wbr) has an assigned ULA prefix. If the test host is IPv4-only, it calls t.Skip() so CI runs on older/unsupported runners do not fail.

e2etests/cvd/network_tests/main_test.go
sferrogoo: Done. Removed network_tests/ and rolled the IPv6 tests directly into e2etests/cvd/networking_tests/ipv6_test.go so they share the existing test framework and Bazel targets.

base/debian/control
sferrogoo: Fixed!

sferrogoo:
Pushed commit 457a75c addressing all review comments:

1. Removed duplicate nftables dependency in debian/control.
2. Moved tests to e2etests/cvd/networking_tests/ipv6_test.go and integrated with Bazel.
3. Factorized polling and ping logic into reusable helpers with explicit timeouts.
4. Split into two tests: TestIPv6DualStackConnectivity and TestIPv6OnlyMode.
5. Added host bridge check to skip gracefully on IPv4-only test hosts.

sferrogoo:
@dxapd @3405691582 PTAL, addressed all feedback in commit 457a75c.

base/debian/cuttlefish-base.cuttlefish-host-resources.init
```diff
@@ -228,6 +242,11 @@ destroy_interface() {
     ipv6_prefix="${4}"
     ipv6_prefix_length="${5}"
 
+    if [ -n "${ipv6_prefix}" -a -n "${ipv6_prefix_length}" ]; then
+        if [ "${ipv6_enable_nat66:-1}" = "1" ] || [ "${ipv6_enable_nat66}" = "true" ]; then
```
3405691582: please test the init script's behavior when ipv6_enable_nat66 is not provided. if this fails, you'll need a `...:-1` here as well.

PR #3111, opened 2026-08-28
WebRTC streamer and vhost-user-input backend processes communicate with a client-server protocol

jemoreira:
`cf_vhost_user_input` listens on a UNIX socket server it inherits from `run_cvd`. `webRTC` connects to this server to inject input events (mouse, touch, keyboard, etc) to the device backends. This enables the input backend to receive events from sources outside the streamer process.

In addition to the protocol change, the device backend ensures all groups of events sent to the guest driver end in a `SYN_REPORT`. These `SYN_REPORT` events are used to group events from multiple sources so that the driver receives a valid flow of events.

The behavior is enabled by run_cvd by the flags it launches `webRTC` and `cf_vhost_user_input` with. When the `run_cvd` binary is not substituted, events are sent via a socket pair as before. It would be a problem if `run_cvd` was substituted but `webRTC` or `cf_vhost_user_input` was not, but this situation is extremely unlikely since the latter have been substituted for a while now.

Bug: b/552079861

### DIFF
--- a/base/cvd/cuttlefish/host/commands/run_cvd/launch/input_paths_provider.cc
+++ b/base/cvd/cuttlefish/host/commands/run_cvd/launch/input_paths_provider.cc
@@ -13,10 +13,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "cuttlefish/host/commands/run_cvd/launch/input_connections_provider.h"
+#include "cuttlefish/host/commands/run_cvd/launch/input_paths_provider.h"
 
-#include <fcntl.h>
-#include <sys/socket.h>
+#include <sys/socket.h>  // IWYU pragma: keep: SOCK_STREAM
 
 #include <regex>
 #include <string>
@@ -49,40 +48,28 @@ namespace {
 
 // Holds all sockets related to a single vhost user input device process.
 struct DeviceSockets {
-  // Device end of the connection between device and streamer.
-  SharedFD device_end;
-  // Streamer end of the connection between device and streamer.
-  SharedFD streamer_end;
+  // Path to the events server socket.
+  std::string events_server_path;
+  // The events server fd. It's created and held at the CommandSource level to
+  // ensure it already exists by the time the streamer attempts to connect.
+  SharedFD events_server;
   // Unix socket for the server to which the VMM connects to. It's created and
   // held at the CommandSource level to ensure it already exists by the time the
   // VMM runs and attempts to connect.
   SharedFD vhu_server;
 };
 
-Result<DeviceSockets> NewDeviceSockets(const std::string& vhu_server_path) {
-  DeviceSockets ret;
-  CF_EXPECTF(
-      SharedFD::SocketPair(AF_UNIX, SOCK_STREAM, 0, &ret.device_end,
-                           &ret.streamer_end),
-      "Failed to create connection sockets (socket pair) for input device: {}",
-      ret.device_end->StrError());
-
-  // The webRTC process currently doesn't read status updates from input
-  // devices, so the vhost processes will write that to /dev/null.
-  // These calls shouldn't return errors since we already know these are a newly
-  // created socket pair.
-  CF_EXPECTF(ret.device_end->Shutdown(SHUT_WR) == 0,
-             "Failed to close input connection's device for writes: {}",
-             ret.device_end->StrError());
-  CF_EXPECTF(ret.streamer_end->Shutdown(SHUT_RD) == 0,
-             "Failed to close input connection's streamer end for reads: {}",
-             ret.streamer_end->StrError());
-
-  ret.vhu_server =
-      SharedFD::SocketLocalServer(vhu_server_path, false, SOCK_STREAM, 0600);
-  CF_EXPECTF(ret.vhu_server->IsOpen(),
-             "Failed to create vhost user socket for device: {}",
-             ret.vhu_server->StrError());
+Result<DeviceSockets> NewDeviceSockets(const std::string& evt_server_path,
+                                       const std::string& vhu_server_path) {
+  DeviceSockets ret{
+      .events_server_path = evt_server_path,
+      .events_server = CF_EXPECT(
+          Fd::SocketLocalServer(evt_server_path, false, SOCK_STREAM, 0600),
+          "Failed to create event server for device"),
+      .vhu_server = CF_EXPECT(
+          Fd::SocketLocalServer(vhu_server_path, false, SOCK_STREAM, 0600),
+          "Failed to create vhost user socket for device"),
+  };
 
   return ret;
 }
@@ -93,9 +80,7 @@ Command NewVhostUserInputCommand(const DeviceSockets& device_sockets,
   cmd.AddParameter("--verbosity=DEBUG");
   cmd.AddParameter("--socket-fd=", device_sockets.vhu_server);
   cmd.AddParameter("--device-config=", spec);
-  cmd.RedirectStdIO(Command::StdIoChannel::kStdIn, device_sockets.device_end);
-  cmd.RedirectStdIO(Command::StdIoChannel::kStdOut,
-                    Fd::Open("/dev/null", O_WRONLY).value_or(Fd()));
+  cmd.AddParameter("--server-fd=", device_sockets.events_server);
   return cmd;
 }
 
@@ -118,8 +103,7 @@ std::string BuildTouchSpec(const std::string& spec_template,
 }
 
 // Creates the commands for the vhost user input devices.
-class VhostInputDevices : public CommandSource,
-                          public InputConnectionsProvider {
+class VhostInputDevices : public CommandSource, public InputPathsProvider {
  public:
   INJECT(VhostInputDevices(const CuttlefishConfig::InstanceSpecific& instance,
                            LogTeeCreator& log_tee))
@@ -237,41 +221,41 @@ class VhostInputDevices : public CommandSource,
     return commands;
   }
 
-  // InputConnectionsProvider
-  SharedFD RotaryDeviceConnection() const override {
-    return rotary_sockets_.streamer_end;
+  // InputPathsProvider
+  std::string RotaryDevicePath() const override {
+    return rotary_sockets_.events_server_path;
   }
 
-  SharedFD MouseConnection() const override {
-    return mouse_sockets_.streamer_end;
+  std::string MousePath() const override {
+    return mouse_sockets_.events_server_path;
   }
 
-  SharedFD GamepadConnection() const override {
-    return gamepad_sockets_.streamer_end;
+  std::string GamepadPath() const override {
+    return gamepad_sockets_.events_server_path;
   }
 
-  SharedFD KeyboardConnection() const override {
-    return keyboard_sockets_.streamer_end;
+  std::string KeyboardPath() const override {
+    return keyboard_sockets_.events_server_path;
   }
 
-  SharedFD SwitchesConnection() const override {
-    return switches_sockets_.streamer_end;
+  std::string SwitchesPath() const override {
+    return switches_sockets_.events_server_path;
   }
 
-  std::vector<SharedFD> TouchscreenConnections() const override {
-    std::vector<SharedFD> conns;
+  std::vector<std::string> TouchscreenPaths() const override {
+    std::vector<std::string> conns;
     conns.reserve(touchscreen_sockets_.size());
     for (const DeviceSockets& sockets : touchscreen_sockets_) {
-      conns.emplace_back(sockets.streamer_end);
+      conns.emplace_back(sockets.events_server_path);
     }
     return conns;
   }
 
-  std::vector<SharedFD> TouchpadConnections() const override {
-    std::vector<SharedFD> conns;
+  std::vector<std::string> TouchpadPaths() const override {
+    std::vector<std::string> conns;
     conns.reserve(touchpad_sockets_.size());
     for (const DeviceSockets& sockets : touchpad_sockets_) {
-      conns.emplace_back(sockets.streamer_end);
+      conns.emplace_back(sockets.events_server_path);
     }
     return conns;
   }
@@ -281,34 +265,43 @@ class VhostInputDevices : public CommandSource,
   std::string Name() const override { return "VhostInputDevices"; }
   std::unordered_set<SetupFeature*> Dependencies() const override { return {}; }
   Result<void> ResultSetup() override {
-    rotary_sockets_ = CF_EXPECT(NewDeviceSockets(RotarySocketPath(instance_)),
-                                "Failed to setup sockets for rotary device");
+    rotary_sockets_ =
+        CF_EXPECT(NewDeviceSockets(RotaryEventsServerPath(instance_),
+                                   RotarySocketPath(instance_)),
+                  "Failed to setup sockets for rotary device");
     if (instance_.enable_mouse()) {
-      mouse_sockets_ = CF_EXPECT(NewDeviceSockets(MouseSocketPath(instance_)),
-                                 "Failed to setup sockets for mouse device");
+      mouse_sockets_ =
+          CF_EXPECT(NewDeviceSockets(MouseEventsServerPath(instance_),
+                                     MouseSocketPath(instance_)),
+                    "Failed to setup sockets for mouse device");
     }
     if (instance_.enable_gamepad()) {
       gamepad_sockets_ =
-          CF_EXPECT(NewDeviceSockets(GamepadSocketPath(instance_)),
+          CF_EXPECT(NewDeviceSockets(GamepadEventsServerPath(instance_),
+                                     GamepadSocketPath(instance_)),
                     "Failed to setup sockets for gamepad device");
     }
     keyboard_sockets_ =
-        CF_EXPECT(NewDeviceSockets(KeyboardSocketPath(instance_)),
+        CF_EXPECT(NewDeviceSockets(KeyboardEventsServerPath(instance_),
+                                   KeyboardSocketPath(instance_)),
                   "Failed to setup sockets for keyboard device");
     switches_sockets_ =
-        CF_EXPECT(NewDeviceSockets(SwitchesSocketPath(instance_)),
+        CF_EXPECT(NewDeviceSockets(SwitchesEventsServerPath(instance_),
+                                   SwitchesSocketPath(instance_)),
                   "Failed to setup sockets for switches device");
     touchscreen_sockets_.reserve(instance_.display_configs().size());
     for (int i = 0; i < instance_.display_configs().size(); ++i) {
       touchscreen_sockets_.emplace_back(
-          CF_EXPECTF(NewDeviceSockets(instance_.touch_socket_path(i)),
+          CF_EXPECTF(NewDeviceSockets(instance_.touch_events_server_path(i),
+                                      instance_.touch_socket_path(i)),
                      "Failed to setup sockets for touchscreen {}", i));
     }
     touchpad_sockets_.reserve(instance_.touchpad_configs().size());
     for (int i = 0; i < instance_.touchpad_configs().size(); ++i) {
       int idx = touchscreen_sockets_.size() + i;
       touchpad_sockets_.emplace_back(
-          CF_EXPECTF(NewDeviceSockets(instance_.touch_socket_path(idx)),
+          CF_EXPECTF(NewDeviceSockets(instance_.touch_events_server_path(idx),
+                                      instance_.touch_socket_path(idx)),
                      "Failed to setup sockets for touchpad {}", i));
     }
     return {};
@@ -327,10 +320,10 @@ class VhostInputDevices : public CommandSource,
 
 }  // namespace
 fruit::Component<fruit::Required<const CuttlefishConfig::InstanceSpecific>,
-                 InputConnectionsProvider, LogTeeCreator>
+                 InputPathsProvider, LogTeeCreator>
 VhostInputDevicesComponent() {
   return fruit::createComponent()
-      .bind<InputConnectionsProvider, VhostInputDevices>()
+      .bind<InputPathsProvider, VhostInputDevices>()
       .addMultibinding<CommandSource, VhostInputDevices>()
       .addMultibinding<SetupFeature, VhostInputDevices>();
 }
--- a/base/cvd/cuttlefish/host/commands/run_cvd/launch/streamer.cpp
+++ b/base/cvd/cuttlefish/host/commands/run_cvd/launch/streamer.cpp
@@ -25,6 +25,7 @@
 #include <vector>
 
 #include "absl/log/log.h"
+#include "absl/strings/str_join.h"
 #include "fruit/component.h"
 #include "fruit/fruit_forward_decls.h"
 #include "fruit/macro.h"
@@ -32,7 +33,7 @@
 #include "cuttlefish/common/libs/fs/shared_fd.h"
 #include "cuttlefish/common/libs/utils/files.h"
 #include "cuttlefish/host/commands/run_cvd/launch/enable_multitouch.h"
-#include "cuttlefish/host/commands/run_cvd/launch/input_connections_provider.h"
+#include "cuttlefish/host/commands/run_cvd/launch/input_paths_provider.h"
 #include "cuttlefish/host/commands/run_cvd/launch/sensors_socket_pair.h"
 #include "cuttlefish/host/commands/run_cvd/launch/webrtc_controller.h"
 #include "cuttlefish/host/commands/run_cvd/reporting.h"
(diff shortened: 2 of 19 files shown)


base/cvd/cuttlefish/host/commands/run_cvd/launch/input_paths_provider.cc
```diff
@@ -280,34 +268,43 @@ class VhostInputDevices : public CommandSource,
   std::string Name() const override { return "VhostInputDevices"; }
   std::unordered_set<SetupFeature*> Dependencies() const override { return {}; }
   Result<void> ResultSetup() override {
-    rotary_sockets_ = CF_EXPECT(NewDeviceSockets(RotarySocketPath(instance_)),
-                                "Failed to setup sockets for rotary device");
+    rotary_sockets_ =
+        CF_EXPECT(NewDeviceSockets(RotaryEventsServerPath(instance_),
+                                   RotarySocketPath(instance_)),
+                  "Failed to setup sockets for rotary device");
```
Databean, line 271: Confirming my understanding: `run_cvd` creates the unix server sockets and then launches webrtc with command line flags with the server path. This preserves the ordering guarantee that the server sockets are created before the client sockets try to connect to them.

What is gained by passing paths rather than file descriptors to webrtc? Does webrtc now retry connections to the server?

base/cvd/cuttlefish/host/commands/run_cvd/launch/input_paths_provider.cc
```diff
@@ -48,37 +47,29 @@ namespace {
...
-  ret.vhu_server =
-      SharedFD::SocketLocalServer(vhu_server_path, false, SOCK_STREAM, 0600);
+Result<DeviceSockets> NewDeviceSockets(const std::string& evt_server_path,
+                                       const std::string& vhu_server_path) {
+  DeviceSockets ret{
+      .events_server_path = evt_server_path,
+      .events_server = SharedFD::SocketLocalServer(evt_server_path, false,
+                                                   SOCK_STREAM, 0600),
+      .vhu_server = SharedFD::SocketLocalServer(vhu_server_path, false,
+                                                SOCK_STREAM, 0600),
```
Databean: nit: this could be written as
```c++
.events_server = CF_EXPECT(Fd::SocketLocalServer(...)),
.vhu_server = CF_EXPECT(Fd::SocketLocalServer(...)),
```
saving the `->IsOpen()` calls later.

base/cvd/cuttlefish/host/commands/run_cvd/launch/streamer.cpp
```diff
@@ -104,48 +104,48 @@ std::vector<Command> LaunchCustomActionServers(
...
+           input_paths_provider_.TouchpadPaths()) {
+        touch_paths.push_back(touchpad_path);
       }
-      cmd.AddParameter("-touch_fds=", touch_connections[0]);
-      for (int i = 1; i < touch_connections.size(); ++i) {
-        cmd.AppendToLastParameter(",", touch_connections[i]);
+      cmd.AddParameter("-touch_server_paths=", touch_paths[0]);
+      for (int i = 1; i < touch_paths.size(); ++i) {
+        cmd.AppendToLastParameter(",", touch_paths[i]);
       }
```
Databean: This had an awkward construction before to take advantage of `Command`'s special handling of `SharedFD`, but that is no longer needed. This can now be

`cmd.AddParameter("-touch_server_paths=", absl::StrJoin(touch_paths, ","));`

base/cvd/cuttlefish/host/commands/vhost_user_input/src/buf_reader.rs
```diff
@@ -7,42 +8,62 @@ use log::trace;
...
+            size: 0,
+            reader,
+        }
     }
 
-    /// Reads available bytes from the underlying reader.
-    pub fn read_ahead(&mut self) -> Result<()> {
+    /// Reads available bytes from the underlying reader. Returns number of bytes read during this
+    /// call (which may be less than the resulting size of the buffer) or 0 on EOF.
+    pub fn read_ahead(&mut self) -> Result<usize> {
```
Databean: `read_ahead` and `consume` look very similar to methods in [`std::io::BufRead`](https://doc.rust-lang.org/std/io/trait.BufRead.html). Is it possible to implement this trait instead of defining a custom API for these methods?

Implementing `BufRead` is not mutually exclusive with exposing additional `pub` functions in a separate `impl`, so `buffer` and `reader` can still be exposed.

base/cvd/cuttlefish/host/commands/vhost_user_input/src/event_source.rs
```diff
@@ -0,0 +1,279 @@
...
+use std::sync::{Arc, Mutex};
+
+use anyhow::{Context, Result};
+use log::{error, warn};
+use vmm_sys_util::epoll::{ControlOperation, Epoll, EpollEvent, EventSet};
+
+use crate::buf_reader::BufReader;
+use crate::vio_input::{trim_to_event_size_multiple, VIRTIO_INPUT_EVENT_SIZE};
+
+/// Provides input events to the vhost-user backend. Implementations may can get these events from
```
Databean: nit: "may" or "can" but not both.

base/cvd/cuttlefish/host/commands/vhost_user_input/src/event_source.rs
```diff
@@ -0,0 +1,279 @@
...
+/// an event device (e.g /dev/input/event0), a pipe, a server, etc.
+/// The fd returned by as_fd() can be watched with a poll-like function, it will become readable
+/// when new events are available. No assumption should be made about the nature of this fd, in
+/// particular it should not be assumed this is an eventfd and the caller should never attempt to
+/// read from it directly.
+pub trait EventSource: Sync + Send + AsFd {
```
Databean: nit: some places are `Sync + Send` but most are `Send + Sync`, can it be the same throughout?

base/cvd/cuttlefish/host/commands/vhost_user_input/src/event_source.rs
```diff
@@ -0,0 +1,279 @@
...
+    /// Gets all input events available in this source. Input events are always delivered in groups
+    /// ending with the SYN_REPORT type event. The returned list may be empty if only a partial
+    /// groups was read.
```
Databean: nit: "a partial group"

base/cvd/cuttlefish/host/frontend/webrtc/main.cpp
```diff
@@ -316,49 +323,109 @@ int CuttlefishMain() {
...
+           "displays and touchpads";
+    for (int i = 0; i < touch_paths.size(); i++) {
+      const std::string& touch_path = touch_paths[i];
+      // Displays are listed first, then touchpads
+      auto label_prefix =
+          i < display_count ? kTouchscreenPrefix : kTouchpadPrefix;
+      int device_idx = i < display_count ? i : i - display_count;
+      std::string device_label = fmt::format("{}{}", label_prefix, device_idx);
+      SharedFD touch_fd =
+          SharedFD::SocketLocalClient(touch_path, false, SOCK_STREAM);
```
Databean: The new `SharedFD::Xyz(...)` calls can also be written as `Fd::Xyz(...).value_or(Fd())` and automatic conversion of `Fd` to `SharedFD` should handle the rest.

base/cvd/cuttlefish/host/frontend/webrtc/main.cpp
```diff
@@ -58,13 +58,20 @@
...
 DEFINE_bool(multitouch, true,
             "Whether to send multi-touch or single-touch events");
-DEFINE_string(touch_fds, "",
-              "A list of fds to listen on for touch connections.");
-DEFINE_int32(mouse_fd, -1, "An fd to listen on for mouse connections.");
-DEFINE_int32(gamepad_fd, -1, "An fd to listen on for gamepad connections.");
-DEFINE_int32(rotary_fd, -1, "An fd to listen on for rotary connections.");
-DEFINE_int32(keyboard_fd, -1, "An fd to listen on for keyboard connections.");
-DEFINE_int32(switches_fd, -1, "An fd to listen on for switch connections.");
+DEFINE_string(touch_fds, "", "DEPRECATED: Use --touch_server_paths instead.");
```
Databean, line 61: The old flags are still there for the android source tree builds of  `run_cvd`?

base/cvd/cuttlefish/host/commands/vhost_user_input/src/main.rs
```diff
@@ -46,6 +52,53 @@ fn init_logging(verbosity: &str) -> Result<()> {
...
+            vmm_sys_util::epoll::EventSet::IN,
+            EVENTS_AVAILABLE as u64,
+        )
+        .context("Failed to register epoll handler")?;
+
+    let listener = {
+        // SAFETY: Safe because we just dupped this fd and don't use it anywhwere else.
+        // Listener takes ownership and ensures it's properly closed when finished with it.
+        unsafe { Listener::from_raw_fd(server_fd.into_raw_fd()) }
+    };
```
Databean: There's [`From<UnixListener>`](https://docs.rs/crate/vhost/0.17.0/source/src/vhost_user/connection.rs#93) implemented on `Listener` and [`From<OwnedFd>`](https://doc.rust-lang.org/std/os/unix/net/struct.UnixListener.html#impl-From%3COwnedFd%3E-for-UnixListener) on `UnixListener`, does this need the `unsafe` code?

base/cvd/cuttlefish/host/commands/vhost_user_input/src/main.rs
```diff
@@ -46,6 +52,53 @@ fn init_logging(verbosity: &str) -> Result<()> {
...
+    info!("Created vhost-user daemon");
+    daemon
+        .start(listener)
+        .map_err(|e| anyhow!("Failed to start vhost-user daemon: {:?}", e))?;
```
Databean: Can this use `.with_context()`? The [anyhow documentation](https://docs.rs/anyhow/latest/anyhow/) gives an example that uses it with a `std::io::Result` instance.

base/cvd/cuttlefish/host/commands/vhost_user_input/src/main.rs
```diff
@@ -46,6 +52,53 @@ fn init_logging(verbosity: &str) -> Result<()> {
...
+    info!("Accepted connection in vhost-user daemon");
+    if let Err(e) = daemon.wait() {
+        // This will print an error even when the frontend disconnects to do a restart.
+        error!("Error: {:?}", e);
```
Databean: I feel like this message will get missed, is it ok to not propagate this error?

base/cvd/cuttlefish/host/commands/vhost_user_input/src/main.rs
```diff
@@ -65,45 +118,31 @@ fn main() -> Result<()> {
     let device_config = VirtioInputConfig::from_json(device_config_str.as_str())
         .context("Unable to parse config file")?;
 
-    // SAFETY: No choice but to trust the caller passed a valid fd representing a unix socket.
-    let server_fd = inherited_fd::take_fd_ownership(args.socket_fd)
+    // SAFETY: No choice but to trust the caller passed valid fds representing unix sockets.
```
Databean: This is a `SAFETY` comment without an `unsafe` block, it necessary?

base/cvd/cuttlefish/host/commands/vhost_user_input/src/event_source.rs
```diff
@@ -0,0 +1,279 @@
...
+    fn get_events(&mut self) -> Vec<u8>;
+    /// Send status feedback such as keyboard LED states back to the source.
+    fn send_status_feedback(&mut self, status_buffer: Vec<u8>);
```
Databean: It seems like a deficiency of the callers that they can't handle these functions returning errors. Rather than force every implementation to panic or log, can the implementations return errors normally and have the one caller deal with converting the errors to aborts or logs?

In particular the one caller of `send_status_feedback` already returns a `Result`.

base/cvd/cuttlefish/host/commands/vhost_user_input/src/event_source.rs
```diff
@@ -0,0 +1,279 @@
...
+        match std::io::copy(&mut self.events_fd, &mut std::io::sink()) {
+            Err(e) if e.kind() == ErrorKind::WouldBlock => {}
+            Err(e) => {
+                panic!("Failed to read from background thread's channel: {:?}", e);
+            }
+            Ok(_) => {
+                panic!("Background thread's channel closed unexpectedly");
+            }
+        }
+        std::mem::take(&mut *self.events.lock().unwrap())
```
Databean: nit: there is also [`Vec::clear`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.clear).

base/cvd/cuttlefish/host/commands/vhost_user_input/src/event_source.rs
```diff
@@ -0,0 +1,279 @@
...
+        let _ = self.events_fd.write_all(&[0u8; 1]);
+    }
+}
+
+fn server_loop(
+    listener: UnixListener,
+    mut events_fd: UnixStream,
+    events: Arc<Mutex<Vec<u8>>>,
+    status: Arc<Mutex<Vec<u8>>>,
+) {
```
Databean, line 143: `anyhow::Error` is `Send` and `Sync` so it is possible to return error values rather than aborting everywhere.

base/cvd/cuttlefish/host/commands/vhost_user_input/src/event_source.rs
```diff
@@ -0,0 +1,279 @@
...
+            .set_nonblocking(true)
+            .context("Unable to set non-blocking")?;
+        bg_events_fd
+            .set_nonblocking(true)
+            .context("Unable to set non-blocking")?;
+        let events = Arc::new(Mutex::new(Vec::new()));
+        let events_clone = events.clone();
+        let status = Arc::new(Mutex::new(Vec::new()));
+        let status_clone = events.clone();
+        std::thread::spawn(move || server_loop(listener, bg_events_fd, events_clone, status_clone));
```
Databean, line 95: Should the `JoinHandle` for this thread be associated with the `UnixSocketEventSource`? Should this thread outlive the `UnixSocketEventSource`?

base/cvd/cuttlefish/host/commands/vhost_user_input/src/event_source.rs
```diff
@@ -0,0 +1,279 @@
...
+            error!("Error reading from event source: {}", e);
+            0usize
+        }
+    };
+    if read_len == 0 {
+        // EOF reached or error
+        return ReadResult::SourceClosed;
+    }
+    // These events were checked on previous calls.
+    let skip_len = trim_to_event_size_multiple(reader.buffer().len() - read_len);
```
Databean: Shouldn't events on previous calls be removed from consideration using `consume`?

base/cvd/cuttlefish/host/commands/run_cvd/launch/input_paths_provider.cc
jemoreira: Your understanding is correct.

WebRTC doesn't retry connections. The main benefit is that if an instance of webRTC crashes mid write it will corrupt that connection since all future events will be read misaligned by the server. The problem goes away by making each instance of webRTC have their own independent connection. This isn't that big a deal right now because a crashed webRTC brings down the entire device, but that may not be the case in the future.

base/cvd/cuttlefish/host/frontend/webrtc/main.cpp
jemoreira: yes

base/cvd/cuttlefish/host/commands/vhost_user_input/src/event_source.rs
jemoreira: Events from previous reads are only removed from the buffer after receiving the `SYN_REPORT` event. If this code is running on a following read, that means `SYN_REPORT` was not found in the previously read events so those are skipped in the search.

base/cvd/cuttlefish/host/commands/run_cvd/launch/input_paths_provider.cc
jemoreira: Done

base/cvd/cuttlefish/host/commands/run_cvd/launch/streamer.cpp
jemoreira: done

base/cvd/cuttlefish/host/commands/vhost_user_input/src/buf_reader.rs
jemoreira: The functions are similar but not the same, for example BufRead's consume doesn't return the consumed buffer. To implement BufRead I also need to implement Read, which doesn't add value to this class.

base/cvd/cuttlefish/host/commands/vhost_user_input/src/event_source.rs
jemoreira: done

base/cvd/cuttlefish/host/commands/vhost_user_input/src/event_source.rs
jemoreira: done

base/cvd/cuttlefish/host/commands/vhost_user_input/src/event_source.rs
jemoreira: done

base/cvd/cuttlefish/host/commands/vhost_user_input/src/event_source.rs
jemoreira: Vec::clear doesn't return the old vector

base/cvd/cuttlefish/host/commands/vhost_user_input/src/event_source.rs
jemoreira: done

base/cvd/cuttlefish/host/commands/vhost_user_input/src/event_source.rs
jemoreira: I could, but it would be more complicated, and there is no real need for it: The thread will exit when it sees the `events_fd` close on the other side. It doesn't have to return anything, so the handle would only be used for joining.

The worst that could happen if the thread isn't joined is that a similar thread is started with a dup of the UnixStream before this one exits and this one wins an accept call, which ends up dropping that connection. That accept going to the wrong place and the connection being dropped is still possible even if the main thread calls join.

base/cvd/cuttlefish/host/commands/vhost_user_input/src/event_source.rs
jemoreira: There is only one `panic!` call in this function: If there is an error reading from the socket pair connecting this thread to the main one different than EAGAIN and EINTR. With a socket pair this is practically impossible, so a panic seems like the right reaction to me.

base/cvd/cuttlefish/host/commands/vhost_user_input/src/main.rs
jemoreira: We're using version 0.13.0 in which `Listener` doesn't implement `From<UnixListener>`.

base/cvd/cuttlefish/host/commands/vhost_user_input/src/main.rs
jemoreira: This one is a `vhost_user_backend::Error` for which `anyhow::Context` isn't implemented.

base/cvd/cuttlefish/host/commands/vhost_user_input/src/main.rs
jemoreira: Propagate how? exit? Won't `run_cvd` treat this process as essential and bring the whole device down in that case? That would break guest-initiated restarts.

base/cvd/cuttlefish/host/commands/vhost_user_input/src/main.rs
jemoreira: removed

base/cvd/cuttlefish/host/frontend/webrtc/main.cpp
jemoreira: Why is that better than `SharedFD::Xyz()`? I mean, it should probably use `Fd` instead of `SharedFD` altogether, but that's a bigger change that doesn't fit this PR.

Databean: Because fewer calls to the `static` functions in `SharedFD` means we're closer to deleting those functions.

jemoreira: done

base/cvd/cuttlefish/host/commands/vhost_user_input/src/main.rs
Databean: `anyhow::Context` is implemented for any `Result<T, E>` where `E: std::error::Error + Send + Sync + 'static`, it is unfortunate if that condition is not satisfied by `vhost_user_backend::Error`.

base/cvd/cuttlefish/host/commands/vhost_user_input/src/main.rs
Databean: Should `error!` be reserved for exceptional conditions only then, rather than being guaranteed to happen as a result of normal users actions like `adb reboot`?

base/cvd/cuttlefish/host/commands/vhost_user_input/src/event_source.rs
```diff
@@ -0,0 +1,279 @@
...
+                        }
+                        Ok(_) => {
+                            // the other end of events_fd must be closed if io::copy succeeded
+                            return;
+                        }
+                    }
+                    let status = std::mem::take(&mut *status.lock().unwrap());
+                    for client in clients.values_mut() {
+                        // The client's reader is a UnixStream, so it can also write
+                        match client.reader().write_all(&status) {
```
Databean: Should it be called something other than `reader()` if it is used multiple ways?

base/cvd/cuttlefish/host/commands/vhost_user_input/src/buf_reader.rs
Databean: It's still unintuitive that a `BufReader` is neither `std::io::Read` nor `std::io::BufRead`. I'd rather see this type either more closely match the standard IO traits or be renamed and explicitly targeted at virtio events, like maybe `read_events` could be part of the public interface and `buffer` / `read_ahead` / `consume` could be internal-only.

base/cvd/cuttlefish/host/commands/vhost_user_input/src/main.rs
Databean: Leave a TODO comment for when we update?

It looks like internally 0.15, 0.16, and 0.23 are available if you want to update sooner rather than later.

Databean (approved):

base/cvd/cuttlefish/host/commands/vhost_user_input/src/event_source.rs
jemoreira: renamed as `inner`

base/cvd/cuttlefish/host/commands/vhost_user_input/src/main.rs
jemoreira: done

base/cvd/cuttlefish/host/commands/vhost_user_input/src/buf_reader.rs
jemoreira: Renamed as `EventReader`

base/cvd/cuttlefish/host/commands/vhost_user_input/src/main.rs
jemoreira: Detected the case where daemon returns because frontend disconnected and treated it differently than other errors

base/cvd/cuttlefish/host/commands/vhost_user_input/src/event_source.rs
```diff
@@ -16,37 +16,38 @@ use crate::buf_reader::EventReader;
...
     fn send_status_feedback(&mut self, status_buffer: Vec<u8>);
+    /// Creates a new independenlty owned handle to the same underlying event source.
+    fn try_clone(&self) -> Result<Self>;
 }
 
-/// Reads events from a pipe
-pub struct PipeEventSource<R: Read + AsFd + Send + Sync, W: Write + Send + Sync> {
-    reader: EventReader<R>,
-    writer: W,
+/// Reads events from a stdin, writes status feedback to stdout. Typically these channels will be
```
Databean: nit: s/a stdin/stdin/

Databean (approved):

base/cvd/cuttlefish/host/commands/vhost_user_input/src/event_source.rs
jemoreira: done

PR #3114, opened 2026-08-28
Set networking mode before configuration.

3405691582:
ConfigureNetworkSettings reads external_network_mode, so if we override it, we should set it beforehand.


dxapd (approved):

Databean (approved):

Databean:
Kokoro is [failing internally](http://sponge2/0ad1b8d9-94e5-4f78-866f-3d9ecd987d2f) on [cl/972824507](http://cl/972824507) which only has this change and the changelog update.

Databean:
It merged successfully in [cl/973614070](http://cl/973614070).

PR #3120, opened 2026-08-29
vhu_input: Add an endpoint to capture events

jemoreira:
The vhost-user-input backend accepts connections on an extra unix
socket. Events received through the main server are sent to any client
connected to this server in addition to the frontend. At most one
connection is allowed through this server at a time.

Bug: b/554190370

### DIFF
--- a/base/cvd/cuttlefish/host/commands/vhost_user_input/src/main.rs
+++ b/base/cvd/cuttlefish/host/commands/vhost_user_input/src/main.rs
@@ -7,6 +7,7 @@ mod vhu_input;
 mod vio_input;
 
 use std::fs;
+use std::io::ErrorKind;
 use std::os::fd::{AsRawFd, FromRawFd, IntoRawFd, OwnedFd};
 use std::os::unix::net::UnixListener;
 use std::str::FromStr;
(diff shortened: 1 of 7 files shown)


base/cvd/cuttlefish/host/commands/vhost_user_input/src/main.rs
```diff
@@ -135,14 +139,31 @@ fn main() -> Result<()> {
...
         // close them when dropped, so dups of the original fds are used in each iteration.
         let socket_dup = socket_fd.try_clone().context("Failed to clone socket fd")?;
         match server_fd {
             Some(ref fd) => {
                 let server_dup = fd.try_clone().context("Failed to clone server fd")?;
-                let event_source = UnixSocketEventSource::new(UnixListener::from(server_dup))?;
+                let event_source = if args.capture_server_path.is_empty() {
+                    UnixSocketEventSource::new(UnixListener::from(server_dup))?
+                } else {
+                    match fs::remove_file(&args.capture_server_path) {
```
Databean: nit: could also be
```rust
if let Err(e) = fs::remove_file(&args.capture_server_path) && e.kind() != ErrorKind::NotFound {
  ...
}
```

Databean (approved):

Databean:
Reviewed only the final commit.

base/cvd/cuttlefish/host/commands/vhost_user_input/src/main.rs
jemoreira: apparently not allowed in the rust version we use

PR #3121, opened 2026-08-30
vhost_user_media: Worker POLLHUP handling and shmem_unmap fix

changyeon-jo:
### Summary

This PR contains standalone fixes and cleanup for `vhost_user_media` drivers:

1. **Baseline Formatting** (Commit 1):
   - Runs `cargo fmt` on the baseline `v4l2_stream_proxy` crate so subsequent functional changes remain strictly scoped and free of cosmetic noise.

2. **Shmem Unmap Message `len`** (Commit 2):
   - Sets `len` to `0` in `VhostUserMMap` for `shmem_unmap` in `vhu_media/src/lib.rs`, with an explanatory comment noting that the frontend (`crosvm`) removes mappings keyed solely by `shm_offset` and does not read `len`.

3. **POLLHUP Handling in Worker Loop** (Commit 3):
   - Handles `POLLHUP` alongside `POLLIN` in `v4l2_stream_proxy/src/worker.rs` when polling the video stream FIFO. When a FIFO writer disconnects and the buffer is drained, Linux `poll()` reports only `POLLHUP` (without `POLLIN`). Handling `POLLHUP` triggers `read()`, consuming remaining bytes and receiving `Ok(0)` (EOF) to cleanly transition to `WorkerState::Unopened` rather than spinning in a 100% CPU poll busy-loop.

### Testing
- `bazel test //cuttlefish/host/commands/vhost_user_media/...` (Passed)


base/cvd/cuttlefish/host/commands/cvd/cli/parser/instance/cf_media_configs.cpp
```diff
@@ -53,16 +54,27 @@ Result<std::vector<std::string>> GenerateMediaFlags(
           const V4l2StreamProxy& v4l2_stream_proxy = device.v4l2_stream_proxy();
 
           flag += "v4l2_stream_proxy";
-          flag += ":input_path=" + v4l2_stream_proxy.input_path();
-          flag +=
-              ":input_width=" + std::to_string(v4l2_stream_proxy.input_width());
-          flag += ":input_height=" +
-                  std::to_string(v4l2_stream_proxy.input_height());
-          flag += ":input_fps=" + v4l2_stream_proxy.input_fps();
+          if (!v4l2_stream_proxy.input_path().empty()) {
```
bridadan: The validation that's done here drops invalid values silently. That's not the desired behavior as all of these options are required by the vhost user v4l2_stream_proxy binary ([there are no defaults](https://github.com/google/android-cuttlefish/blob/3bd35b0f831649e30e8b22e13557047b4ddf3258/base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/main.rs#L33-L52)).

There's already some of this validation being done on CLI arguments here: http://github.com/google/android-cuttlefish/blob/3bd35b0f831649e30e8b22e13557047b4ddf3258/base/cvd/cuttlefish/host/libs/config/media.cpp#L87-L113

@ser-io if configuration is specified in JSON files instead of CLI arguments, are they still validated by the code I linked to above? This is the flow I have in my head, is that accurate?

```mermaid
flowchart TD
    A[JSON Config] -->|Serialzied to CLI args| B
    D[Terminal] -->|CLI args provided by user| B
    B[CLI Arg String] --> C[Split and validated by host/libs/config/media.cpp]
```

If so, then I'd recommend all the validation should be done in `host/libs/config/media.cpp`.

base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
```diff
@@ -1360,7 +1360,14 @@ Result<CuttlefishConfig> InitializeCuttlefishConfiguration(
...
     auto media_configs = media_configs_bindings[0]->GetConfigs();
-    instance.set_media_configs(media_configs);
+    std::vector<CuttlefishConfig::MediaConfig> instance_media_configs;
+    for (const auto& config : media_configs) {
+      if (!config.instance_index.has_value() ||
+          config.instance_index.value() == instance_index) {
+        instance_media_configs.push_back(config);
+      }
+    }
+    instance.set_media_configs(instance_media_configs);
```
bridadan: Is there a fundamental limitation right now only allowing us to have a single VM use a media config? It seems like it might come from how were using the `injector.getMultibindings<...>`  flow (all the other configs that use it also seem to limit it to 1 binding).

Might be a better question for @ser-io

bridadan (commented):
Thanks for the fixes! I think the final commit adding multi-instance support could be its own PR, the commits proceeding it are nice standalone fixes that I'd like to see merged.

changyeon-jo:
Thanks @bridadan! I've updated this PR to only contain the 4 standalone driver and vhost-user fixes. I'll open a separate follow-up PR for the multi-instance fleet configuration and validation changes.

changyeon-jo:
> Thanks for the fixes! I think the final commit adding multi-instance support could be its own PR, the commits proceeding it are nice standalone fixes that I'd like to see merged.

https://github.com/google/android-cuttlefish/pull/3136 is created. :)

base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc
changyeon-jo: This change is moved to https://github.com/google/android-cuttlefish/pull/3136/changes/5262fa50f1bfc76a6c231edce65dced7836eef04

base/cvd/cuttlefish/host/commands/cvd/cli/parser/instance/cf_media_configs.cpp
changyeon-jo: As suggested, I've spun off the fleet/config changes into follow-up PR #3136 and addressed this there:

Removed silent dropping in cf_media_configs.cpp: The if guards have been removed so all fields are serialized unconditionally into the --media flag.
Centralized validation in host/libs/config/media.cpp: Added explicit CF_EXPECT non-empty checks for input_path and input_fps (alongside the existing input_width > 0 and input_height > 0 checks).

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
```diff
@@ -809,4 +812,135 @@ where
             ..Default::default()
         })
     }
+
+    fn query_ext_ctrl(
```
bridadan: @ser-io Are you ok with lens facing support being added to the `v4l2_stream_proxy` device now?

bridadan (approved):
LGTM, but let's wait for @ser-io

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
ser-io: Why is this needed? Adding this only for the purpose of being aligned with other device implementations is not a valid reason. What's the use case to present this new device as a front/back camera?

ser-io (changes requested):

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
bridadan: > Adding this only for the purpose of being aligned with other device implementations is not a valid reason

I would argue keeping consistent support for `lens_facing` across the device implementations _is_ a valid reason. It reduces cognitive load for our users, allowing them to swap implementations while maintaining a consistent level of base functionality. And from the guest VM's perspective, it doesn't care what the host side implementation looks like, they're all V4L2 devices at the end of the day.

I understand wanting to keep implementations lean, but what's the reason for treating this implementation differently from the existing ones?

base/cvd/cuttlefish/host/commands/vhost_user_media/vhu_media/src/lib.rs
```diff
@@ -157,7 +157,7 @@ impl VirtioMediaHostMemoryMapper for HostMemoryMapper {
             padding: [0, 0, 0, 0, 0, 0, 0],
             fd_offset: 0,
             shm_offset: shm_offset,
-            len: 1,
+            len: end - shm_offset + 1,
```
ser-io: Did you have an production issue you can reference because of this?

From reading the crosvm code they don't use the `len` field: https://github.com/google/crosvm/blob/60f812ebfbf9cec7c4686a07d3f39e7d3b27dbb1/devices/src/virtio/vhost_user_frontend/handler.rs#L94. Please confirm this field is not being read on the frontend (crosvm) and if so, let's leave it rather empty and add a comment explaining it's not needed.

ser-io (changes requested):

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
ser-io: I used the wrong wording, we don't seek alignment among device implementations. Devices should implement relevant controls/features based on actual use cases and not because other device implements it. If there's an actual use case where the v4l2_stream_proxy acts as the front/back camera of a cuttlefish target, you can always implement the control. You shouldn't implement the control just because the mplane implementation does it. The mplane device acts as the front and back cameras of the phone target, hence the implementation of that control. 

If you were to implement a control for the v4l2_stream_proxy to switch the streaming origin, there's no reason we also need to implement that control in the mplane device because is irrelevant.

The "emulated_splane" device is deprecated, it can now be fully replaced by the `emulated_mplane`. there are only two production implementations.

bridadan: > The mplane device acts as the front and back cameras of the phone target, hence the implementation of that control

You _could_ use a `v4l2_stream_proxy` device instead of an `mplane` device for a phone target though right? I suppose that's the main thing I'm getting at. In fact, that could be pretty useful for certain test cases when you want to mock a camera feed with a video file from the host.

> If you were to implement a control for the v4l2_stream_proxy to switch the streaming origin, there's no reason we also need to implement that control in the mplane device because is irrelevant.

Totally agree here. That kind of control is specific to `v4l2_stream_proxy` and wouldn't make sense for `mplane`. Sorry if I made it sound like all controls should be blindly copied without care for how relevant they are. I was specifically advocating for `lens_facing` to work for all devices since it doesn't affect how the frame data is generated.

ser-io: >You could use a v4l2_stream_proxy device instead of an mplane device for a phone target though right?

No, a production cuttlefish phone target configuration shouldn't use v4l2_stream_proxy as back and front cameras because they use static data. You cannot replace the ability to compute sensor data programatically with streaming a static video file in this case as you need full control of the generated data in real time.

Please see the current production emulated sensor codebase used in cuttlefish phone targets: https://cs.android.com/android/platform/superproject/+/android-latest-release:hardware/google/camera/devices/EmulatedCamera/hwl/EmulatedSensor.cpp to have more context  on emulation logic, `emulated_mplane` needs to reach feature parity with that codebase, we are not there yet. From my understanding `v4l2_stream_proxy` was not created for this.

base/cvd/cuttlefish/host/commands/vhost_user_media/vhu_media/src/lib.rs
changyeon-jo: You're right—we verified crosvm's frontend implementation ([handler.rs:108](https://github.com/google/crosvm/blob/60f812ebfbf9cec7c4686a07d3f39e7d3b27dbb1/devices/src/virtio/vhost_user_frontend/handler.rs#L108)) and confirmed that `shared_mapper_state.mapper.remove_mapping(req.shm_offset)` looks up the region solely by `shm_offset`, without inspecting `req.len`.

The issue we ran into during early integration was actually that `vmm_vhost` in crosvm was missing the handler for opcode `BackendReq::SHMEM_UNMAP = 10`. The change to `len: end - shm_offset + 1` was made as a spec-accuracy adjustment when inspecting the unmap call site, rather than due to a frontend length check.

Happy to leave `len: 0` (or `len: 1`) and add a comment explaining that the frontend identifies mappings by `shm_offset` alone.

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
changyeon-jo: Thanks for clarifying the context regarding `emulated_mplane` and `EmulatedSensor.cpp` parity for phone targets.

What do you envision as the primary purpose and scope of `v4l2_stream_proxy`?

From our end, our concrete use case is emulating front and back cameras from media files and external video feeds via `v4l2_stream_proxy`:

1. **XR / Smart Glasses (`cf_x86_64_glasses`)**: For wearable devices, synthetic colored bars / test patterns from `EmulatedSensor` are insufficient because guest perception services (Visual-Inertial Odometry, SLAM, and AR navigation) require realistic real-world camera input. We stream pre-recorded dataset video files (or live camera feeds) into the guest VM via `v4l2_stream_proxy`.
2. **Guest Camera HAL Binding**: Android's Camera HAL inside the guest queries `V4L2_CID_CAMERA_LENS_FACING` to determine whether the device represents a front-facing or back/world-facing camera. Without this control, the HAL cannot correctly bind the video stream to the expected camera role.
3. **Automated Replay Testing**: Even on standard virtual devices, testing camera-dependent apps with reproducible, pre-recorded video files rather than synthetic patterns is a common testing requirement.

Given that `v4l2_stream_proxy` is the mechanism for feeding external media files into Cuttlefish V4L2 devices, adding `lens_facing` enables it to back front or back camera roles in these targets. Would this use case align with how you see the device being utilized?

changyeon-jo: Another way to look at this is through Android's Camera HAL architecture and hardware levels:

1. **`FULL` vs. `LIMITED` Hardware Levels**:
   Under [`CameraCharacteristics.INFO_SUPPORTED_HARDWARE_LEVEL`](https://developer.android.com/reference/android/hardware/camera2/CameraCharacteristics#INFO_SUPPORTED_HARDWARE_LEVEL), devices aiming for `FULL` / `LEVEL_3` require programmatic per-frame sensor physics simulation (manual exposure time, sensitivity/noise calculation, RAW readout—which `EmulatedSensor.cpp` and `emulated_mplane` target).
   However, Android also officially defines `LIMITED` and `EXTERNAL` hardware levels for devices that stream pre-processed frames without per-frame manual sensor computation. Crucially, Android fully supports `LENS_FACING_FRONT` and `LENS_FACING_BACK` on `LIMITED` hardware level devices. In that architectural sense, `v4l2_stream_proxy` acts as a camera device at `LIMITED` (or `EXTERNAL`) level.

2. **Android Emulator Analogy**:
   This mirrors Android Emulator's (Goldfish) `camera-<facing>` options (`emulator -help-camera-back`):
```
  Use -camera-back <mode> to control emulation of a camera facing back.
  Valid values for <mode> are:

     emulated             -> camera will be emulated using software ('fake') camera emulation
     webcam<N>            -> camera will be emulated using a webcamera connected to the host
     environment          -> If the feature is enabled, camera will render a virtual scene that's defined in environment.ini file of the AVD
     imagefile:<filename> -> Camera will render an image file.
     videofile:<filename> -> Camera will render a video file.
     image360:<filename>  -> Camera will render a panoramic image file.
     none                 -> camera emulation will be disabled
```
   There, both synthetic programmatic generation (`emulated`) and streaming from a video file (`videofile`) are supported as complementary ways to back front or back cameras.

In Cuttlefish, `emulated_mplane` fits the `FULL`/`emulated` role, while `v4l2_stream_proxy` nicely fills the `LIMITED`/`videofile` role. In both cases, knowing the lens facing direction (`front` vs `back`) helps the guest camera stack expose and route the stream properly.

Hope this helps provide some extra context on how we were picturing these devices working together!

bridadan: > 2\. **Guest Camera HAL Binding**: Android's Camera HAL inside the guest queries `V4L2_CID_CAMERA_LENS_FACING` to determine whether the device represents a front-facing or back/world-facing camera. Without this control, the HAL cannot correctly bind the video stream to the expected camera role.

@changyeon-jo can you confirm which HAL implementation you're using? I think the only HAL implementation that queries this control is the VirtioMedia implementation in the cuttlefish device (`device/google/cuttlefish/guest/hals/virtio_media/`). Or are you working on your own? I think the existing ExternalCameraProvider implementation does not require the lens_facing property to be set (but it does hard code it `EXTERNAL`)

base/cvd/cuttlefish/host/commands/vhost_user_media/vhu_media/src/lib.rs
ser-io: Let's do `len: 0` and add relevant comment. Thanks.

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
ser-io: >What do you envision as the primary purpose and scope of v4l2_stream_proxy?

@changyeon-jo `v4l2_stream_proxy` was implemented by the SDV team (@bridadan) with the main purpose of replacing guest v4l2_streamer guest command (for now), see ag/41877998, this doesn't require implementing `CID_LENS_FACING` control. 

Now, for `cf_x86_64_glasses` target use cases. Implementing `CID_LENS_FACING` control triggers the use of new Virtio Media HAL which is still **under development** and lacks CI support, see b/543095818. Please wait till the new HAL is stable and discard this change for now. You are now cc'd in that bug.

Please schedule a meeting to better understand your use case for implementing custom `CID_LENS_FACING` for `v4l2_stream_proxy`.

ser-io (changes requested):

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
changyeon-jo: My comment assumes `device/google/cuttlefish/guest/hals/virtio_media/` is the HAL running on virtio-media + `v4l2_stream_proxy`.

base/cvd/cuttlefish/host/commands/vhost_user_media/vhu_media/src/lib.rs
changyeon-jo: Updated commit 3 to set `len: 0` with the explanatory comment and pushed. Thanks!

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
ser-io: I see. No; for now `v4l2_stream_proxy` can solve its objectives working nicely with the stable reference external camera HAL `hardware/interfaces/camera/provider/default/`.

base/cvd/cuttlefish/host/commands/vhost_user_media/vhu_media/src/lib.rs
```diff
@@ -157,7 +157,9 @@ impl VirtioMediaHostMemoryMapper for HostMemoryMapper {
             padding: [0, 0, 0, 0, 0, 0, 0],
             fd_offset: 0,
             shm_offset: shm_offset,
-            len: 1,
+            // The frontend (e.g. crosvm) removes mappings keyed solely by shm_offset
```
ser-io: nit: As we only work with crosvm for the virtio-media devices, let's this comment very specific. 

```
The frontend (crosvm) removes ...
```

Also `and does not use len, so len is set to 0.` seems redundant. We can keep it as.

```
The frontend (crosvm) removes mappings keyed solely by shm_offset, `len` is not used.
```

ser-io (changes requested):

base/cvd/cuttlefish/host/commands/vhost_user_media/v4l2_stream_proxy/src/device.rs
changyeon-jo: Understood. I'll remove `CID_LENS_FACING` implementation from this PR and switch to the external camera HAL until virtio-media HAL is ready.

base/cvd/cuttlefish/host/commands/vhost_user_media/vhu_media/src/lib.rs
changyeon-jo: Updated the comment to match your wording and removed `CID_LENS_FACING` from this PR. Thanks!

ser-io (approved):

ser-io:
Please update the PR summary: https://github.com/google/android-cuttlefish/pull/3121#issue-5293754993 after your changes.

changyeon-jo:
> Please update the PR summary: [#3121 (comment)](https://github.com/google/android-cuttlefish/pull/3121#issue-5293754993) after your changes.

Missed this comment - summary has been updated.

changyeon-jo:
@ser-io when you have a moment, could you please add this rebased PR to the merge queue again? Thanks!

changyeon-jo:
@ser-io I've rebased the branch onto the latest main. Could you please let me know why the previous merge queue attempts failed to merge, and whether there is any action I should take on my end?

PR #3122, opened 2026-08-31
Enable IP forwarding on cuttlefish-podcvd.init

0405ysj:
The debian package `cuttlefish-podcvd` requires to enable IP forwarding. Note that this way follows how `cuttlefish-host-resources` service enabled IP forwarding.

For most cases which I listed below, it was okay to skip enabling IP forwarding from `cuttlefish-podcvd.init` as it's already turned on.
- When `cuttlefish-base` is installed already.
- When `dockerd` is running.

Context: b/504804568

### DIFF
--- a/container/debian/cuttlefish-podcvd.cuttlefish-podcvd.init
+++ b/container/debian/cuttlefish-podcvd.cuttlefish-podcvd.init
@@ -53,6 +53,10 @@ gen_cert() {
 }
 
 start() {
+    # Enable ip forwarding
+    echo 1 > /proc/sys/net/ipv4/ip_forward
+    echo 1 > /proc/sys/net/ipv6/conf/all/forwarding
+
     modprobe vhost_net || true
     modprobe vhost_vsock || true
 


container/debian/cuttlefish-podcvd.cuttlefish-podcvd.init
```diff
@@ -53,6 +53,10 @@ gen_cert() {
 }
 
 start() {
+    # Enable ip forwarding
+    echo 1 > /proc/sys/net/ipv4/ip_forward
+    echo 1 > /proc/sys/net/ipv6/conf/all/forwarding
```
k311093, line 58: For clarification:

I wandered that it looks like changing global configuration. But in PR message, does the configuration already turned on by another pakcage?

k311093 (approved):

container/debian/cuttlefish-podcvd.cuttlefish-podcvd.init
0405ysj: Yes, some other packages turned on such configuration. However, either `docker` or `cuttlefish-base` aren't dependencies of `cuttlefish-podcvd`. It would be fine for most current cases as those might have either one of them, however I newly found the edge case where `cuttlefish-podcvd` should be installed but not `docker` and `cuttlefish-base`.

k311093: Thanks for the clarification!

PR #3124, opened 2026-08-31
bazel: Enable --incompatible_strict_action_env by default

Databean:
Enables `--incompatible_strict_action_env` in `.bazelrc` to prevent spurious cache invalidations and full rebuilds when switching between different client environments (such as interactive terminal shells and IDE / language server subshells).

Without this option, Bazel inherits client environment variables like `PATH` and `LD_LIBRARY_PATH` into action cache keys, causing full cache invalidations when environment variables differ between clients.

When enabled, Bazel uses a static, hermetic value for `PATH` (`/bin:/usr/bin:/usr/local/bin`) and strips `LD_LIBRARY_PATH` across all build actions.

Reference:
http://bazel.build/reference/command-line-reference#param-no-incompatible-strict-action-env

Assisted-by: Jetski:GeminiNext
Bug: b/555354159

### DIFF
--- a/base/cvd/.bazelrc
+++ b/base/cvd/.bazelrc
@@ -1,5 +1,6 @@
 common --vendor_dir=vendor_src
 
+build --incompatible_strict_action_env
 build --build_tag_filters=-clang_tidy,-clang-tidy
 build --copt=-fdiagnostics-color=always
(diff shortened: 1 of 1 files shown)


jemoreira (approved):

jemoreira (approved):
There are more bazel modules under frontend and e2etests

PR #3126, opened 2026-09-01
GCE detection should not error when outside GCE.

3405691582:
HttpGetToString will error out when not in GCE. We should not push the error if the initial GET fails, but instead catch it gracefully and interpret this as being in an environment that is not GCE.

### DIFF
--- a/base/cvd/cuttlefish/metrics/host/gce_environment.cc
+++ b/base/cvd/cuttlefish/metrics/host/gce_environment.cc
@@ -46,13 +46,17 @@ constexpr std::string_view kZone = "instance/zone";
 
 // https://docs.cloud.google.com/compute/docs/instances/detect-compute-engine
 Result<bool> IsGceEnvironment(HttpClient& http_client) {
-  const HttpResponse<std::string> response =
-      CF_EXPECT(HttpGetToString(http_client, std::string(kDetectUrl)));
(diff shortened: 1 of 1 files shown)


base/cvd/cuttlefish/metrics/host/gce_environment.cc
```diff
@@ -46,13 +46,16 @@ constexpr std::string_view kZone = "instance/zone";
 
 // https://docs.cloud.google.com/compute/docs/instances/detect-compute-engine
 Result<bool> IsGceEnvironment(HttpClient& http_client) {
-  const HttpResponse<std::string> response =
-      CF_EXPECT(HttpGetToString(http_client, std::string(kDetectUrl)));
-  CF_EXPECTF(response.HttpSuccess(),
+  auto response = HttpGetToString(http_client, std::string(kDetectUrl));
+  if (!response.has_value()) {
```
ser-io, line 50: Let's add a comment similar to the PR description mentioning that this request will fail if not in  GCE.

ser-io (approved):

base/cvd/cuttlefish/metrics/host/gce_environment.cc
3405691582: Done.

PR #3127, opened 2026-09-01
Add nftables RAII library + unit tests

dxapd:


base/cvd/allocd/net/nft_rule.h
```diff
@@ -0,0 +1,55 @@
...
+namespace cuttlefish {
+
+class NftRule {
+ public:
+  // `tag` is a caller-supplied unique identifier (e.g. an interface name).
+  static Result<NftRule> Create(Nftables& nft, std::string_view family,
+                                std::string_view table, std::string_view chain,
+                                std::string_view content, std::string_view tag);
+
+  NftRule() = delete;
```
Databean: nit: the default no-arg constructor is already deleted implicitly if another constructor is defined.

base/cvd/allocd/net/nft_rule.h
```diff
@@ -0,0 +1,55 @@
...
+  NftRule(Nftables* nft, std::string_view family, std::string_view table,
+          std::string_view chain, std::string comment);
```
Databean: Can this constructor be `private` to avoid bypassing `Create`?

base/cvd/allocd/net/nftables_nft.cc
```diff
@@ -0,0 +1,202 @@
...
+  return *path;
+}
+
+Result<void> NftablesNft::EnsureTable(std::string_view family,
+                                      std::string_view table) {
+  Command cmd{CF_EXPECT(BinaryPath())};
+  cmd.AddParameter("add");
+  cmd.AddParameter("table");
+  cmd.AddParameter(std::string(family));
+  cmd.AddParameter(std::string(table));
```
Databean: Are these explicit conversions to `std::string` necessary? Seems like a deficiency in `Command`.

Databean: nit: can also be written as

```
Command cmd =
    Command(CF_EXPECT(BinaryPath()))
        .AddParameter(...)
        .AddParameter(...);

File: base/cvd/allocd/test/mock_nftables.h [hunk 4]
```diff
@@ -0,0 +1,54 @@
...
+#include <gmock/gmock.h>
+
+#include <string_view>
+
+#include "allocd/net/nftables.h"
+
+namespace cuttlefish {
+
+class MockNftables : public Nftables {
+ public:
```
Databean: What's the motivation for creating both a mock and a fake?

Based on [go/choose-test-double#prefer-fakes](http://go/choose-test-double#prefer-fakes) it seems like the order of preference is

1. Real implementation
2. Fake
3. Mock/Stub

so is the mock necessary?

base/cvd/allocd/test/fake_nftables.h
```diff
@@ -0,0 +1,124 @@
...
+#include "cuttlefish/result/result.h"
+
+namespace cuttlefish {
+
+// A minimal, stateful in-memory Nftables test double.
+//
+// It models a single (family, table, chain) and ignores those arguments. 
+// Its purpose is to reflect the one behaviour that we need statefulness for
+// (handle numbers).
+class FakeNftables : public Nftables {
```
Databean, line 36: Note [go/fakes#need-tests](http://go/fakes#need-tests) recommends creating tests for fakes as well.

> A fake should usually have its own tests to ensure that it conforms to the API of its corresponding real implementation. A fake without tests might initially provide realistic behavior, but without tests, this behavior can diverge over time as the real implementation evolves.

base/cvd/allocd/test/fake_nftables.h
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  // Deleting the table drops its rules and restarts the handle counter, exactly
+  // as nft does when the table is later recreated.
+  Result<void> DeleteTable(std::string_view /*family*/,
+                           std::string_view /*table*/) override {
+    rules_.clear();
+    next_handle_ = 1;
+    return {};
+  }
```
Databean: nit: can the implementation of this be moved to a `.cc` file?

The [style guide](https://google.github.io/styleguide/cppguide.html#Defining_Functions_in_Header_Files) writes

> Only define a function at its public declaration if it is short, say, 10 lines or fewer. Put longer function bodies in the .cc file unless they must be in the header for performance or technical reasons.

My personal preference is to trim header files to the bare minimum, but since most of these functions are short it is defensible to keep them in the header.

In this case, the only benefit from having them in the header is reducing some boilerplate. Since these are all `override` functions anyway, they're probably not possible to inline since they would be accessed only through dynamic dispatch on a `Nftables&` or `Nftables*` value.

base/cvd/allocd/net/nftables_nft.cc
```diff
@@ -0,0 +1,202 @@
...
+    const Json::Value& rule = item["rule"];
+    if (rule.isMember("comment") && rule.isMember("handle") &&
+        rule["comment"].asString() == comment) {
+      handles.push_back(rule["handle"].asUInt64());
+    }
+  }
+
+  for (uint64_t handle : handles) {
+    CF_EXPECT(DeleteRule(family, table, chain, handle));
+  }
```
Databean, line 213: Should this make a best-effort attempt to delete all rules even if one of them fails to be deleted?

base/cvd/allocd/net/nftables_nft.cc
```diff
@@ -0,0 +1,202 @@
...
+  Result<std::string> stdout_str = RunAndCaptureStdout(std::move(cmd));
+  if (!stdout_str.has_value()) {
+    LOG(INFO) << "nft list chain failed, treating as no-op: family=" << family
+              << ", table=" << table << ", chain=" << chain;
+    return {};
+  }
+
+  Json::Value json = CF_EXPECT(ParseJson(*stdout_str));
+  CF_EXPECT(json.isMember("nftables") && json["nftables"].isArray(),
+            "Invalid JSON output from nft: " << *stdout_str);
```
Databean: nit: prefer `CF_EXPECTF` over `CF_EXPECT` with `<<`, see [b/285968197](http://b/285968197)

base/cvd/allocd/net/nftables_nft.cc
```diff
@@ -0,0 +1,202 @@
...
+}
+
+}  // namespace
+
+Result<std::string> NftablesNft::BinaryPath() {
+  static const absl::NoDestructor<std::string> path(
+      SearchForNft().value_or(""));
+  CF_EXPECT(!path->empty(), "could not find nft binary");
+  return *path;
+}
```
Databean, line 63: Can this be a free function in the anonymous namespace rather than a `static` member of `NftablesNft`?

base/cvd/allocd/net/nftables_nft.cc
```diff
@@ -0,0 +1,202 @@
...
+                                               std::string_view table,
+                                               std::string_view chain,
+                                               std::string_view comment) {
+  Command cmd{CF_EXPECT(BinaryPath())};
+  cmd.AddParameter("-j");
+  cmd.AddParameter("list");
+  cmd.AddParameter("chain");
+  cmd.AddParameter(std::string(family));
+  cmd.AddParameter(std::string(table));
+  cmd.AddParameter(std::string(chain));
```
Databean: The [nftables wiki](https://wiki.nftables.org/wiki-nftables/index.php/Simple_rule_management#Removing_all_the_rules_in_a_chain) claims there is
```bash
$ nft flush chain filter output
```
Are there frequently cases where a comment is used to distinguish different categories of rules in a chain? Would it be easier to break the rules into more chains so they don't need to be distinguished by comment within one chain?

base/cvd/allocd/test/BUILD.bazel
```diff
@@ -0,0 +1,24 @@
...
+)
+
+cf_cc_library(
+    name = "mock_nftables",
+    hdrs = ["mock_nftables.h"],
+    deps = [
+        "//allocd/net:nftables",
+        "@googletest//:gtest",
+    ],
+)
```
Databean, line 16: Should this also be `testonly`?

base/cvd/allocd/net/nft_rule.cc
```diff
@@ -0,0 +1,74 @@
...
+
+}  // namespace
+
+Result<NftRule> NftRule::Create(Nftables& nft, std::string_view family,
+                                std::string_view table, std::string_view chain,
+                                std::string_view content,
+                                std::string_view tag) {
+  std::string comment = absl::StrCat(kCvdallocCommentPrefix, tag);
+  std::string full_content =
+      absl::StrCat(content, " comment \"", comment, "\"");
```
Databean: Rather than at the `NftRule` level, should this happen at the `NftNftables` level? Then `FakeNftables` doesn't have to deal with parsing these strings.

base/cvd/allocd/test/fake_nftables.h
```diff
@@ -0,0 +1,124 @@
...
+
+  Result<void> DeleteRulesByComment(std::string_view /*family*/,
+                                    std::string_view /*table*/,
+                                    std::string_view /*chain*/,
+                                    std::string_view comment) override {
+    // A missing match is a no-op, matching idempotent teardown semantics.
+    rules_.erase(
+        std::remove_if(rules_.begin(), rules_.end(),
+                       [&](const Rule& r) { return r.comment == comment; }),
+        rules_.end());
```
Databean: The [erase-remove idiom](https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom) has now been collapsed into [`std::erase`](https://en.cppreference.com/cpp/container/vector/erase2) as of C++20.

base/cvd/allocd/net/nftables_nft.cc
3405691582: Yeah -- you could add an overload as a separate change.

base/cvd/allocd/net/nftables_nft.cc
```diff
@@ -0,0 +1,202 @@
...
+#include "cuttlefish/process/subprocess.h"
+#include "cuttlefish/result/result.h"
+
+namespace cuttlefish {
+
+namespace {
+
+constexpr std::string_view kNftBinary = "nft";
+
+// Searches PATH, then the usual sbin locations, for the nft binary.
```
3405691582: redundant. Instead, you should (concisely) remark that we do the explicit search in case it's not on PATH, so we don't bail prematurely.

base/cvd/allocd/net/BUILD.bazel
```diff
@@ -19,3 +19,38 @@ cf_cc_library(
         "@abseil-cpp//absl/log",
     ],
 )
+
+cf_cc_library(
```
3405691582, line 23: Limit to Linux with 

```
    target_compatible_with = [
        "@platforms//os:linux",
    ],
```

base/cvd/allocd/net/nft_rule.cc
dxapd: yeah

base/cvd/allocd/net/nftables_nft.cc
dxapd: the main reason for using comments here is to avoid accidentally deleting someone else's handle upon table refresh. on top of that i think it would be a bit more complicated to read if a chain was maintained for every instance. so since we can just keep track of whatever state we want in the binary, it seems cleanest to me to just use comments.

base/cvd/allocd/net/nftables_nft.cc
dxapd: this will be used in one of the followup PRs to check if the environment is valid

base/cvd/allocd/net/nftables_nft.cc
dxapd: good call

base/cvd/allocd/net/BUILD.bazel
dxapd: yep

base/cvd/allocd/test/fake_nftables.h
dxapd: yep

base/cvd/allocd/test/BUILD.bazel
dxapd: yep

base/cvd/allocd/test/fake_nftables.h
dxapd: yep

base/cvd/allocd/net/nftables_nft.cc
dxapd: yep

base/cvd/allocd/test/fake_nftables.h
dxapd: yep

base/cvd/allocd/test/mock_nftables.h
dxapd: moved to a fake as discussed

base/cvd/allocd/net/nftables_nft.cc
dxapd: yep

base/cvd/allocd/net/nft_rule.h
dxapd: yep

base/cvd/allocd/net/nft_rule.h
dxapd: yep

base/cvd/allocd/net/nftables_nft.cc
dxapd: yep

base/cvd/allocd/test/fake_nftables.cc
```diff
@@ -0,0 +1,160 @@
...
+                                                std::string_view table,
+                                                std::string_view chain,
+                                                std::string_view comment) {
+  // A missing table/chain or a comment that matches nothing is a no-op,
+  // matching the idempotent teardown semantics of the real implementation.
+  Table* t = FindTable(family, table);
+  if (t == nullptr) {
+    return {};
+  }
+  auto chain_it = t->chains.find(std::string(chain));
```
Databean: If `Table::chains` used a `std::less<void>` comparator, i.e. it is defined as `std::map<std::string, std::vector<Rule>, std::less<void>`, then it would be possible to use `find` with a `std::string_view` without the `std::string` allocation.

The only way to achieve something comparable for `tables_` is if it was a `std::map<std::string, std::map<std::string, ...>>` which I'm not sure is worth it.

base/cvd/allocd/test/fake_nftables.cc
```diff
@@ -0,0 +1,160 @@
...
+Result<void> FakeNftables::DeleteRule(std::string_view family,
+                                      std::string_view table,
+                                      std::string_view chain, uint64_t handle) {
+  Table* t = FindTable(family, table);
+  CF_EXPECTF(t != nullptr, "no such table: family={}, table={}", family, table);
+  auto chain_it = t->chains.find(std::string(chain));
+  CF_EXPECTF(chain_it != t->chains.end(),
+             "no such chain: family={}, table={}, chain={}", family, table,
+             chain);
+  auto removed = std::erase_if(
```
Databean: nit: `auto` is helpful for iterators as it is used elsewhere, but should not be used for integeres.

See https://abseil.io/tips/232

base/cvd/allocd/test/fake_nftables_test.cc
```diff
@@ -0,0 +1,130 @@
...
+  EXPECT_THAT(fake_.AddRule("ip", "t", "missing", "content", "cmt"), IsError());
+}
+
+TEST_F(FakeNftablesTest, HandlesAreUniqueAndMonotonic) {
+  Result<uint64_t> h1 = fake_.AddRule("ip", "t", "c", "content1", "a");
+  Result<uint64_t> h2 = fake_.AddRule("ip", "t", "c", "content2", "b");
+  ASSERT_THAT(h1, IsOk());
+  ASSERT_THAT(h2, IsOk());
+  EXPECT_NE(*h1, *h2);
+  EXPECT_LT(*h1, *h2);
```
Databean: nit: "not equal" is redundant with the next assertion, they can't be equal if one is less than the other.

base/cvd/allocd/test/fake_nftables_test.cc
```diff
@@ -0,0 +1,130 @@
...
+  ASSERT_THAT(first, IsOk());
+
+  ASSERT_THAT(fake_.DeleteTable("ip", "t"), IsOk());
+  EXPECT_FALSE(fake_.HasTable("ip", "t"));
+
+  ASSERT_THAT(fake_.EnsureTable("ip", "t"), IsOk());
+  ASSERT_THAT(fake_.EnsureChain("ip", "t", "c", ""), IsOk());
+  Result<uint64_t> again = fake_.AddRule("ip", "t", "c", "content", "a");
+  ASSERT_THAT(again, IsOk());
+  EXPECT_EQ(*first, *again);
```
Databean: nit: `EXPECT_THAT(again, IsOkAndValue(*first))`

base/cvd/allocd/test/fake_nftables_test.cc
```diff
@@ -0,0 +1,130 @@
...
+#include "cuttlefish/result/result_matchers.h"
+
+namespace cuttlefish {
+namespace {
+
+class FakeNftablesTest : public ::testing::Test {
+ protected:
+  void SetUp() override {
+    ASSERT_THAT(fake_.EnsureTable("ip", "t"), IsOk());
+    ASSERT_THAT(fake_.EnsureChain("ip", "t", "c", ""), IsOk());
```
Databean: nit: [go/unit-testing-practices#setUp](http://go/unit-testing-practices#setUp) writes

> Use setup methods to define implicit defaults for tests, but never let tests explicitly rely on values defined in setup methods.

Databean (approved):

3405691582 (approved):

base/cvd/allocd/test/fake_nftables.cc
dxapd: TIL, thats neat (the syntax is kind of unfortunate but hey its c++)

base/cvd/allocd/test/fake_nftables.cc
dxapd: yep

base/cvd/allocd/test/fake_nftables_test.cc
dxapd: yep

base/cvd/allocd/test/fake_nftables_test.cc
dxapd: moved to an explicitly called function

base/cvd/allocd/test/fake_nftables_test.cc
dxapd: ye

PR #3131, opened 2026-09-02
Add cvd input_devices subcommand

jemoreira:
Bug: b/554190370

### DIFF
--- a/base/cvd/cuttlefish/host/commands/cvd/cli/commands/event_devices.cpp
+++ b/base/cvd/cuttlefish/host/commands/cvd/cli/commands/event_devices.cpp
@@ -1,0 +1,269 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cuttlefish/host/commands/cvd/cli/commands/event_devices.h"
+
+#include <fcntl.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#include <chrono>
+#include <iostream>
+#include <optional>
+#include <string>
+#include <thread>
+#include <utility>
+#include <vector>
+
+#include "absl/log/log.h"
+#include "absl/strings/str_join.h"
+
+#include "cuttlefish/common/libs/fs/fd.h"
+#include "cuttlefish/flag_parser/flag.h"
+#include "cuttlefish/flag_parser/gflags_compat.h"
+#include "cuttlefish/host/commands/cvd/cli/command_request.h"
+#include "cuttlefish/host/commands/cvd/cli/help_format.h"
+#include "cuttlefish/host/commands/cvd/cli/selector/selector.h"
+#include "cuttlefish/host/commands/cvd/instances/instance_manager.h"
+#include "cuttlefish/host/commands/cvd/instances/local_instance.h"
+#include "cuttlefish/io/io.h"
+#include "cuttlefish/io/read_exact.h"
+#include "cuttlefish/io/write_exact.h"
+#include "cuttlefish/result/result.h"
+
+namespace cuttlefish {
+
+namespace {
+const char* kListCommand = "list";
+const char* kLsAlias = "ls";
+const char* kCaptureCommand = "capture";
+const char* kInjectCommand = "inject";
+
+struct __attribute__((packed)) Event {
+  // The virtio_input_event structure actually contains 2 16 bits integers and
+  // one 32 bits one, in little endian representation. However the only
+  // operation on these events is going to be comparing the first two integers
+  // with zero, for which endianness doesn't matter and can be done in a single
+  // comparisson using 32 bits.
+  uint32_t ev_type;
+  uint32_t value;
+};
+static_assert(sizeof(Event) == 8,
+              "Event structure doesn't match virtio_input_event size");
+
+Result<void> SaveEventsToFile(Reader& conn, Writer& out) {
+  std::chrono::steady_clock::time_point start =
+      std::chrono::steady_clock::now();
+  std::vector<Event> events;
+  for (;;) {
+    Event event = CF_EXPECT(ReadExactBinary<Event>(conn),
+                            "Failed to read complete event");
+    events.push_back(event);
+    if (event.ev_type != 0) {
+      continue;
+    }
+    uint64_t timestamp = std::chrono::duration_cast<std::chrono::milliseconds>(
+                             std::chrono::steady_clock::now() - start)
+                             .count();
+    CF_EXPECT(WriteExactBinary(out, timestamp),
+              "Unable to write timestamp to file");
+    CF_EXPECT(WriteExact(out, reinterpret_cast<const char*>(events.data()),
+                         events.size() * sizeof(Event)),
+              "Unable to write events to file");
+    events.clear();
+    // The loop never ends. The command stops when the user interrupts it with
+    // CTRL-C.
+  }
+  return {};
+}
+
+Result<void> InjectEventsFromFile(Writer& conn, Reader& in) {
+  std::chrono::steady_clock::time_point start =
+      std::chrono::steady_clock::now();
+  for (;;) {
+    std::optional<uint64_t> delay =
+        CF_EXPECT(ReadExactBinaryOrEof<uint64_t>(in));
+    if (!delay) {
+      break;
+    }
+    std::this_thread::sleep_until(start + std::chrono::milliseconds(*delay));
+    Event event;
+    do {
+      event = CF_EXPECT(ReadExactBinary<Event>(in));
+      CF_EXPECT(WriteExactBinary(conn, event));
+    } while (event.ev_type != 0);
+  }
+  return {};
+}
+
+}  // namespace
+
+CvdInputDevicesHandler::CvdInputDevicesHandler(
+    InstanceManager& instance_manager)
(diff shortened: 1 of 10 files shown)


base/cvd/cuttlefish/host/commands/cvd/cli/commands/input_devices.cpp
```diff
@@ -0,0 +1,282 @@
...
+      events.clear();
+    }
+    // The loop ends when the user interrupts the command with CTRL-C
+  }
+  return {};
+}
+
+// Attemps to read a delay from the events file, returns false if it can't
+// because of EOF.
+Result<bool> ReadDelay(Fd& in, uint64_t& delay) {
```
Databean: nit: can this accept a `Reader&` since it only calls `Read`?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/input_devices.cpp
```diff
@@ -0,0 +1,282 @@
...
+  char* ptr = reinterpret_cast<char*>(&delay);
+  size_t to_read = sizeof(delay);
+  do {
+    size_t read = CF_EXPECT(in.Read(ptr + sizeof(delay) - to_read, to_read));
+    if (read == 0) {
+      return false;
+    }
```
Databean: This function is "[`ReadExact`](https://github.com/google/android-cuttlefish/blob/cfdadf95a8b8101ca2651bc08e7d3299e531ff1e/base/cvd/cuttlefish/io/read_exact.cc#L27), but distinguish EOF from other errors."

Do you think this is worthwhile to keep somewhere in `cuttlefish/io`?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/input_devices.cpp
```diff
@@ -0,0 +1,282 @@
...
+  // operation on these events is going to be comparing the first two integers
+  // with zero, for which endianness doesn't matter and can be done in a single
+  // comparisson using 32 bits.
+  uint32_t ev_type;
+  uint32_t value;
+};
+static_assert(sizeof(Event) == 8,
+              "Event structure doesn't match virtio_input_event size");
+
+Result<void> SaveEventsToFile(Fd conn, Fd out) {
```
Databean: Can this accept `Reader& conn, Writer& out`? It doesn't rely on any other capabilities of these instances.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/input_devices.cpp
```diff
@@ -0,0 +1,282 @@
...
+    size_t read = CF_EXPECT(in.Read(ptr + sizeof(delay) - to_read, to_read));
+    if (read == 0) {
+      return false;
+    }
+    to_read -= read;
+  } while (to_read > 0);
+  return true;
+}
+
+Result<void> InjectEventsFromFile(Fd conn, Fd in) {
```
Databean: Can this accept `Writer& conn, Reader& in`? It doesn't rely on any other capabilities of these instances.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/event_devices.cpp
```diff
@@ -0,0 +1,282 @@
...
+    cmd = std::move(args[0]);
+    args.erase(args.begin());
+  }
+  if (cmd == kListCommand || cmd == kLsAlias) {
+    CF_EXPECT(ListDevices(request, std::move(args)));
+  } else if (cmd == kCaptureCommand) {
+    CF_EXPECT(CaptureEvents(request, std::move(args)));
+  } else if (cmd == kInjectCommand) {
+    CF_EXPECT(InjectEvents(request, std::move(args)));
+  } else {
```
Databean, line 183: If only `CommandHandler` had a `CanHandle` method, then these all could have been separate `CommandHandler`s.

Maybe they still can, the inner subcommands could still be separate `CommandHandler` classes that are only created inside the `CvdInputDevicesHandler`.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/input_devices.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+Result<void> CvdInputDevicesHandler::CaptureEvents(
+    const CommandRequest& request, std::vector<std::string> args) {
+  std::vector<Flag> flags = CF_EXPECT(Flags(request));
+  CF_EXPECT(ConsumeFlags(flags, args));
+  CF_EXPECTF(args.size() < 2,
+             "Too many standalone arguments provided ({}), expected "
+             "`EVENTS_FILE` only",
+             absl::StrJoin(args, " "));
+  CF_EXPECT(args.size() == 1, "Missing events file name");
```
Databean: `CF_EXPECT_EQ(args.size(), 1, "message")` maybe? It'll print the expected and actual values.

Same for `InjectEvents`.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/input_devices.cpp
```diff
@@ -0,0 +1,282 @@
...
+  CF_EXPECT(!flags_.device_name.empty(),
+            "A device name is required to capture events");
+  if (args[0] == "-") {
+    args[0] = "/proc/self/fd/1";
+  }
+  const auto [instance, _] =
```
Databean: `_` isn't a placeholder in C++20 like it is in Rust. Here it's a regular variable declaration. You'll have to wait for [name-independent declarations](https://en.cppreference.com/cpp/language/conflicting_declarations#Multiple_declarations_of_the_same_entity) in C++26.

Same issue in `InjectEvents`.

Databean (commented):
Reviewed only the last commit.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/input_devices.cpp
Databean: Would also be nice if it returned the output rather than passed it through a reference, maybe something like `Result<std::optional<T>> ReadExactOrEof(Reader&)`.

base/cvd/cuttlefish/host/commands/cvd/cli/commands/input_devices.cpp
jemoreira: done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/input_devices.cpp
jemoreira: done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/input_devices.cpp
jemoreira: done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/event_devices.cpp
jemoreira: > If only CommandHandler had a CanHandle method, then these all could have been separate CommandHandlers.

But then the arguments parsing would have to be repeated 3 times (at worst)

> Maybe they still can, the inner subcommands could still be separate CommandHandler classes that are only created inside the CvdInputDevicesHandler.

That would complicate the help output though. I'd rather have three different commands: `list_input_devices`, `capture_input_events` and `inject_input_events`. But I still prefer this implementation over that one. If you're worried about the file being too long I can make it a directory and split it like monitor does. WDYT?

base/cvd/cuttlefish/host/commands/cvd/cli/commands/input_devices.cpp
jemoreira: done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/input_devices.cpp
jemoreira: done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/input_devices.cpp
jemoreira: done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/event_devices.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+Result<void> SaveEventsToFile(Reader& conn, Writer& out) {
+  std::chrono::steady_clock::time_point start =
+      std::chrono::steady_clock::now();
+  std::vector<Event> events;
+  for (;;) {
+    Event event = CF_EXPECT(ReadExactBinary<Event>(conn),
+                            "Failed to read complete event");
+    events.push_back(event);
+    if (event.ev_type == 0) {
```
Databean: nit: could also be written as
```c++
if (event.ev_type != 0) {
  continue;
}
...
```

base/cvd/cuttlefish/host/commands/cvd/cli/commands/event_devices.cpp
```diff
@@ -0,0 +1,268 @@
...
+              std::chrono::steady_clock::now() - start)
+              .count();
+      CF_EXPECT(WriteExactBinary(out, timestamp),
+                "Unable to write timestamp to file");
+      CF_EXPECT(WriteExact(out, reinterpret_cast<const char*>(events.data()),
+                           events.size() * sizeof(Event)),
+                "Unable to write events to file");
+      events.clear();
+    }
+    // The loop ends when the user interrupts the command with CTRL-C
```
Databean: This line kind of implies that the code in the loop catches the signal. It's not that the loop ends and control returns to the caller, but instead that execution stops right?

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/cli/commands/event_devices.cpp
jemoreira: done

base/cvd/cuttlefish/host/commands/cvd/cli/commands/event_devices.cpp
jemoreira: correct. Updated the comment.

PR #3134, opened 2026-09-03
More SharedFD::Open -> Fd::Open and WriteExact

Databean:
Bug: b/545278508

### DIFF
--- a/base/cvd/cuttlefish/files/copy.cc
+++ b/base/cvd/cuttlefish/files/copy.cc
@@ -22,49 +22,54 @@
 #include <sys/types.h>
 
 #include <string>
+#include <utility>
 
 #include "absl/log/check.h"
 #include "absl/log/log.h"
 
-#include "cuttlefish/common/libs/fs/shared_fd.h"
+#include "cuttlefish/common/libs/fs/fd.h"
+#include "cuttlefish/result/result_type.h"
 
 namespace cuttlefish {
 
 bool Copy(const std::string& from, const std::string& to) {
-  SharedFD fd_from = SharedFD::Open(from, O_RDONLY);
-  SharedFD fd_to = SharedFD::Open(to, O_WRONLY | O_CREAT | O_TRUNC, 0644);
+  Result<Fd> fd_from_res = Fd::Open(from, O_RDONLY);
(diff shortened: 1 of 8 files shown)


base/cvd/cuttlefish/files/copy.cc
```diff
@@ -26,59 +26,59 @@
...
 namespace cuttlefish {
 
 bool Copy(const std::string& from, const std::string& to) {
-  SharedFD fd_from = SharedFD::Open(from, O_RDONLY);
-  SharedFD fd_to = SharedFD::Open(to, O_WRONLY | O_CREAT | O_TRUNC, 0644);
+  Fd fd_from = Fd::Open(from, O_RDONLY).value_or(Fd());
+  Fd fd_to = Fd::Open(to, O_WRONLY | O_CREAT | O_TRUNC, 0644).value_or(Fd());
 
-  if (!fd_from->IsOpen() || !fd_to->IsOpen()) {
+  if (!fd_from.IsOpen() || !fd_to.IsOpen()) {
```
cjreynol: NIT:

We could avoid the `.value_or(Fd())` and `.IsOpen()` calls with something like:

```cpp
...
#include <utility>
...

Result<Fd> fd_from_result = Fd::Open(from, O_RDONLY);
Result<Fd> fd_to_result = Fd::Open(to, O_WRONLY | O_CREAT | O_TRUNC, 0644);

if (!fd_from_result.has_value() || !fd_to_result.has_value()) {
  return false;
}

Fd fd_from = std::move(*fd_from_result);
Fd fd_to = std::move(*fd_to_result);
```

Which puts us closer to eventually converting `Copy` to a `Result` return value, and moves us away from using `Fd()` sentinel values and needing to expose `IsOpen` for that purpose.

But, we do this elsewhere, so I am fine if you close this and move forward as-is.

cjreynol (approved):

base/cvd/cuttlefish/files/copy.cc
Databean: Sounds reasonable, made this change.

Databean:
Thanks for reviewing!

PR #3137, opened 2026-09-03
Move host_tests sandboxing harness into base/cvd

dxapd:
Most of this PR is mechanical stuff (moving files and adding the go toolchain) but it was also a convenient spot to add in the cvdalloc.go that will be needed with the final wiring change as well as the dnsmasq shim that goes with it (since dnsmasq itself cannot run in the namespaces we are making, but we don't really need it to).

So it also poses a good opportunity to review that file in isolation while we're at it. It's not wired up to any test in this PR, but if you want to see an example of it being used: https://github.com/dxapd/android-cuttlefish/blob/a8e54dc2a4e629bd1faab6487712e56cc9ec13f4/base/cvd/host_tests/cvdalloc_test/main_test.go


base/cvd/MODULE.bazel
```diff
@@ -62,3 +64,21 @@ include("//cuttlefish/host/commands/append_squashfs_overlay:append_squashfs_over
...
+    "com_github_google_go_cmp",
+)
+
+new_local_repository = use_repo_rule("@bazel_tools//tools/build_defs/repo:local.bzl", "new_local_repository")
+
+new_local_repository(
+    name = "cuttlefish_debian",
+    build_file_content = 'exports_files(glob(["**"]))\n',
+    path = "../debian",
+)
```
Databean, line 88: The rest of this file is `bazel_dep`s and `include`s, can this be moved elsewhere and be referenced with `include`?

base/cvd/host_tests/common/cvdalloc.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	if _, err := os.Stat(bin); err != nil {
+		t.Fatalf("cvdalloc binary %q is not usable: %v", bin, err)
+	}
+	log.Printf("cvdalloc binary under test: %s", bin)
+
+	// Prepare a tmpfs for the rundir.
+	if _, err := s.Run("sh", "-c", fmt.Sprintf("mkdir -p %s && mount -t tmpfs tmpfs %s", cvdallocRunDir, cvdallocRunDir)); err != nil {
+		t.Fatalf("failed to prepare cvdalloc run dir %q: %v", cvdallocRunDir, err)
+	}
```
Databean, line 68: There's [`os.MkdirTemp`](https://pkg.go.dev/os#MkdirTemp) whic hcan replace `mkdir -p` but within the context of a test it would be better to use [`TempDir` in `testing`](https://pkg.go.dev/testing#T.TempDir) which is automatically removed at the end of the test .

base/cvd/host_tests/common/dnsmasq_shim.sh
```diff
@@ -0,0 +1,39 @@
...
+	esac
+done
+
+setsid /bin/sh -c '
+	pidfile="$1"
+	# Close descriptors above stderr (e.g. the caller control socket) so the
+	# daemon does not keep them open for its whole lifetime.
+	for fd in 3 4 5 6 7 8 9; do
+		eval "exec ${fd}>&-" 2>/dev/null || true
+	done
```
Databean, line 22: Should this also try to capture higher numbers than 9? This [stackexchange post](https://unix.stackexchange.com/a/123430) has some suggestions.

base/cvd/host_tests/common/dnsmasq.go
```diff
@@ -0,0 +1,36 @@
...
+	"path/filepath"
+	"testing"
+)
+
+//go:embed dnsmasq_shim.sh
+var dnsmasqShimScript []byte
+
+func installDnsmasqShim(t *testing.T, s *Sandbox) {
+	dir := filepath.Join(s.tempdir, "shim")
+	if err := os.MkdirAll(dir, 0o755); err != nil {
```
Databean, line 29: Can this use [`T.TempDir()`](https://pkg.go.dev/testing#T.TempDir)?

base/cvd/host_tests/common/cvdalloc.go
```diff
@@ -0,0 +1,229 @@
...
+	stdout *bytes.Buffer
+	stderr *bytes.Buffer
+}
+
+// StartInstance launches "cvdalloc --id=<id> --socket=<fd>" inside the sandbox:
+// 1. A socketpair is created
+// 2. the peer end is passed to the child as fd 3 (inherited through nsenter)
+// 3. this call blocks until the child signals that allocation is complete.
+func (c *Cvdalloc) StartInstance(id int) (*Instance, error) {
+	fds, err := syscall.Socketpair(syscall.AF_UNIX, syscall.SOCK_STREAM, 0)
```
Databean, line 120: There's also `go/x/net/socket` which has [`NewConn`](https://pkg.go.dev/golang.org/x/net/internal/socket#NewConn), is it easy to include this library or will that be a hassle?

base/cvd/host_tests/common/cvdalloc.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+// NewCvdalloc resolves the cvdalloc binary under test and prepares the sandbox
+// to run it.
+//
+// Resolution order:
+//  1. $CVDALLOC_BIN, if set, is used verbatim (an explicit override).
+//  2. Otherwise the binary is located through the Bazel runfiles of the
+//     //cuttlefish/host/commands/cvdalloc:cvdalloc data dependency, whose
+//     runfiles path is passed in $CVDALLOC_RLOCATION.
```
Databean, line 54: I don't see this data dependency included yet, is this supposed to be in `host/tests/common/BUILD.bazel`?

base/cvd/MODULE.bazel
dxapd: yeah i think so

base/cvd/host_tests/common/cvdalloc.go
dxapd: it's included in one of the follow up PRs (the one that wires all of the cvdalloc-side stuff up)

base/cvd/host_tests/common/dnsmasq.go
dxapd: s.tempdir actually comes from T.TempDir() when the sandbox is constructed, so that depends on if there's a reason for this function and the sandbox to call it separately.

base/cvd/host_tests/common/cvdalloc.go
dxapd: i don't think we can use net/internal, right?

base/cvd/host_tests/common/cvdalloc.go
dxapd: that mkdir runs inside the sandbox to create a mirrored path that is hardcoded by the software under test, so i don't think we can use TempDir. it _will_ create that dir on the host if it doesn't exist, though, but i don't think that's a problem

base/cvd/host_tests/common/dnsmasq_shim.sh
dxapd: don't think we can adopt that suggestion without using bash here, but also i'm not 100% sure if there will be such numbers in the context this will be used in.

PR #3139, opened 2026-09-03
Add switches to the host init script to toggle legacy behaviour

dxapd:
add two new flags for toggling static resource initialization as well as running cvdalloc's init (which is to be implemented in the final PR in this series). 

due to the way nftables works, cvdalloc also needs to do a small amount of install-time init (namely, creating its tables and chains). that code is better off living inside the cvdalloc binary, so we have the init script, which is our one hook point into install and uninstall, invoke it at exactly those times.

this change, as is, should be a no-op. the code enclosed within the `if [ "${allocate_static_resources}" = "1" ]; then` branches is the code we hope to delete one day, but until that day it will remain.

### DIFF
--- a/base/debian/cuttlefish-base.cuttlefish-host-resources.default
+++ b/base/debian/cuttlefish-base.cuttlefish-host-resources.default
@@ -36,3 +36,20 @@
 #wifi_ipv6_prefix_length=
 #ethernet_ipv6_prefix=
 #ethernet_ipv6_prefix_length=
+
+# Behavioural settings during the transition to cvdalloc.
+#
+# allocate_static_resources sets up all of the resources
+# expected by the legacy flow. initialize_cvdalloc simply invokes
+# cvdalloc to perform its own per-install setup.
+# Note that for cvd itself to use cvdalloc, it will need the
+# use_cvdalloc flag to be set on its own end.
+#
+# Static resources are disjoint to those allocated by cvdalloc.
+# Thus, the legacy behaviour is maintained by the below defaults.
+#
+# The only invalid combination is `allocate_static_resources=0`
+# with the cvd flag `use_cvdalloc` flipped off.
+#
+#allocate_static_resources=1
+#initialize_cvdalloc=0
--- a/base/debian/cuttlefish-base.cuttlefish-host-resources.init
+++ b/base/debian/cuttlefish-base.cuttlefish-host-resources.init
@@ -44,6 +44,8 @@ ipv4_bridge=${ipv4_bridge:-1}
 ipv6_bridge=${ipv6_bridge:-1}
 dns_servers=${dns_servers:-8.8.8.8,8.8.4.4}
 dns6_servers=${dns6_servers:-2001:4860:4860::8888,2001:4860:4860::8844}
+allocate_static_resources=${allocate_static_resources:-1}
+initialize_cvdalloc=${initialize_cvdalloc:-0}
 
 readonly CUTTLEFISH_RUN_DIR="/run/cuttlefish"
 
@@ -109,18 +111,30 @@ destroy_tap() {
 setup_nftables() {
     mkdir -p "$CUTTLEFISH_RUN_DIR"
 
-    # Ensure nftables tables and chains exist.
-    nft add table ip cuttlefish_nat
-    nft add chain ip cuttlefish_nat postrouting '{ type nat hook postrouting priority 100 ; }'
+    if [ "${initialize_cvdalloc}" = "1" ] && [ -x /usr/lib/cuttlefish-common/bin/cvdalloc ]; then
+        /usr/lib/cuttlefish-common/bin/cvdalloc --setup
+    fi
 
-    nft add table bridge cuttlefish_bridge
-    nft add chain bridge cuttlefish_bridge prerouting '{ type filter hook prerouting priority -250 ; }'
-    nft add chain bridge cuttlefish_bridge forward '{ type filter hook forward priority 0 ; }'
+    if [ "${allocate_static_resources}" = "1" ]; then
+        # Ensure nftables tables and chains exist.
+        nft add table ip cuttlefish_nat
+        nft add chain ip cuttlefish_nat postrouting '{ type nat hook postrouting priority 100 ; }'
+
+        nft add table bridge cuttlefish_bridge
+        nft add chain bridge cuttlefish_bridge prerouting '{ type filter hook prerouting priority -250 ; }'
+        nft add chain bridge cuttlefish_bridge forward '{ type filter hook forward priority 0 ; }'
+    fi
 }
 
 delete_nftables() {
-    nft delete table ip cuttlefish_nat
-    nft delete table bridge cuttlefish_bridge
+    if [ "${initialize_cvdalloc}" = "1" ] && [ -x /usr/lib/cuttlefish-common/bin/cvdalloc ]; then
+        /usr/lib/cuttlefish-common/bin/cvdalloc --teardown
+    fi
+
+    if [ "${allocate_static_resources}" = "1" ]; then
+        nft delete table ip cuttlefish_nat
+        nft delete table bridge cuttlefish_bridge
+    fi
 }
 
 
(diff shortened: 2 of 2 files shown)


3405691582:
> due to the way nftables works, cvdalloc also needs to do a small amount of install-time init (namely, creating its tables and chains).

Can we do this without having to do this at init? Can we interrogate nftables and see if our tables and chains exist, and if not, install them then?

dxapd:
> > due to the way nftables works, cvdalloc also needs to do a small amount of install-time init (namely, creating its tables and chains).
> 
> Can we do this without having to do this at init? Can we interrogate nftables and see if our tables and chains exist, and if not, install them then?

that can theoretically be done for setup, but for teardown, we'd need to do something similar to refcounting to reason out that it's safe to tear down all of the resources. i don't think that machinery is worth it in this case, given that the service stopping is a pretty natural hook to clean up the tables and chains. and then (IMO) once we have that teardown path, it's nice and simple to just expand it a bit to have a simple setup routine to go with it. 

which i guess boils down to  - it's possible, but imo this is simpler and we're handed the lifecycle for free by debian itself

3405691582:
> > > due to the way nftables works, cvdalloc also needs to do a small amount of install-time init (namely, creating its tables and chains).
> > 
> > 
> > Can we do this without having to do this at init? Can we interrogate nftables and see if our tables and chains exist, and if not, install them then?
> 
> that can theoretically be done for setup, but for teardown, we'd need to do something similar to refcounting to reason out that it's safe to tear down all of the resources. i don't think that machinery is worth it in this case, given that the service stopping is a pretty natural hook to clean up the tables and chains. and then (IMO) once we have that teardown path, it's nice and simple to just expand it a bit to have a simple setup routine to go with it.
> 
> which i guess boils down to - it's possible, but imo this is simpler and we're handed the lifecycle for free by debian itself

Ideally, we shouldn't need the init script at all. I don't know if we need to _fully_ refcount; for example, if I remember correctly the existing implementation checks the bridge interface's associated addresses and only tears the bridge down if it's the last. I might not be fully having all the context at hand with nftables though.

If that's a bigger lift, nonetheless we should at least init dynamically and teardown statically and revisit stateless teardown later.

dxapd:
i don't think i agree about this case; there is a handful of things that we will need to keep the init script around for (unless there is a plan to move off of the debian package that i didn't know about), so these hook points will always be there for the taking 

and even beyond that, it's just very simple to reason about (imo) "run the setup routine when the program is installed, then run the teardown routine when it is removed" vs. depending on each instance at runtime to have some machinery to construct its own environment statelessly.

for the teardown case, there's two ways I'm familiar of:
- nftables doesn't have any direct way to count rules, so the way around that would be to get it to output the chain in json and count the amount of rules in there. if only the current instance's rule, then clean up.
- on the other hand, to do it in a refcounting manner, yeah you'd need to track that count state somewhere.

for the init case, sure, it's possible to make that idempotent. just make the expected environment exist at init time. but is that needed, given that we have the ideal hook points available already, and we're already using one of them? dunno, this one is less defensible to me than my reasoning for teardown above, but i do think it makes it very easy to understand how the assumptions are set up. either cvdalloc is installed or it isn't. unless of course there are pitfalls to the way debian handles it that i don't know about.

that aside, i do feel more strongly about teardown being run at uninstall time, just as a matter of hygiene. we're uninstalling the application, so we make sure we've cleaned house. i'd want that to be there regardless, i think.

so it's not really about how much of a lift it is; i just think it's cleaner, as well as easier to understand what assumptions the allocator is operating under like this. but perhaps there's an angle to this i haven't thought about yet.

Databean:
Adding to @3405691582's points, part of the motivation of cvdalloc in the first place was to make the debian packages less intrusive to have installed, to make them more palatable for debian upstream. The average public user that installs the packages may be interested in launching a Cuttlefish device on occasion, but is less likely to be launching one every day. It's nicer to them to have fewer runtime system resources allocated while they're not actively engaging with our software.

dxapd:
> Adding to @3405691582's points, part of the motivation of cvdalloc in the first place was to make the debian packages less intrusive to have installed, to make them more palatable for debian upstream. The average public user that installs the packages may be interested in launching a Cuttlefish device on occasion, but is less likely to be launching one every day. It's nicer to them to have fewer runtime system resources allocated while they're not actively engaging with our software.

yes, i understand that, but it's not clear to me what the line between intrusive and standard is, especially since a single nftables table won't clutter any command outputs like static resources does.

that said! all of this discussion got me thinking, and I'm quite optimistic that making a fresh nftables table for every instance might actually be super clean and let us delete setup and teardown modes. i think that would resolve all of our concerns simultaneously since there would be no special running modes, no state to leak during uninstall, and the logic would also be dead simple... I'll look into whether or not it's possible

dxapd:
Got rid of initialize_cvdalloc since I conrfirmed that we can make it work with fresh nftables every time.

base/debian/cuttlefish-base.cuttlefish-host-resources.default
```diff
@@ -36,3 +36,15 @@
...
+#
+# allocate_static_resources sets up all of the resources
+# expected by the legacy flow.
+#
+# Static resources are disjoint to those allocated by cvdalloc.
+# Thus, the legacy behaviour is maintained by the below default.
+# It is invalid to flip the below flag off while use_cvdalloc is
+# also disabled.
+#
+#allocate_static_resources=1
```
Databean, line 50: nit: newline at end of file

Databean (approved):

base/debian/cuttlefish-base.cuttlefish-host-resources.default
```diff
@@ -36,3 +36,15 @@
 #wifi_ipv6_prefix_length=
 #ethernet_ipv6_prefix=
 #ethernet_ipv6_prefix_length=
+
+# Behavioural settings during the transition to cvdalloc.
+#
```
3405691582:
```suggestion
```

base/debian/cuttlefish-base.cuttlefish-host-resources.default
```diff
@@ -36,3 +36,15 @@
...
+# allocate_static_resources sets up all of the resources
+# expected by the legacy flow.
```
3405691582:
```suggestion
# allocate_static_resources creates resources like tap devices 
# statically for instance networking at boot.
```

base/debian/cuttlefish-base.cuttlefish-host-resources.default
```diff
@@ -36,3 +36,15 @@
...
+#
+# Static resources are disjoint to those allocated by cvdalloc.
+# Thus, the legacy behaviour is maintained by the below default.
+# It is invalid to flip the below flag off while use_cvdalloc is
```
3405691582: Is it? It won't be helpful, but is it actually invalid? Why?

base/debian/cuttlefish-base.cuttlefish-host-resources.init
```diff
@@ -385,33 +391,35 @@ start() {
...
-      "${wifi_ipv6_prefix}" "${wifi_ipv6_prefix_length}"
-    for i in $(seq ${num_cvd_accounts}); do
-        tap="$(printf cvd-wifiap-%02d $i)"
-        if [ $i -lt 65 ]; then
-            destroy_interface $tap 192.168.94 $i
-        elif [ $i -lt 129 ]; then
-            destroy_interface $tap 192.168.95 $(($i - 64))
-        fi
-    done
+    if [ "${allocate_static_resources}" = "1" ]; then
```
3405691582: I think you could reverse sense here and the nesting will drop: if allocate_static_resources is not 1, delete nftables, return, fi, destroy interfaces, etc.

base/debian/cuttlefish-base.cuttlefish-host-resources.init
```diff
@@ -331,43 +336,44 @@ start() {
...
-      "${wifi_ipv6_prefix}" "${wifi_ipv6_prefix_length}"
-    for i in $(seq ${num_cvd_accounts}); do
-        tap="$(printf cvd-wifiap-%02d $i)"
-        if [ $i -lt 65 ]; then
-            create_interface $tap 192.168.94 $i
-        elif [ $i -lt 129 ]; then
-            create_interface $tap 192.168.95 $(($i - 64))
-        fi
-    done
+    if [ "${allocate_static_resources}" = "1" ]; then
```
3405691582: You could reverse sense here too to avoid the indentation (see below).

PR #3142, opened 2026-09-04
Update composite disk mtime when disk spec matches existing

Elie-Kh:
When CreateOrUpdateCompositeDisk determines that an existing composite disk already matches the target partition layout, it skips regenerating the disk file. However, if a component partition (such as persistent_vbmeta.img) was regenerated or touched during instance setup, the composite disk's modification time would remain older than that component. This causes subsequent assemble_cvd checks to falsely detect that a component was updated and fail with an mtime mismatch error.

Update the modification time of output_composite_path using utimensat when the existing disk spec is retained, ensuring the composite disk mtime remains up to date.

Bug: 554545294

### DIFF
--- a/base/cvd/cuttlefish/host/libs/image_aggregator/image_aggregator.cc
+++ b/base/cvd/cuttlefish/host/libs/image_aggregator/image_aggregator.cc
@@ -23,6 +23,7 @@
 #include <fcntl.h>
 #include <stdint.h>
 #include <string.h>
+#include <sys/stat.h>
 
 #include <fstream>
 #include <ios>
@@ -31,6 +32,7 @@
 #include <string>
 #include <vector>
 
+#include "absl/log/log.h"
 #include "google/protobuf/util/message_differencer.h"
 #include <zlib.h>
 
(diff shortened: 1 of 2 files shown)


base/cvd/cuttlefish/host/libs/image_aggregator/image_aggregator.cc
```diff
@@ -361,7 +364,12 @@ Result<void> CreateOrUpdateCompositeDisk(
       google::protobuf::util::MessageDifferencer::Equals(
           composite_proto, composite_image_res->GetCompositeDisk())) {
     // The existing composite disk matches the given partitions, no need to
-    // regenerate
+    // regenerate, but update its modification time so it is not older than
+    // any recently updated component images.
+    if (utimensat(AT_FDCWD, output_composite_path.c_str(), nullptr, 0) != 0) {
+      LOG(WARNING) << "Failed to update modification time for \""
+                   << output_composite_path << "\": " << strerror(errno);
```
Databean: nit: can use `PLOG` instead of `<< ":" << strerror(errno)`

Databean (approved):
This change is [failing on linters](https://github.com/google/android-cuttlefish/actions/runs/33894692896/job/101215947958?pr=3142)

```
/usr/local/google/home/schuffelen/.cache/bazel/_bazel_schuffelen/e5255ea4f0651a267b9880f9e2ad5869/sandbox/linux-sandbox/801/execroot/_main/cuttlefish/host/libs/image_aggregator/image_aggregator.cc:371:67: error: no header providing "errno" is directly included [misc-include-cleaner,-warnings-as-errors]
   23 |                    << output_composite_path << "\": " << strerror(errno);
      |                                                                   ^
```

base/cvd/cuttlefish/host/libs/image_aggregator/image_aggregator.cc
Elie-Kh: done

PR #3146, opened 2026-09-04
Remove full output comparison assertion in create_from_images_zip_test

ser-io:
Bug: b/557197628

### DIFF
--- a/e2etests/orchestration/create_from_images_zip_test/main_test.go
+++ b/e2etests/orchestration/create_from_images_zip_test/main_test.go
@@ -76,7 +76,7 @@ func TestCreateInstance(t *testing.T) {
 		EnvConfig: envConfig,
 	}
 
-	got, createErr := srv.CreateCVD(createReq, &hoclient.AccessTokenBuildAPICreds{})
+	res, createErr := srv.CreateCVD(createReq, &hoclient.AccessTokenBuildAPICreds{})
 
 	if err := common.DownloadHostBugReport(srv, group_name); err != nil {
 		t.Errorf("failed creating bugreport: %s\n", err)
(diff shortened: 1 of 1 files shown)


e2etests/orchestration/create_from_images_zip_test/main_test.go
```diff
@@ -76,28 +80,32 @@ func TestCreateInstance(t *testing.T) {
...
+	if err := adbH.Connect(cvd.ADBSerial); err != nil {
+		t.Fatal(err)
+	}
+	fingerprint, err := adbH.ExecShellCommand(cvd.ADBSerial, []string{"getprop", "ro.build.fingerprint"})
+	if err != nil {
+		t.Fatal(err)
 	}
-	if diff := cmp.Diff(want, got); diff != "" {
-		t.Errorf("response mismatch (-want +got):\n%s", diff)
+	if diff := cmp.Diff("generic/aosp_cf_x86_64_only_phone/vsoc_x86_64_only:16/BP4A.251205.006/15150359:userdebug/test-keys\n", fingerprint); diff != "" {
```
jemoreira: This is going to be a problem when those artifacts change. It seems to me this is checking that `adb shell` works (which can be done without comparing the fingerprint, just ensure there is some value in that property) and that the booted device is in fact the one in the artifacts (which I don't think is necessary, but can be done by ensuring no devices are running before booting this one)

jemoreira (approved):

e2etests/orchestration/create_from_images_zip_test/main_test.go
ser-io: Good point, removed asserting on the fingerprint.

PR #3148, opened 2026-09-04
Host Orchestrator endpoints to list event devices and inject events

jemoreira:
Assisted-by: Gemini-Next

Bug: b/554190370


jemoreira:
This PR is based on https://github.com/google/android-cuttlefish/pull/3131. Only the top two commits are new.

frontend/src/host_orchestrator/api/v1/messages.go
```diff
@@ -154,3 +154,7 @@ type DisplayScreenshotResponse struct {
 type ListScreenRecordingsResponse struct {
 	ScreenRecordings []string `json:"screen_recordings"`
 }
+
+type ListInputDevicesResponse struct {
```
ser-io: should it be "EventDevices" across the board?

frontend/src/host_orchestrator/api/v1/messages.go
```diff
@@ -154,3 +154,7 @@ type DisplayScreenshotResponse struct {
...
+	InputDevices []string `json:"event_devices"`
```
ser-io: should we have a new struct "EventDevice" to be future proof and avoid using `[]string`. Wondering whether in the future we might have to return other properties other than name, for example `type`.

ser-io (changes requested):

frontend/src/host_orchestrator/api/v1/messages.go
jemoreira: done

frontend/src/host_orchestrator/api/v1/messages.go
jemoreira: done

frontend/src/host_orchestrator/orchestrator/cvd/cvd_test.go
```diff
@@ -42,5 +44,111 @@ func TestSliceItoa(t *testing.T) {
...
+				return
+			}
+			if diff := cmp.Diff(tc.want, got); diff != "" {
+				t.Errorf("result mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
 
+func TestInjectInputEvents(t *testing.T) {
```
ser-io: Can we add an orchestration e2e tests for injecting events in this PR? if so I think we can get rid of this unit test.

frontend/src/host_orchestrator/orchestrator/cvd/cvd_test.go
```diff
@@ -42,5 +44,111 @@ func TestSliceItoa(t *testing.T) {
 			t.Errorf("result mismatch (-want +got):\n%s", diff)
 		}
 	}
+}
+
+func TestParseListEventDevicesOutput(t *testing.T) {
```
ser-io: if underlying "cvd event_devices" list sub commands changes its output format this test won't capture that. 

Can we rather have an e2e tests? I'm not sure whether we could cover all this scenarios in a single e2e tests though.

frontend/src/host_orchestrator/orchestrator/listeventdevicesaction_test.go
ser-io, line 1: If adding a e2e tests I think we no longer need this unite tests.

ser-io (changes requested):

frontend/src/host_orchestrator/orchestrator/listeventdevicesaction_test.go
jemoreira: done

frontend/src/host_orchestrator/orchestrator/cvd/cvd_test.go
jemoreira: done

frontend/src/host_orchestrator/orchestrator/cvd/cvd_test.go
jemoreira: done

e2etests/orchestration/inject_input_events_test/main_test.go
```diff
@@ -0,0 +1,67 @@
...
+	devices, err := srv.ListEventDevices(cvd.Group, cvd.Name)
+	if err != nil {
+		t.Fatalf("Failed to list event devices: %v", err)
+	}
+	const deviceName = "keyboard"
+	if !slices.Contains(devices, hoapi.EventDevice{Name: deviceName}) {
+		t.Fatalf("Device %q not found in event devices list: %v", deviceName, devices)
+	}
+
+	if err := srv.InjectInputEvents(cvd.Group, cvd.Name, deviceName, bytes.NewReader(events)); err != nil {
```
ser-io, line 64: just curious if there might be a way to verify the events were injected via "adb"?

ser-io (approved):

e2etests/orchestration/inject_input_events_test/main_test.go
jemoreira: Not easily: I can check that the events were delivered with `adb shell getevent -lqc 8`, but that command won't exit until 8 events are delivered to "some" input device so  detecting a failure is not trivial.

jemoreira: Gave it a try, but I can't find an elegant way to ensure event injection happens after `getevent` starts listening. I uploaded the diff to https://paste.googleplex.com/4591301870354432 in case you want to see how it looks like.

PR #3151, opened 2026-09-07
Print stderr message without any confusing logs

0405ysj:
So far, `podcvd create` printed some stderr messages however it doesn't fit to `podcvd` UX. This PR aims to modify stderr message, to make users & agents understand the log better.

Before:
```
  Run `cvd logs` to report paths to device logs.
  Point your browser to https://localhost:8443 to interact with the device.
  Serial console is disabled; use -console=true to enable it.
  Launcher Build ID:
```

After:
```
  Run `podcvd logs` to report paths to device logs.
  Point your browser to https://192.168.112.1:11443/devices/cvd_1-1-1/files/client.html to interact with the device.
  Serial console is disabled; use -console=true to enable it.
  Launcher Build ID: 
```

Context: b/555922254

### DIFF
--- a/base/cvd/cuttlefish/host/commands/run_cvd/launch/streamer.cpp
+++ b/base/cvd/cuttlefish/host/commands/run_cvd/launch/streamer.cpp
@@ -18,7 +18,7 @@
 #include <errno.h>
 #include <sys/socket.h>
 
-#include <sstream>
+#include <optional>
 #include <string>
 #include <unordered_set>
 #include <utility>
@@ -26,11 +26,13 @@
 
 #include "absl/log/log.h"
 #include "absl/strings/str_join.h"
+#include "fmt/format.h"
 #include "fruit/component.h"
 #include "fruit/fruit_forward_decls.h"
 #include "fruit/macro.h"
 
 #include "cuttlefish/common/libs/fs/shared_fd.h"
+#include "cuttlefish/common/libs/utils/environment.h"
 #include "cuttlefish/common/libs/utils/files.h"
 #include "cuttlefish/host/commands/run_cvd/launch/enable_multitouch.h"
 #include "cuttlefish/host/commands/run_cvd/launch/input_paths_provider.h"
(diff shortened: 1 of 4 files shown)


base/cvd/cuttlefish/host/commands/run_cvd/launch/streamer.cpp
```diff
@@ -218,10 +220,16 @@ class WebRtcServer : public virtual CommandSource,
     if (!Enabled()) {
       return {};
     }
-    std::ostringstream out;
-    out << "Point your browser to https://localhost:"
-        << config_.sig_server_proxy_port() << " to interact with the device.";
-    return {out.str()};
+    std::string url;
+    if (StringFromEnv("CVD_INVOKER").value_or("") == "podcvd") {
```
0405ysj, line 224: @Databean I want your review as this PR is not only for `podcvd` change, but also a bit of `run_cvd`.

This change is required for podcvd and won't be harmful as the container image always include both `cuttlefish-base` and `cuttlefish-user`. However speaking of `cvd`, I'm uncertain whether `https://localhost:8443` is still the primary URL we want to show to users or not, and I guess not.

Databean: The operator server isn't running in the container though right, the user is still interacting with the one provided by `cuttlefish-user` on the host?

Databean (approved):

base/cvd/cuttlefish/host/commands/run_cvd/launch/streamer.cpp
0405ysj: In `podcvd` case, there's always `cuttlefish-user` in the container and its port(1443) is exposed to the outside.

My wondering was if relying on 1443 is only acceptable for `podcvd`, or `cvd` as well. For example, I wondered there's still some cases that only installed `cuttlefish-base`, not `cuttlefish-user`. Or, I'm thinking about removing conditional statement(`if (StringFromEnv("CVD_INVOKER").value_or("") == "podcvd")`).

Databean: Is it exposed at port `1443` of the host though? Not every container can simultaneously expose `1443` at `1443` in the host, they would conflict with each other and with the equivalent `cuttlefish-user` service of the host.

0405ysj: No, port 1443 of each container is exposed to `<assigned_ip_for_container>:11443` today. Thus, no conflicts.

Previously, it printed `https://localhost:8443`. With this commit, cvd in the container prints `https://localhost:1443/devices/cvd_1-1-1/files/client.html`. And podcvd captures the stderr of cvd execution and modify as `https://192.168.112.1:11443/devices/cvd_1-1-1/files/client.html`.

It seems my change at `streamer.cpp` is a bit confusing. Let me add explanation comments there when I get back to the office.

0405ysj: Left explanation comment in the code. Thanks for the review!

PR #3153, opened 2026-09-08
Add SDV 1-VM and 2-VM presubmit E2E tests 

SuperStrongDinosaur:
This PR introduces end-to-end presubmit test coverage for SDV targets under `e2etests/cvd/sdv_tests`.                                                                
                                                                                                                                                                                                    
The test suite validates SDV device lifecycle operations (status querying, graceful stop, start, and restart) and multi-VM isolation based on the device-host interaction baseline tests:                                                                                                                                                                      
    - Single-VM lifecycle operations (`sdv_fw_baseline_dhi_1vm_test.py`)                                                                                                                            
    - Two-VM multi-instance lifecycle and isolation (`sdv_fw_baseline_dhi_2vm_test.py`)                                                                                                             
                                                                                                                                                                                                   
BUG: b/559633154


chihchiachen:
Instead of treating SDV CF as an outlier, a better approach is to generalize the new test cases so they can optionally support non-SDV CF variants.

So I recommend the following changes:
- Generalize `sdv_1vm_test` as `lifecycle_test` and use it to replace `cvd_create_tests`.
- Generalize `sdv_2vm_test` as `isolation_test`, apply it to `sdv_media` + `sdv_core`, and then clean up the dupplicate config from `cvd_load_tests`.

@Databean please let me know your thoughts.

e2etests/cvd/common/common.go
```diff
@@ -131,6 +133,43 @@ func (tc *TestContext) RunAdbWaitForDevice() error {
...
 }
 
+// Waits for a specific device serial to connect to adb and become reachable.
+func (tc *TestContext) RunAdbWaitForDeviceSerial(serial string) error {
+	return tc.WaitForDeviceOnline(serial, 45)
+}
+
+// Checks if adb shell is reachable for a given device serial.
+func (tc *TestContext) IsAdbShellReachable(serial string) bool {
+	res, err := tc.RunCmd("timeout", "5s", "adb", "-s", serial, "shell", "echo", "ping")
```
Databean: Sleep is not recommended in tests ([go/tott/303](http://go/tott/303)) whether using `timeout` as a shell command or `time.Sleep` below.

In the case of trying to connect with adb, the biggest source of issues I usually see is that the Cuttlefish host tools internally attempt to `adb connect` the device at regular intervals, but only if the `adb` server is already running. That means that the first `adb shell` may fail if it starts the server, and a later one succeeds because the host tools have had a chance to `adb connect`.

If the problem is specifically that issue, it should be possible to run `adb devices` first, then wait for `cvd create` / `cvd start` to return, then the device should be there.

e2etests/cvd/common/common.go
```diff
@@ -131,6 +133,43 @@ func (tc *TestContext) RunAdbWaitForDevice() error {
...
+func (tc *TestContext) WaitForDeviceOnline(serial string, timeoutSeconds int) error {
+	tc.RunCmd("adb", "connect", serial)
+	for i := 0; i < timeoutSeconds; i++ {
+		if tc.IsAdbShellReachable(serial) {
+			return nil
+		}
+		time.Sleep(1 * time.Second)
+		if i%5 == 0 {
+			tc.RunCmd("adb", "connect", serial)
+		}
```
Databean: The cuttlefish host tools will already attempt to `adb connect` in [`adb_connector`](https://github.com/google/android-cuttlefish/blob/efa10f841686d72d7bcf4be31eca796686050a2c/base/cvd/cuttlefish/host/frontend/adb_connector/adb_connection_maintainer.cpp#L222).

e2etests/cvd/common/common.go
```diff
@@ -233,21 +272,144 @@ func (tc *TestContext) CVDCreate(args CreateArgs) (CommandOutput, error) {
...
-		log.Printf("Failed to stop instance(s): %w", err)
+// Performs `cvd stop <args>`.
+func (tc *TestContext) CVDStop(args ...string) error {
+	stopCmd := append([]string{"stop"}, args...)
+	if _, err := tc.RunCVD(stopCmd...); err != nil {
+		log.Printf("Failed to stop instance(s): %v", err)
+		return err
+	}
+	return nil
+}
```
Databean: Can all the `CVD*` methods be moved out to free functions in a new file that accept a `TestContext` as the first argument? This file is getting kind of large and they all only rely on `tc.RunCVD` and not any other `TestContext` implementation details.

e2etests/cvd/common/common.go
```diff
@@ -233,21 +272,144 @@ func (tc *TestContext) CVDCreate(args CreateArgs) (CommandOutput, error) {
...
+	InstanceDir    string `json:"instance_dir"`
+	InstanceName   string `json:"instance_name"`
+	Status         string `json:"status"`
+	WebAccess      string `json:"web_access"`
+	WebRtcDeviceID string `json:"webrtc_device_id"`
+}
+
+// ParseCVDStatusJSON parses the JSON output returned by `cvd status --print`.
+// It handles both single-instance array `[...]` and group object `{"instances": [...]}` formats.
+func ParseCVDStatusJSON(output string) ([]CVDInstanceStatusEntry, error) {
```
Databean: Since the only way tests get this json output is to call `cvd status --print` and since no test calls `cvd status` without `--print`, can this be called internally from `CVDInstanceStatus` and `CVDStatus`?

Databean (commented):
> a better approach is to generalize the new test cases so they can optionally support non-SDV CF variants.

Most of the tests do look pretty general, except for the `GetSyspropStringForDevice` parts.

e2etests/cvd/common/common.go
SuperStrongDinosaur: Addressed. We I run adb start-server early in SetUp() so that the ADB daemon is already active before Cuttlefish boots. This allows Cuttlefish's adb_connector to handle connections automatically. I removed the │ manual adb connect retries, replaced the external timeout shell command with Go context.WithTimeout, and replaced sleep loops with adb wait-for-device and adb wait-for-disconnect.

e2etests/cvd/common/common.go
SuperStrongDinosaur: Done. CVDStatus and CVDInstanceStatus now automatically pass --print

e2etests/cvd/common/common.go
SuperStrongDinosaur: Done. Created e2etests/cvd/common/cvd.go and moved CVDStop, CVDStart, CVDRestart, CVDStatus, CVDInstanceStop, CVDInstanceStart, CVDInstanceRestart, and CVDInstanceStatus to free functions

SuperStrongDinosaur:
> Instead of treating SDV CF as an outlier, a better approach is to generalize the new test cases so they can optionally support non-SDV CF variants.
> 
> So I recommend the following changes:
> 
> * Generalize `sdv_1vm_test` as `lifecycle_test` and use it to replace `cvd_create_tests`.
> * Generalize `sdv_2vm_test` as `isolation_test`, apply it to `sdv_media` + `sdv_core`, and then clean up the dupplicate config from `cvd_load_tests`.
> 
> @Databean please let me know your thoughts.

Thanks for the feedback @chihchiachen. Generalizing the lifecycle and multi-instance isolation tests across other CF targets is definitely a right long-term direction. In this PR, my goal is to have the foundational presubmit coverage required specifically for SDV targets without affecting existing CI pipelines                                                  
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
Now that the CVD lifecycle methods are decoupled into free functions under e2etests/cvd/common/cvd.go, we can build a generalized lifecycle_test and isolation_test in a dedicated follow-up PR once we define common identity checks for standard CF

e2etests/cvd/common/common.go
SuperStrongDinosaur: Ok. I removed the manual adb connect call and the periodic reconnect loop

e2etests/cvd/sdv_tests/sdv_2vm_test.go
```diff
@@ -0,0 +1,246 @@
...
+
+		log.Printf("Starting %s...", instance2Name)
+		if err := c.CVDInstanceStart(instance2Name); err != nil {
+			t.Fatalf("CVDInstanceStart(%s) failed: %v", instance2Name, err)
+		}
+
+		log.Printf("Waiting for %s (%s) to come back online...", instance2Name, instance2Serial)
+		if err := c.WaitForDeviceOnline(instance2Serial, 45); err != nil {
+			t.Fatalf("Instance 2 (%s) did not come back online after start: %v", instance2Serial, err)
+		}
```
Databean, line 206: In practice, are you seeing that devices aren't immediately accessible with `adb shell` after `cvd start` or `cvd load` returns? The intended contract is that the devices are supposed to be ready at this point.

In the successful case, `adb wait-for-device` should be free if the device is ready and responsive. In the unsuccessful case, I'm not confident that waiting with some timeout will improve anything.

Databean (approved):

e2etests/cvd/sdv_tests/sdv_2vm_test.go
SuperStrongDinosaur: Yes, in practice I tested removing wait-for-device, and adb shell getprop immediately failed with exit status 1

PR #3154, opened 2026-09-08
Prevent run_cvd abort and keep monitor socket alive on graceful VM sh…

SuperStrongDinosaur:
## Summary                                                                                                                                                                                    
                                                                                                                                                                                                
Fixes b/534717429: "Shutdown/Power on VMs does not work on cuttlefish"                                                                                                                        
                                                                                                                                                                                                
Previously, when an Android guest VM initiated a clean shutdown the virtual machine manager exited with status code 0. ProcessMonitor treated any termination of a critical process as an error, causing run_cvd to abort with "process monitor has died". This caused cvd fleet and cvd status to report the device as Unreachable, and prevented subsequent cvd restart commands from reconnecting and booting the instance.                                                                                                                                                        
                                                                                                                                                                                                
This PR enables run_cvd to differentiate graceful guest VM shutdown from unexpected process crashes, keep the launcher monitor socket active and listening in a powered-off standby state, and allow cvd restart to cleanly boot the instance back up.       
                                                                                                                                                                                                                                                                                                                   
## Key Changes                                                                                                                                                                                
                                                                                                                                                                                                
1. Differentiate Graceful VMM Exit from Process Failures                                                                                                                                  
                                                                                                                                                                                              
• process_monitor.cc: In MonitorLoop(), check the exit status of the VMM process. If a critical VMM process exits with WIFEXITED and status 0, log the clean shutdown and stop all other monitored subprocesses gracefully without returning an error.                                                                                                                                 
• If a critical process exits due to a signal, with a non-zero exit code, or is an unexpected non-VMM daemon exiting prematurely, MonitorLoop() continues to treat it as an error and reports the failure.                                                                                                                                                                                  
• In MonitorRoutine(), monitor socket communication disconnects between parent and child during shutdown are logged as warnings rather than failing an otherwise clean shutdown.              
                                                                                                                                                                                              
2. Precise VMM Process Identification & Auxiliary VM Handling                                                                                                                             
                                                                                                                                                                                              
• Added IsVmmCommand() to accurately inspect whether a process is the main virtual machine manager.                                                                      
• When binaries are wrapped by process_restarter, inspect the full command line to identify the underlying executable, preventing non-VMM daemons from matching.                                                                                                                                                                                     
• Explicitly exclude auxiliary VMs so that independent exits of auxiliary components do not trigger guest VM shutdown handling.                                      
                                                                                                                                                                                              
3. Signal Safety & Clean Subprocess Termination                                                                                                                                           
                                                                                                                                                                                              
• Resume Suspended Subprocesses: In StopSubprocesses(), broadcast SIGCONT to process groups before executing process stoppers. Paused processes cannot otherwise process termination signals or exit hooks, which previously caused cvd restart to hang on suspended devices.                                               
• Signal Interruption Protection: Wrapped both non-blocking and blocking waitpid() calls in ProcessMonitor::StopMonitoredProcesses() with TEMP_FAILURE_RETRY to protect against EINTR caused  
by incoming signals like SIGCHLD.                                                                                                                                                             
• SIGPIPE Isolation: In Command::Start(), explicitly reset signal after fork() to ensure child subprocesses do not inherit run_cvd's SIG_IGN disposition across execvpe().  
• Null Safety: Added explicit null checks in StopSubprocesses() to avoid dereferencing null subprocess pointers.                                                                              
                                                                                                                                                                                              
4. Device Status Management & Action Guarding                                                                                                                                             
                                                                                                                                                                                              
• server_loop_impl.h / server_loop_impl.cpp:
    • Added DeviceStatus::kGuestOff to track when the guest VM has powered off cleanly while run_cvd remains alive.
    • In ServerLoopImpl::Run(), transition device_status_ to kGuestOff upon graceful monitor exit.
    • Guarded extended launcher actions: reject invalid operations when the device is not active or is powered off.
    • Documented LauncherAction::kStatus behavior and protocol trade-offs regarding cvd status reporting.

## Verification
• Verified lifecycle workflows:
  • Graceful guest VM shutdown via VePSM / adb reboot -p leaves launcher_monitor.sock active and eliminates zombie processes.
  • cvd restart successfully connects, reaps state, and re-executes run_cvd.
  • cvd stop cleanly terminates run_cvd with exit code 0.
  • cvd status and cvd fleet successfully respond without connection errors.
  
Bug: b/534717429


base/cvd/cuttlefish/host/commands/run_cvd/server_loop_impl.cpp
```diff
@@ -116,16 +116,27 @@ Result<void> ServerLoopImpl::Run() {
...
+      read_set.Set(process_monitor.status());
+    }
 
     Select(&read_set, nullptr, nullptr, nullptr);
 
-    if (read_set.IsSet(process_monitor.status())) {
-      return CF_ERR("process monitor has died");
+    if (process_monitor_active && read_set.IsSet(process_monitor.status())) {
+      LOG(INFO) << "Process monitor has exited (guest VM shut down). Server "
+                   "loop continuing to listen for status/restart.";
```
Databean: Does this assume that every process monitor exit implies a graceful VM shutdown? There's also the scenario that crosvm or another [critical](https://github.com/google/android-cuttlefish/blob/efa10f841686d72d7bcf4be31eca796686050a2c/base/cvd/cuttlefish/host/libs/process_monitor/process_monitor.cc#L143) process exits unexpectedly and an error should be reported.

SuperStrongDinosaur: I see. Yes, right now, `run_cvd` assumes any process monitor termination is a graceful shutdown, which would mask an unexpected crash of `crosvm` or something else.
  
We can differentiate between the two cases using the exit status of the monitor process:

1. Graceful Shutdown: The `crosvm` exits cleanly with code 0 . In this case, `ProcessMonitor` exits with code 0, and `run_cvd` keeps the launcher socket open to listen for `cvd restart`/`status`.
2. Unexpected Crash: A critical process exits with a signal (`WIFSIGNALED`) or a non-zero exit code. In this case, `ProcessMonitor` exits with code 1, and `run_cvd` reports an error .
  
Would you prefer updating `ProcessMonitor` to propagate this exit status so `run_cvd` can abort on unexpected crashes and only stay alive on 0 exits?

base/cvd/cuttlefish/host/libs/process_monitor/process_monitor.cc
```diff
@@ -77,6 +79,27 @@ Result<void> SendEmptyResponse(Channel& channel, uint32_t type) {
   return {};
 }
 
+bool IsVmmCommand(const Command& cmd) {
```
Databean: Rather than match on the name, this should be tracked the same way `is_critical` is tracked now. Maybe processes can be categorized as "vmm", "critical support", and "non-critical support"?

base/cvd/cuttlefish/host/libs/process_monitor/process_monitor.cc
```diff
@@ -138,12 +161,34 @@ Result<void> MonitorLoop(std::atomic_bool& running,
...
         if (running.load() && is_critical) {
-          LOG(ERROR) << "Stopping all monitored processes due to unexpected "
-                        "exit of critical process";
           running.store(false);
-          break;
+          if (is_vmm && WIFEXITED(wstatus) && WEXITSTATUS(wstatus) == 0) {
+            LOG(INFO)
+                << "Stopping all monitored processes due to graceful exit "
+                   "of critical process "
+                << name;
```
Databean, line 152: Does this mean that a graceful exit of the openWRT VM will trigger a shutdown of the android VM? That also seems undesirable.

base/cvd/cuttlefish/host/commands/run_cvd/server_loop_impl.cpp
```diff
@@ -116,16 +116,29 @@ Result<void> ServerLoopImpl::Run() {
...
 
-    if (read_set.IsSet(process_monitor.status())) {
-      return CF_ERR("process monitor has died");
+    if (process_monitor_active && read_set.IsSet(process_monitor.status())) {
+      process_monitor_active = false;
+      auto stop_result = process_monitor.StopMonitoredProcesses();
+      if (!stop_result.has_value()) {
+        return CF_ERR(
+            "process monitor exited unexpectedly: " << stop_result.error());
+      }
```
Databean: This is taking an error `Result`, converting it to a string, and packaging that inside another error `Result`. It would be simpler and better report the stack trace to run

```c++
CF_EXPECT(process_monitor.StopMonitoredProcesses());
```

base/cvd/cuttlefish/host/commands/run_cvd/server_loop_impl.cpp
```diff
@@ -183,6 +196,8 @@ Result<void> ServerLoopImpl::HandleExtended(
   switch (action_info.extended_action.actions_case()) {
     case ActionsCase::kSuspend: {
       VLOG(0) << "Run_cvd received suspend request.";
+      CF_EXPECT(device_status_.load() != DeviceStatus::kGuestOff,
+                "Device is powered off, cannot suspend");
```
Databean: Here and below, there is `CF_EXPECT_NE` and `CF_EXPECT_EQ`. It may require that `DeviceStatus` implements a [`format_as`](https://fmt.dev/latest/api/#formatting-user-defined-types) function ([example](https://github.com/google/android-cuttlefish/blob/01a542780acce4f28a0c831b75d361deb69054d2/base/cvd/cuttlefish/host/libs/config/gpu_mode.cc#L56)) so that it can be rendered correctly for the comparison.

base/cvd/cuttlefish/host/commands/run_cvd/server_loop_impl.cpp
Databean: Sounds good.

base/cvd/cuttlefish/host/libs/process_monitor/process_monitor.cc
SuperStrongDinosaur: Done. Introduced `ProcessCategory` in `command_source.h`, stored it in `MonitorCommand` and `MonitorEntry`. Removed the `IsVmmCommand` name-matching helper completely.

base/cvd/cuttlefish/host/libs/process_monitor/process_monitor.cc
SuperStrongDinosaur: Done. With `ProcessCategory`, auxiliary VMs like OpenWrt are configured as `kCriticalSupport` rather than `kVmm`. Only a clean exit of the primary `kVmm` process initiates the graceful VM shutdown  sequence; an exit of OpenWrt will not trigger a guest VM shutdown.

base/cvd/cuttlefish/host/commands/run_cvd/server_loop_impl.cpp
SuperStrongDinosaur: Done. Simplified to `CF_EXPECT(process_monitor.StopMonitoredProcesses());` to preserve the original error result and stack trace

base/cvd/cuttlefish/host/commands/run_cvd/server_loop_impl.cpp
SuperStrongDinosaur: Done. Implemented `format_as`, `operator<<` and converted the precondition checks in `HandleExtended` to use `CF_EXPECT_NE` and `CF_EXPECT_EQ`.

base/cvd/cuttlefish/host/libs/process_monitor/process_monitor.cc
```diff
@@ -153,7 +177,15 @@ Result<void> MonitorLoop(std::atomic_bool& running,
 
 Result<void> StopSubprocesses(std::vector<MonitorEntry>& monitored) {
   VLOG(0) << "Stopping monitored subprocesses";
+  for (const auto& it : monitored) {
+    if (it.proc) {
+      (void)it.proc->SendSignalToGroup(SIGCONT);
+    }
+  }
```
Databean: Can this be removed? There's no `SIGSTOP` as far as I can see, so what is `SIGCONT` covering?

Databean (approved):

base/cvd/cuttlefish/host/libs/process_monitor/process_monitor.cc
SuperStrongDinosaur: Done, removed. You are right, monitored processes are never sent SIGSTOP/SIGTSTP during normal operation  or shutdown

PR #3160, opened 2026-09-09
Execute podman system migrate by podcvd-setup, only when it's required

0405ysj:
When `podcvd-setup` is executed, all podman containers are stopped due to execution of `podman system migrate`.  However `podman system migrate` was required to launch any rootless podman instance, so it should be one-time setup right after setting subuid/subgid properly.

Context: b/558905779


jemoreira (approved):

0405ysj:
Thanks for the review!

PR #3161, opened 2026-09-09
Operator can now delegate its infra_config values.

3405691582:
Instead of statically allocating the output of the infra_config handler, we allow the operator to optionally delegate its contents to a sidecar command that can supply the values returned by this handler. This means that if a user wishes to supply alternate STUN or TURN servers for running webRTC, it can do so.

Why not a static configuration loaded from a file? For TURN for example, the credential returned from the TURN server may have a limited TTL, so we need to be able to request the TURN credential on-demand. We default however to letting the operator set the credential TTL for the moment, and may revisit that behavior in the future.

If the delegate is specified and fails for whatever reason, then we fall back to the existing, static configuration so as not to break existing clients.

TURN support is particularly useful, since it means that large port-forward ranges are no longer necessary.

### DIFF
--- a/frontend/src/liboperator/operator/operator.go
+++ b/frontend/src/liboperator/operator/operator.go
@@ -26,8 +26,11 @@ import (
 	"net/http/httputil"
 	"net/url"
 	"os"
+	"os/exec"
 	"strconv"
 	"strings"
+	"sync"
+	"time"
 
 	apiv1 "github.com/google/android-cuttlefish/frontend/src/liboperator/api/v1"
 	"github.com/gorilla/mux"
@@ -101,6 +104,85 @@ func SetupControlEndpoint(pool *DevicePool, path string) (func() error, error) {
 	}, nil
 }
 
+type InfraConfigProvider interface {
+	Get() (apiv1.InfraConfig, error)
+}
+
+type StaticInfraConfigProvider struct {
+	config apiv1.InfraConfig
+}
+
+func NewStaticInfraConfigProvider(
+	config apiv1.InfraConfig,
+) *StaticInfraConfigProvider {
+	return &StaticInfraConfigProvider{config: config}
+}
+
+func (s *StaticInfraConfigProvider) Get() (apiv1.InfraConfig, error) {
+	return s.config, nil
+}
+
+const DefaultDelegateTimeout = 10 * time.Second
+
+type DelegateInfraConfigProvider struct {
+	staticConfig apiv1.InfraConfig
+	delegatePath string
+	defaultTTL   time.Duration
+	timeout      time.Duration
+
+	mu        sync.Mutex
+	cfg       apiv1.InfraConfig
+	expiresAt time.Time
+}
+
+func NewDelegateInfraConfigProvider(
+	staticConfig apiv1.InfraConfig,
+	delegatePath string,
+	defaultTTL, timeout time.Duration,
+) *DelegateInfraConfigProvider {
+	if timeout <= 0 {
+		timeout = DefaultDelegateTimeout
+	}
+	return &DelegateInfraConfigProvider{
+		staticConfig: staticConfig,
+		delegatePath: delegatePath,
+		defaultTTL:   defaultTTL,
+		timeout:      timeout,
+	}
+}
+
+func (p *DelegateInfraConfigProvider) Get() (apiv1.InfraConfig, error) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	if time.Now().Before(p.expiresAt) {
+		return p.cfg, nil
+	}
+
+	p.cfg = p.runDelegate()
+	p.expiresAt = time.Now().Add(p.defaultTTL)
+	return p.cfg, nil
+}
+
+func (p *DelegateInfraConfigProvider) runDelegate() apiv1.InfraConfig {
+	ctx, cancel := context.WithTimeout(context.Background(), p.timeout)
+	defer cancel()
+
+	cmd := exec.CommandContext(ctx, p.delegatePath)
+	out, err := cmd.Output()
+	if err != nil {
+		log.Printf("runDelegate failed: %v", err)
+		return p.staticConfig
+	}
+
+	var cfg apiv1.InfraConfig
+	if err := json.Unmarshal(out, &cfg); err != nil {
+		log.Printf("runDelegate failed: %v", err)
+		return p.staticConfig
+	}
+	return cfg
+}
+
 // Creates a router with handlers for the following endpoints:
 // GET  /infra_config
 // GET  /devices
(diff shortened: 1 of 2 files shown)


frontend/src/liboperator/operator/operator.go
```diff
@@ -122,10 +125,83 @@ func SetupControlEndpoint(pool *DevicePool, path string) (func() error, error) {
 // The maybeIntercept parameter is a function that accepts the
 // requested device file and returns a path to a file to be returned instead or
 // nil if the request should be allowed to proceed to the device.
```
0405ysj: This comment is about `CreateHttpHandlers` function. Could we move it into proper places?

frontend/src/operator/main.go
```diff
@@ -91,6 +92,9 @@ func main() {
 	webUiUrlStr := flag.String("webui_url", DefaultWebUIUrl, "WebUI URL.")
 	address := flag.String("listen_addr", DefaultListenAddress, "IP address to listen for requests.")
 	logFile := flag.String("log_file", "", "Path to file to write logs to.")
+	infraConfigDelegate := flag.String("infra_config_delegate", "", "Path to delegate binary returning infrastructure config JSON.")
+	infraConfigTTL := flag.Duration("infra_config_ttl", 5*time.Minute, "Default TTL for cached infrastructure config.")
+	infraConfigTimeout := flag.Duration("infra_config_timeout", operator.DefaultDelegateTimeout, "Timeout for infrastructure config delegate execution.")
```
0405ysj: Like any other flags here, I expect introducing const variable right below `DefaultListenAddress` would be better?

frontend/src/liboperator/operator/operator.go
```diff
@@ -122,10 +125,83 @@ func SetupControlEndpoint(pool *DevicePool, path string) (func() error, error) {
...
+
+func (p *DelegateInfraConfigProvider) runDelegate() apiv1.InfraConfig {
+	ctx, cancel := context.WithTimeout(context.Background(), p.timeout)
+	defer cancel()
+
+	cmd := exec.CommandContext(ctx, p.delegatePath)
+	out, err := cmd.Output()
+	if err != nil {
+		log.Printf("runDelegate failed: %v", err)
+		out = []byte("{}")
```
0405ysj: We can probably return `p.staticConfig` early?

frontend/src/liboperator/operator/operator.go
```diff
@@ -122,10 +125,83 @@ func SetupControlEndpoint(pool *DevicePool, path string) (func() error, error) {
...
+	}
+
+	cfg := p.staticConfig
```
0405ysj: `var cfg apiv1.InfraConfig`?

frontend/src/operator/main.go
```diff
@@ -91,6 +92,9 @@ func main() {
 	webUiUrlStr := flag.String("webui_url", DefaultWebUIUrl, "WebUI URL.")
 	address := flag.String("listen_addr", DefaultListenAddress, "IP address to listen for requests.")
 	logFile := flag.String("log_file", "", "Path to file to write logs to.")
+	infraConfigDelegate := flag.String("infra_config_delegate", "", "Path to delegate binary returning infrastructure config JSON.")
```
0405ysj, line 97: Q: Could I get any example of such binaries?

frontend/src/liboperator/operator/operator.go
3405691582: We could, but since Unmarshal will merge into p.staticConfig anyway, the effect is the same. If you'd prefer to be explicit though, let me know.

frontend/src/operator/main.go
3405691582: Any shell script that just dumps JSON in the right form to stdout will do the trick.

frontend/src/liboperator/operator/operator.go
3405691582: The default value is the static config. Unmarshal here will merge its contents on the static config.

The delegate could supply the entire config, but that's a more invasive change when we just want a way to add extra additional configuration in some cases.

0405ysj: I wasn't aware of the intention that merging with static config. Leaving my thoughts behind.

As this PR's intention is to deliver TURN server configuration into `infra_config` someday IIUC, most existing configurations including `IceServers` will be overridden, not reused. So I expect merging config is not beneficial according to current `apiv1.InfraConfig` type. If you desire to merge configs properly such as appending elements into the slice, you probably want https://pkg.go.dev/dario.cat/mergo.

frontend/src/liboperator/operator/operator.go
0405ysj: Not strong opinion, but I'm leaning on returning it early for better readability.

frontend/src/liboperator/operator/operator.go
3405691582: Done.

frontend/src/liboperator/operator/operator.go
3405691582: Done. I've switched behavior such that the delegate must construct a full InfraConfig.

frontend/src/liboperator/operator/operator.go
3405691582: Done. I've switched behavior such that the delegate must construct a full InfraConfig.

frontend/src/operator/main.go
3405691582: Done.

0405ysj (approved):

frontend/src/liboperator/operator/operator.go
```diff
@@ -101,6 +104,85 @@ func SetupControlEndpoint(pool *DevicePool, path string) (func() error, error) {
...
+	p.cfg = p.runDelegate()
+	p.expiresAt = time.Now().Add(p.defaultTTL)
+	return p.cfg, nil
+}
+
+func (p *DelegateInfraConfigProvider) runDelegate() apiv1.InfraConfig {
+	ctx, cancel := context.WithTimeout(context.Background(), p.timeout)
+	defer cancel()
+
+	cmd := exec.CommandContext(ctx, p.delegatePath)
```
ser-io, line 171: From a security standpoint I'm concerned how this binary is provided? Does it ship along the operator binary in the frontend package? Should we hardcode its SHA-256 in order to verify the integrity of the binary to be run?

ser-io (changes requested):

frontend/src/liboperator/operator/operator.go
3405691582: This binary must be provided from the client side and I am not proposing shipping any new binaries as part of the packages we provide on GitHub. This change merely provides the mechanism to properly configure the operator's behavior.

ser-io (approved):

PR #3167, opened 2026-09-10
Add support for Venus GPU mode

valentineburley:
Based on the original implementation for Mesa CI by @ao2.

This adds a new GPU mode using the Venus Vulkan driver, ANGLE, SkiaVK,
and drm_hwcomposer, utilizing the cross-domain context type.

It supports Intel (both i915 and xe) and Qualcomm (msm) platforms, using
ANV and Turnip as the host Vulkan drivers.

Bug: b/549217582
Test: cvd create --gpu_mode=venus

The first commit fixes kHostVulkanIsNonSoftwareRenderer on integrated GPUs.

The following three patches add the virglrenderer, venus_protocol, and minigbm dependencies.

The final commit introduces the new GPU mode.

### DIFF
--- a/base/cvd/cuttlefish/host/commands/assemble_cvd/graphics_flags.cc
+++ b/base/cvd/cuttlefish/host/commands/assemble_cvd/graphics_flags.cc
@@ -234,10 +234,15 @@ GetGpuModeRequirementsMap() {
           [](const CommonState& common) {
             const auto& availability = common.graphics_availability;
             if (availability.has_vulkan() &&
-                !availability.vulkan().physical_devices().empty() &&
-                (availability.vulkan().physical_devices(0).type() !=
-                 ::gfxstream::proto::VulkanPhysicalDevice::TYPE_DISCRETE_GPU)) {
-              return false;
+                !availability.vulkan().physical_devices().empty()) {
+              const auto type =
+                  availability.vulkan().physical_devices(0).type();
+              if (type != ::gfxstream::proto::VulkanPhysicalDevice::
+                              TYPE_DISCRETE_GPU &&
+                  type != ::gfxstream::proto::VulkanPhysicalDevice::
+                              TYPE_INTEGRATED_GPU) {
+                return false;
+              }
             }
             return true;
           },
--- a/base/cvd/cuttlefish/host/graphics_detector/graphics_detector.proto
+++ b/base/cvd/cuttlefish/host/graphics_detector/graphics_detector.proto
@@ -66,6 +66,8 @@ message VulkanPhysicalDevice {
   enum Type {
     TYPE_OTHER = 0;
     TYPE_DISCRETE_GPU = 1;
+    TYPE_INTEGRATED_GPU = 2;
+    TYPE_CPU = 3;
   }
   optional Type type = 3;
 
--- a/base/cvd/cuttlefish/host/graphics_detector/graphics_detector_vk.cpp
+++ b/base/cvd/cuttlefish/host/graphics_detector/graphics_detector_vk.cpp
@@ -35,10 +35,24 @@ gfxstream::expected<Ok, vk::Result> PopulateVulkanAvailabilityImpl(
 
     const auto props = physicalDevice.getProperties();
     outPhysicalDevice->set_name(std::string(props.deviceName));
-    outPhysicalDevice->set_type(
-        props.deviceType == vk::PhysicalDeviceType::eDiscreteGpu
-            ? ::gfxstream::proto::VulkanPhysicalDevice::TYPE_DISCRETE_GPU
-            : ::gfxstream::proto::VulkanPhysicalDevice::TYPE_OTHER);
+    switch (props.deviceType) {
+      case vk::PhysicalDeviceType::eDiscreteGpu:
+        outPhysicalDevice->set_type(
+            ::gfxstream::proto::VulkanPhysicalDevice::TYPE_DISCRETE_GPU);
+        break;
+      case vk::PhysicalDeviceType::eIntegratedGpu:
+        outPhysicalDevice->set_type(
+            ::gfxstream::proto::VulkanPhysicalDevice::TYPE_INTEGRATED_GPU);
+        break;
+      case vk::PhysicalDeviceType::eCpu:
+        outPhysicalDevice->set_type(
+            ::gfxstream::proto::VulkanPhysicalDevice::TYPE_CPU);
+        break;
+      default:
+        outPhysicalDevice->set_type(
+            ::gfxstream::proto::VulkanPhysicalDevice::TYPE_OTHER);
+        break;
+    }
 
     const auto exts =
         VK_EXPECT_RV(physicalDevice.enumerateDeviceExtensionProperties());


base/debian/control
```diff
@@ -13,6 +13,7 @@ Build-Depends: bazelisk [amd64 arm64],
                libcap-dev,
                libclang-dev,
                libcurl4-openssl-dev,
+               libepoxy-dev,
```
jmacnak: I noticed that the change in `crosvm_manager.cpp` has `{"androidboot.hardware.egl", "angle"},`. 

IIRC, epoxy was the GL func pointer loader/helper.

Does venus not support running in a VK only mode? How much work would it be to carve that out and remove the epoxy dep?

base/cvd/cuttlefish/metrics/external_proto/cf_flags.proto
```diff
@@ -95,6 +95,7 @@ message CuttlefishFlags {
     CUTTLEFISH_FLAGS_GPU_MODE_GUEST_VIRGL_RENDERER = 8;
     CUTTLEFISH_FLAGS_GPU_MODE_NONE = 9;
     CUTTLEFISH_FLAGS_GPU_MODE_GUEST_LAVAPIPE = 10;
+    CUTTLEFISH_FLAGS_GPU_MODE_GUEST_VENUS = 11;
```
jmacnak: Drop the `GUEST`. Looks like some of the other `GUEST_` prefixes might just be a copy paste fail.

jmacnak (commented):
(note for internal build: cl/979356492)

base/cvd/cuttlefish/metrics/external_proto/cf_flags.proto
valentineburley: Right, done. Thanks!

base/debian/control
valentineburley: Thanks for flagging this.

I gave it a try and it works fine after removing epoxy.

I added a few stubs to the Bazel build for now to share a quick proof of concept, but I intend to properly take this upstream to virglrenderer.

base/cvd/cuttlefish/package/BUILD.bazel
```diff
@@ -82,8 +86,10 @@ package_files(
         "bin/health": "//cuttlefish/host/commands/health",
         "bin/kernel_log_monitor": "//cuttlefish/host/commands/kernel_log_monitor",
         "bin/libmem_overrides.so": "//cuttlefish/host/graphics/mem_overrides:libmem_overrides.so",
+        "bin/libgbm.so.1": "@minigbm//:libgbm.so.1",
         "bin/libgfxstream_backend.so": "@gfxstream//host:gfxstream_backend",
-        "bin/libvulkan_lvp.so": "@mesa//:vulkan_lvp",
+        "bin/libvirglrenderer.so.1": "@virglrenderer//:libvirglrenderer.so.1",
```
jmacnak, line 91: Instead of putting more libs in the `bin` subdirectory (I believe the gfxstream and VK library ones were just due to backwards compat with android built host tools), could we keep these in the `lib64` subdirectory and potentially add an additional `rpath` pointing to the `lib64` subdirectory (I believe that would be `../lib64` for crosvm and rutabaga in `crosvm.MODULE.bazel`?

base/cvd/cuttlefish/host/commands/assemble_cvd/graphics_flags.cc
```diff
@@ -631,7 +664,7 @@ Result<bool> SelectGpuVhostUserMode(const GpuMode gpu_mode,
             gpu_vhost_user_mode_arg == kGpuVhostUserModeOn ||
             gpu_vhost_user_mode_arg == kGpuVhostUserModeOff);
   if (gpu_vhost_user_mode_arg == kGpuVhostUserModeAuto) {
-    if (IsGuestRenderingMode(gpu_mode)) {
+    if (IsGuestRenderingMode(gpu_mode) || gpu_mode == GpuMode::Venus) {
```
jmacnak, line 667: Is there a particular reason for disallowing vhost-user-gpu for venus?

(IIRC, venus works with the strong full crosvm sandboxing as well? Potentially https://github.com/google/android-cuttlefish/blob/e761fd1b4cae66dc46e5c2df55c4c95a8354b8e2/base/cvd/cuttlefish/host/commands/assemble_cvd/flags.cc#L1226 could be updated to allow the sandbox when running with venus?)

jmacnak (commented):
Overall looks good. 2 minor questions

base/cvd/cuttlefish/host/commands/assemble_cvd/graphics_flags.cc
valentineburley: Looking into these, thanks.

jmacnak: Checked in with Yiwei and he mentioned the need for fixed-blob-mapping and how that might not be supported in vhost-user-mode. Feel free to ignore this thread and keep as is.

zzyiwei: > IIRC, venus works with the strong full crosvm sandboxing as well?

...and Valentine has confirmed venus in Cuttlefish does still work with full crosvm sandbox enabled. So yeah, let's avoid vhost-user-gpu for now.

base/debian/control
valentineburley: Upstream MR: https://gitlab.freedesktop.org/virgl/virglrenderer/-/merge_requests/1682

I'll update virglrenderer once it lands. This will also allow removing gallium from the build here.

base/cvd/cuttlefish/host/commands/assemble_cvd/graphics_flags.cc
valentineburley: Updated to enable full sandboxing. I had some issues with Cuttlefish picking up an older local seccomp directory, but Venus indeed works well with full sandboxing when using Cuttlefish's newer policies with `seccomp_policy_dir=""`.

base/debian/control
valentineburley: Done.

base/cvd/cuttlefish/package/BUILD.bazel
valentineburley: Done. Thanks fore the `rpath` pointer, that's what I was missing.

jmacnak:
(conflicts need to be resolved before we can approve presubmit workflows)

valentineburley:
Rebased, thanks!

jmacnak:
(rebased to pull in some changes that I believe are needed to make kokoro happy >_>)

jmacnak:
(rebased to re-try kokoro again)

valentineburley:
I don't have access to the kokoro logs, so if there's something actionable on my side please let me know.

jmacnak:
> * kokoro

I see an abort in `netsimd` which looks unrelated. Rebasing and retrying ;)

PR #3171, opened 2026-09-10
bazel 7 readiness test

grandpaul:
Hi. I added bazel-7 readiness test. This is based on the Debian package we have in Debian unstable. 
The patchset is reduced to only do the readiness test. Also the mock_repos are reduced too. 
Please review the commits. 
Thanks.


tools/testutils/bazel-7-readiness/patches/0025-remove-unused-build-external-modules.patch
```diff
@@ -0,0 +1,130 @@
+--- a/base/cvd/MODULE.bazel	2026-09-10 21:48:00.155816566 +0000
++++ b/base/cvd/MODULE.bazel	2026-09-10 21:48:00.156377436 +0000
```
Databean, line 2: The changes to `MODULE.bazel` and `build_external.MODULE.bazel` seem very brittle, like we'll run into merge conflicts with these patches on practically any change to these files. Is there any way to reduce the change surface here?

It's not a quick fix but I was looking at converting some  of the `git_repository` dependencies to `bazel_dep`s in https://github.com/google/android-cuttlefish/pull/3172 and https://github.com/google/android-cuttlefish/pull/3173 which might align better with the `override_module` setup in `run_bazel_7_readiness_test.sh‎`.

tools/testutils/bazel-7-readiness/patches/0016-cuttlefish-Provide-fallback-wrapper-for-symbolic-mac.patch
```diff
@@ -0,0 +1,27 @@
+--- a/base/cvd/cuttlefish/bazel/rules.bzl	2026-09-10 21:46:27.729380206 +0000
++++ b/base/cvd/cuttlefish/bazel/rules.bzl	2026-09-10 21:46:27.730110696 +0000
+@@ -30,6 +30,24 @@
+ COPTS = BUILD_VAR_COPTS
+ LINKOPTS = BUILD_VAR_LINKOPTS
+ 
++def _fallback_macro(inherit_attrs = None, attrs = {}, implementation = None):
```
Databean, line 7: Rather than as a patch file, can this be in `rules.bzl`?

If I understand correctly, https://github.com/bazel-contrib/bazel_features can be used, e.g.

```
load("@bazel_features//:features.bzl", "bazel_features")
if not bazel_features.globals.macro:
    def ...
```

tools/testutils/bazel-7-readiness/run_bazel_7_readiness_test.sh
```diff
@@ -0,0 +1,173 @@
...
+
+DEB_HOST_MULTIARCH=`dpkg-architecture -qDEB_HOST_MULTIARCH`
+HOME1=`mktemp -d`
+CACHE1=`mktemp -d`
+MOCKREPOS=`mktemp -d`
+python3 `pwd`/tools/testutils/bazel-7-readiness/generate_mock_repos.py ${MOCKREPOS}
+
+for i in `pwd`/tools/testutils/bazel-7-readiness/patches/*.patch; do
+    patch -p1 < $i
+done
```
Databean, line 12: One of the nice things about the bazel 9 readiness test is that it is possible to reproduce issues locally with something like `OVERRIDE_BAZEL_VERSION=9.0.0 bazel build <something>`, but this script looks a lot more invasive.

What's the intended workflow here if a developer sees a presubmit failure on the github action? Running this script looks like it will make a lot of changes to their local git repository that can only be reset by `git restore`.

tools/testutils/bazel-7-readiness/patches/0025-remove-unused-build-external-modules.patch
grandpaul: OK. I'll see if I can convert those dependencies needed here.

PR #3175, opened 2026-09-11
broadcastradio: Add Audio Generator and PCM Stream Server for Virtual Tuner

abinba:
Introduce the audio emulation pipeline for the Cuttlefish Virtual Tuner:
- AudioGenerator: Generates 48 kHz, 16-bit stereo PCM audio frames (white noise while tuned/playing, silence while stopped/untuned).
- PcmStreamServer: Multi-client PCM stream server delivering 48 kHz 16-bit stereo audio frames over a UNIX domain stream socket to Cuttlefish's virtio-snd virtual sound card backend (CrosVM).
- Unit and integration tests covering audio frame synthesis and socket streaming.

Test: bazel test //cuttlefish/host/commands/virtual_tuner_daemon:... 
Bug: 521329213
TAG=agy
CONV=809829e8-ac68-4c32-b9b7-632f61743e47

### DIFF
--- a/base/cvd/cuttlefish/host/commands/virtual_tuner_daemon/pcm_stream_server.cpp
+++ b/base/cvd/cuttlefish/host/commands/virtual_tuner_daemon/pcm_stream_server.cpp
@@ -1,0 +1,217 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cuttlefish/host/commands/virtual_tuner_daemon/pcm_stream_server.h"
+
+#include <sys/types.h>
+#include <unistd.h>
+
+#include <atomic>
+#include <cerrno>
+#include <chrono>
+#include <cstddef>
+#include <cstdint>
+#include <memory>
+#include <mutex>
+#include <ratio>
+#include <string>
+#include <thread>
+#include <utility>
+#include <vector>
+
+#include "absl/log/log.h"
+
+#include "cuttlefish/common/libs/fs/fd.h"
+#include "cuttlefish/common/libs/fs/shared_fd.h"
+#include "cuttlefish/host/commands/virtual_tuner_daemon/audio_generator.h"
+#include "cuttlefish/host/commands/virtual_tuner_daemon/tuner_state.h"
+#include "cuttlefish/result/result.h"
+
+namespace cuttlefish {
+namespace virtualtuner {
+namespace {
+
+constexpr mode_t kSocketMode = 0600;
+
+using ChunkDuration =
+    std::chrono::duration<int64_t, std::ratio<kChunkFrames, kSampleRate>>;
+
+constexpr auto kMaxSchedulingLag = std::chrono::milliseconds(500);
+constexpr int kPrebufferChunks = 2;
+
+bool SendAll(const SharedFD& fd, const void* data, size_t size) {
+  const char* cursor = static_cast<const char*>(data);
+  size_t remaining = size;
+  while (remaining > 0) {
+    const ssize_t written = fd->Send(cursor, remaining, MSG_NOSIGNAL);
+    if (written > 0) {
+      cursor += written;
+      remaining -= static_cast<size_t>(written);
+    } else if (written < 0 && fd->GetErrno() == EINTR) {
+      continue;
+    } else {
+      return false;
+    }
+  }
+  return true;
+}
+
+}  // namespace
+
+PcmStreamServer::PcmStreamServer(TunerState* tuner_state,
+                                 std::string socket_path)
+    : tuner_state_(tuner_state), socket_path_(std::move(socket_path)) {}
+
+PcmStreamServer::~PcmStreamServer() { Stop(); }
+
+Result<void> PcmStreamServer::Start() {
+  if (is_running_) {
+    return {};
+  }
+
+  ::unlink(socket_path_.c_str());
+  server_fd_ = SharedFD::SocketLocalServer(
+      socket_path_, /* is_abstract= */ false, SOCK_STREAM, kSocketMode);
+  CF_EXPECT(server_fd_->IsOpen(),
+            "Failed to open PCM streaming server UNIX socket at "
+                << socket_path_ << ": " << server_fd_->StrError());
+
+  LOG(INFO) << "PCM streaming server listening on " << socket_path_;
+  is_running_ = true;
+  accept_thread_ = std::thread(&PcmStreamServer::AcceptLoop, this);
+  return {};
+}
+
+void PcmStreamServer::Stop() {
+  if (!is_running_.exchange(false)) {
+    return;
+  }
+
+  if (server_fd_->IsOpen()) {
+    server_fd_->Shutdown(SHUT_RDWR);
+    server_fd_->Close();
+  }
+  if (accept_thread_.joinable()) {
+    accept_thread_.join();
+  }
+
+  std::vector<std::unique_ptr<ClientSession>> sessions;
+  {
+    std::lock_guard<std::mutex> lock(clients_mutex_);
+    sessions.swap(clients_);
+  }
+  for (const auto& session : sessions) {
+    if (session->fd->IsOpen()) {
+      session->fd->Shutdown(SHUT_RDWR);
+    }
+  }
+  for (auto& session : sessions) {
+    if (session->thread.joinable()) {
+      session->thread.join();
+    }
+  }
+
+  ::unlink(socket_path_.c_str());
+  LOG(INFO) << "PCM streaming server stopped.";
+}
+
+void PcmStreamServer::AcceptLoop() {
+  while (is_running_) {
+    SharedFD client_fd = Fd::Accept(*server_fd_).value_or(Fd());
+    if (!client_fd->IsOpen()) {
+      if (!is_running_) {
+        break;
+      }
+      continue;
+    }
+
+    std::lock_guard<std::mutex> lock(clients_mutex_);
+    ReapFinishedClients();
+    if (clients_.size() >= kMaxConcurrentClients) {
+      LOG(WARNING) << "Refusing PCM client: already serving " << clients_.size()
+                   << " clients.";
+      client_fd->Close();
+      continue;
+    }
+
+    LOG(INFO) << "PCM streaming server accepted a client.";
+    auto session = std::make_unique<ClientSession>();
+    session->fd = std::move(client_fd);
+    ClientSession* session_ptr = session.get();
+    clients_.push_back(std::move(session));
+    session_ptr->thread =
+        std::thread(&PcmStreamServer::StreamClient, this, session_ptr);
+  }
+}
+
+void PcmStreamServer::ReapFinishedClients() {
+  std::erase_if(clients_, [](const std::unique_ptr<ClientSession>& session) {
+    if (!session->finished.load(std::memory_order_acquire)) {
+      return false;
+    }
+    if (session->thread.joinable()) {
+      session->thread.join();
+    }
+    return true;
+  });
+}
+
+void PcmStreamServer::StreamClient(ClientSession* session) {
+  const SharedFD& client_fd = session->fd;
+  AudioGenerator generator;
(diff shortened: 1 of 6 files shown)


base/cvd/cuttlefish/host/commands/virtual_tuner_daemon/pcm_stream_server.h
```diff
@@ -0,0 +1,71 @@
...
+    SharedFD fd;
+    std::thread thread;
+    std::atomic<bool> finished = false;
+  };
+
+  void AcceptLoop();
+  void StreamClient(ClientSession* session);
+  void ReapFinishedClients();
+
+  TunerState* const tuner_state_;
```
jemoreira, line 61: nit: use `TunerState&`

base/cvd/cuttlefish/host/commands/virtual_tuner_daemon/pcm_stream_server.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+bool SendAll(const SharedFD& fd, const void* data, size_t size) {
+  const char* cursor = static_cast<const char*>(data);
+  size_t remaining = size;
+  while (remaining > 0) {
+    const ssize_t written = fd->Send(cursor, remaining, MSG_NOSIGNAL);
+    if (written > 0) {
+      cursor += written;
+      remaining -= static_cast<size_t>(written);
+    } else if (written < 0 && fd->GetErrno() == EINTR) {
```
jemoreira, line 63: `Fd::Send` handles `EINTR` for you, this check is unnecessary.

base/cvd/cuttlefish/host/commands/virtual_tuner_daemon/pcm_stream_server.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+constexpr mode_t kSocketMode = 0600;
+
+using ChunkDuration =
+    std::chrono::duration<int64_t, std::ratio<kChunkFrames, kSampleRate>>;
+
+constexpr auto kMaxSchedulingLag = std::chrono::milliseconds(500);
+constexpr int kPrebufferChunks = 2;
+
+bool SendAll(const SharedFD& fd, const void* data, size_t size) {
```
jemoreira, line 55: There is already a `SendAll` function in `common/libs/fs/shared_buf.*`, please add an overload that takes `void*, size_t` instead of `std::string_view` and have the existing one call the new one to avoid repetition.

While you're at it, please add the `flags` parameter as well with a default value of `MSG_NOSIGNAL` to make it more flexible without changing the current behavior.

base/cvd/cuttlefish/host/commands/virtual_tuner_daemon/pcm_stream_server.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+PcmStreamServer::~PcmStreamServer() { Stop(); }
+
+Result<void> PcmStreamServer::Start() {
+  if (is_running_) {
+    return {};
+  }
+
+  ::unlink(socket_path_.c_str());
+  server_fd_ = SharedFD::SocketLocalServer(
```
jemoreira, line 86: Because of the way cuttlefish spawns processes there is no guarantee that this socket will be created before the client attempts to connect to it. Our way of dealing with this race is to have the parent process `run_cvd` create the socket and let the child process (this one) inherit it. The actual fd is passed by command line parameter.

There are several examples under `host/commands/run_cvd/launch` that you can use for guidance.

base/cvd/cuttlefish/host/commands/virtual_tuner_daemon/pcm_stream_server.cpp
```diff
@@ -0,0 +1,217 @@
...
+  return {};
+}
+
+void PcmStreamServer::Stop() {
+  if (!is_running_.exchange(false)) {
+    return;
+  }
+
+  if (server_fd_->IsOpen()) {
+    server_fd_->Shutdown(SHUT_RDWR);
```
jemoreira, line 104: why `Shutdown`? Is `Close` not enough?

base/cvd/cuttlefish/host/commands/virtual_tuner_daemon/pcm_stream_server.cpp
```diff
@@ -0,0 +1,217 @@
...
+}
+
+void PcmStreamServer::AcceptLoop() {
+  while (is_running_) {
+    SharedFD client_fd = Fd::Accept(*server_fd_).value_or(Fd());
+    if (!client_fd->IsOpen()) {
+      if (!is_running_) {
+        break;
+      }
+      continue;
```
jemoreira, line 138: nit: you may want to log why this fd is not open, the reason is likely in `client_fd->StrError()`.

base/cvd/cuttlefish/host/commands/virtual_tuner_daemon/pcm_stream_server.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+  ::unlink(socket_path_.c_str());
+  LOG(INFO) << "PCM streaming server stopped.";
+}
+
+void PcmStreamServer::AcceptLoop() {
+  while (is_running_) {
+    SharedFD client_fd = Fd::Accept(*server_fd_).value_or(Fd());
+    if (!client_fd->IsOpen()) {
+      if (!is_running_) {
```
jemoreira, line 135: nit: you could check this even when `client_fd` is open, it will close without sharing any data after going out of scope.

base/cvd/cuttlefish/host/commands/virtual_tuner_daemon/pcm_stream_server.cpp
```diff
@@ -0,0 +1,217 @@
...
+      }
+      continue;
+    }
+
+    std::lock_guard<std::mutex> lock(clients_mutex_);
+    ReapFinishedClients();
+    if (clients_.size() >= kMaxConcurrentClients) {
+      LOG(WARNING) << "Refusing PCM client: already serving " << clients_.size()
+                   << " clients.";
+      client_fd->Close();
```
jemoreira, line 146: nit: Explicit close is not needed, the `continue` will cause `client_fd` to go out of scope and close on its destructor.

base/cvd/cuttlefish/host/commands/virtual_tuner_daemon/pcm_stream_server.cpp
```diff
@@ -0,0 +1,217 @@
...
+      continue;
+    }
+
+    LOG(INFO) << "PCM streaming server accepted a client.";
+    auto session = std::make_unique<ClientSession>();
+    session->fd = std::move(client_fd);
```
jemoreira, line 152: nit: `client_fd` is a `SharedFD`, it's typically copied instead of moved, just like a `shared_ptr`.

base/cvd/cuttlefish/host/commands/virtual_tuner_daemon/pcm_stream_server.cpp
```diff
@@ -0,0 +1,217 @@
...
+      return false;
+    }
+    if (session->thread.joinable()) {
+      session->thread.join();
+    }
+    return true;
+  });
+}
+
+void PcmStreamServer::StreamClient(ClientSession* session) {
```
jemoreira, line 172: nit: use `ClientSession&` for the parameter

base/cvd/cuttlefish/host/commands/virtual_tuner_daemon/pcm_stream_server.cpp
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  auto anchor = std::chrono::steady_clock::now();
+  int64_t chunks_since_anchor = 0;
+
+  while (connected && is_running_) {
+    ++chunks_since_anchor;
+    const auto deadline = anchor + ChunkDuration(chunks_since_anchor);
+    std::this_thread::sleep_until(deadline);
+    if (!is_running_) {
```
jemoreira, line 194: nit: If you use `std::condition_variable::wait_until` you can notify the variable from the stop routine and not have to wait until this sleep ends.

base/cvd/cuttlefish/host/commands/virtual_tuner_daemon/pcm_stream_server.cpp
```diff
@@ -0,0 +1,217 @@
...
+    return true;
+  });
+}
+
+void PcmStreamServer::StreamClient(ClientSession* session) {
+  const SharedFD& client_fd = session->fd;
+  AudioGenerator generator;
+  std::vector<int16_t> buffer(kChunkFrames * kChannels);
+
+  const auto write_next_chunk = [&] {
```
jemoreira, line 177: please write the type for this lambda since it's not obvious that it returns bool without looking at other source locations.

base/cvd/cuttlefish/host/commands/virtual_tuner_daemon/virtual_tuner_audio_test.cpp
```diff
@@ -0,0 +1,263 @@
...
+#include "cuttlefish/host/commands/virtual_tuner_daemon/audio_generator.h"
+#include "cuttlefish/host/commands/virtual_tuner_daemon/pcm_stream_server.h"
+#include "cuttlefish/host/commands/virtual_tuner_daemon/tuner_state.h"
+#include "cuttlefish/result/result.h"
+
+namespace cuttlefish {
+namespace virtualtuner {
+namespace {
+
+bool ReadExactly(const SharedFD& fd, std::span<uint8_t> buffer) {
```
jemoreira, line 39: We already have ReadExact in shared_buf.h

jemoreira (commented):
Please rebase your branch on top of main instead of creating merge commits.

base/cvd/cuttlefish/host/commands/virtual_tuner_daemon/virtual_tuner_audio_test.cpp
abinba: Calling ReadExact and casting it to char* here.

base/cvd/cuttlefish/host/commands/virtual_tuner_daemon/pcm_stream_server.cpp
abinba: Done. Added an overload with `void*, size_t` and the existing one calling the new one. Using default value of `MSG_NOSIGNAL`.

base/cvd/cuttlefish/host/commands/virtual_tuner_daemon/pcm_stream_server.cpp
abinba: We need `Shutdown()` because `accept_thread_` is sleeping inside blocking `Fd::Accept()` when `Stop()` is called. If `accept_thread_` weren't sleeping inside `Accept()`, `Close()` alone would be sufficient.

base/cvd/cuttlefish/host/commands/virtual_tuner_daemon/pcm_stream_server.cpp
abinba: Added the injection of `server_fd` into the PCM streaming server.

The changes to the `run_cvd` for injection would be in the subsequent PR to keep this PR minimal.

base/cvd/cuttlefish/common/libs/fs/shared_buf.cc
```diff
@@ -74,20 +74,25 @@ ssize_t WriteAll(SharedFD fd, const std::vector<char>& buf) {
...
+  size_t total_written = 0;
   if (!sock->IsOpen()) {
     return false;
   }
-  while (total_written < static_cast<ssize_t>(msg.size())) {
-    auto just_written = sock->Send(msg.data() + total_written,
-                                   msg.size() - total_written, MSG_NOSIGNAL);
+  const char* cursor = static_cast<const char*>(buf);
+  while (total_written < size) {
+    auto just_written =
```
jemoreira: nit: Please avoid `auto` except when the type is mentioned very near or when it's too onerous/impossible to state explicitly, as described in the [style guide](https://google.github.io/styleguide/cppguide.html#Type_deduction). I see that's how it was before, but since you're changing it now might as well fix it.

base/cvd/cuttlefish/common/libs/fs/shared_buf.cc
```diff
@@ -74,20 +74,25 @@ ssize_t WriteAll(SharedFD fd, const std::vector<char>& buf) {
...
+        sock->Send(cursor + total_written, size - total_written, flags);
     if (just_written <= 0) {
       return false;
     }
-    total_written += just_written;
+    total_written += static_cast<size_t>(just_written);
```
jemoreira: nit: The static_cast wouldn't be needed if `total_written` was still `ssize_t`.

base/cvd/cuttlefish/host/commands/virtual_tuner_daemon/pcm_stream_server.cpp
jemoreira: SG

jemoreira (approved):

PR #3186, opened 2026-09-15
build_packages.sh doesn't build cuttlefish-podcvd anymore

0405ysj:
Context: b/561649046


Databean:
Looks like this change will have to be updated for https://github.com/google/android-cuttlefish/pull/3188/ when that goes in.

Databean (approved):

ser-io (changes requested):
This PR is no longer relevant:

See https://github.com/google/android-cuttlefish/pull/3184 and https://github.com/google/android-cuttlefish/pull/3188

PR #3192, opened 2026-09-17
Pad vbmeta images during cvd fetch

3405691582:
The guest reads the vbmeta images with libavb, which expects to be able to read the maximum vbmeta size, so we pad the images in place to match this or the read will fail.

This implies that the image location must be writable, which is problematic for particular use-cases. It is worthwhile noting that images are also modified during cvd fetch to de-sparse images before the images are quiescent on disk. While this change does not alleviate the need for in-place image modification, it does ensure that the images need not be modifiable after the cvd fetch occurs.

assemble_cvd behavior is unchanged -- indeed it must not change, since it will need to operate correctly on images not sourced from cvd fetch.


base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_context.cc
```diff
@@ -53,6 +55,28 @@ static constexpr mode_t kRwxAllMode = S_IRWXU | S_IRWXG | S_IRWXO;
...
 using android::base::Dirname;
 
+namespace {
+
+bool IsVbmetaImage(std::string_view path) {
+  return absl::EndsWith(path, "/vbmeta.img") ||
+         absl::EndsWith(path, "/vbmeta_system.img") ||
+         absl::EndsWith(path, "/vbmeta_system_dlkm.img") ||
+         absl::EndsWith(path, "/vbmeta_vendor_dlkm.img");
+}
```
Databean, line 65: Not a huge fan of extra behavior for specific filenames, but this does look like the simplest change.

The `vbmeta_system` call at least is made explicitly
https://github.com/google/android-cuttlefish/blob/01a542780acce4f28a0c831b75d361deb69054d2/base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_cvd.cc#L286-L295
but the other case is buried inside `ExtractAll`.

What do you think of changing `ExtractAll` to return the paths of extracted files, so that `fetch_cvd.cc` can do postprocessing like padding?

Databean (approved):

base/cvd/cuttlefish/host/commands/cvd/fetch/fetch_context.cc
3405691582: We could do this but there's some subtlety in both the shape and semantics of the API (paths versus filenames) and the types being returned (bare strings, or more structured data), so it probably deserves a separate change.

PR #3196, opened 2026-09-17
Start sharding presubmit cvd e2e tests.

ser-io:
Add the initial two first jobs. The new jobs would start running as regular presubmit jobs.

Bug: b/562957494


.kokoro/presubmit_cvd_e2etest_runner_1.cfg
```diff
@@ -0,0 +1,10 @@
+build_file: "android-cuttlefish/.kokoro/presubmit.sh"
```
jmacnak: Could we instead default to this to a new `presubmit_noop.sh` until we have the sharding infrastructure in place to avoid the extra 1.5 hour runs?

jmacnak (changes requested):

.kokoro/presubmit_cvd_e2etest_runner_1.cfg
ser-io: Done.

jmacnak (approved):

PR #3197, opened 2026-09-17
Support OpenWRT and wmediumd snapshot/restore

Elie-Kh:
Enable proper snapshot save and restore for OpenWRT AP instances and their vhost-user mac80211_hwsim connections in wmediumd:

- Patch external wmediumd to implement vhost-user device state save (SLEEP + GET_VRING_BASE + DEVICE_STATE) and restore (DEVICE_STATE + SET_VRING_ADDR + SET_VRING_BASE + SET_VRING_KICK + SET_VRING_CALL + WAKE) callbacks so existing vhost-user clients can reconnect and resume without losing station/queue state or spinning on stale EPOLLHUP events.
- Update CvdBootStateMachine to explicitly resume the OpenWRT crosvm instance (`crosvm resume <sock> --full`) upon guest restore completion, replacing the previous `service network restart` LuciRPC workaround.
- Check that the OpenWRT snapshot directory exists before passing `--restore` to crosvm when launching OpenWrt, falling back to a cold boot with a warning if absent.
- Cache the discovered OpenWRT IP address in OpenwrtControlServiceImpl so IP lookups continue to succeed after restoring from a snapshot when launcher.log no longer emits fresh boot logs.

Bug: 335361533
Test: Verified snapshot take and restore with OpenWRT Wi-Fi enabled
TAG=agy
CONV=5df97b60-361a-4803-9dc9-6c486b536214


base/cvd/build_external/wmediumd/0001-wmediumd-Implement-device-state-save-and-restore-for.patch
```diff
@@ -0,0 +1,750 @@
+From 4d439a3686758dde1a6707781988f2a4efdd57f3 Mon Sep 17 00:00:00 2001
+From: Elie Kheirallah <khei@google.com>
+Date: Fri, 4 Sep 2026 17:40:12 +0000
+Subject: [PATCH] wmediumd: Implement device state save and restore for
+ snapshot
```
Databean, line 5: Is there a plan to send this wmediumd upstream as well? Honestly I'm not sure what the upstream is at this point, https://github.com/ramonfontes/wmediumd looks like it has the most recent changes, but the code imported to Android is based on this version: https://github.com/bcopeland/wmediumd , and that's also what this `build_external` code references.

base/cvd/cuttlefish/host/commands/run_cvd/boot_state_machine.cc
Databean: Since this anonymous function is becoming kind of large, can it be a top-level function? The top-level function can accept the file descriptors as arguments, and `std::thread` can forward the values, i.e.

```c++
void RestoreComplete(SharedFD restore_complete_stop_read, SharedFD restore_complete_stop_write) {
  ...
}
...
std::thread(RestoreComplete restore_complete_stop_read, restore_complete_stop_write)
```

If it needs to access a lot of members of `this`, [`std::bind_front`](https://en.cppreference.com/cpp/utility/functional/bind_front) can capture the object reference. See [abseil.io/tips/108](https://abseil.io/tips/108) for more on `std::bind_front`.

base/cvd/cuttlefish/host/commands/run_cvd/boot_state_machine.cc
```diff
@@ -387,27 +382,60 @@ class CvdBootStateMachine : public SetupFeature, public KernelLogPipeConsumer {
...
+                if (guest_snapshot_dir_suffix.has_value()) {
+                  const auto restore_path = snapshot_dir_path + "/" +
+                                            *guest_snapshot_dir_suffix + "/" +
+                                            kGuestSnapshotBase + "_openwrt";
+                  openwrt_restored = DirectoryExists(restore_path);
+                }
+              }
+            }
+            if (openwrt_restored) {
+              const auto openwrt_sock = instance_.OpenwrtCrosvmSocketPath();
```
Databean: We have [`WaitForUnixSocket`](https://github.com/google/android-cuttlefish/blob/01a542780acce4f28a0c831b75d361deb69054d2/base/cvd/cuttlefish/common/libs/utils/wait_for_unix_socket.cpp#L37) with a `ListeningWithoutConnect` variation. The difference is one will periodically try to connect, and the other will not.

base/cvd/cuttlefish/host/commands/run_cvd/boot_state_machine.cc
```diff
@@ -387,27 +382,60 @@ class CvdBootStateMachine : public SetupFeature, public KernelLogPipeConsumer {
...
+                      "--full",
+                  },
+                  SubprocessOptions(), WEXITED);
+              CHECK(infop.has_value())
+                  << "Failed to execute crosvm resume for OpenWRT: "
+                  << infop.error();
+              CHECK_EQ(infop->si_code, CLD_EXITED);
+              CHECK_EQ(infop->si_status, 0)
+                  << "crosvm resume for OpenWRT returned non-zero code "
+                  << infop->si_status;
```
Databean: This can use the fewer-arguments version of `Execute` which has comparable behavior and a simpler return type:
https://github.com/google/android-cuttlefish/blob/01a542780acce4f28a0c831b75d361deb69054d2/base/cvd/cuttlefish/process/execute.cc#L33-L42

Databean (approved):
Added @jemoreira for the wmediumd changes for vhost-user experience

jemoreira (approved):

base/cvd/build_external/wmediumd/0001-wmediumd-Implement-device-state-save-and-restore-for.patch
Elie-Kh: It's probably worth investigating deltas. If we aren't too divergent from https://github.com/ramonfontes/wmediumd, it might be worthwhile consolidating.

Databean, line 5: It occurs to me also that because we don't have an internal build of wmediumd, this PR might cause problems for Google employees trying to use the internal Cuttlefish host tools build to launch non-`git_main` devices like AOSP builds because they'll be using the latest run_cvd but an older build of wmediumd.

Elie-Kh: Good point.. Have we had a precedence of this?

Databean: Two similar cases I can think of, neither with a satisfying / relevant conclusion:

- `secure_env` has had additional feature work on git_main, and I'm pushing to get that ported to the github project.
- `netsim` has had additional feature work on git_main. We're trying to temporarily roll back shipping netsim in our debian package until netsim finalizes their own github release that we can rely on.

PR #3205, opened 2026-09-18
Split the Reader type into a new file

Databean:
To avoid a circular dependency, this removes the default implementation of Reader::visit.

This change attempts to limit the blast radius of changes made to common headers.

Bug: b/562645911


jemoreira (approved):

Databean:
Thanks for reviewing!

PR #3208, opened 2026-09-21
Exclude `cuttlefish-podcvd` from deployment

0405ysj:
With https://github.com/google/android-cuttlefish/pull/3184, `cuttlefish-podcvd` debian package is no longer built from `build-debian-packages` GitHub action, however our deployment procedure relies on it. Due to that, [postsubmit workflow](https://github.com/google/android-cuttlefish/actions/workflows/postsubmit.yaml) is broken. To apply immediate fix on postsubmit workflow, deployment should exclude `cuttlefish-podcvd` and include it again afterwards.

Context: b/561649046


0405ysj:
Note: Will bypass presubmit check as this is only for postsubmit workflow change.

ikicha (approved):

PR #3211, opened 2026-09-21
cvd host_bugreport: don't discard the whole archive on a directory entry

shuowanghsu:
`AddFileAt` guarded its input only with `FileExists`, which is a `stat` check and therefore succeeds for anything that exists, including directories. libzip does not read the source until `WritableZip::Finalize`, where reading a directory fails with `EISDIR`. libzip then calls `zip_source_rollback_write`, which discards the **entire** archive rather than just the offending entry.

`AddNetsimdLogs` hits this: it lists the netsimd temp directory with a non-recursive `DirectoryContents`, so when netsimd runs with `--pcap` it hands the `pcap/` directory straight to `AddFileAt`. The result is that `cvd host_bugreport` produces an archive containing almost nothing, and the netsimd capture files are never collected either.

Against the unpatched library the new test reproduces it exactly:

```
Expected "zip_close(raw_zip)" == "0" but was -1 vs 0. Read error: Is a directory
```

### Changes

1. **Require a regular file in `AddFileAt`.** This is the important one — today any caller that passes a directory silently destroys the archive it was building, and does so at finalization time, far from the offending call site. After this change a bad entry can only lose itself. The check is `S_ISREG` rather than just a directory test, because a FIFO or a socket reaches `Finalize` the same way, and opening one there can fail or block indefinitely.
2. **Collect netsimd logs recursively**, naming each entry by its path relative to the netsimd directory, so the capture files under `pcap/` are actually included. Per-entry errors are logged rather than returned, so one bad file cannot abort the rest of the walk.
3. **Add `WalkDirectoryFiles`** next to the existing `WalkDirectory` in `common/libs/utils/files.h`, reporting only non-directories. Symlinks are not followed when deciding whether to recurse, so a symlink loop cannot trap the walk.
4. **Add regression tests** for both halves: that `AddFileAt` rejects directories and FIFOs, that a rejected entry no longer takes unrelated entries down with it at `Finalize`, and that `WalkDirectoryFiles` recurses while `WalkDirectory` keeps reporting directories.

### Notes

- `WalkDirectoryFiles` is a new sibling rather than a change to `WalkDirectory`, because `SymlinkHostPackage`, `WriteFsConfig` and `LinkOrCopyDirectoryContentsRecursively` all rely on `WalkDirectory` invoking the callback for directories. `WriteFsConfig` in particular would silently drop directory entries from `filesystem_config.txt`. Happy to consolidate them differently if you'd prefer.
- `logs/`, `tombstones/` and `recording/` are left non-recursive. They are flat in practice, and after change 1 an unexpected subdirectory there is merely logged and skipped rather than fatal. Can extend `WalkDirectoryFiles` to them if you'd like uniformity.
- Per-entry errors remain non-fatal, consistent with the existing contract that "a partial bug report is still useful."

### Testing

```
bazel test  //cuttlefish/host/libs/zip:zip_file_test \
            //cuttlefish/common/libs/utils:files_test
bazel build //cuttlefish/host/commands/host_bugreport:cvd_internal_host_bugreport
```

All new test cases pass with the fix; the `zip_file_test` cases fail without it.


Voyagerroc-Lab (commented):
Review summary

The regular-file guard prevents directories, FIFOs, and sockets from reaching deferred libzip reads; the recursive netsimd walk collects nested capture files without allowing one bad entry to abort the rest. The new tests cover archive preservation, FIFO rejection, directory/file traversal semantics, and the build target includes the new utility dependency.

I found no blocking correctness issue in the reviewed diff. The security scan checks are green; the focused Bazel validation is documented by the author.

base/cvd/cuttlefish/host/libs/zip/zip_file.cc
```diff
@@ -53,7 +52,14 @@ Result<void> AddFile(WritableZip& zip, const std::string& fs_path) {
...
 Result<void> AddFileAt(WritableZip& zip, const std::string& fs_path,
                        const std::string& zip_path) {
-  CF_EXPECTF(FileExists(fs_path), "No file in the filesystem at '{}'", fs_path);
+  // libzip does not read the source until `WritableZip::Finalize`, where a
+  // directory fails with EISDIR and a FIFO can block indefinitely. libzip then
+  // rolls back and discards the *entire* archive, so reject anything that is
+  // not a regular file here, where the error can only lose this one entry.
+  struct stat file_stat;
+  CF_EXPECTF(stat(fs_path.c_str(), &file_stat) == 0,
+             "No file in the filesystem at '{}'", fs_path);
```
Databean, line 61: Can this use the [`Stat` wrapper](https://github.com/google/android-cuttlefish/blob/5a467328d69a4caa2715ad0da3bf1edd2b8c47ba/base/cvd/cuttlefish/posix/stat.h#L27)?

Databean (approved):

Databean:
Two tests failed:
```
 INFO: Build completed, 2 tests FAILED, 45888 total actions
  //cuttlefish/common/libs/utils:files_test_format_test_C++_with_clang-format FAILED in 0.1s
    /home/runner/.cache/bazel/_bazel_runner/a1ea073c0d21ddc5ae89652639f74223/execroot/_main/bazel-out/k8-fastbuild/testlogs/cuttlefish/common/libs/utils/files_test_format_test_C++_with_clang-format/test.log
  //cuttlefish/host/commands/host_bugreport:cvd_internal_host_bugreport_format_test_C++_with_clang-format FAILED in 0.1s
    /home/runner/.cache/bazel/_bazel_runner/a1ea073c0d21ddc5ae89652639f74223/execroot/_main/bazel-out/k8-fastbuild/testlogs/cuttlefish/host/commands/host_bugreport/cvd_internal_host_bugreport_format_test_C++_with_clang-format/test.log
```

PR #3212, opened 2026-09-22
Enable use_random_serial as default on `podcvd create`

0405ysj:
Context: b/564412239


0405ysj:
@Databean `CF_DEFAULTS_USE_RANDOM_SERIAL` is set as false in `cvd`. Would you see any issues when flipping that boolean value? Personally, I'm believing using random serial is better than fixed value. If it sounds okay, I'm thinking of dropping commits here and set it from `cvd`.

Reference: https://android-review.googlesource.com/c/device/google/cuttlefish/+/1347468

Databean:
Thanks for the reference to [aosp/1347468](https://android-review.googlesource.com/c/device/google/cuttlefish/+/1347468). I think the issue described in b/159852618 is mostly applicable now for snapshots, where the snapshot-restore launch shouldn't have a different serial number than the original device. I'm not sure what the interaction is between the two features right now.

0405ysj:
According to your reply, I expect it's better to replace current commits into changing `CF_DEFAULTS_USE_RANDOM_SERIAL` from `cvd`. I've looked around `cvd resume` right before, there were no flags around device serial. And I assume device serial should be kept after resuming regardless of enabling/disabling `use_random_serial`, or it's a bug of `cvd`.

And have one question, correct if I'm wrong. I'm believing one of the major use cases of suspend/resume is suspending one, copying the snapshot, and resuming several. Then I expect all CF instances will have same device serial, would this be ideal? @Databean

Databean:
> I expect it's better to replace current commits into changing `CF_DEFAULTS_USE_RANDOM_SERIAL` from cvd. I've looked around cvd resume right before, there were no flags around device serial. And I assume device serial should be kept after resuming regardless of enabling/disabling `use_random_serial`, or it's a bug of cvd.

I believe the problem is that `assemble_cvd --use_random_serial` randomizes the serial every time it runs, and restoring from a snapshot still requires running `assemble_cvd` again, with the same flags it had the first time.

> I'm believing one of the major use cases of suspend/resume is suspending one, copying the snapshot, and resuming several. Then I expect all CF instances will have same device serial, would this be ideal?

No, the instance numbers currently are baked into the instance state and must be exclusive per instance, so it is not possible to resume with a different number. There is also currently no infrastructure to move snapshots between machines where they wouldn't conflict.

0405ysj:
> I believe the problem is that assemble_cvd --use_random_serial randomizes the serial every time it runs, and restoring from a snapshot still requires running assemble_cvd again, with the same flags it had the first time.

I'm speculating we have 2 routes for using `cvd` with snapshot feature. For both routes, I've verified the device serial is preserved. Also I've found the device serial is captured under `cuttlefish_config.json`. It doesn't seem to be executing assemble_cvd when restoring, correct me if I'm wrong. And I'm a bit uncertain whether we have more user journeys or not, and I think I can check if there's any.

```
cvd create --use_random_serial=true
adb shell getprop ro.serialno
cvd suspend
cvd resume
adb shell getprop ro.serialno
```

```
cvd create --gpu_mode=guest_swiftshader --enable_sandbox=false --enable_virtiofs=false --use_random_serial=true
adb shell getprop ro.serialno
cvd snapshot_take --auto_suspend --snapshot_path=/tmp/cf_snap
cvd stop
cvd start --snapshot_path=/tmp/cf_snap
adb shell getprop ro.serialno
```

> No, the instance numbers currently are baked into the instance state and must be exclusive per instance, so it is not possible to resume with a different number. There is also currently no infrastructure to move snapshots between machines where they wouldn't conflict.

I might have misunderstood how snapshot feature works.. I thought snapshot feature supports restoring after `cvd remove`, but it looks not. Never mind on my prev comment, and thanks!

